PR #15, opened 2019-05-07
Developer Interface

Sam-Jeston:
This feature determined the developer interface, and provided functional examples, linked from the README.

I settled on a class with static and instance methods. It felt "right" from the examples.

The following was implemented to facilitate this:
 - A Provider interface for interactions with the protocol
 - A mocked provider for examples and testing
 - The Wallet interface was abstracted behind a higher order function as it should get a reference to the provider when the CardanoSdk class is instantiated
- verifyMessage was moved to the static Util module, as it didn't make sense as part of Wallet
- The return type of signMessage now reflects what a client would expect
- Cleanup to the transaction fee logic, based on Sebastien's recommendations


src/Transaction/Transaction.ts
```diff
@@ -28,27 +29,23 @@ export function Transaction (inputs: TransactionInput[], outputs: TransactionOut
   })
 
   return {
-    estimateFee: () => estimateTransactionFee(transactionBuilder),
-    validateAndMake: () => validateAndMake(transactionBuilder),
+    estimateNetworkFee: (feeAlgorithm = LinearFeeAlgorithm.default()): string => {
```
rhyslbw:
```suggestion
    estimateNetworkFee: (): string => {
```

rhyslbw: I think I prefer `estimateFee` here, or we should make `fee`: `networkFee`

src/Transaction/Transaction.spec.ts
```diff
@@ -29,49 +30,66 @@ describe('Transaction', () => {
     expect(() => Transaction(inputs, invalidOutputType)).to.throw(/Invalid value/)
   })
 
-  describe('Finalize', () => {
-    it('throws if a transaction has more input than output', () => {
-      const inputs = [
-        { pointer: { id: '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef', index: 1 }, value: { address: 'addressWithFunds1', value: '1000000' } },
-        { pointer: { id: 'fedcba9876543210fedcba9876543210fedcba9876543210fedcba9876543210', index: 0 }, value: { address: 'addressWithFunds2', value: '5000000' } }
-      ]
+  it('throws if a transaction has more output than input', () => {
```
rhyslbw:
```suggestion
  it('throws if a transaction has more combined output value than input value', () => {
```

src/Transaction/Transaction.spec.ts
```diff
@@ -29,49 +30,66 @@ describe('Transaction', () => {
...
-    })
+    expect(() => Transaction(inputs, outputs)).to.throw(InsufficientTransactionInput)
+  })
 
-    it('throws if a transaction has more output than input', () => {
-      const inputs = [
-        { pointer: { id: '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef', index: 1 }, value: { address: 'addressWithFunds1', value: '1' } },
-        { pointer: { id: 'fedcba9876543210fedcba9876543210fedcba9876543210fedcba9876543210', index: 0 }, value: { address: 'addressWithFunds2', value: '1' } }
-      ]
+  it('does not throw if a transaction has more input than output', () => {
```
rhyslbw:
```suggestion
  it('accepts more combined input value than output, to cover fees', () => {
```

src/Transaction/Transaction.spec.ts
```diff
@@ -29,49 +30,66 @@ describe('Transaction', () => {
...
+    let outputs = [
+      { address: 'Ae2tdPwUPEZCEhYAUVU7evPfQCJjyuwM6n81x6hSjU9TBMSy2YwZEVydssL', value: '10000' }
+    ]
 
-      expect(() => Transaction(inputs, outputs).validateAndMake()).to.throw(TransactionOverweight)
-    })
+    expect(() => Transaction(inputs, outputs)).to.not.throw()
+  })
+
+  it('allows access to a transaction as hex when the transaction is balanced', () => {
```
rhyslbw:
```suggestion
  it('allows access to a transaction as hex', () => {
```

src/Transaction/Transaction.ts
```diff
@@ -1,19 +1,35 @@
-import { Transaction as CardanoTransaction, TransactionBuilder as CardanoTransactionBuilder, LinearFeeAlgorithm as CardanoLinearFeeAlgorithm } from 'cardano-wallet'
 import { getBindingsForEnvironment } from '../lib/bindings'
-import { TransactionOverweight, TransactionUnderweight } from './errors'
+import { InsufficientTransactionInput } from './errors'
```
rhyslbw: 👍

src/test/MemoryKeyManager.spec.ts
```diff
@@ -0,0 +1,17 @@
+import { expect } from 'chai'
+import { Utils, InMemoryKeyManager, connect } from '..'
+import { generateTestTransaction } from './utils/test_transaction'
+import { mockProvider } from './utils/mock_provider'
+
+describe('Example: Memory Key Manager', () => {
+  const password = 'secure'
+
+  it('allows a user to create a new memory key from a valid mnemonic, sign a transaction and submit it to the network', () => {
```
rhyslbw:
```suggestion
  it('allows a user to create a key manager in memory from a valid mnemonic, sign a transaction and submit it to the network', () => {
```

src/Provider/HttpProvider.ts
```diff
@@ -1,6 +1,8 @@
-export function Api (_uri: string) {
+import { Provider } from '.'
+
+export function HttpProvider (_uri: string): Provider {
   return {
     submitTransaction: (_signedTransaction: string) => new Error('Not yet implemented'),
     queryUtxo: (_utxos: string[]) => new Error('Not yet implemented')
-  }
+  } as any
```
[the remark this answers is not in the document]
rhyslbw: In future it's probably best to leave this out until we can implement fully. There's already a stub provider for testing

PR #20, opened 2019-05-10
Wallet interface

Sam-Jeston:
The wallet interface faciliates:
- Determing next receipt and change addresses
- Querying transactions for a BIP44 public account
- Querying utxos for a BIPP44 public account
- Selecting inputs for transaction outputs using the WASM FirstMatchFirst bindings


README.md
```diff
@@ -2,10 +2,10 @@
 JavaScript SDK for interacting with Cardano, providing various key management options including support for popular hardware wallets
 
 ## Features
-*This SDK is a work in progress. Feature state as follows:*
+*This SDK is a work in progress and should not be used in production. Feature state as follows:*
```
rhyslbw: 👍

src/Wallet/Wallet.spec.ts
```diff
@@ -0,0 +1,152 @@
+import { expect } from 'chai'
+import { Utils, InMemoryKeyManager } from '..'
+import { AddressType, addressDiscoveryWithinBounds } from '.'
+import { mockProvider, seedTransactionSet, seedUtxoSet, generateTestTransaction, generateTestUtxos } from '../test/utils'
+import { Wallet } from './Wallet'
+
+describe('Wallet', () => {
+  it('getNextChangeAddress', async () => {
```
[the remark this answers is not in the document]
Sam-Jeston: I've already tested the extended behaviour of the lib functions in the module itself (i.e. returning the first internal address of a new wallet vs a wallet with existing utxos). Doing so again  here would be duplication. The desire of these specs was just to get coverage across the top level paths, where we instantiate the wallet with the provider, and then call the returned function with an account.

So totally agree regarding the refactor to describe and beforeEach for clarity regarding construction of tests, but more "it" blocks here would just duplicate the lib tests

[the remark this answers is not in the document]
rhyslbw: Fair point. If you move the wallet instantiation to the `beforeEach` block, then it's communicating the intent a little better. ie this test suite doesn't cover the complex scenarios of state variations.

src/Wallet/config.ts
```diff
@@ -0,0 +1,2 @@
+export const MAX_TRANSACTION_INPUTS = 50
```
rhyslbw: This is also not being used anywhere

src/Wallet/lib/get_next_address.ts
```diff
@@ -0,0 +1,70 @@
...
+  })
+
+  const transactions = await provider.queryTransactionsByAddress(addresses.map(a => a.address))
+
+  if (transactions.length === 0) {
+    return addresses[0]
+  }
+
+  // Group transactions by address, if they have outputs for that address, it means they have had a utxo in the past.
+  // The transactions are now garunteed to be in address order, as they are grouped against the
```
rhyslbw:
```suggestion
  // The transactions are now guaranteed to be in address order, as they are grouped against the
```

src/Wallet/lib/get_next_address.ts
```diff
@@ -0,0 +1,70 @@
...
+    type
+  })
+
+  const transactions = await provider.queryTransactionsByAddress(addresses.map(a => a.address))
+
+  if (transactions.length === 0) {
+    return addresses[0]
+  }
+
+  // Group transactions by address, if they have outputs for that address, it means they have had a utxo in the past.
```
rhyslbw:
```suggestion
  // Group transactions by address, if they have outputs for that address, it means they have had a UTXO in the past.
```

PR #29, opened 2019-06-13
Cardano Primitives

Sam-Jeston:
To improve the flexibility of this SDK, we abstract the cryptographic primitives available through the rust wasm binding behind an interface. This allows us to use other modules for these primitives as they become available.

So consequences of this refactor:
- Upgrade `cardano-wallet` dependency to the most recent version. This changed some interfaces, and the transaction.balance.is_negative behaviour has changed as well. I have not looked into this further as I think its outside our scope. Possibly we just require that inputs - outputs is a positive number, and let the network determine if the fee is appropriate. This is the web3 way
- Relevant top level interfaces now take `Cardano` as an optional parameter. These are currently defaulted to `RustCardano`


src/lib/Ledger/index.spec.ts
```diff
@@ -1,12 +1,12 @@
 import { LedgerKeyManager } from '.'
-import { KeyManager } from '../KeyManager'
+import { KeyManager } from '../../KeyManager/KeyManager'
```
rhyslbw: As a general rule I think `index.ts` for `KeyManager` within a directory named `KeyManager` is a good style, or we could decide to go with exporting the module as a default from the index. Either way, let's make it consistent

src/index.ts
```diff
@@ -1,18 +1,35 @@
-import Transaction from './Transaction'
+import Transaction, { TransactionInput, TransactionOutput } from './Transaction'
 import Wallet from './Wallet'
 import { Provider, HttpProvider } from './Provider'
 import * as Utils from './Utils'
-import { InMemoryKeyManager } from './KeyManager'
-
-export { Transaction, InMemoryKeyManager, Utils }
+import { InMemoryKeyManager, LedgerKeyManager, RustCardano } from './lib'
```
rhyslbw: 👍 nice and clean

src/lib/Ledger/index.spec.ts
Sam-Jeston: Ahhh that was just the editor

Sam-Jeston: > As a general rule I think index.ts for KeyManager within a directory named KeyManager is a good style
100%

PR #71, opened 2021-08-20
chore: auto-generate docs with typedoc

sweeetland:
# Context

- docs, typedoc

# Proposed Solution

This PR adds a script `yarn docs` which will auto-generate documentation with typedoc.

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@
 
 A Yarn Workspace containing packages to collectively form the SDK, written in TypeScript.
 
+- [@cardano-sdk/cardano-graphql-provider](packages/cardano-graphql-provider)
 - [@cardano-sdk/core](./packages/core)
 - [@cardano-sdk/golden-test-generator](./packages/golden-test-generator)
 
@@ -52,6 +53,10 @@ yarn cleanup
 ### Publish to npm.org
 ```console
 ./scripts/publish.sh
+```
+### Generate Docs
+```console
+yarn docs
 ```
 
 [img_src_CI]: https://github.com/input-output-hk/cardano-js-sdk/actions/workflows/continuous-integration.yaml/badge.svg
--- a/README.md
+++ b/README.md
@@ -59,5 +59,9 @@ yarn cleanup
 yarn docs
 ```
 
+<p align="center">
+  <a href="https://input-output-hk.github.io/cardano-js-sdk">:book: Documentation</a>
+</p>
+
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -14,6 +14,7 @@
 
 A Yarn Workspace containing packages to collectively form the SDK, written in TypeScript.
 
+- [@cardano-sdk/cardano-graphql-provider](packages/cardano-graphql-provider)
```
sweeetland, line 17: These links are broken when on the docs homepage, I'm not sure how to fix?

rhyslbw (changes requested):
1. It was unspecified, but let's gitignore the docs output as the primary purpose is to deploy to GH pages.
2. Please add a section to the README Generate Docs:

image

README.md
rhyslbw: I've fixed this by disabling the injection of the readme, and just have the module index as the landing page. We don't need this to be a fully self-contained site, as the codebase links to these docs

```
readme": "none"
```

packages/cardano-graphql-provider/src/cardanoGraphqlProvider.ts
```diff
@@ -3,6 +3,12 @@ import { gql, GraphQLClient } from 'graphql-request';
 import { TransactionSubmitResponse } from '@cardano-graphql/client-ts';
 import { Schema as Cardano } from '@cardano-ogmios/client';
 
+/**
```
rhyslbw, line 6: @sweeetland 

We should aim to provide a description and example for most functions or modules

packages/core/src/Provider/CardanoProvider.ts
```diff
@@ -4,6 +4,7 @@ import { Schema as Cardano } from '@cardano-ogmios/client';
 type Tx = { hash: Cardano.Hash16 } & Cardano.Tx;
 
 export interface CardanoProvider {
+  /** @param signedTransaction signed and serialized cbor */
```
rhyslbw, line 7: Also adding documentation here for more detail on the param since the type does help. We should probably improve the type directly, but it's not critical at this point.

typedoc.json
```diff
@@ -0,0 +1,5 @@
+{
```
rhyslbw, line 1: Lifted the config to it's own file, to avoid polluting `tsconfig.json`

rhyslbw (approved):
For clarity, I left notes to detail my changes.

PR #72, opened 2021-08-23
Refactor/ Rescope and rename `cardano-graphql-provider` package

rhyslbw:
# Context
Just a quick refactor to ensure the package is scoped and named for future intentions.

# Proposed Solution
:pushpin: **refactor(cardano-graphql-provider): rescope and rename package**
- Rescopes the package to not specifically be a provider, so we can add
other modules in the future if needed.
- Renames to reserve `cardano-graphql` for an API not based on `cardano-db-sync`
or the Hasura-influenced model.
- Override unicorn lint rule demanding abbreviations

:pushpin: **chore: remove superfluous eslintignore entry**
Also adds IDE config to only lint `.ts`


sweeetland (approved):
👍

PR #73, opened 2021-08-24
feat(blockfrost): create new provider called blockfrost

sweeetland:
# Context

- [blockfrost](https://blockfrost.io/)

# Proposed Solution

This PR creates a new package @cardano-sdk/blockfrost with a provider that can connect to the blockfrost service.

### DIFF
--- a/packages/blockfrost/src/blockfrostProvider.ts
+++ b/packages/blockfrost/src/blockfrostProvider.ts
@@ -1,0 +1,73 @@
+import { CardanoProvider } from '@cardano-sdk/core';
+import { Schema as Cardano } from '@cardano-ogmios/client';
+
+import { BlockFrostAPI } from '@blockfrost/blockfrost-js';
+import { blockfrostOutputToCardanoTxOut, blockfrostTxContentUtxoToCardanoTx } from './utils';
+import { Options } from '@blockfrost/blockfrost-js/lib/types';
+
+/**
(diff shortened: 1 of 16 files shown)


rhyslbw (changes requested):
The testnet service requires an extra config option, so we should include it for reference despite the mock not requiring it.

packages/blockfrost/src/blockfrostProvider.ts
```diff
@@ -0,0 +1,73 @@
+import { CardanoProvider } from '@cardano-sdk/core';
+import { Schema as Cardano } from '@cardano-ogmios/client';
+
+import { BlockFrostAPI } from '@blockfrost/blockfrost-js';
+import { blockfrostOutputToCardanoTxOut, blockfrostTxContentUtxoToCardanoTx } from './utils';
+import { Options } from '@blockfrost/blockfrost-js/lib/types';
+
+/**
+ * Connect to the [blockfrost service](https://docs.blockfrost.io/)
```
rhyslbw, line 9:
```suggestion
 * Connect to the [Blockfrost service](https://docs.blockfrost.io/)
```

rhyslbw (changes requested):

rhyslbw (changes requested):
Missing the yarn.lock changes

rhyslbw (approved):

PR #79, opened 2021-09-03
Feature/cardano serialization lib translations

rhyslbw:
# Context
We need a way to easily move between our base types (Ogmios) and the format required for transaction construction.

# Proposed Solution
:pushpin: **feature(cardano-serialization-lib): add basic package and loader based on environment**

:pushpin: **feature(cardano-serialization-lib): add Ogmios to CardanoWasm translator**
The SDK uses Ogmios types as it's base, however easily converting to CardanoWasm instances is required for transaction construction and to convert between formats.

:pushpin: **ci: fixup copy/paste error in naming**


sweeetland (approved):
Nice 👍

PR #80, opened 2021-09-03
Refactor/blockfrost provider

rhyslbw:
# Context
While looking at the option of converting the provider types to the `cardano-serialization-lib` objects  I noticed the utils didn't contain all mappings, as the simple translations were inline in the provider. It would be best to keep these all in one file for clarity. Also, we're depending on the `@cardano-ogmios/client` package when the `schema` is suffice.

# Proposed Solution
:pushpin: **refactor(blockfrost): slight rework of the Ogmios translations**
- existing tests cover this object
- change to `@cardano-ogmios/schema`

:pushpin: **refactor(core): slight rework of the module structure**
- change to `@cardano-ogmios/schema`


sweeetland (approved):
👍

PR #81, opened 2021-09-03
Refactor/window injection

rhyslbw:
# Context
_See commit messages_

# Proposed Solution
:pushpin: **refactor(cip30): lift window injection behaviour out of class**
In hindsight my direction to do this inside the class (and particularly
the constructor) was wrong, when we can simply use a function and keep
the concerns of modifying window state separate. I suspect we'll end
up refactoring this further once the allow list is improved, but for
now this is an isolated change that includes more testing.

:pushpin: **refactor(cip30): pick properties from the Wallet for WalletPublic**

:pushpin: **docs(cip30): add short intro to package README**

### DIFF
--- a/packages/cip30/src/Wallet/Wallet.ts
+++ b/packages/cip30/src/Wallet/Wallet.ts
@@ -4,6 +4,7 @@ import { WalletApi } from './WalletApi';
 import { ApiError, APIErrorCode } from '../errors';
 import { dummyLogger, Logger } from 'ts-log';
 import { WalletPublic } from './WalletPublic';
+import { WindowMaybeWithCardano } from '../injectWindow';
 
 /**
  * CIP30 Specification version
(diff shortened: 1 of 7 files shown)


packages/cip30/src/Wallet/Wallet.ts
```diff
@@ -81,47 +53,19 @@ export class Wallet {
...
-        {
-          module: 'Wallet',
-          walletName: this.name
-        },
-        'Cardano global scope exists'
-      );
-    }
-
-    const walletPublic: WalletPublic = {
+  public getPublicApi(window: WindowMaybeWithCardano): WalletPublic {
```
rhyslbw, line 58: New method

packages/cip30/src/Wallet/Wallet.ts
```diff
@@ -130,7 +74,7 @@ export class Wallet {
     if (this.options.persistAllowList) {
       const currentList = this.getAllowList();
       // Todo: Encrypt
-      this.options.storage?.setItem(window.location.hostname, JSON.stringify([...currentList, appName]));
+      this.options.storage?.setItem(this.name, JSON.stringify([...currentList, appName]));
```
rhyslbw, line 77: This is a bug right?

packages/cip30/src/Wallet/Wallet.ts
```diff
@@ -81,47 +53,19 @@ export class Wallet {
...
-        globalCardanoScope: this.window.cardano,
-        allowList: this.allowList
-      },
-      'Constructed'
-    );
   }
 
   private getAllowList(): string[] {
-    return JSON.parse(this.options.storage?.getItem(window.location.hostname)) || [];
+    return JSON.parse(this.options.storage?.getItem(this.name)) || [];
```
rhyslbw, line 68: ... also bug here?

packages/cip30/src/Wallet/Wallet.ts
sweeetland: I think so

sweeetland (approved):

PR #82, opened 2021-09-06
feat(cip30): create Messaging interface

sweeetland:
# Context

For the [Dapp Connector CIP-30](https://github.com/cardano-foundation/CIPs/blob/b473884c9d208ed085414fb51fb339d09f4443e2/CIP-DappConnector/CIP-dapp-connector.md) we will need a way to communicate between a webpage and the browser extension.

# Proposed Solution

Create a Messaging interface to describe the API to communicate between the webpage and browser extension.

### DIFF
--- a/packages/cip30/src/BrowserExtension/handleMessages.ts
+++ b/packages/cip30/src/BrowserExtension/handleMessages.ts
@@ -1,0 +1,21 @@
+import { dummyLogger, Logger } from 'ts-log';
+import browser from 'webextension-polyfill';
+import { Message } from './types';
+import { WalletApi } from '../Wallet';
+
+export const handleMessages = (walletApi: WalletApi, logger: Logger = dummyLogger): void => {
+  browser.runtime.onMessage.addListener(async (msg: Message) => {
+    logger.debug('new message received: ', msg);
+
+    const walletMethod = walletApi[msg.method];
+
+    if (!walletMethod) {
+      logger.error(`No method implemented for ${msg.method}`);
+
+      return;
+    }
+
+    // eslint-disable-next-line consistent-return
+    return walletMethod(...msg.arguments);
+  });
+};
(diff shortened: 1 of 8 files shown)


packages/cip30/src/BrowserExtension/handleMessages.ts
```diff
@@ -0,0 +1,14 @@
+import browser from 'webextension-polyfill';
+import { WalletApi } from '..';
+import { Message } from './Message';
+
+export const handleMessages = (walletApi: WalletApi): void => {
+  browser.runtime.onMessage.addListener(async (msg: Message) => {
+    const walletMethod = walletApi[msg.method];
+
+    if (!walletMethod) return; // todo: what to do if method does not exist on WalletApi?
+
```
rhyslbw, line 10: Let's add logging here

packages/cip30/src/BrowserExtension/handleMessages.ts
```diff
@@ -0,0 +1,14 @@
+import browser from 'webextension-polyfill';
+import { WalletApi } from '..';
+import { Message } from './Message';
+
+export const handleMessages = (walletApi: WalletApi): void => {
+  browser.runtime.onMessage.addListener(async (msg: Message) => {
+    const walletMethod = walletApi[msg.method];
+
+    if (!walletMethod) return; // todo: what to do if method does not exist on WalletApi?
```
rhyslbw, line 9: Silently ignoring is fine IMO

rhyslbw, line 9: Silently ignoring is fine IMO

packages/cip30/src/BrowserExtension/uiWallet.ts
```diff
@@ -0,0 +1,80 @@
...
+  SubmitTx,
+  WalletApi
+} from '..';
+import browser from 'webextension-polyfill';
+import { Message } from './Message';
+
+const getUtxos: GetUtxos = (...args) => {
+  const msg: Message = { method: 'getUtxos', arguments: args };
+
+  return browser.runtime.sendMessage(msg);
```
rhyslbw, line 19: Let's add a `sendMessage` function so we can include a log line on the send side

rhyslbw (changes requested):

rhyslbw:
@sweeetland I've pushed 3 commits for review that should be squashed prior to merging.

rhyslbw (approved):

PR #88, opened 2021-09-14
chore(cardano-serialization-lib): dynamic import for @emurgo/cardano-serialization-lib-nodejs

mkazlauskas:
# Context

`package.json` of `@emurgo/cardano-serialization-lib-nodejs` doesn't have a `browser` field, which means it will be bundled for browser builds when used with webpack.

# Proposed Solution

Dynamic import

Re-export OgmiosToCardanoWasm from cardano-serialization-lib package main

### DIFF
--- a/packages/cardano-serialization-lib/src/loadCardanoSerializationLib.ts
+++ b/packages/cardano-serialization-lib/src/loadCardanoSerializationLib.ts
@@ -1,5 +1,3 @@
-import CardanoSerializationLibNodeJs from '@emurgo/cardano-serialization-lib-nodejs';
-
 export const isNodeJs = (): boolean => {
   try {
     return !!process;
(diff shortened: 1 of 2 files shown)


packages/cardano-serialization-lib/src/loadCardanoSerializationLib.ts
```diff
@@ -8,12 +6,11 @@ export const isNodeJs = (): boolean => {
...
 
-export type CardanoSerializationLib = typeof CardanoSerializationLibNodeJs;
+export type CardanoSerializationLib = typeof import('@emurgo/cardano-serialization-lib-nodejs');
 
 /**
- * Loads the environment-specific library.
- * The type of each complete library is the same, so one is statically imported for the return type.
- * Dynamically loads the browser library.
+ * Dynamically loads the environment-specific library.
+ * The type of each complete library is the same.
```
rhyslbw, line 13: This comment might be misleading, since we discovered there is a difference which could be a bug that we need to follow-up. Let's document the situation here

rhyslbw (changes requested):

packages/cardano-serialization-lib/src/loadCardanoSerializationLib.ts
mkazlauskas: I'll need to try to reproduce the original issue again. I've just tried to reproduce with a minimal test case and TypeScript seems to be happy about that:

```ts
type CardanoSerializationLib = typeof import('@emurgo/cardano-serialization-lib-nodejs');
type CardanoSerializationLibBrowser = typeof import('@emurgo/cardano-serialization-lib-browser');

let libNodejs: CardanoSerializationLib;
let libBrowser: CardanoSerializationLibBrowser;

const doSthWithLibNodejs = (lib: CardanoSerializationLib): void => {}
const doSthWithLibBrowser = (lib: CardanoSerializationLibBrowser): void => {}

doSthWithLibNodejs(libBrowser);
doSthWithLibBrowser(libNodejs);
```

rhyslbw: Great!

rhyslbw (approved):

PR #90, opened 2021-09-20
feature: add maxTxSize to `ProtocolParametersRequiredByWallet`

rhyslbw:
# Context
This parameter is required for input selection

# Proposed Solution
Adds the property to the narrowed type and it's implementations.


mkazlauskas (approved):
Changes looks good. However since we are no longer creating the transaction during input selection (leaving it up to the caller), we should probably leave transaction size validation up to the caller as well. Either way this parameter is still needed.

PR #95, opened 2021-09-24
Feature/ UTxO repository and single-address wallet

rhyslbw:
# Context
We need to provide a single-address wallet that utilises the `CardanoProvider` to access information and submit transactions to the network.

# Proposed Solution
- A UTxO repository interface and in-memory implementation
- Helper function to create transaction internals (currently very limited to achieve simple payment transactions)
- Extend the key manager interface to include required behaviour and properties.
- This implementation uses hard-coded return values for the selection constraints,
to be replaced by a later scope of work.

- This PR also cleans up the `cardano-serialization-lib` dependencies, so that only `@cardano-sdk/cardano-serialization-lib` depends on the Emurgo package, and makes the `InMemoryKeyManager` isomorphic.

### DIFF
--- a/packages/core/src/Ogmios/OgmiosToCardanoWasm.ts
+++ b/packages/core/src/Ogmios/OgmiosToCardanoWasm.ts
@@ -19,7 +19,7 @@ export const OgmiosToCardanoWasm = {
     ),
   value: (ogmios: OgmiosSchema.Value): CardanoWasm.Value => {
     const value = CardanoWasm.Value.new(CardanoWasm.BigNum.from_str(ogmios.coins.toString()));
-    const assets = Object.entries(ogmios.assets);
+    const assets = ogmios.assets !== undefined ? Object.entries(ogmios.assets) : [];
     if (assets.length === 0) {
       return value;
     }
--- a/packages/core/src/Ogmios/OgmiosToCardanoWasm.ts
+++ b/packages/core/src/Ogmios/OgmiosToCardanoWasm.ts
@@ -1,40 +1,34 @@
-import CardanoWasm from '@emurgo/cardano-serialization-lib-nodejs';
+import { CSL } from '@cardano-sdk/cardano-serialization-lib';
 import OgmiosSchema from '@cardano-ogmios/schema';
 import * as Asset from '../Asset';
 
 export const OgmiosToCardanoWasm = {
-  txIn: (ogmios: OgmiosSchema.TxIn): CardanoWasm.TransactionInput =>
-    CardanoWasm.TransactionInput.new(
-      CardanoWasm.TransactionHash.from_bytes(Buffer.from(ogmios.txId, 'hex')),
-      ogmios.index
-    ),
-  txOut: (ogmios: OgmiosSchema.TxOut): CardanoWasm.TransactionOutput =>
-    CardanoWasm.TransactionOutput.new(
-      CardanoWasm.Address.from_bech32(ogmios.address),
-      OgmiosToCardanoWasm.value(ogmios.value)
-    ),
-  utxo: (ogmios: OgmiosSchema.Utxo): CardanoWasm.TransactionUnspentOutput[] =>
+  txIn: (ogmios: OgmiosSchema.TxIn): CSL.TransactionInput =>
+    CSL.TransactionInput.new(CSL.TransactionHash.from_bytes(Buffer.from(ogmios.txId, 'hex')), ogmios.index),
+  txOut: (ogmios: OgmiosSchema.TxOut): CSL.TransactionOutput =>
+    CSL.TransactionOutput.new(CSL.Address.from_bech32(ogmios.address), OgmiosToCardanoWasm.value(ogmios.value)),
+  utxo: (ogmios: OgmiosSchema.Utxo): CSL.TransactionUnspentOutput[] =>
     ogmios.map((item) =>
-      CardanoWasm.TransactionUnspentOutput.new(OgmiosToCardanoWasm.txIn(item[0]), OgmiosToCardanoWasm.txOut(item[1]))
+      CSL.TransactionUnspentOutput.new(OgmiosToCardanoWasm.txIn(item[0]), OgmiosToCardanoWasm.txOut(item[1]))
     ),
-  value: (ogmios: OgmiosSchema.Value): CardanoWasm.Value => {
-    const value = CardanoWasm.Value.new(CardanoWasm.BigNum.from_str(ogmios.coins.toString()));
+  value: (ogmios: OgmiosSchema.Value): CSL.Value => {
+    const value = CSL.Value.new(CSL.BigNum.from_str(ogmios.coins.toString()));
     const assets = ogmios.assets !== undefined ? Object.entries(ogmios.assets) : [];
     if (assets.length === 0) {
       return value;
     }
-    const multiAsset = CardanoWasm.MultiAsset.new();
+    const multiAsset = CSL.MultiAsset.new();
     const policies = [...new Set(assets.map(([assetId]) => Asset.util.policyIdFromAssetId(assetId)))];
     for (const policy of policies) {
       const policyAssets = assets.filter(([assetId]) => Asset.util.policyIdFromAssetId(assetId) === policy);
-      const wasmAssets = CardanoWasm.Assets.new();
+      const wasmAssets = CSL.Assets.new();
       for (const [assetId, assetQuantity] of policyAssets) {
         wasmAssets.insert(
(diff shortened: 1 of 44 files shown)


packages/core/src/Ogmios/OgmiosToCardanoWasm.ts
```diff
@@ -1,40 +1,34 @@
-import CardanoWasm from '@emurgo/cardano-serialization-lib-nodejs';
+import { CSL } from '@cardano-sdk/cardano-serialization-lib';
 import OgmiosSchema from '@cardano-ogmios/schema';
 import * as Asset from '../Asset';
 
 export const OgmiosToCardanoWasm = {
```
mkazlauskas, line 5: Should inject `CardanoSerializationLib` because we want to use the same wasm object everywhere. Not sure if that's in scope, but if it is might as well rename this to `OgmiosToCSL` at the same time because the interface will change

packages/core/src/Transaction/WithHash.ts
```diff
@@ -0,0 +1,3 @@
+import OgmiosSchema from '@cardano-ogmios/schema';
+
+export type WithHash = { hash: OgmiosSchema.Hash16 } & OgmiosSchema.Tx;
```
mkazlauskas, line 3: I used different style for exporting types for a module in `cip2`, using one `types.ts` instead of file per type. We should use one style of the other for consistency, I don't mind either. Let me know if you want me to refactor `cip2`.

packages/in-memory-key-manager/test/InMemoryKeyManager.test.ts
```diff
@@ -1,33 +1,41 @@
...
+    csl = await loadCardanoSerializationLib();
     const mnemonic = util.generateMnemonic();
     keyManager = createInMemoryKeyManager({
+      csl,
       mnemonic,
+      networkId: Cardano.NetworkId.testnet,
       password: '123'
     });
-    expect(keyManager.publicKey).toBeInstanceOf(CardanoSerializationLib.PublicKey);
+    expect(keyManager.publicKey).toBeInstanceOf(CSL.PublicKey);
```
mkazlauskas, line 20: This passes because both `csl` and `CSL` are the same in nodejs environment, as CSL is exported from nodejs version. However a better test would be  `.toBeInstanceOf(csl.PublicKey)`

packages/wallet/src/InMemoryUtxoRepository.ts
```diff
@@ -0,0 +1,75 @@
...
+  public async selectInputs(
+    outputs: CSL.TransactionOutputs,
+    constraints: SelectionConstraints
+  ): Promise<SelectionResult> {
+    if (this.#utxoSet.size === 0) {
+      this.#logger.debug('Local UTxO set is empty. Syncing...');
+      await this.sync();
+    }
+    return this.#inputSelector.select({
+      utxo: Ogmios.OgmiosToCardanoWasm.utxo([...this.#utxoSet.values()]),
```
mkazlauskas, line 58: I'd reuse the existing method: `utxo: Ogmios.OgmiosToCardanoWasm.utxo(this.allUtxos()),`

packages/wallet/test/util.ts
```diff
@@ -0,0 +1,8 @@
+import { SelectionConstraints } from '@cardano-sdk/cip2';
+
+export const NO_CONSTRAINTS: SelectionConstraints = {
```
mkazlauskas, line 3: These are technically not "no constraints". There is a [similar object](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cip2/test/RoundRobinRandomImprove.test.ts#L17) in `cip2` tests, should probably flag it to move to `dev-utils` package when we create it.

packages/wallet/test/jest.config.js
```diff
@@ -2,6 +2,7 @@ const { pathsToModuleNameMapper } = require('ts-jest/utils')
 const { compilerOptions } = require('./tsconfig')
```
mkazlauskas, line 3: We could have a shared config files directory in monorepo root and either use those files directly (if no changes needed and it doesn't need to be under package root), or extending those files. Similarly to how tsconfig works right now.

packages/core/src/Transaction/WithHash.ts
rhyslbw: I'll conform to what we already have

packages/core/src/Ogmios/OgmiosToCardanoWasm.ts
rhyslbw: Yes, I also just picked this up. Will fix and rename

packages/wallet/test/util.ts
rhyslbw: Right, I just kept the same name to make a refactor easy

packages/core/src/Transaction/WithHash.ts
rhyslbw: Actually, will defer to another PR as there's others to align

packages/wallet/test/jest.config.js
rhyslbw: Will also defer this to another PR since it touches the `cip2` package

packages/wallet/test/util.ts
rhyslbw: Done in c33edc69a400bc39617f62ee2d810f7b0e54f5c2

packages/wallet/src/InMemoryUtxoRepository.ts
rhyslbw: Done in 3fd77aac4b2ea57f54fe8177e430739610b89e59

packages/in-memory-key-manager/test/InMemoryKeyManager.test.ts
rhyslbw: Done in https://github.com/input-output-hk/cardano-js-sdk/pull/95/commits/12f0171f8cc9dcd73a2ed9426bd92e65887ea538

packages/core/src/Ogmios/OgmiosToCardanoWasm.ts
rhyslbw: Done in e9a242170cdca968c39f281b254edd2bef297b65

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -0,0 +1,100 @@
...
+              outputs,
+              inputs: utxo,
+              change,
+              fee: 0n
+            },
+            validityInterval
+          });
+          const witnessSet = await keyManager.signTransaction(transactionInternals.hash);
+          const tx = csl.Transaction.new(transactionInternals.body, witnessSet);
+          return BigInt(
```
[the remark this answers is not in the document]
rhyslbw: Possibly, although this is one area of improvement we can offer. The use of strings is a workaround for target environments we're not supporting, so it would be great to shield our users.

packages/core/test/Ogmios/ogmiosToCsl.test.ts
```diff
@@ -0,0 +1,29 @@
...
+  value: { coins: 10, assets: { '2a286ad895d091f2b3d168a6091ad2627d30a72761a5bc36eef00740': 20n } }
+};
+
+describe('ogmiosToCsl', () => {
+  let csl: CardanoSerializationLib;
+  beforeAll(async () => {
+    csl = await loadCardanoSerializationLib();
+  });
+  test('txIn', () => {
+    expect(ogmiosToCsl(csl).txIn(txIn)).toBeInstanceOf(CSL.TransactionInput);
```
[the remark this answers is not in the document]
rhyslbw: Done in d98c896dad0d2aeebd307a71d44360d5658932e9

mkazlauskas (dismissed):

packages/wallet/src/InMemoryUtxoRepository.ts
```diff
@@ -0,0 +1,83 @@
...
+    csl: CardanoSerializationLib,
+    provider: CardanoProvider,
+    keyManager: KeyManager,
+    inputSelector: InputSelector,
+    logger?: Logger
+  ) {
+    this.#csl = csl;
+    this.#logger = logger ?? dummyLogger;
+    this.#provider = provider;
+    this.#utxoSet = new Set();
```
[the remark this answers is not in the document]
rhyslbw: Captured in a Jira ticket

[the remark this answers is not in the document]
rhyslbw: Captured in a Jira ticket

PR #97, opened 2021-09-27
docs: generate code coverage reports

mkazlauskas:
# Context

We aim to have high code coverage. Generating coverage reports on each build would be nice.

# Proposed Solution

Generate and upload coverage reports to gh-pages together with docs.


rhyslbw (approved):
Great

PR #99, opened 2021-09-28
Feature: CardanoProvider network info and stake pool stats

rhyslbw:
# Context
The light wallet requires some network and stake pool information.

# Proposed Solution
Adds two new CardanoProvider properties: `networkInfo` and `stakePoolStats`. UTC dates are localised.

### DIFF
--- a/packages/blockfrost/test/blockfrostProvider.test.ts
+++ b/packages/blockfrost/test/blockfrostProvider.test.ts
@@ -3,7 +3,7 @@
 import { BlockFrostAPI, Responses } from '@blockfrost/blockfrost-js';
 import { blockfrostProvider } from '../src';
 import { Schema as Cardano } from '@cardano-ogmios/client';
-import { NetworkInfo, Transaction } from '@cardano-sdk/core';
+import { NetworkInfo, StakePoolStats, Transaction } from '@cardano-sdk/core';
 jest.mock('@blockfrost/blockfrost-js');
 
 describe('blockfrostProvider', () => {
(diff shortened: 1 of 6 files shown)


packages/blockfrost/test/blockfrostProvider.test.ts
```diff
@@ -3,12 +3,107 @@
...
+      },
+      supply: {
+        max: '45000000000000000',
+        total: '40267211394073980',
+        circulating: '42064399450423723',
+        locked: '6161981104458'
+      }
+    } as Responses['network'];
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
mkazlauskas, line 39: Probably don't need to disable this rule here

packages/blockfrost/test/blockfrostProvider.test.ts
```diff
@@ -340,7 +435,7 @@ describe('blockfrostProvider', () => {
       confirmations: 0
     } as Responses['block_content'];
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
mkazlauskas, line 437: Probably don't need to disable this rule here

packages/cardano-graphql-db-sync/src/cardanoGraphqlDbSyncProvider.ts
```diff
@@ -42,6 +42,170 @@ export const cardanoGraphqlDbSyncProvider = (uri: string): CardanoProvider => {
...
+          currentEpoch {
+            lastBlockTime
+            number
+            startedAt
+          }
+        }
+      }
+    `;
+
+    type Response = {
```
mkazlauskas, line 72: I'm wondering if there's a way to generate response type from graphql schema+query

rhyslbw: We can use https://www.graphql-code-generator.com/, and it's probably at the point where we can consider introducing it.

mkazlauskas (approved):

PR #100, opened 2021-09-28
feat(cip2): add functions to create selection constraints

mkazlauskas:
# Context

Since the CIP2 package is dependent on the CSL, we may as well include functions to cover the selection constraints:

- computeMinimumCost
- tokenBundleSizeExceedsLimit
- computeMinimumCoinQuantity
- computeSelectionLimit

# Proposed Solution

Implement default selection constraints and export them from cip2 package. Use as defaults in coin selection if applicable.

### DIFF
--- a/packages/cip2/src/selectionConstraints.ts
+++ b/packages/cip2/src/selectionConstraints.ts
@@ -56,8 +56,6 @@ export const tokenBundleSizeExceedsLimit =
     return value.to_bytes().length > maxValueSize;
   };
 
-// TODO: move this to core package and test it by comparing
-// the result to result of serializing the transaction via cardano-cli
 const getTxSize = (tx: CSL.Transaction) => tx.to_bytes().length;
 
 /**
--- a/packages/cip2/src/selectionConstraints.ts
+++ b/packages/cip2/src/selectionConstraints.ts
@@ -1,12 +1,12 @@
 import { CardanoSerializationLib, CSL, ProtocolParametersRequiredByWallet } from '@cardano-sdk/core';
 import { ComputeSelectionLimit, SelectionConstraints, TokenBundleSizeExceedsLimit } from '.';
 import { ComputeMinimumCoinQuantity, EstimateTxFee, SelectionSkeleton } from './types';
-import { MAX_U64 } from './util';
+import { maxBigNum } from './util';
 
 export type BuildTx = (selection: SelectionSkeleton) => Promise<CSL.Transaction>;
 export interface DefaultSelectionConstraintsProps {
   csl: CardanoSerializationLib;
-  protocolParams: ProtocolParametersRequiredByWallet;
+  protocolParameters: ProtocolParametersRequiredByWallet;
   buildTx: BuildTx;
 }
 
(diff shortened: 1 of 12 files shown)


packages/cip2/src/types.ts
```diff
@@ -53,7 +54,7 @@ export type TokenBundleSizeExceedsLimit = (tokenBundle: CSL.MultiAsset) => boole
 /**
  * @returns minimum lovelace amount in a UTxO
  */
-export type ComputeMinimumCoinQuantity = (assetQuantities: CSL.MultiAsset) => bigint;
+export type ComputeMinimumCoinQuantity = (assetQuantities?: CSL.MultiAsset) => bigint;
```
rhyslbw, line 57: Good catch

packages/cip2/src/selectionConstraints.ts
```diff
@@ -0,0 +1,51 @@
...
+  };
+
+//
+export const computeMinimumCoinQuantity =
+  (
+    csl: CardanoSerializationLib,
+    { coinsPerUtxoWord }: ProtocolParametersRequiredByWallet
+  ): ComputeMinimumCoinQuantity =>
+  (multiasset) => {
+    const minUTxOValue = CSL.BigNum.from_str((BigInt(coinsPerUtxoWord) * 29n).toString());
```
rhyslbw, line 34:
```suggestion
    const minUTxOValue = CSL.BigNum.from_str((coinsPerUtxoWord * 29).toString());
```
Does this need to be a `bigint`?

rhyslbw (commented):
Looks good so far :+1:

rhyslbw (changes requested):
Need to also export the constraints from the package index:
```
export * as selectionConstraints from './selectionConstraints';
```

packages/cip2/src/selectionConstraints.ts
```diff
@@ -0,0 +1,90 @@
...
+    );
+  };
+
+export const computeMinimumCoinQuantity =
+  (
+    csl: CardanoSerializationLib,
+    { coinsPerUtxoWord }: ProtocolParametersRequiredByWallet
+  ): ComputeMinimumCoinQuantity =>
+  (multiasset) => {
+    const minUTxOValue = CSL.BigNum.from_str((coinsPerUtxoWord * 29).toString());
```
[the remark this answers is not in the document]
mkazlauskas: This is actually now a tsc error after I rebased on master.  No more mistakes like that

mkazlauskas:
@rhyslbw I'm finished with this for the time being. Please review and let me know if you want me to do any changes.

I've changed my mind about needing serialization/size tests right now. It's probably out of scope for this repo and we should be trusting CSL. Let me know if you think otherwise.

Also, do you think there's any new code in this PR worth moving to core package? I could think of a few pieces but I'm not sure they'd ever be used. I'm thinking to refactor as needed if we start duplicating code, like with `ogmiosToCsl.value()`.

rhyslbw (approved):
Great work @mkazlauskas 

> I've changed my mind about needing serialization/size tests right now. It's probably out of scope for this repo and we should be trusting CSL. Let me know if you think otherwise

Agree. These were always going to be sanity checks, but this can also be done as once-off manual QA.

> Also, do you think there's any new code in this PR worth moving to core package? I could think of a few pieces but I'm not sure they'd ever be used. I'm thinking to refactor as needed if we start duplicating code, like with ogmiosToCsl.value().

Nothing jumps out, so let's do it on-demand.

PR #102, opened 2021-09-29
Refactor/consolidate in-memory-key-manager with wallet

rhyslbw:
# Context
There's no practical benefit of keeping the `InMemoryKeyManager` separate from the wallet, in the same way
a particular hardware wallet would have, as most wallets will use the in-memory key manager
as the default.

# Proposed Solution
- Consolidate the `in-memory-key-manager` package with `wallet`.
- Restructures the tests to reflect the source structure, removing source path mapping.
- https://github.com/input-output-hk/cardano-js-sdk/pull/102/commits/955079143cbd564e3e1452b39a6a73c371f16bd2


mkazlauskas (approved):
Looks good

rhyslbw:
> Looks good. The tests are probably failing because it's missing a dependency: `"bip39": "^3.0.4"`

Yep, missed the failure since my local state had the package installed.

rhyslbw:
Additional important change listed above @mkazlauskas

mkazlauskas (dismissed):

mkazlauskas:
> Additional important change listed above @mkazlauskas

Is CSL not following semantic versioning? Either way I think it's ok to pin to specific CSL version because cardano-js-sdk users are supposed to get CSL indirectly anyway.

rhyslbw:
> Is CSL not following semantic versioning? Either way I think it's ok to pin to specific CSL version because cardano-js-sdk users are supposed to get CSL indirectly anyway.

I'm not sure, but given the critical nature of this dep, I'd prefer we lock it and remove the chance of unexpected changes.

mkazlauskas (approved):

PR #103, opened 2021-09-30
fix(cip2): exclude fee from change bundles

mkazlauskas:
# Context

CIP2 spec didn't contain information about how to return min fee and we made a mistake of including fee quantity in change bundles. This is not the desired behavior and quantities in change bundles are used as transaction outputs to the return address.

# Proposed Solution

After computing initial change bundles, deduct the estimated min fee and re-balance+re-validate change. There might be an exception when change becomes less than min utxo value at this point, in which case need to select an extra utxo and re-estimate the fee.

### DIFF
--- a/packages/cip2/src/RoundRobinRandomImprove/change.ts
+++ b/packages/cip2/src/RoundRobinRandomImprove/change.ts
@@ -5,11 +5,11 @@ import { InputSelectionError, InputSelectionFailure } from '../InputSelectionErr
 import { TokenMap, OgmiosValue, ogmiosValueToCslValue } from '../util';
 import {
   assetQuantitySelector,
-  assetTotalsQuantitySelector,
+  assetWithValueQuantitySelector,
   getCoinQuantity,
-  getCoinTotalsQuantity,
+  getWithValuesCoinQuantity,
   UtxoSelection,
-  UtxoWithTotals
+  UtxoWithValue
 } from './util';
 
 type EstimateTxFeeWithOriginalOutputs = (utxo: CSL.TransactionUnspentOutput[], change: CSL.Value[]) => Promise<bigint>;
@@ -31,10 +31,10 @@ interface ChangeComputationResult {
   fee: bigint;
 }
 
-const getLeftoverAssets = (utxoSelected: UtxoWithTotals[], uniqueOutputAssetIDs: string[]) => {
+const getLeftoverAssets = (utxoSelected: UtxoWithValue[], uniqueOutputAssetIDs: string[]) => {
   const leftovers: Record<string, Array<bigint>> = {};
   for (const {
-    totals: { assets }
+    value: { assets }
   } of utxoSelected) {
     if (assets) {
       const leftoverAssetKeys = Object.keys(assets).filter((id) => !uniqueOutputAssetIDs.includes(id));
@@ -54,7 +54,7 @@ const getLeftoverAssets = (utxoSelected: UtxoWithTotals[], uniqueOutputAssetIDs:
  *   coalesce the smallest quantities together.
  */
 const redistributeLeftoverAssets = (
-  utxoSelected: UtxoWithTotals[],
+  utxoSelected: UtxoWithValue[],
   requestedAssetChangeBundles: OgmiosValue[],
   uniqueOutputAssetIDs: string[]
 ) => {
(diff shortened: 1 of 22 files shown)


packages/cip2/src/RoundRobinRandomImprove/change.ts
```diff
@@ -3,14 +3,21 @@ import { orderBy } from 'lodash-es';
...
+  UtxoWithTotals
+} from './util';
 
 type EstimateTxFeeWithOriginalOutputs = (utxo: CSL.TransactionUnspentOutput[], change: CSL.Value[]) => Promise<bigint>;
 
 interface ChangeComputationArgs {
   csl: CardanoSerializationLib;
   utxoSelection: UtxoSelection;
-  outputsWithTotals: OutputWithTotals[];
+  outputsQuantities: ValueQuantities[];
```
rhyslbw, line 20:
```suggestion
  outputValues: Value[];
```

rhyslbw: We should stay close to the Ogmios type name, as we will soon be implementing it directly:
```
export interface Value {
    coins: Lovelace;
    assets?: {
        [k: string]: AssetQuantity;
    };
}
```

packages/cip2/src/RoundRobinRandomImprove/change.ts
```diff
@@ -113,23 +123,23 @@ const createBundlePerOutput = (
  */
 const computeRequestedAssetChangeBundles = (
   utxoSelected: UtxoWithTotals[],
-  outputsWithTotals: OutputWithTotals[],
-  uniqueOutputAssetIDs: string[]
+  outputsQuantities: ValueQuantities[],
```
rhyslbw, line 126:
```suggestion
  outputValues: Value[],
```

packages/cip2/src/RoundRobinRandomImprove/change.ts
```diff
@@ -234,9 +262,31 @@ const computeChangeBundles = (
   }
   return { changeBundles, ...utxoSelection };
 };
+
+const changeBundlesToValues = (
+  csl: CardanoSerializationLib,
+  changeBundles: ValueQuantities[],
+  tokenBundleSizeExceedsLimit: TokenBundleSizeExceedsLimit
+) => {
+  const values = changeBundles.map((bundle) => valueQuantitiesToValue(bundle, csl));
```
rhyslbw, line 271: I think renaming this so we know it' explicit behaviour prior to the removal when the Ogmios type is updated. 
```suggestion
  const values = changeBundles.map((bundle) => ogmiosValueToCslValue(bundle, csl));
```

packages/cip2/src/RoundRobinRandomImprove/change.ts
```diff
@@ -234,9 +262,31 @@ const computeChangeBundles = (
...
+      throw new InputSelectionError(InputSelectionFailure.UtxoFullyDepleted);
+    }
+  }
+  return values;
+};
+
 /**
- * 1. Compute change bundles and coalesce them to cover min ADA requirement. Select additional UTxO if needed.
- * 2. Compute fee and verify that change coin covers it. If not, select an additional UTxO and retry.
+ * 1. Compute change bundles with fee included and coalesce them to cover min ADA requirement.
```
rhyslbw, line 284: :+1:

packages/cip2/src/RoundRobinRandomImprove/change.ts
```diff
@@ -245,53 +295,57 @@ export const computeChangeAndAdjustForFee = async ({
...
-    const multiasset = value.multiasset();
-    if (!multiasset) continue;
-    if (tokenBundleSizeExceedsLimit(multiasset)) {
-      // Algorithm could be improved to attempt to rebalance the bundles
-      throw new InputSelectionError(InputSelectionFailure.UtxoFullyDepleted);
-    }
-  }
+  });
 
+  // Calculate fee with change outputs that include fee.
```
rhyslbw, line 310: :+1:

rhyslbw (changes requested):

packages/cip2/src/RoundRobinRandomImprove/change.ts
mkazlauskas: I renamed the type we're using to `OgmiosValue`. When we implement this type directly I suggest we leave `OgmiosValue` as re-export for internal use, because having 2 types with same name is confusing (`CSL.Value` is the other one).

rhyslbw (approved):
Great work @mkazlauskas

PR #106, opened 2021-10-01
chore: config improvements

mkazlauskas:
# Primary changes

- [x] enable 'strict' TypeScript option
- [x] hoist `jest.config` to workspace root
- [x] align code structure to use `types.ts` where appropriate
- [x] remove duplicated `ogmiosToCsl.value`
- [x] refactor cip2 to use Set instead of arrays 
- [x] move some of the `cip2` test utils to a new `util-dev` package 

## Other changes

- Disable a few eslint rules
- Refactor cip30 `Wallet.ts` constructor and `WalletPublic` type export
- Update CSL to 9.1.0
- Add util export to core/CSL for max u64 value


packages/blockfrost/src/BlockfrostToOgmios.ts
```diff
@@ -19,9 +19,10 @@ export const BlockfrostToOgmios = {
   // without `as OgmiosSchema.Utxo` above TS thinks the return value is (OgmiosSchema.TxIn | OgmiosSchema.TxOut)[][]
 
   blockToTip: (block: Responses['block_content']): OgmiosSchema.Tip => ({
-    blockNo: block.height,
+    // Review: if height and slot be actually be null we should check it and either throw or return something else
```
rhyslbw, line 22: `slot` and `height` are indeed nullable in Ouroboros Classic (a legacy protocol that we have to consider in a uniform model) since there were special epoch boundary blocks, but these would never be returned for a tip query. We can simply throw this as an exception

packages/cardano-graphql-db-sync/src/CardanoGraphqlToOgmios.ts
```diff
@@ -56,8 +56,9 @@ export const CardanoGraphqlToOgmios = {
   }),
 
   tip: (tip: CardanoGraphQlTip): Cardano.Tip => ({
-    blockNo: tip.number,
+    // Review: if number and slot can actually be undefined we should check it and either throw or return something else
```
rhyslbw, line 59: ditto

packages/cip2/src/selectionConstraints.ts
```diff
@@ -4,16 +4,28 @@ import { ComputeMinimumCoinQuantity, EstimateTxFee, SelectionSkeleton } from './
 import { maxBigNum } from './util';
 
 export type BuildTx = (selection: SelectionSkeleton) => Promise<CSL.Transaction>;
+
+// Review: we might want to use this for ProtocolParametersRequiredByWallet instead
```
rhyslbw: Something like ?

``` TypeScript
export type ProtocolParametersRequiredByWallet = ProtocolParametersRequiredBySelectionConstraints && NonNullableFields<
  Pick<
    ProtocolParametersAlonzo,
    | 'stakeKeyDeposit'
    | 'minPoolCost'
    | 'poolDeposit'
    | 'protocolVersion'
  >
>;
```

packages/cip2/src/selectionConstraints.ts
```diff
@@ -32,10 +38,7 @@ export const computeMinimumCost =
   };
 
 export const computeMinimumCoinQuantity =
-  (
-    csl: CardanoSerializationLib,
-    { coinsPerUtxoWord }: ProtocolParametersRequiredByWallet
-  ): ComputeMinimumCoinQuantity =>
+  (csl: CardanoSerializationLib, coinsPerUtxoWord: number): ComputeMinimumCoinQuantity =>
```
rhyslbw, line 41:
```suggestion
  (csl: CardanoSerializationLib, coinsPerUtxoWord: ProtocolParametersRequiredByWallet['coinsPerUtxoWord']): ComputeMinimumCoinQuantity =>
```
It's more verbose but is less brittle and improves documentation

packages/cip2/src/selectionConstraints.ts
```diff
@@ -46,12 +49,12 @@ export const computeMinimumCoinQuantity =
   };
 
 export const tokenBundleSizeExceedsLimit =
-  (csl: CardanoSerializationLib, { maxValueSize }: ProtocolParametersRequiredByWallet): TokenBundleSizeExceedsLimit =>
+  (csl: CardanoSerializationLib, maxValueSize: number): TokenBundleSizeExceedsLimit =>
```
rhyslbw, line 52:
```suggestion
  (csl: CardanoSerializationLib, maxValueSize: ProtocolParametersRequiredByWallet['maxValueSize']): TokenBundleSizeExceedsLimit =>
```

packages/cip2/src/selectionConstraints.ts
```diff
@@ -66,23 +69,30 @@ const getTxSize = (tx: CSL.Transaction) => tx.to_bytes().length;
  * @returns {ComputeSelectionLimit} constraint that returns txSize <= maxTxSize ? utxo[].length : utxo[].length+1
  */
 export const computeSelectionLimit =
-  ({ maxTxSize }: ProtocolParametersRequiredByWallet, buildTx: BuildTx): ComputeSelectionLimit =>
+  (maxTxSize: number, buildTx: BuildTx): ComputeSelectionLimit =>
```
rhyslbw, line 72:
```suggestion
  (maxTxSize: ProtocolParametersRequiredByWallet['maxTxSize'], buildTx: BuildTx): ComputeSelectionLimit =>
```

packages/golden-test-generator/src/AddressBalance/getOnChainAddressBalances.ts
```diff
@@ -8,11 +9,12 @@ import {
   isMaryBlock,
   Schema,
   ConnectionConfig,
-  createInteractionContext,
+  createInteractionContext
 } from '@cardano-ogmios/client';
 import { GeneratorMetadata } from '../Content';
 import { isByronStandardBlock } from '../util';
 import { applyValue } from './applyValue';
+import { Intersection } from '@cardano-ogmios/client/dist/ChainSync';
```
rhyslbw, line 17:
```suggestion
import { ChainSync } from '@cardano-ogmios/client';
```

packages/golden-test-generator/src/AddressBalance/getOnChainAddressBalances.ts
```diff
@@ -47,7 +50,8 @@ export const getOnChainAddressBalances = (
           compactGenesis: await StateQuery.genesisConfig(
             await createInteractionContext(reject, logger.info, { connection: options.ogmiosConnectionConfig })
           ),
-          intersection: undefined
+          // Review: this can't be undefined acccording to type
+          intersection: undefined as unknown as Intersection
```
rhyslbw, line 54:
```suggestion
          intersection: undefined as unknown as ChainSync.Intersection
```

packages/golden-test-generator/src/Block/getBlocks.ts
```diff
@@ -13,22 +14,22 @@ import {
 import { GeneratorMetadata } from '../Content';
 
 import { isByronStandardBlock } from '../util';
+import { Intersection } from '@cardano-ogmios/client/dist/ChainSync';
```
rhyslbw, line 17:
```suggestion
import { ChainSync } from '@cardano-ogmios/client';
```

packages/golden-test-generator/src/Block/getBlocks.ts
```diff
@@ -39,7 +40,8 @@ export const getBlocks = async (
           compactGenesis: await StateQuery.genesisConfig(
             await createInteractionContext(reject, logger.info, { connection: options.ogmiosConnectionConfig })
           ),
-          intersection: undefined
+          // Review: this can't be undefined acccording to type
+          intersection: undefined as unknown as Intersection
```
rhyslbw, line 44:
```suggestion
          intersection: undefined as unknown as ChainSync.Intersection
```

packages/golden-test-generator/src/AddressBalance/getOnChainAddressBalances.ts
```diff
@@ -47,7 +50,8 @@ export const getOnChainAddressBalances = (
           compactGenesis: await StateQuery.genesisConfig(
             await createInteractionContext(reject, logger.info, { connection: options.ogmiosConnectionConfig })
           ),
-          intersection: undefined
+          // Review: this can't be undefined acccording to type
```
rhyslbw, line 53: We could type the constant as a partial, then coerce it after the intersection is set?
``` TypeScript
const response: Partial<AddressBalancesResponse> = {
...
if (atBlocks.includes(currentBlock)) {
  response.balances[currentBlock] = { ...trackedAddressBalances };
    if (atBlocks[atBlocks.length - 1] === currentBlock) {
      draining = true;
      await syncClient.shutdown();
      return resolve(response as ChainSync.Intersection);
    }
  }
```

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -5,6 +5,7 @@ import { dummyLogger, Logger } from 'ts-log';
 import { defaultSelectionConstraints } from '@cardano-sdk/cip2';
 
 export type InitializeTxProps = {
+  // Review: should this be Set<> too?
```
rhyslbw, line 8: :+1: 
Yes, let's keep it consistent

rhyslbw (changes requested):

packages/cip2/src/selectionConstraints.ts
mkazlauskas: I changed this to use `ProtocolParametersRequiredByWallet` and throwing an exception in `defaultSelectionConstraints` if any param is null

packages/cip2/src/selectionConstraints.ts
mkazlauskas: So it would be better to check for null and throw in each constraint instead of in `defaultSelectionConstraints`?

packages/golden-test-generator/src/Block/getBlocks.ts
mkazlauskas: I don't like the double cast, the expected type is non-nullable so it's a potential bug when using this

packages/cip2/src/selectionConstraints.ts
rhyslbw: As we want this package to be standalone I think we should still make a smaller type for coin selection though.

packages/cip2/src/selectionConstraints.ts
rhyslbw: My suggestion here is just to use an alias for the function parameter type, so we know arguments being passed in are expected to be part of the `ProtocolParametersRequiredByWallet` type

packages/golden-test-generator/src/AddressBalance/getOnChainAddressBalances.ts
mkazlauskas: I tried that and it wasn't pretty. I think leaving the double cast is best there.

rhyslbw (approved):
Nice work

PR #108, opened 2021-10-02
Feature: wallet name

rhyslbw:
# Context
We're missing the wallet name property, and the `createSingleAddressWallet` function interface could be improved before simply adding another property.

# Proposed Solution
Restructure to group dependencies in an object, then add another as the first to carry wallet props.

### DIFF
--- a/packages/wallet/src/SingleAddressWallet.ts
+++ b/packages/wallet/src/SingleAddressWallet.ts
@@ -14,6 +14,7 @@ export type InitializeTxProps = {
 export interface SingleAddressWallet {
   address: Schema.Address;
   initializeTx: (props: InitializeTxProps) => Promise<TxInternals>;
+  name: string;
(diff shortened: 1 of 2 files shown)


mkazlauskas (approved):
Looks good, although I don't understand why we need the name in the sdk (as opposed to in the wallet implementation)

rhyslbw:
You raise a good point @mkazlauskas.Given it's a single-address wallet the address can be used to identify it, but with a descriptive name we can improve our logging output, and this property is certainly desirable as we know so to me it makes sense to model it here. Based on your feedback though I think it will be more appropriate to be an optional prop, as we can always fallback to the address to handle the aforementioned use-case.

mkazlauskas:
Alternatively if it's only needed for logging the wallet could pass the name to an instance of logger

PR #109, opened 2021-10-06
Feat/delegation

mkazlauskas:
# Context

A single transaction should support the following scenarios:

- Register a stake key and delegate to a pool 
- Deregister a stake key
- Change delegation from one pool to another

# Proposed Solution

- [x] Add an optional 'certificates' prop to `InitializeTxProps` and `CreateTxInternalsProps`
- [ ] Export new utility from `wallet` package that takes in KeyManager and provides functions to create Certificates that then can be used to initialize Tx 


packages/wallet/src/Certificate/factory.ts
```diff
@@ -0,0 +1,21 @@
+import { KeyManager } from '../KeyManagement';
+import { StakeKeyRegistration, StakeKeyDeregistration, StakeDelegation, Bech32Key } from './types';
+
+export const factory = (keyManager: KeyManager) => {
+  const delegator = keyManager.stakeKey.to_bech32();
+  return {
+    registerStakeKey: {
+      certType: 'STAKE_KEY_REGISTRATION',
+      delegator
```
rhyslbw, line 9:
```suggestion
      stakeKey
```

packages/wallet/src/Certificate/factory.ts
```diff
@@ -0,0 +1,21 @@
...
+    } as StakeKeyRegistration,
+    unregisterStakeKey: {
+      certType: 'STAKE_KEY_DEREGISTRATION',
+      delegator
```
rhyslbw, line 13:
```suggestion
      stakeKey
```

packages/wallet/src/Certificate/factory.ts
```diff
@@ -0,0 +1,21 @@
...
+    } as StakeKeyDeregistration,
+    delegate: (delegatee: Bech32Key): StakeDelegation => ({
+      certType: 'STAKE_DELEGATION',
+      delegator,
```
rhyslbw, line 17:
```suggestion
      stakeKey,
```
```suggestion
      delegator,
```

packages/wallet/src/createTransactionInternals.ts
```diff
@@ -36,6 +38,13 @@ export const createTransactionInternals = async (
   if (props.validityInterval.invalidBefore !== undefined) {
     body.set_validity_start_interval(props.validityInterval.invalidBefore);
   }
+  if (props.certificates?.length) {
```
rhyslbw, line 41: This actually impacts the minimum fee since it increases the tx size, so we need to think about how attaching certificates interoperates with input selection :thinking:

rhyslbw (changes requested):
We're missing the key deposit on registration, and reclaim on deregistration. The value is: `ProtocolParametersRequiredByWallet.stakeKeyDeposit`. This needs to be added prior to the input selection. It's in the csl codebase as `key_deposit`, however I'm not sure of the specifics when wishing to set this outside of the `tx_builder`.

packages/wallet/src/createTransactionInternals.ts
mkazlauskas: Input selection constraints takes in `buildTx` and we're using `createTransactionInternals` for it in SingleAddressWalet,  so it already takes this into account for the minimum fee

packages/wallet/src/Delegation/withdrawal.ts
```diff
@@ -0,0 +1,20 @@
...
+  quantity: CSL.BigNum;
+};
+
+export const withdrawal = (
+  csl: CardanoSerializationLib,
+  keyManager: KeyManager,
+  quantity: Lovelace,
+  network: number = Cardano.NetworkId.mainnet
+): Withdrawal => ({
+  // Review: are withdrawals supposed to go to stake address or
```
rhyslbw, line 16: - Funds cannot be sent by network participants to reward addresses, only the protocol can credit these accounts.
- We don't encourage wallets to implement an explicit _withdrawal_ action, as it serves no purpose and costs fees.
- Instead, reward accounts are simply considered funding sources for regular payment transactions in addition to UTxOs, and wallets should design a UX to withdraw all available rewards when making a payment transaction. 
- The SDK need not make any assumptions as to which address they should be sent, so unconsumed balance should just be returned as change.

rhyslbw, line 16: - Funds cannot be sent by network participants to reward addresses, only the protocol can credit these accounts.
- We don't encourage wallets to implement an explicit _withdrawal_ action, as it serves no purpose and costs fees. Instead, reward accounts are simply considered funding sources for regular payment transactions in addition to UTxOs, and wallets should design a UX to withdraw all available rewards when making a payment transaction, returning the remaining as change.

packages/wallet/src/Delegation/withdrawal.ts
```diff
@@ -0,0 +1,20 @@
...
+  // to change address that we currently have as keyManager.deriveAddress(0, 0)?
+  address: csl.RewardAddress.new(network, csl.StakeCredential.from_keyhash(keyManager.stakeKey.hash())),
```
rhyslbw, line 18: :+1:

rhyslbw, line 18: :+1:

packages/wallet/test/Delegation/computeImplicitCoin.test.ts
```diff
@@ -0,0 +1,25 @@
+import { loadCardanoSerializationLib, ProtocolParametersRequiredByWallet } from '@cardano-sdk/core';
+import { testKeyManager } from '../testKeyManager';
+import { Delegation, InitializeTxProps } from '../../src';
+
+describe('Delegation.computeImplicitCoin', () => {
+  it('computes correctly', async () => {
```
rhyslbw, line 6: Please elaborate on what "correctly" means :smile:

packages/wallet/src/Delegation/computeImplicitCoin.ts
```diff
@@ -0,0 +1,31 @@
+import { ImplicitCoin } from '@cardano-sdk/cip2';
+import { ProtocolParametersRequiredByWallet } from '@cardano-sdk/core';
+import { sum } from 'lodash-es';
+import { InitializeTxProps } from '../types';
+
+/**
+ * Implementation is the same as in csl.get_implicit_input() and csl.get_deposit().
+ * Alternatively could build transaction body and use CSL to compute these.
```
rhyslbw, line 8: For now I'd prefer if we use this approach and just use the CSL for serialization.

packages/wallet/test/Transaction/computeImplicitCoin.test.ts
```diff
@@ -0,0 +1,26 @@
+import { loadCardanoSerializationLib, ProtocolParametersRequiredByWallet } from '@cardano-sdk/core';
+import { testKeyManager } from '../testKeyManager';
+import { Transaction } from '../../src';
+import { InitializeTxProps } from '../../src/Transaction';
+
+describe('Transaction.computeImplicitCoin', () => {
+  it('sums registrations for deposit, withdrawals and deregistrations for input', async () => {
```
rhyslbw, line 7: :+1:

rhyslbw (changes requested):
Looks great @mkazlauskas. Please squash down into minimal commits, then we're good to merge :+1:

PR #113, opened 2021-10-11
docs(wallet): add comment to highlight fees being a function of the tx body

rhyslbw:
# Context
We're missing some direction when it comes to knowing the calculated fees when initialising a transaction.

# Proposed Solution
Add comment to highlight fees being a function of the tx body


mkazlauskas (approved):
We should probably add more usage examples. I'm adding examples for new features, but it's still missing basic examples.

PR #114, opened 2021-10-13
feat(wallet): balance

mkazlauskas:
# Context

Need to have a convenient way to get available balance of coin, assets and rewards.

# Proposed Solution

- [x] add `UtxoRepository.availableRewards` (lock in the same way as locking UTxO until transaction is confirmed)
- [x] add `UtxoRepositoryEvent.Changed`
- [x] add `SingleAddressWallet.balance`, something like this:
  ```typescript
  interface Balance extends Ogmios.Value {
    rewards: Ogmios.Lovelace;
  }
  interface Balances {
      total: Balance;
      available: Balance;
  }
  ```  

- export Lovelace=bigint from core/Ogmios and use it where appropriate
- export ogmios types directly from core/Ogmios (`Ogmios.util.Type` -> `Ogmios.Type`)
- add `core/cslUtil.bytewiseEquals` utility function
- fix a potential bug with UTxO locking: in-memory UTxO set might be incorrect if user calls `UtxoRepository.sync` while transaction is in pending state. To mitigate this, instead of computing new UTxO set in-memory, `UtxoRepository` will always call `sync()` after transaction is confirmed or if the confirmation fails. As a result, removing `UtxoRepositoryEvent.TransactionUntracked` and adding `UtxoRepositoryEvent.OutOfSync`.

### DIFF
--- a/packages/core/test/CSL/util.test.ts
+++ b/packages/core/test/CSL/util.test.ts
@@ -1,0 +1,25 @@
+import { cslUtil } from '../../src';
+
+describe('cslUtil', () => {
+  describe('bytewiseEquals', () => {
+    it('returns true if the two CSL objects have equal byte representation', () => {
+      expect(
+        cslUtil.bytewiseEquals({ to_bytes: () => new Uint8Array([1, 2]) }, { to_bytes: () => new Uint8Array([1, 2]) })
+      ).toBe(true);
+    });
+
+    describe('false', () => {
+      it('returns false if the two CSL objects have different byte lengths', () => {
+        expect(
+          cslUtil.bytewiseEquals({ to_bytes: () => new Uint8Array([1, 2]) }, { to_bytes: () => new Uint8Array([1]) })
+        ).toBe(false);
+      });
+
+      it('returns false if the two CSL objects have different byte representations', () => {
+        expect(
+          cslUtil.bytewiseEquals({ to_bytes: () => new Uint8Array([1, 2]) }, { to_bytes: () => new Uint8Array([1, 3]) })
+        ).toBe(false);
+      });
+    });
+  });
+});
--- a/packages/wallet/test/InMemoryUtxoRepository.test.ts
+++ b/packages/wallet/test/InMemoryUtxoRepository.test.ts
@@ -8,7 +8,8 @@ import {
   KeyManagement,
   TransactionTrackerEvent,
   UtxoRepository,
-  UtxoRepositoryEvent
+  UtxoRepositoryEvent,
+  UtxoRepositoryFields
 } from '../src';
 import { MockTransactionTracker } from './mockTransactionTracker';
 import { ogmiosToCsl } from '@cardano-sdk/core/src/Ogmios';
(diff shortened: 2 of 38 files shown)


packages/core/test/CSL/util.test.ts
```diff
@@ -0,0 +1,25 @@
+import { cslUtil } from '../../src';
+
+describe('cslUtil', () => {
+  describe('bytewiseEquals', () => {
+    it('same bytes => true', () => {
```
rhyslbw, line 5:
```suggestion
    it('returns true if the two CSL objects have equal byte representation', () => {
```

packages/core/test/CSL/util.test.ts
```diff
@@ -0,0 +1,25 @@
...
+      expect(
+        cslUtil.bytewiseEquals({ to_bytes: () => new Uint8Array([1, 2]) }, { to_bytes: () => new Uint8Array([1, 2]) })
+      ).toBe(true);
+    });
+
+    describe('false', () => {
+      it('different lengths => false', () => {
```
rhyslbw, line 12:
```suggestion
      it('returns false if the two CSL objects have different byte lengths', () => {
```

packages/core/test/CSL/util.test.ts
```diff
@@ -0,0 +1,25 @@
...
+        expect(
+          cslUtil.bytewiseEquals({ to_bytes: () => new Uint8Array([1, 2]) }, { to_bytes: () => new Uint8Array([1]) })
+        ).toBe(false);
+      });
+
+      it('different byte => false', () => {
```
rhyslbw, line 18:
```suggestion
      it('returns false if the two CSL objects have different byte representations', () => {
```

packages/wallet/test/InMemoryUtxoRepository.test.ts
```diff
@@ -48,20 +57,38 @@ describe('InMemoryUtxoRepository', () => {
...
+      allUtxos: utxo,
+      availableUtxos: utxo,
+      allRewards: rewards,
+      availableRewards: rewards,
+      delegation: delegate
+    };
+    expect(utxoRepository).toMatchObject(expectedFields);
+    expect(syncedHandler).toBeCalledTimes(1);
+    expect(syncedHandler).toBeCalledWith(expectedFields);
+    // await expect(utxoRepository.allUtxos.length).toBe(3);
```
rhyslbw, line 89: Why are these commented out?

packages/wallet/test/InMemoryUtxoRepository.test.ts
```diff
@@ -109,64 +150,92 @@ describe('InMemoryUtxoRepository', () => {
...
+      });
 
-      // transaction rejected
-      await completeConfirmation!();
-      expect(onTransactionUntracked).not.toBeCalled();
-      expect(utxoRepository.availableUtxos).toHaveLength(numUtxoPreTransaction);
-      expect(utxoRepository.availableUtxos).toHaveLength(utxoRepository.allUtxos.length);
-      expect(utxoRepository.availableUtxos).toContain(transactionUtxo);
+      afterEach(async () => {
+        // Review: is it ok to reuse code by utilizing afterEach, or should we only use it for test cleanup?
```
rhyslbw, line 208: I will always lean towards clear test structure for best communication over convenience of writing the tests! :smiley: 

Best to make the primary assertions within the test case, so that a developer can easy comprehend the behaviour. I would expect assertions used within pre/post blocks to be solely based on setup or cleanup states.

packages/wallet/test/InMemoryUtxoRepository.test.ts
```diff
@@ -109,64 +153,90 @@ describe('InMemoryUtxoRepository', () => {
...
+        expect(utxoRepository.availableUtxos).toHaveLength(numUtxoPreTransaction - 1);
+        expect(utxoRepository.availableUtxos).not.toContain(transactionUtxo);
+        expect(utxoRepository.availableRewards).toBe(rewardsPreTransaction - transactionWithdrawal);
+        expect(onOutOfSync).not.toBeCalled();
+      };
+
+      it('transaction confirmed', async () => {
+        await trackTransaction(new Promise<void>((resolve) => (completeConfirmation = resolve)));
+        await confirmTxAndAssertUtxoRepositorySynced();
+        assertThereAreNoPendingTransactionsOrRewards();
```
rhyslbw, line 211: :+1:

rhyslbw (approved):
:tada:

PR #117, opened 2021-10-15
refactor: always use nodejs version of csl (BREAKING)

mkazlauskas:
# Context

Dynamic imports only work with es modules and our build target is commonjs

# Proposed Solution

Do not use dynamic imports and ditch all the CSL injections, relying on sdk user to use tooling like webpack to replace CSL module with the one targeting browser


rhyslbw (approved):
This improves the APIs, as while it takes an extra step to get the browser env configured, the dev experience is improved. Good job

rhyslbw (approved):

PR #118, opened 2021-10-18
feat: get stakepool info by id

mkazlauskas:
# Context

Wallets need to display stake pool information

# Proposed Solution

Create a new `StakePoolSearchProvider` interface and `cardano-graphql` implementation

### DIFF
--- a/packages/util-dev/package.json
+++ b/packages/util-dev/package.json
@@ -22,6 +22,7 @@
     "@cardano-sdk/cip2": "0.1.4"
   },
   "dependencies": {
-    "@cardano-sdk/core": "0.1.4"
+    "@cardano-sdk/core": "0.1.4",
+    "delay": "^5.0.0"
   }
 }
--- a/packages/util-dev/src/createStubStakePoolSearchProvider.ts
+++ b/packages/util-dev/src/createStubStakePoolSearchProvider.ts
@@ -1,0 +1,43 @@
+import { StakePoolSearchProvider } from '@cardano-sdk/core';
+import { StakePool } from '@cardano-sdk/core/src/Cardano';
+import delay from 'delay';
+
+export const somePartialStakePools: StakePool[] = [
+  {
+    id: 'pool1euf2nh92ehqfw7rpd4s9qgq34z8dg4pvfqhjmhggmzk95gcd402',
+    hexId: 'cf12a9dcaacdc09778616d60502011a88ed4542c482f2ddd08d8ac5a',
+    metadata: {
+      name: 'Keiths PiTest',
+      description: 'Keiths Pi test pool',
+      ticker: 'KPIT',
+      homepage: ''
+    }
+  },
+  {
+    id: 'pool1fghrkl620rl3g54ezv56weeuwlyce2tdannm2hphs62syf3vyyh',
+    hexId: '4a2e3b7f4a78ff1452b91329a7673c77c98ca96dece7b55c37869502',
+    metadata: {
+      name: 'VEGASPool',
+      description: 'VEGAS TestNet(2) ADA Pool',
+      ticker: 'VEGA2',
+      homepage: 'https://www.ada.vegas'
+    }
+  }
+] as StakePool[];
+
+/**
+ * Good source for testnet pools: https://testnet.adatools.io/pools
+ */
+export const createStubStakePoolSearchProvider = (
+  stakePools: StakePool[] = somePartialStakePools,
+  delayMs?: number
+): StakePoolSearchProvider => ({
+  queryStakePools: async (fragments) => {
+    if (delayMs) await delay(delayMs);
+    return stakePools.filter(({ id, metadata }) =>
+      fragments.some(
+        (fragment) => id.includes(fragment) || metadata?.name.includes(fragment) || metadata?.ticker.includes(fragment)
+      )
+    );
+  }
+});
--- a/packages/util-dev/src/index.ts
+++ b/packages/util-dev/src/index.ts
@@ -2,3 +2,4 @@ export * as AssetId from './assetId';
 export * as CslTestUtil from './cslTestUtil';
 export * as SelectionConstraints from './selectionConstraints';
 export * from './util';
+export * from './createStubStakePoolSearchProvider';
--- a/packages/util-dev/test/createStubStakePoolSearchProvider.test.ts
+++ b/packages/util-dev/test/createStubStakePoolSearchProvider.test.ts
@@ -1,0 +1,33 @@
+import { StakePoolSearchProvider } from '@cardano-sdk/core';
+import { createStubStakePoolSearchProvider } from '../src/createStubStakePoolSearchProvider';
+
+describe('createStubStakePoolSearchProvider', () => {
+  describe('queryStakePools', () => {
+    const ID_TO_MATCH = 'id-to-match';
+    let provider: StakePoolSearchProvider;
+    beforeEach(() => {
+      provider = createStubStakePoolSearchProvider([
+        { id: ID_TO_MATCH, metadata: { name: 'pool1', ticker: 'TICKR' } },
+        { id: 'other-id' }
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      ] as any);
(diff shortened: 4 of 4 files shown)


packages/core/src/Provider/types/StakePool.ts
```diff
@@ -0,0 +1,136 @@
...
+     * Stake pool's fixed cost
+     */
+    margin: Percent;
+  };
+  /**
+   * Stake pool stats
+   */
+  stats: {
+    /**
+     * Total blocks minted by the pool
```
rhyslbw, line 97:
```suggestion
     * Total blocks created by the pool
```
We're avoiding the terminology clash with _minting tokens_, and the negative connotation surrounding _forging_

packages/core/src/Provider/types/StakePool.ts
```diff
@@ -0,0 +1,136 @@
...
+     */
+    blocksMinted: number;
```
rhyslbw, line 99:
```suggestion
    blocksCreated: number;
```

packages/core/src/Provider/types/StakePool.ts
```diff
@@ -0,0 +1,136 @@
...
+     */
+    size: {
+      live: number;
+      active: number;
+    };
+    liveSaturation: Percent;
+    liveDelegators: number;
+  };
+  /**
+   * List of stake pool owners addresses
```
rhyslbw, line 119:
```suggestion
   * List of stake pool owners reward accounts
```

packages/core/src/Provider/types/StakePool.ts
```diff
@@ -0,0 +1,136 @@
...
+   * optional, 68 Characters
+   */
+  extVkey?: Hash16; // Review: is this the correct type alias?
+  /**
+   * Extended metadata defined by CIP-6
+   */
+  ext?: ExtendedStakePoolMetadata;
+}
+
+export interface StakePool {
```
rhyslbw, line 57: We should model the on-chain aspect as closely to the ledger model as possible, i.e extend the Ogmios `PoolParameters` interface.

packages/core/src/Provider/types/StakePool.ts
```diff
@@ -0,0 +1,136 @@
...
+ * Within range [0; 1]
+ */
+export type Percent = number;
+
+export type TransactionId = Hash16;
+
+/**
+ * https://github.com/cardano-foundation/CIPs/blob/master/CIP-0006/CIP-0006.md#on-chain-referenced-main-metadata-file
+ */
+export interface StakePoolMetadata {
```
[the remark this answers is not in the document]
mkazlauskas: I was planning to put these types to `wallet/StakePoolSearch`. It's unlikely that it would be used from any other package than `wallet`.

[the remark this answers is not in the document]
rhyslbw: It will be used in the backend service, and that's not in a wallet context.

packages/cardano-graphql-sps-provider/test/CardanoGraphQLStakePoolSearchProvider.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    fetchExtendedMetadata.mockReset();
+  });
+
+  describe('queryStakePools', () => {
+    const stakePools = [{}, { metadata: { extDataUrl: 'some-url' } }, { metadata: {} }];
+    beforeEach(() => {
+      queryStakePoolsWithMetadata.mockResolvedValueOnce(stakePools);
+    });
+    describe('[fetchExt = false]', () => {
+      it('queries and returns stake pools without fetching extended metadata', async () => {
```
rhyslbw: We can probably remove `fetchExt` param and just treat it as the service's responsibility to maintain a local cache that includes it if available, for clients to naturally request alongside the other properties.

packages/core/src/WalletProvider/types.ts
```diff
@@ -50,7 +50,7 @@ export type NetworkInfo = {
   stake: StakeSummary;
 };
 
-export interface CardanoProvider {
+export interface WalletProvider {
   ledgerTip: () => Promise<Cardano.Tip>;
   networkInfo: () => Promise<NetworkInfo>;
   stakePoolStats?: () => Promise<StakePoolStats>;
```
rhyslbw, line 56: Please add a todo here: move `stakePoolStats` out to other provider type, since it's not required for wallet operation.

scripts/publish.sh
```diff
@@ -3,8 +3,7 @@
 set -euo pipefail
 
 npm publish --cwd ./packages/blockfrost && \
-npm publish --cwd ./packages/cardano-graphql-db-sync && \
-npm publish --cwd ./packages/cardano-graphql-sps-provider && \
+npm publish --cwd ./packages/cardano-graphql && \
```
rhyslbw, line 6: Sorry @mkazlauskas , I probably wasn't clear and it's not obvious, but the intent is to have two packages since the first gen API isn't being brought forward. It's better to think of the two as separate services:

1. `cardano-graphql-db-sync` :point_right:  Gen 1
2. `cardano-graphql` :point_right: Gen 2 with our new stake pool search API

packages/core/src/StakePoolSearchProvider/types/StakePool.ts
```diff
@@ -99,6 +95,14 @@ export interface StakePool extends Omit<PoolParameters, 'pledge' | 'metadata'> {
    * Declared pledge quantity.
    */
   pledge: Ogmios.Lovelace;
+  /**
+   * Fixed stake pool running cost
+   */
+  cost: Ogmios.Lovelace;
```
rhyslbw, line 101: :+1:

packages/cardano-graphql/src/StakePoolSearchProvider/StakePoolSearchClient.ts
```diff
@@ -0,0 +1,96 @@
+import { ProviderError, ProviderFailure, StakePool } from '@cardano-sdk/core';
+import { gql, GraphQLClient } from 'graphql-request';
+import { StakePoolsQueryResponse } from './types';
+
+export class StakePoolSearchClient {
```
rhyslbw, line 5: This could just be a stateless function, separating the type and calling it `createStakePoolSearchClient`

packages/cardano-graphql/src/StakePoolSearchProvider/query.gql
```diff
@@ -0,0 +1,75 @@
+# Review: this is not used anywhere right now,
+# but would be great to keep the schema that server uses
+# and SDK typescript schema in sync.
+# Perhaps import this npm package in server?
```
rhyslbw, line 4: Yes, this is the design I have in mind. We may be better off deriving the GraphQL schema from TS types though for better composition and likely less duplication. Let's chat about this.

packages/cardano-graphql/src/StakePoolSearchProvider/types.ts
```diff
@@ -0,0 +1,74 @@
+// Review: this may be a good time to set up generating query response types
```
rhyslbw, line 1: Agree

packages/core/src/StakePoolSearchProvider/types/StakePool.ts
```diff
@@ -0,0 +1,119 @@
+import { Hash16, PoolParameters, PoolMetadata } from '@cardano-ogmios/schema';
```
rhyslbw, line 1: I'd like to see the `StakePool` type not bounded to the search provider, as it's more of a general concept in Cardano

packages/cardano-graphql-sps-provider/test/CardanoGraphQLStakePoolSearchProvider.test.ts
rhyslbw: > were you thinking to fetch it server-side

Yes, as it's just an extension of fetching the standard metadata from SMASH. We can make the service configurable to allow the operator control over this, but either way the client can be shielded quite easily here.

PR #126, opened 2021-10-28
feat(wallet): observable interface & transaction tracker

mkazlauskas:
# Context

1. Using `wallet` package with it's current interface has a lot of complexity. Similarly, implementing new services internally is complex because it's difficult to synchronize state between dependencies. 
2. Need a new utility to conveniently get transaction history.
3. Core types and CSL types representing the same data are both used in wallet interfaces.

# Proposed Solution

1. Introduce Observables via RxJS and manage all source (provider) subscriptions within SDK. Expose clean observable-based interface.
2. Expose transaction history and events via observable-based interface on `Wallet` object.
3. Keep only core types in user-facing wallet interfaces

- Rename TxAlonzo `metadata` to `auxiliaryData`
- Add `coreToCsl.txInputs`
- Make `ProtocolParametersRequiredByWallet` interface fields required
- Add dependency on [rxjs](https://www.npmjs.com/package/rxjs) and [backoff-rxjs](https://www.npmjs.com/package/backoff-rxjs)
- Add `StakePool.status` property
- Replace `jest.setup.js` with `moduleNameMapper` config option


packages/wallet/test/services/util/ProviderTrackerSubject.test.ts
```diff
@@ -47,6 +47,9 @@ describe('ProviderTrackerSubject', () => {
     });
   });
 
+  // Review: this is a list of tests needed to test all paths.
```
rhyslbw, line 50: Happy for us to merge with this as a todo, and we can decide to drop completely or spent the effort when possible

packages/wallet/test/services/TransactionsTracker.test.ts
```diff
@@ -0,0 +1,133 @@
...
+import { createTestScheduler } from '../testScheduler';
+import { queryTransactionsResult } from '../mocks';
+
+describe('createTransactionsTracker', () => {
+  // both of these variables are not relevant for tests, because
+  // they're using mock transactionsSource$
+  let config: SourceTrackerConfig;
+  let transactionsProvider: SimpleProvider<DirectionalTransaction[]>;
+
+  it('observable properties behaves correctly on successful transaction', async () => {
```
rhyslbw, line 24:
```suggestion
  it('observable properties behave correctly on successful transaction', async () => {
```

rhyslbw (changes requested):

packages/core/src/Cardano/types/Transaction.ts
```diff
@@ -42,8 +44,11 @@ export interface Redeemer {
   executionUnits: Cardano.ExUnits;
 }
 
-export type Witness = Omit<Partial<BlockBodyAlonzo['witness']>, 'redeemers'> & {
+export type Witness = Omit<Partial<BlockBodyAlonzo['witness']>, 'redeemers' | 'signatures'> & {
   redeemers?: Redeemer[];
+  signatures: Partial<{
```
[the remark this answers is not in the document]
mkazlauskas: I think it's safer when using indexer. TS makes you check if object key exists before using it. It's easy to do with optional chaining (or non-null assertion operator if you're sure it exists). Anyway I'm good with non-partial type if you don't think it's a good idea.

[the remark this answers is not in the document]
rhyslbw: Ok let's leave it as there's a benefit

packages/wallet/src/services/RewardsTracker.ts
```diff
@@ -31,7 +31,7 @@ export const createRewardsProvider$ =
         .then(({ delegationAndRewards: { rewards } }) => rewards || 0n)
     );
 
-const getWithdrawalQuantity = (tx: Cardano.NewTxAlonzo, keyManager: KeyManager): Cardano.Lovelace => {
+const getWithdrawalQuantity = (tx: Cardano.NewTxAlonzo, addresses: string[]): Cardano.Lovelace => {
```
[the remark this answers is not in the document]
mkazlauskas: I'm removing `addresses` argument as you can't withdraw from another wallet: all withdrawal quantities should be unavailable until transaction goes through or fails.

packages/wallet/test/services/RewardsTracker.test.ts
```diff
@@ -16,8 +16,8 @@ describe('createRewardsTracker', () => {
             body: {
               withdrawals: [
                 {
-                  address,
-                  quantity: 100n
+                  quantity: 100n,
+                  stakeAddress: address
```
rhyslbw, line 20: On line 9:

``` TypeScript
    const address = keyManager.deriveAddress(0, 0);
```
It should be:
``` TypeScript
    const stakeAddress = keyManager.stakeKey.to_bech32();
```

rhyslbw (changes requested):

packages/wallet/test/services/RewardsTracker.test.ts
mkazlauskas: Good catch. This exposed another issue that rewards provider didn't even need addreses[] that it used for calling `walletProvider.utxoDelegationAndRewards`

rhyslbw (changes requested):
Looks good @mkazlauskas

Just need to squash commits now

rhyslbw (approved):
Fantastic work here @mkazlauskas :tada:

PR #130, opened 2021-11-05
feat: rewards history & delegation preference

mkazlauskas:
# Context

- Add support to query rewards history of stake addresses
- Get the stake pool the wallet is currently delegating

# Proposed Solution

- [x] add `WalletProvider.rewardsHistory`
- [x] implement `rewardsHistory` for Blockfrost wallet provider
- [x] add `Wallet.networkInfo$` to get current epoch
- [x] add `Wallet.delegation` with `{rewardsHistory$: {all, lastReward, avgReward, lifetimeRewards}, stakePool$}`
- [x]  only re-fetch the following data when there is a new block on `tip$`:
  - utxo
  - transactions
  - networkInfo
  - rewardsHistory

- [x] do not emit provider data until it changes 
- [ ] add `KeyManager.stakeAddress`


packages/blockfrost/test/blockfrostProvider.test.ts
```diff
@@ -409,4 +409,43 @@ describe('blockfrostProvider', () => {
       slot: 37_767_194
     });
   });
+
+  test('rewardsHistory', async () => {
+    const pool_id = 'pool1pu5jlj4q9w9jlxeu370a3c9myx47md5j5m2str0naunn2q3lkdy';
+    const mockedResponse = [
```
rhyslbw, line 415: This test isn't covering the multiple page scenario. See: https://github.com/input-output-hk/cardano-js-sdk/blob/02f73783e39c7cb2c0e56d675bd2c5a3ae175903/packages/blockfrost/test/blockfrostProvider.test.ts#L70-L101

rhyslbw (changes requested):
>  only re-fetch the following data when there is a new block on tip$: `rewardsHistory`

Rewards history (and delegation) only needs to be checked at the start of each epoch

mkazlauskas:
> > only re-fetch the following data when there is a new block on tip$: `rewardsHistory`
> 
> Rewards history (and delegation) only needs to be checked at the start of each epoch

Makes sense for rewards. I'm planning to use `transactions.history.outgoing$` and just taking the last delegation certificate as a source for pool id, so it's not going to refetch until that changes.

packages/wallet/src/services/DelegationTracker/RewardsHistory.ts
```diff
@@ -0,0 +1,56 @@
...
+  walletProvider: WalletProvider,
+  keyManager: KeyManagement.KeyManager,
+  transactionsTracker: Transactions,
+  { getRewardsHistory = (props) => from(walletProvider.rewardsHistory(props)) }: RewardsHistoryProviderInternals = {}
+) => {
+  const firstDelegationEpoch$ = certificateTransactions(transactionsTracker, [
+    Cardano.CertificateType.StakeDelegation
+  ]).pipe(
+    map((transactions) => Math.min(...transactions.map(({ blockHeader: { slot } }) => slot))),
+    map(getEpoch),
```
rhyslbw, line 34: Delegation preferences take effect in n+2 epochs from the submitted certificate

packages/wallet/src/services/DelegationTracker/RewardsHistory.ts
```diff
@@ -0,0 +1,56 @@
...
+import { KeyManagement, ProviderWithArg, Transactions } from '../..';
+import { certificateTransactions } from './util';
+import { filter, from, map, mergeMap, take } from 'rxjs';
+
+interface RewardsHistoryProviderInternals {
+  getRewardsHistory?: ProviderWithArg<EpochRewards[], RewardHistoryProps>;
+}
+
+export const getEpoch = (slotNo: number) => {
+  // Review: we should probably be getting these constants from the provider?
```
rhyslbw, line 11: What about if we instead just lookup the block to get the epoch? These constants are going to be awkward to maintain in our own providers, and likely impossible to get from 3rd party providers such as Blockfrost.

mkazlauskas: Make a WalletProvider request to look up the block? That's an option, but not great for something that can be computed locally. I think something like `WalletProvider { slotEpoch(slot): Promise }` would be best, then if the provider supported something like  `{ eras: { firstEpoch; firstSlot; slotLength; }[] }` it could be implemented to compute locally.

rhyslbw: Yes, that design also occurred to me, but if we're calling a backend service for the constants why not just keep it consistent? This function is only being called off the back of a delegation being made right? In that case it will be very rarely called.

mkazlauskas: I guess it boils down to saving a few requests vs a more generic WalletProvider. Can we keep this hardcoded implementation for now and do this as a separate PR?

rhyslbw: These constants are network-specific, so to pursue this approach we'd need some type of config to correlate the network.

packages/blockfrost/src/blockfrostProvider.ts
```diff
@@ -318,12 +327,92 @@ export const blockfrostProvider = (options: Options, logger = dummyLogger): Wall
...
+  };
+
+  const queryBlocksByHashes: WalletProvider['queryBlocksByHashes'] = async (hashes) => {
+    const responses = await Promise.all(hashes.map((hash) => blockfrost.blocks(hash)));
+    // Review: a bunch of nullable fields here. Is there a reason for them to be nullable?
+    // Maybe just throw if any of the fields are not set?
+    return responses.map((response) => ({
+      confirmations: response.confirmations,
+      date: new Date(response.time * 1000),
+      epoch: response.epoch!,
```
rhyslbw, line 386: There are two type of blocks that curiously do not belong to epochs. We do not need to be concerned with these special blocks though, as they are only applicable to the Byron-era and do not contain transactions. Can throw if not set

packages/blockfrost/src/blockfrostProvider.ts
```diff
@@ -318,12 +327,92 @@ export const blockfrostProvider = (options: Options, logger = dummyLogger): Wall
...
+      // why is this nullable?
+      epochSlot: response.epoch_slot!,
```
rhyslbw, line 388: See above

packages/blockfrost/src/blockfrostProvider.ts
```diff
@@ -318,12 +327,92 @@ export const blockfrostProvider = (options: Options, logger = dummyLogger): Wall
...
+      // why is this nullable?
+      fees: BigInt(response.fees || '0'),
```
rhyslbw, line 390: Not sure, could also be related to the special Byron-era blocks. This handling is fine though

packages/blockfrost/src/blockfrostProvider.ts
```diff
@@ -318,12 +327,92 @@ export const blockfrostProvider = (options: Options, logger = dummyLogger): Wall
...
+      header: {
+        blockHash: response.hash,
+        // why is this nullable?
+        blockHeight: response.height!,
```
rhyslbw, line 394: The same special Byron-era blocks also do not have block numbers! :upside_down_face: Can throw if not set

packages/blockfrost/src/blockfrostProvider.ts
```diff
@@ -318,12 +327,92 @@ export const blockfrostProvider = (options: Options, logger = dummyLogger): Wall
...
+        slot: response.slot! // why is this nullable?
```
rhyslbw, line 395: Ditto

packages/blockfrost/src/blockfrostProvider.ts
```diff
@@ -318,12 +327,92 @@ export const blockfrostProvider = (options: Options, logger = dummyLogger): Wall
...
+      }, // why is this nullable?
+      nextBlock: response.next_block || undefined,
```
rhyslbw, line 397: If you get the tip, then the next block would be undefined

packages/blockfrost/src/blockfrostProvider.ts
```diff
@@ -318,12 +327,92 @@ export const blockfrostProvider = (options: Options, logger = dummyLogger): Wall
...
+      previousBlock: response.previous_block || undefined,
+      size: response.size,
+      slotLeader: response.slot_leader,
+      totalOutput: BigInt(response.output || '0'),
+      txCount: response.tx_count,
+      vrf: response.block_vrf! // why is this nullable?
```
rhyslbw, line 403: This will only be present in the post Byron-era blocks

packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
```diff
@@ -0,0 +1,84 @@
...
+    switchMap((transactions) =>
+      blockEpochProvider(transactions.map((tx) => tx.blockHeader.blockHash)).pipe(
+        map((epochs) => transactions.map((tx, txIndex) => ({ epoch: epochs[txIndex], tx })))
+      )
+    ),
+    share()
+  );
+
+// Review: all individual pieces of this are tested and it doesn't contain any logic
+// besides initializing the dependencies. Is it worth writing the test?
```
rhyslbw, line 54: I agree the test would just create more overhead with little benefit. I'm happy if it builds :wink:

rhyslbw (changes requested):
Looks good @mkazlauskas. Since we're relying on the external e2e testing I'm happy to see this go through to the next pre-release, with the additional todos handled separately. Just need to squash commits.

rhyslbw (approved):
:tada:

PR #131, opened 2021-11-11
refactor: cleanup after initial observable interface implementation

mkazlauskas:
# Context

Some chores after initial observable interface implementation. See below.

# Proposed Solution

- Use (cold) `Observable` instead of `() => Observable` for provider source
- `Wallet.sync()` will now only force-sync `tip$`. The rest of the wallet is synced only if `tip$` changes.
- Hoist `isNotNil` util to core package
- Add some missing tests

## Other changes
- Rename `providerStub` to `mockWalletProvider` since it's no longer a stub and uses `jest.fn()` for it's functions.


rhyslbw (approved):
Great work here @mkazlauskas, there are some nice simplifications compared to the initial impl :+1:

PR #133, opened 2021-11-12
fix(blockfrost): interpret 404s in Blockfrost provider and optimise batching

rhyslbw:
# Context
Currently 404s for address-based lookups are allowed to fall through to an error state,
however, they need to be interpreted as expected results since it indicates the address is unused.

# Proposed Solution
- Refactors the Blockfrost client implementation to remove `addressesUtxosAll`
in favour of `addressesUtxos`, and `addressesTransactionsAll` with `addressesTransactions`
as each of the former sends a batch of requests asynchronously, often over-fetching and
over-consuming the API quota. The new implementation fetches results sequentially as needed.

- Replaces the `utxoDelegationAndRewards` parameter `stakeKeyHash` with `stakeAddress`

### DIFF
--- a/packages/blockfrost/src/blockfrostProvider.ts
+++ b/packages/blockfrost/src/blockfrostProvider.ts
@@ -9,8 +9,8 @@ import {
   WalletProvider
 } from '@cardano-sdk/core';
 import { BlockFrostAPI, Error as BlockfrostError, Responses } from '@blockfrost/blockfrost-js';
-import { BlockfrostToCore } from './BlockfrostToCore';
-import { Options } from '@blockfrost/blockfrost-js/lib/types';
+import { BlockfrostToCore, BlockfrostTransactionContent, BlockfrostUtxo } from './BlockfrostToCore';
+import { Options, PaginationOptions } from '@blockfrost/blockfrost-js/lib/types';
 import { dummyLogger } from 'ts-log';
 import { flatten, groupBy } from 'lodash-es';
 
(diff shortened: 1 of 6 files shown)


packages/blockfrost/src/blockfrostProvider.ts
```diff
@@ -66,6 +66,35 @@ const toProviderError = (error: unknown) => {
 export const blockfrostProvider = (options: Options, logger = dummyLogger): WalletProvider => {
   const blockfrost = new BlockFrostAPI(options);
 
+  const fetchByAddressSequentially = async <Item, Response>(
```
mkazlauskas, line 69: Nice function. Could be made even more generic and used for non-address queries, e.g. `const fetchSequentially = async <Item, Arg, Response>(`. Not a requested change for this PR, just a note.

packages/blockfrost/test/blockfrostProvider.test.ts
```diff
@@ -118,96 +136,95 @@ describe('blockfrostProvider', () => {
...
-        tx_hash: '0f3abbc8fc19c2e61bab6059bf8a466e6e754833a08a62a6c56fe0e78f19d9d5',
-        tx_index: 0
-      },
-      {
-        amount: [
-          {
-            quantity: '1097647',
-            unit: 'lovelace'
-          }
+  describe('utxoDelegationAndRewards', () => {
```
mkazlauskas, line 139: I think it would be cleaner to test`fetchByAddressSequentially` separately and then mock it for all Blockfrost requests that uses it. Not a requested change, just a note.

mkazlauskas (changes requested):

packages/blockfrost/src/blockfrostProvider.ts
```diff
@@ -122,21 +151,32 @@ export const blockfrostProvider = (options: Options, logger = dummyLogger): Wall
...
-    return { delegationAndRewards, utxo };
+    try {
+      const accountResponse = await blockfrost.accounts(rewardAccount);
+      const delegationAndRewards = {
+        delegate: accountResponse.pool_id || undefined,
+        rewards: BigInt(accountResponse.withdrawable_amount)
+      };
+      return { delegationAndRewards, utxo };
+    } catch (error) {
+      if (error.status_code === 404) {
```
mkazlauskas, line 175: Getting a ts error here since `error` is of type `unknown`

packages/blockfrost/src/blockfrostProvider.ts
rhyslbw: I almost went to that design, but decided to keep it address-specific until we need more capabilities.

packages/blockfrost/test/blockfrostProvider.test.ts
rhyslbw: Good point. I'll make a ticket to follow-up

packages/blockfrost/src/blockfrostProvider.ts
rhyslbw: Hmm that's odd, I didn't see the compiler complaining :thinking:

mkazlauskas: `yarn test` works without errors for me too, but the ts error is shown by my vscode :thinking:

mkazlauskas: Tried it. There must be a difference between our setups. Anyway I think this an actual issue to fix because we don't want to be using `unknown` type without explicitly casting it.

mkazlauskas (approved):
Looks good!

PR #136, opened 2021-11-15
feat: add Balance.deposit and Wallet.delegation.rewardAccounts$

mkazlauskas:
# Context

Stake key deposit can be used as implicit input, so it can be considered as part of balance. This information should be available in Wallet API.

# Proposed Solution

- [x] add `Balance.deposit`.
- [x] add `Wallet.delegation.rewardAccounts$` that has stake key status field, used to compute balance deposit.

- [x] add e2e test for delegation, find out that it doesn't work, fix all the issues, primarily:
  - Input selection has to always select at least 1 UTxO, even if implicit input is enough to cover output quantities.
  - Stake key certificates were using invalid stake address
  - Transaction builder wasn't adding the certificates to serialized transaction


packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -186,22 +188,24 @@ export class SingleAddressWallet implements Wallet {
...
               )
           );
         })
       )
     );
   }
-  async finalizeTx({ body, hash }: TxInternals, auxiliaryData?: Cardano.AuxiliaryData): Promise<Cardano.NewTxAlonzo> {
-    const signatures = await this.#keyManager.signTransaction(hash);
+  async finalizeTx(tx: TxInternals, auxiliaryData?: Cardano.AuxiliaryData): Promise<Cardano.NewTxAlonzo> {
+    const signatures = await this.#keyManager.signTransaction(tx);
```
rhyslbw, line 204: :open_mouth:

packages/wallet/src/services/BalanceTracker.ts
```diff
@@ -4,6 +4,11 @@ import { TrackerSubject } from './util';
 import { TransactionalObservables } from '..';
 import { combineLatest, map } from 'rxjs';
 
+// TODO: subtract deposit quantity from total utxo coin as it can't be spent
+// Review: not sure how to represent this. Is it 'balance.available$'? 'balance.total$'? a new one?
```
rhyslbw: How about representing `balance.deposited` explicitly? That will best communicate these allocated funds using domain terminology

packages/wallet/test/e2e/SingleAddressWallet.test.ts
```diff
@@ -26,6 +28,48 @@ describe('SingleAddressWallet', () => {
     expect(wallet.addresses[0].bech32.startsWith('addr')).toBe(true);
   });
 
+  test('delegation', async () => {
+    const delegateeBefore1stTx = await firstValueFrom(wallet.delegation.delegatee$);
+    // swap poolId if it's already delegating to one of the pools
```
rhyslbw: :sunglasses:

packages/wallet/test/e2e/SingleAddressWallet.test.ts
```diff
@@ -26,6 +28,48 @@ describe('SingleAddressWallet', () => {
...
+    const tx1Internals = await wallet.initializeTx({
+      certificates: [
+        ...(isStakeKeyRegistered
+          ? []
+          : ([
+              { __typename: Cardano.CertificateType.StakeRegistration, address: rewardAccount }
+            ] as Cardano.Certificate[])),
+        { __typename: Cardano.CertificateType.StakeDelegation, address: rewardAccount, epoch, poolId }
+      ],
+      // TODO: make outputs optional. Coin selection has to select at least 1 utxo for change output in this case.
```
rhyslbw: :+1:

packages/wallet/src/services/BalanceTracker.ts
```diff
@@ -1,28 +1,52 @@
...
+  );
 
 export const createBalanceTracker = (
+  protocolParameters$: Observable<ProtocolParametersRequiredByWallet>,
   utxoTracker: TransactionalObservables<Cardano.Utxo[]>,
-  rewardsTracker: TransactionalObservables<Cardano.Lovelace>
+  rewardsTracker: TransactionalObservables<Cardano.Lovelace>,
+  delegationTracker: Delegation
 ): TransactionalTracker<Balance> => {
+  const depositTotal$ = createDepositTracker(
```
rhyslbw, line 34: Yep, this is a good representation. It uses the domain terminology to make an explicit categorisation of the unusable funds

packages/wallet/src/services/types.ts
```diff
@@ -41,6 +42,7 @@ export interface FailedTx {
   reason: TransactionFailure;
 }
 
+// TODO: rename -> TransactionsTracker
```
rhyslbw, line 45: :+1:

packages/wallet/src/services/types.ts
```diff
@@ -71,8 +73,23 @@ export interface Delegatee {
...
+  Registering = 'REGISTERING',
+  Registered = 'REGISTERED',
+  Unregistering = 'UNREGISTERING',
+  Unregistered = 'UNREGISTERED'
+}
+
+export interface RewardAccount {
+  address: Cardano.Address;
+  keyStatus: DelegationKeyStatus;
+  // Maybe add rewardsHistory$ and delegatee$ for each reward account too
```
rhyslbw, line 86: :+1:

packages/wallet/src/services/types.ts
```diff
@@ -71,8 +73,23 @@ export interface Delegatee {
...
+}
+
+// TODO: rename -> DelegationTracker
```
rhyslbw, line 89: :+1:

packages/wallet/src/services/types.ts
```diff
@@ -71,8 +73,23 @@ export interface Delegatee {
   nextNextEpoch: Cardano.StakePool | null;
 }
 
+export enum DelegationKeyStatus {
```
rhyslbw, line 76:
```suggestion
export enum StakeKeyStatus {
```
This should be a bit more generic to handle stake pool registration, which is a separate action to delegating stake. Could also be `StakeKeyRegistrationStatus` if the specificity is desired.

packages/cip2/src/RoundRobinRandomImprove/roundRobin.ts
```diff
@@ -105,5 +105,9 @@ export const roundRobinSelection = ({
       }
     }
   }
+  if (utxoSelected.length === 0) {
```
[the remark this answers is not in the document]
mkazlauskas: It just isn't one of the properties it checks. Also wasn't a concern with initial implementation before adding support for implicit coin. Not an obvious one too, intuitively tx with only implicit coin sounds ok to me.

packages/wallet/src/services/types.ts
mkazlauskas: I think we should leave this out of scope of this PR and do it when we need multi-address support

packages/wallet/test/services/DelegationTracker/transactionCertificates.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  transactionHasAnyCertificate,
+  transactionStakeKeyCertficates
+} from '../../../src';
+import { createTestScheduler } from '../../testScheduler';
+
+describe('transactionCertificates', () => {
+  test('transactionStakeKeyCertficates', () => {
+    const certificates = transactionStakeKeyCertficates({
+      certificates: [
+        { __typename: Cardano.CertificateType.StakeDelegation } as Cardano.Certificate,
```
rhyslbw, line 15: I've just picked this up when reviewing the tests. `CertificateType.StakeRegistration` should be `CertificateType.StakeKeyRegistration` for clarity, ditto for deregistration. Without it we're indicating the stake is being registered, which could be confused with the concept of delegation.

rhyslbw (changes requested):

rhyslbw (approved):
Nice work @mkazlauskas 👍

rhyslbw (changes requested):
Just need to squash commits

rhyslbw (approved):

PR #137, opened 2021-11-16
Test/refactor wallet e2e multi wallet

rhyslbw:
# Context
We're currently not running the wallet e2e test in CI. It should at least run prior to releasing. The tests should also retain the bulk of the funds to reduce maintenance of the test wallet.

# Proposed Solution
- Refactor the test to use two wallets, with a 3rd transaction sending funds back to the first wallet.
- Add a pre-release workflow, that must be run manually via a trigger.


.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,43 @@
+name: Pre-release
+
+on:
+  pull_request:
```
rhyslbw, line 4: To remove once the workflow is confirmed

mkazlauskas (changes requested):
The main test of 2nd tx is deposit being returned, so we won't lose much if we wait for 2nd and 3rd tx confirmation in parallel, then assert balances after both txs are confirmed.

packages/wallet/test/e2e/SingleAddressWallet.test.ts
```diff
@@ -32,97 +29,154 @@ const createDelegationCertificates = async (wallet: Wallet, rewardAccount: Carda
...
+    const afterTx2TotalBalance = {
+      one: await firstValueFrom(walletOne.balance.total$),
+      two: await firstValueFrom(walletTwo.balance.total$)
+    };
+
+    // Make a 3rd tx using wallet2 with:
+    // - payment to return the coins received in tx1 for future tests
+
+    const tx3Internals = await walletTwo.initializeTx({
+      outputs: new Set([{ address: walletOne.addresses[0].bech32, value: { coins: tx1OutputCoins } }])
```
mkazlauskas, line 173: Need to send smaller quantity to still have funds for fee if walletTwo was originally empty, which is likely to be the case for new test setups (faucet limits to 1 request per 24h or sth).

rhyslbw: I've loaded both CI wallets with 1000 ada

PR #139, opened 2021-11-16
refactor(wallet): move delegatee$ under RewardAccount

mkazlauskas:
# Context

Currently `delegatee$` picks up delegation from any stake key found in local transactions. Need to have more specific information, delegatee per reward account.

# Proposed Solution

Move it under `Wallet.delegation.rewardAccounts$` and make it optional

## Additional changes

Fix a bug where subscribing to `Wallet.transactions.outgoing.confirmed$` or `failed$` after transaction is submitted wouldn't emit anything.


packages/wallet/src/services/types.ts
```diff
@@ -82,12 +82,12 @@ export enum StakeKeyStatus {
 export interface RewardAccount {
   address: Cardano.Address;
   keyStatus: StakeKeyStatus;
-  // Maybe add rewardsHistory$ and delegatee$ for each reward account too
+  delegatee: Delegatee;
```
rhyslbw, line 85:
```suggestion
  delegatee?: Delegatee;
```
Semanticly reward accounts are not always delegated. This would be the case for a Stake Pool reward account, which is associated with the pool formally, but not delegating to it.

rhyslbw (changes requested):

packages/wallet/src/services/types.ts
mkazlauskas: The fields of `Delegatee` type are nullable:
```
export interface Delegatee {
  currentEpoch: Cardano.StakePool | null;
  nextEpoch: Cardano.StakePool | null;
  nextNextEpoch: Cardano.StakePool | null;
}
```
Should we make `nextNextEpoch` not nullable then?

rhyslbw: Stake pool reward accounts will likely never have a `Delegatee` though :thinking:

mkazlauskas: I did make it optional and `nextNextEpoch` non-nullable as it will always be set if reward account is delegating

rhyslbw (approved):
Nice work @mkazlauskas

PR #140, opened 2021-11-17
refactor(wallet): change KeyManager.deriveAddress return type to Address object

mkazlauskas:
# Context

The current implementation binds the stake key as a property of the key manager, but this is restrictive and doesn't quite align with how payment addresses should be bound to the stake key and derived with it's credential. In short there can be multiple pairs of payment address + stake key derived from a pair of pub/priv keys.

# Proposed Solution

Change `KeyManager.deriveAddress` return type to `Address` object that contains both payment address and reward account.

Refactor wallet services to take in an observable of `Address` objects where either payment address or reward account bech32 string was previously required.


packages/wallet/src/KeyManagement/types.ts
```diff
@@ -1,19 +1,39 @@
...
+  accountIndex: number;
+  addressIndex: number;
+  address: Cardano.Address;
+  rewardAccount: Cardano.RewardAccount;
+}
+
 export interface KeyManager {
-  deriveAddress: (addressIndex: number, index: 0 | 1) => string;
+  // TODO: do not expose CSL objects publicly
+  // Reconsider which keys should be exposed, if any.
```
rhyslbw, line 27: The public keys should be accessible, and `publicAccountKey` is good as-is for now, but `publicStakeKey` should be probably be returned based on the address derivation path. This is how `cardano-wallet` handles it: https://input-output-hk.github.io/cardano-wallet/api/edge/#tag/Keys

We can map this through from the `Wallet` interface to keep the KeyManager an internal concern.

rhyslbw (changes requested):

rhyslbw (approved):
Great work @mkazlauskas

PR #141, opened 2021-11-20
feat: asset info

mkazlauskas:
# Context

Wallets need to display asset info (name, logo, ticker etc.).

This branch is based on #140 branch.

# Proposed Solution

Add `Wallet.assets$` that emits asset info for all assets found in `Wallet.balance.total$`


packages/blockfrost/src/blockfrostWalletProvider.ts
```diff
@@ -8,93 +7,36 @@ import {
...
+  });
 
 /**
  * Connect to the [Blockfrost service](https://docs.blockfrost.io/)
  *
  * @param {Options} options BlockFrostAPI options
  * @returns {WalletProvider} WalletProvider
+ * @throws {ProviderFailure}
  */
 export const blockfrostProvider = (options: Options, logger = dummyLogger): WalletProvider => {
```
rhyslbw, line 37:
```suggestion
export const blockfrostWalletProvider = (options: Options, logger = dummyLogger): WalletProvider => {
```

packages/blockfrost/src/util.ts
```diff
@@ -0,0 +1,80 @@
...
+  throw new ProviderError(ProviderFailure.Unknown, error, `status_code: ${status_code}`);
+};
+
+export const withProviderErrors = <T>(providerImplementation: T) =>
+  Object.keys(providerImplementation).reduce((provider, key) => {
+    provider[key] = (...args: any[]) => (providerImplementation as any)[key](...args).catch(toProviderError);
+    return provider;
+  }, {} as any) as T;
+
+// TODO: test this
```
rhyslbw, line 55:
```suggestion
```

rhyslbw (changes requested):

packages/blockfrost/src/blockfrostWalletProvider.ts
```diff
@@ -306,12 +248,38 @@ export const blockfrostProvider = (options: Options, logger = dummyLogger): Wall
     ];
   };
 
+  const fetchJsonMetadata = async (txHash: Cardano.Hash16): Promise<Cardano.MetadatumMap | null> => {
+    try {
+      const response = await blockfrost.txsMetadata(txHash);
+      return response.reduce((map, metadatum) => {
+        // Not sure if types are correct, missing 'label', but it's present in docs
```
rhyslbw, line 255: It was fixed in https://github.com/blockfrost/blockfrost-js/releases/tag/v2.0.1. [2.0.0]( https://github.com/blockfrost/blockfrost-js/releases/tag/v2.0.0) appears to only have one breaking change, so the upgrade should be routine.

packages/blockfrost/src/blockfrostWalletProvider.ts
```diff
@@ -306,12 +248,38 @@ export const blockfrostProvider = (options: Options, logger = dummyLogger): Wall
...
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const { json_metadata, label } = metadatum as any;
+        if (!json_metadata || !label) return map;
+        // not sure how Blockfrost represents integer metadatum. Might need to convert to bigint here.
```
rhyslbw, line 259: Are we missing a test to cover this? We can dig up an example tx on _testnet_ that includes a large integer to use as a mock.

rhyslbw (changes requested):

packages/blockfrost/src/blockfrostWalletProvider.ts
mkazlauskas: Yep, we are missing a test for this. That would be nice, how does one dig up tx'es?

rhyslbw (approved):
Great work @mkazlauskas

PR #145, opened 2021-11-23
refactor: improve string type safety

mkazlauskas:
# Context

Currently all string types are simple aliases of `string`. We have a lot of different types of strings for keys and hashes and we should utilize TypeScript to ensure that the string we get is the correct type that is not ambiguous.

# Proposed Solution

Convert all string type aliases to more restricted types and validate them when creating. (BREAKING)

* Add dependency to `bech32` package
* Add `ProviderFailure.InvalidResponse`
* Rename certificate fields from `address` to `rewardAccount` where appropriate (BREAKING)
* Convert transaction witness, `Assets` and `TokenMap` to `Map` (BREAKING)

## Follow-up work

- `Address` type doesn't accept base58-encoded Byron addresses and `Address.util.isAddress` does consider it valid. We need to decide whether we want to support it. If we do, there will be more code changes with CSL usage.
- Might be a good idea to write more e2e tests with Blockfrost to ensure hash types are aligned. If we do that, we should dig up testnet transactions with certificates. What is the easiest way to do that?
- There might still some type aliases from Ogmios used implicitly via nested types. I did not check every single type imported from Ogmios, but this PR converted the important string type aliases that we use and I suggest to convert the remainder (if there are any) as we go.


packages/core/src/Cardano/types/Strings/util.ts
```diff
@@ -0,0 +1,59 @@
...
+ */
+export const assertIsBech32WithPrefix = (target: string, prefix: string, expectedDecodedLength?: number): void => {
+  let decoded: Decoded;
+  try {
+    decoded = bech32.decode(target);
+  } catch {
+    throw new InvalidStringError(`expected bech32-encoded string with '${prefix}' prefix`);
+  }
+  if (decoded.prefix !== prefix) {
+    throw new InvalidStringError(`expected bech32 prefix '${prefix}', got '${decoded.prefix}''`);
```
rhyslbw, line 36: This isn't covered by a test currently, and it's not clear to me why the check is required

packages/core/test/Cardano/types/Strings/util.test.ts
```diff
@@ -0,0 +1,33 @@
+import { InvalidStringError, assertIsBech32WithPrefix, assertIsHexString } from '../../../../src/Cardano';
+
+describe('Cardano/types/Strings/util', () => {
+  test('assertIsBech32WithPrefix', () => {
```
rhyslbw, line 4: Best to break this down into more explicit test cases that describe each scenario.

packages/core/test/Cardano/types/Strings/util.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    expect(() => assertIsBech32WithPrefix('', '')).toThrowError(InvalidStringError);
+    expect(() =>
+      assertIsBech32WithPrefix(
+        'pool',
+        'addr_test1qrydm8hsalwjmuqj624cwnyrs554zu6a8n8wg64dxk3zarsxu2hyfhlkwuxupa9d5085eunq2qywy7hvmvej456flknsq3qxgd'
+      )
+    ).toThrowError(InvalidStringError);
+  });
+
+  test('assertIsHexString', () => {
```
rhyslbw, line 24: Ditto.

packages/core/test/Cardano/types/Strings/util.test.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    it('throws when asserting an invalid bech32 (too short)', () => {
+      expect(() => assertIsBech32WithPrefix('pool1', 'pool')).toThrowError(InvalidStringError);
+    });
+
+    it('throws when asserting an empty string', () => {
+      expect(() => assertIsBech32WithPrefix('', '')).toThrowError(InvalidStringError);
+    });
+
+    it('throws when decoded prefix does not match the expected prefix', () => {
```
rhyslbw, line 28: This is throwing at an unexpected execution point. Making the assertion more specific will produce a failing test

image

rhyslbw (changes requested):

rhyslbw (commented):
I'm wondering if we'll have a cleaner model by removing the `types/Strings/` path withing `src/Cardano`, and just have say `src/Cardano/StakePool.ts`, since this module contains more than type definitions, and using the primative type to categorise is a bit redundant for a domain layer. Not a blocker, but just wanted to mention it before proceeding with the others.

image

mkazlauskas:
> I'm wondering if we'll have a cleaner model by removing the `types/Strings/` path withing `src/Cardano`, and just have say `src/Cardano/StakePool.ts`, since this module contains more than type definitions, and using the primitive type to categorise is a bit redundant for a domain layer. Not a blocker, but just wanted to mention it before proceeding with the others.

Sure, let's do that

rhyslbw (approved):

packages/core/src/Cardano/types/AuxiliaryData.ts
```diff
@@ -26,6 +26,6 @@ export interface AuxiliaryDataBody {
 }
 
 export interface AuxiliaryData {
-  hash?: Cardano.Hash16;
+  hash?: Cardano.Hash16; // TODO: Review: need to find an example of this to verify type and length
```
rhyslbw, line 29: Transaction and block hashes are also `Hash16` type, which can be validated as 64 character strings

https://explorer.cardano.org/en/transaction?id=61f85fad0dbabcafa9cf1d3

packages/core/src/Cardano/types/Certificate.ts
```diff
@@ -80,20 +84,22 @@ export interface PoolRetirementCertificate {
...
   __typename: CertificateType.StakeDelegation;
-  address: Address;
+  rewardAccount: RewardAccount;
   poolId: PoolId;
   epoch: Epoch;
 }
 
 export interface MirCertificate {
   __typename: CertificateType.MIR;
+  // Review: need to learn what this cert is and figure out if 'address' is actually an Address or a RewardAccount
```
rhyslbw, line 94: MIR = Move instantaneous rewards
This is a special mechanism that enables funds to be moved from reserves or the treasury into a _reward_ account. It was required as a way to enable rewards generated on side-chains, such as the Incentivized Testnet (ITN), to be applied to participants on the main chain.

packages/core/src/Cardano/types/Transaction.ts
```diff
@@ -21,8 +57,12 @@ export interface TxBodyAlonzo {
   withdrawals?: Withdrawal[];
   certificates?: Cardano.Certificate[];
   mint?: Cardano.TokenMap;
-  scriptIntegrityHash?: Cardano.Hash16;
-  requiredExtraSignatures?: Cardano.Hash16[];
+  scriptIntegrityHash?: Cardano.Hash28ByteBase16;
+  // TODO: need to find an example of this to verify type and length
+  // This is the type we use for witness object keys too, which kinda
+  // makes sense to have as bech32, because you might want to display
+  // it to the user to show who signed the tx
```
sadasant, line 64: (Hello! I’m new to Cardano. I want to start contributing in one way or another! The future of global finances etc etc)

What about making an issue instead of a TODO comment? The issue could be linked to this PR (as a comment on this line) and anybody looking to collaborate could try to pick it up later.

I also think that bech32 sounds good for the reasons stated in this comment, so I wonder: why not just making it bech32?

sadasant:
This PR looks really good! Makes the project way more robust in multiple fronts. I also appreciate the tests!

packages/core/src/Cardano/types/Transaction.ts
rhyslbw: Hi @sadasant, thanks for the contribution :+1: 

We find annotating the code keeps it fresh in mind during this early development phase, with the todos reflected in our internal project management system. We're currently reserving Github issues for external submissions, but we'll move to a more traditional OS model with contributor guidelines once things stabilise. As it stands, code contributions would be considered, but we're just not actively seeking them right now.

mkazlauskas: Hi @sadasant, thank you for a review! It is currently validated as bech32 with `ed25519_pk` prefix.

@rhyslbw some extra context for this TODO: Blockfrost doesn't seem to support `requiredExtraSignatures` field and I don't see it in cardano-graphql docs. This field is from Ogmios types and it's a `Hash16` in there.

packages/core/src/Cardano/types/Certificate.ts
```diff
@@ -80,23 +30,24 @@ export interface PoolRetirementCertificate {
...
   pot: 'reserve' | 'treasury';
 }
 
 export interface GenesisKeyDelegationCertificate {
   __typename: CertificateType.GenesisKeyDelegation;
-  genesisHash: Hash16;
-  genesisDelegateHash: Hash16;
-  vrfKeyHash: Hash16;
+  // TODO: need to find examples of these hashes to figure out type and length
+  genesisHash: Hash32ByteBase16;
```
rhyslbw, line 48: https://github.com/input-output-hk/cardano-configurations/blob/26b6b6de73f90e4777602b372798bf77addcc321/network/testnet/cardano-node/config.json#L7-L15

packages/core/src/Cardano/types/StakePool/StakePool.ts
```diff
@@ -20,11 +22,12 @@ export interface Cip6MetadataFields {
    * 128 Characters Maximum, must be a valid URL
    */
   extSigUrl?: string;
+  // TODO: this is probably wrong, as CIP says it's 68 characters long. Find and validate the correct format.
   /**
    * the public Key for verification
    * optional, 68 Characters
    */
-  extVkey?: Hash16; // TODO: figure out if this is the correct type alias
+  extVkey?: Hash32ByteBase16;
```
rhyslbw, line 30: A 32 byte ed25519 verification key, rather than it's hash, bech32 encoded with prefix `vrf_vk`

https://github.com/cardano-foundation/CIPs/pull/15#discussion_r538230733

https://github.com/cardano-foundation/CIPs/blob/d5842cc0e58d08c4ba10f5a96d9b9ef14e78ff72/CIP-0005/README.md#keys

packages/core/src/Cardano/types/Transaction.ts
rhyslbw: This Ogmios test vector has `b5ae663aaea8e500157bdf4baafd6f5ba0ce5759f7cd4101fc132f54` (56 chars):

https://raw.githubusercontent.com/CardanoSolutions/ogmios/master/server/test/vectors/ChainSync/Response/RequestNext/135.json

rhyslbw (changes requested):

packages/core/src/Cardano/types/Transaction.ts
mkazlauskas: Would you prefer our type to be hex as in ogmios instead of bech32?

packages/core/src/Cardano/types/StakePool/StakePool.ts
mkazlauskas: `vrf_vk` prefix results in bech32 string length of 65 and CIP specifies this field length to be 68, which matches `poolmd_vk` prefix as per [this comment](https://github.com/cardano-foundation/CIPs/pull/15#discussion_r572736672). I think I'll make it accept both prefixes.

mkazlauskas: How about PoolParameters? Ogmios has [this](https://github.com/input-output-hk/cardano-js-sdk/pull/145/files#diff-a9da9e5f1f8b2f32b8a4d95f189ec0afb6126e9b1839eef8a4aa1b29f512d0c8R41) as `Hash16` but since it's the same data maybe it's also not a hash but an actual key?

packages/core/src/Cardano/types/Certificate.ts
mkazlauskas: This verifies that GenesisHash is `Hash32ByteBase16`. Assume the same for `genesisDelegateHash` and `vrfKeyHash`?

packages/core/src/Cardano/types/StakePool/StakePool.ts
rhyslbw: `poolmd_vk` does indeed fit the bill, so let's just validate against this only.

> How about PoolParameters?

Yeah I agree it's likely not a hash, as we would have that in the name + we know the `Hash16` type has been used for non-hash values.

packages/core/src/Cardano/types/Certificate.ts
rhyslbw: Yes that's a safe assumption

packages/core/src/Cardano/types/Transaction.ts
rhyslbw: Sorry, no as you were. In general we should be using the bech32 version if a prefix has been defined in the CIP

mkazlauskas: Prefix `ed25519_pk` that it currently uses is not listed in cip5 and CSL's `PublicKey.to_bech32()` does not accept a prefix argument. I think what we want is `addr_vk` prefix. I already added a dependency to `bech32` npm package so we can create the appropriate  string from `PublicKey.as_bytes()`

packages/core/test/Cardano/types/Block.test.ts
```diff
@@ -1,7 +1,12 @@
-import { BlockId } from '../../../src/Cardano';
+import { BlockId, VrfVkBech32 } from '../../../src/Cardano';
 
 describe('Cardano/types/Block', () => {
   it('BlockId() accepts a valid transaction hash', () => {
     expect(() => BlockId('0dbe461fb5f981c0d01615332b8666340eb1a692b3034f46bcb5f5ea4172b2ed')).not.toThrow();
   });
+
+  it('VrfVkBech32() accepts a valid vrf verification key bech32 string with vrf_vk or poolmd_vk prefix', () => {
```
rhyslbw: Missing a couple of throwing cases: 
1. invalid prefix
2. valid prefix, but truncated string

packages/core/src/Cardano/types/StakePool/StakePool.ts
```diff
@@ -1,11 +1,20 @@
...
 
 /**
  * Within range [0; 1]
  */
 export type Percent = number;
 
+/**
+ * 32 byte ed25519 verification key as bech32 string with 'poolmd_vk' prefix.
+ */
+export type PoolmdVkey = OpaqueString<'PoolmdVk'>;
```
rhyslbw, line 15:
```suggestion
export type PoolmdVkey = OpaqueString<'PoolmdVk'>;
```
```suggestion
export type PoolMdVkey = OpaqueString<'PoolMdVk'>;
```

rhyslbw (changes requested):

rhyslbw:
> Address type doesn't accept base58-encoded Byron addresses and Address.util.isAddress does consider it valid. We need to decide whether we want to support it. If we do, there will be more code changes with CSL usage.

All wallets need capability to send to legacy addresses, as they are still considered valid addresses just with limited utility.

packages/core/src/Cardano/types/Transaction.ts
sadasant: (Thank you for your answers, @rhyslbw and @mkazlauskas 🙏)

packages/core/test/Cardano/types/Asset.test.ts
```diff
@@ -1,16 +1,57 @@
-import { AssetFingerprint, AssetId, AssetName, PolicyId } from '../../../src/Cardano';
+import { AssetFingerprint, AssetId, AssetName, PolicyId, util } from '../../../src/Cardano';
```
rhyslbw, line 1: Nice tests, very expressive.

rhyslbw (approved):
Awesome work here @mkazlauskas :tada: 

Please note the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#specification) standard for representing breaking changes:

image

The reason I've not rejected the commits here is that we're not maintaining a stable API yet and pre-production minor versioning is not based on breaking changes, but moving forward it will be good to get into the habit so these details are second nature post-1.0. `standard-version` interprets this to decide the next version, so it's imperative for automating the release process, in addition to the benefits the standard brings to developers.

PR #146, opened 2021-11-27
chore: set node version to >=14.15.0 <15.0.0

jarkkosyrjala:
# Context

Errors on install when using incompatible version of Node allowed by current engines value.

# Proposed Solution

Set minimum version to `14.15.0`. Also set the maximum to `<15.0.0`, because some of the dependencies are not compatible with Node versions above 14.

## Way to reproduce the error in current master branch:
```
nvm install 14.14.0
yarn install --force
```

Output:
```
error jest@27.0.6: The engine "node" is incompatible with this module. Expected version "^10.13.0 || ^12.13.0 || ^14.15.0 || >=15.0.0". Got "14.14.0"
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "description": "An SDK for interacting with the Cardano blockchain",
   "engines": {
-    "node": "^14"
+    "node": ">=14.15.0 <15.0.0"
   },
   "workspaces": [
     "packages/*"


package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "description": "An SDK for interacting with the Cardano blockchain",
   "engines": {
-    "node": "^14"
+    "node": ">=14.15.0 <15.0.0"
```
rhyslbw, line 7:
```suggestion
    "node": ">=14.17.0 <15.0.0"
```

rhyslbw (commented):
Thanks @jarkkosyrjala. We might be best off keeping in sync with https://github.com/input-output-hk/cardano-js-sdk/blob/f628f37d9233a32800c0b20debd71f285591c3c4/.nvm#L1. Thoughts @mkazlauskas ?

mkazlauskas:
> . Thoughts @mkazlauskas ?

I agree it should be in sync with `.nvm`, but I don't have a problem with `engines` allowing lower version as long as it works.

mkazlauskas (dismissed):

rhyslbw (approved):

jarkkosyrjala:
@mkazlauskas @rhyslbw Rebased to master and signed the commit, because it was previously unverified and prevented the merge. Sorry about that.

rhyslbw (approved):
Thanks @jarkkosyrjala !

PR #147, opened 2021-12-01
feat: key safety and restoration

mkazlauskas:
# Context

The current version of the `InMemoryKeyManager` was designed primarily for testing, with the intent to iterate it into a secure, production-ready implementation.

We're currently holding private key in memory with our `InMemoryKeyManager`. Stealing it with access to someone's laptop would be fairly quick and easy by using Chrome devtools memory heap snapshot, or maybe even just running a script in console. We should probably only decrypt private key when it needs to sign a transaction.

# Proposed Solution

- Implement [EmIP-003](https://github.com/Emurgo/EmIPs/blob/master/specs/emip-003.md) encryption
  - Add tests for Yoroi compatibility:
    - Decrypting Yoroi-encrypted root private key
    - Restoring root private key from Yoroi-generated 15 word mnemonic
- (BREAKING) Refactor key agent to not keep keys in-memory and decrypt on-demand. **Restoring keys from mnemonic will result in a new private key, so any testing/development wallets will be recreated**.
  - Fix an issue that appeared after refactoring some KeyManager functions to be async (`sharedDistinctBlock` -> `distinctBlock` and `sharedDistinctEpoch`->`distinctEpoch`)
  - Implement and export `KeyManagement.cachedGetPassword` to cache password for `InMemoryKeyAgent`. This is needed for any bearable user experience, as pretty much every operation requires decrypting private key.
  - Create custom KeyAgent error type

- Add dependencies to the following npm packages:
  - [pbkdf2](https://www.npmjs.com/package/pbkdf2)
  - [chacha](https://www.npmjs.com/package/chacha)
  - [get-random-values](https://www.npmjs.com/package/get-random-values)
- (BREAKING) Rename `KeyManager` to `KeyAgent` and refactor it to support async operations (planning for hardware wallet support)
- New types in `core` package: `Bip32PrivateKey`, `Bip32PrivateKey`, `Signatures`
- Extend `KeyAgent` interface adding a `serializableData` field that will be used for storage. Make it generic enough to be able to implement generic `KeyAgent` serialization in the future.
- Add `KeyAgentBase` abstract class that specific implementations can extend  to reuse some functionality.


packages/wallet/src/KeyManagement/emip3.ts
```diff
@@ -0,0 +1,41 @@
+// Review: a separate package for emip3 might be a good idea because it can be useful as a standalone package.
```
rhyslbw, line 1: Agree. We can source it at the top level of this SDK for now

packages/wallet/src/KeyManagement/InMemoryKeyAgent.ts
```diff
@@ -0,0 +1,141 @@
...
+    return { publicKey, signature };
+  }
+
+  async derivePublicKey({ index, type }: AccountKeyDerivationPath): Promise<Cardano.Ed25519PublicKey> {
+    const accountPrivateKey = await this.#deriveAccountPrivateKey();
+    const cslPublicKey = accountPrivateKey.derive(type).derive(index).to_public().to_raw_key();
+    return Cardano.Ed25519PublicKey(Buffer.from(cslPublicKey.as_bytes()).toString('hex'));
+  }
+
+  // Review: assuming Bip39 entropy is the actual key
```
rhyslbw, line 103: :+1:

packages/wallet/test/KeyManagement/InMemoryKeyManager.test.ts
```diff
@@ -1,28 +1,26 @@
-import { CSL, Cardano } from '@cardano-sdk/core';
+import { Cardano } from '@cardano-sdk/core';
```
rhyslbw, line 1: Looks like you have this noted given the temporary export as the old name, but just a reminder to rename to `InMemoryKeyAgent.test.ts`

packages/wallet/src/KeyManagement/types.ts
```diff
@@ -13,32 +40,51 @@ export enum AddressType {
...
+  encryptedRootPrivateKeyBytes: number[];
+}
+
+export interface SerializableLedgerKeyAgentData extends SerializableKeyAgentDataBase {
+  __typename: KeyAgentType.Ledger;
+}
+
+export type SerializableKeyAgentData = SerializableInMemoryKeyAgentData | SerializableLedgerKeyAgentData;
+
+// TODO: utility to cache password for specified duration
```
rhyslbw, line 78: Would be good to have a doc block for this function. When I think of the behaviour of authentication, I expect the return value to be a boolean, so maybe this needs a different name.

packages/wallet/src/KeyManagement/emip3.ts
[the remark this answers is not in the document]
rhyslbw: Logged

packages/wallet/src/KeyManagement/types.ts
mkazlauskas: Renamed it to `getPassword`

packages/wallet/src/KeyManagement/types.ts
```diff
@@ -13,32 +36,73 @@ export enum AddressType {
...
+  networkId: Cardano.NetworkId;
+  accountIndex: number;
+}
+
+export interface SerializableInMemoryKeyAgentData extends SerializableKeyAgentDataBase {
+  __typename: KeyAgentType.InMemory;
+  encryptedRootPrivateKeyBytes: number[];
+}
+
+export interface SerializableLedgerKeyAgentData extends SerializableKeyAgentDataBase {
```
rhyslbw, line 68: Maybe drop this and all other Ledger types out until we have a Ledger implementation

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -233,5 +233,22 @@ export class SingleAddressWallet implements Wallet {
     this.protocolParameters$.complete();
     this.genesisParameters$.complete();
     this.#tip$.complete();
+    this.addresses$.complete();
+  }
+
+  #initializeAddress(existingAddress?: GroupedAddress): Observable<GroupedAddress[]> {
```
rhyslbw, line 239: This is good for a single-address wallet but doesn't align with multi-address wallet semantics. I think it's fine to just refactor later given it's a private method though

packages/wallet/test/KeyManagement/InMemoryKeyAgent.test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  test('__typename', () => {
+    expect(typeof keyAgent.__typename).toBe('string');
+  });
+
+  test('accountIndex', () => {
+    expect(typeof keyAgent.accountIndex).toBe('number');
+  });
+
+  test('serializableData', () => {
```
rhyslbw, line 32: A test that demonstrates serializability could be to convert to JSON and back

packages/wallet/src/KeyManagement/types.ts
rhyslbw: The todo can be removed now right?

packages/wallet/test/KeyManagement/cachedGetPassword.test.ts
```diff
@@ -0,0 +1,29 @@
+import { KeyManagement } from '../../src';
+
+jest.useFakeTimers();
+
+describe('cachedGetPassword', () => {
+  it('caches password for specified duration and ignores cache when "noCache=true"', async () => {
```
rhyslbw, line 6: Let's break this into two distinct tests for better readability

rhyslbw (changes requested):

rhyslbw (approved):
Excellent work @mkazlauskas. Great job with the PR description, and conventional commits standards application. :rocket:

PR #148, opened 2021-12-07
feat: InitializeTxProps validation

mkazlauskas:
# Context

Wallets need to know minimum coin quantity on outputs so that they can create a valid transaction.

# Proposed Solution

```typescript
interface MinimumCoinQuantity {
  minimumCoin: Cardano.Lovelace;
  coinMissing: Cardano.Lovelace;
}
type MinimumCoinQuantityPerOutput = Map<Cardano.TxOut, MinimumCoinQuantity>;
interface InitializeTxPropsValidationResult {
  minimumCoinQuantities: MinimumCoinQuantityPerOutput;
}
interface Wallet {
  validateInitializeTxProps(props: InitializeTxProps): Promise<InitializeTxPropsValidationResult>;
}
```

- Update `cip2` package types to use SDK's core types instead of CSL. As a result,
  - Add a few new test utils to `util-dev` package
  - Add `coreToCsl.tokenMap`
- Add `inputSelection` field to result of `wallet.initializeTx`

### DIFF
--- a/packages/cip2/src/RoundRobinRandomImprove/change.ts
+++ b/packages/cip2/src/RoundRobinRandomImprove/change.ts
@@ -1,46 +1,43 @@
-import { CSL, Cardano, coreToCsl } from '@cardano-sdk/core';
+import { Cardano } from '@cardano-sdk/core';
 import { ComputeMinimumCoinQuantity, TokenBundleSizeExceedsLimit } from '../types';
-import {
-  ImplicitCoinBigint,
-  UtxoSelection,
-  UtxoWithValue,
-  assetQuantitySelector,
-  assetWithValueQuantitySelector,
-  getCoinQuantity,
-  getWithValuesCoinQuantity
-} from './util';
 import { InputSelectionError, InputSelectionFailure } from '../InputSelectionError';
+import { UtxoSelection, assetQuantitySelector, getCoinQuantity, toValues } from './util';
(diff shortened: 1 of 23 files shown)


rhyslbw (changes requested):
Nice, https://github.com/input-output-hk/cardano-js-sdk/pull/148/commits/c842fd5e08faad328d412baf874afa7272ac14ad is a great refactor!

It would be great to add a little context to the `feat(wallet): add 'inputSelection' to initializeTx result` commit message, describing the motivation for the change. It's also a breaking change, so needs the conventional commits format.

mkazlauskas:
> Nice, [c842fd5](https://github.com/input-output-hk/cardano-js-sdk/commit/c842fd5e08faad328d412baf874afa7272ac14ad) is a great refactor!

Thanks!

> It's also a breaking change, so needs the conventional commits format.

I thought about this and I understand how it could be a breaking change, but it's only 'slightly breaking' as it's just adding a property to an object.

rhyslbw:
> I thought about this and I understand how it could be a breaking change, but it's only 'slightly breaking' as it's just adding a property to an object.

Oh shoot, yes your original commit format was fine as it's not breaking. I misinterpreted as the original object being nested under a sub-property.

rhyslbw (approved):
Great work @mkazlauskas

PR #155, opened 2021-12-23
Feature/cardano graphql projections

rhyslbw:
# Context
A chain follower is required to sync data from the chain into Dgraph, and the SDK is the logical place to home this component, which can then be imported and run from an external service.

# Proposed Solution
- Implement a client to provision Dgraph with the generated schema

# Todo:
- [ ] Add the chain-sync client

n/a

### DIFF
--- a/packages/cardano-graphql/test/e2e/projection.test.ts
+++ b/packages/cardano-graphql/test/e2e/projection.test.ts
@@ -1,0 +1,20 @@
+import { DgraphClient, createDgraphClient } from '../../src';
+import fs from 'fs';
+import path from 'path';
+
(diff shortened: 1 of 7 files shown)


packages/cardano-graphql/test/e2e/projection.test.ts
```diff
@@ -0,0 +1,20 @@
...
+describe('Projections', () => {
+  let dgraphClient: DgraphClient;
+
+  beforeEach(() => {
+    dgraphClient = createDgraphClient(address, console);
+  });
+
+  describe('setting the schema', () => {
+    it('completes without throwing an error', async () => {
+      await expect(dgraphClient.setSchema(schema)).resolves;
```
rhyslbw, line 17: This is actually failing with `Error: 2 UNKNOWN: while lexing...`

image

PR #158, opened 2022-01-11
fix(blockfrost): added e2e test and fix for collaterals

iadmytro:
# Context

Fixed issue with collaterals in blockfrost

### DIFF
--- a/packages/blockfrost/src/BlockfrostToCore.ts
+++ b/packages/blockfrost/src/BlockfrostToCore.ts
@@ -54,20 +54,9 @@ export const BlockfrostToCore = {
 
   outputs: (outputs: BlockfrostOutputs): Cardano.TxOut[] => outputs.map((output) => BlockfrostToCore.txOut(output)),
 
-  // without `as OgmiosSchema.Utxo` above TS thinks the return value is (OgmiosSchema.TxIn | OgmiosSchema.TxOut)[][]
   transactionUtxos: (utxoResponse: Responses['tx_content_utxo']) => ({
-    collaterals: utxoResponse.inputs
-      .filter((input) => input.collateral)
-      .map((input) => ({
-        ...BlockfrostToCore.txIn(input),
-        address: Cardano.Address(input.address)
-      })),
-    inputs: utxoResponse.inputs
-      .filter((input) => !input.collateral)
-      .map((input) => ({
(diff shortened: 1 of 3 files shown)


packages/blockfrost/src/BlockfrostToCore.ts
```diff
@@ -56,10 +56,18 @@ export const BlockfrostToCore = {
 
   // without `as OgmiosSchema.Utxo` above TS thinks the return value is (OgmiosSchema.TxIn | OgmiosSchema.TxOut)[][]
```
mkazlauskas, line 57: This comment seems obsolete. Could you please remove it?

packages/blockfrost/src/BlockfrostToCore.ts
```diff
@@ -56,10 +56,18 @@ export const BlockfrostToCore = {
...
   transactionUtxos: (utxoResponse: Responses['tx_content_utxo']) => ({
-    inputs: utxoResponse.inputs.map((input) => ({
-      ...BlockfrostToCore.txIn(input),
-      address: Cardano.Address(input.address)
-    })),
+    collaterals: utxoResponse.inputs
+      .filter((input) => input.collateral)
+      .map((input) => ({
+        ...BlockfrostToCore.txIn(input),
```
mkazlauskas, line 62: Seems like `address` is already set in `BlockfrostToCore.txIn`, so you can just use that as map function

mkazlauskas (changes requested):

packages/blockfrost/test/e2e/queryTransactions.test.ts
```diff
@@ -11,5 +11,20 @@ describe('blockfrostWalletProvider', () => {
         '9223372036854775707922337203685477570792233720368547757079223372'
       );
     });
+    it('parses collaterals correctly', async () => {
```
james-iohk, line 14: It would be good to also check that it handles when script fails to validate and collateral is consumed. In this scenario we'd expect to see only collateral UTxOs as input and zero outputs because all collateral is treated as fees

mkazlauskas: Let's address this in a separate PR

mkazlauskas (approved):

PR #164, opened 2022-01-14
feat: slot datetime computation util

mkazlauskas:
# Context

Wallet needs to display transaction date

# Proposed Solution

- add util for computing slot date given ledger time settings in `core` package
- add a new provider type (`TimeSettingsProvider`) in `core` package
- add a stub `TimeSettingsProvider` implementation in `util-dev` package
- add `Wallet.timeSettings$` and an integration test for usage example

## Out of scope

An actual provider implementation is out of scope of this PR. Blockfrost doesn't seem to support this, so will implement it for `cardano-graphql`.

## Other changes

Running `buildSchema.js` script in `cardano-graphql` package broke due to not knowing how to import ES modules from `lodash-es`. Having mixed modules is not great, we should probably set up the packages to build both esm and cjs. For now I added a workaround so that it can build successfully.

### DIFF
--- a/packages/core/test/util/slotTimeCalc.test.ts
+++ b/packages/core/test/util/slotTimeCalc.test.ts
@@ -1,0 +1,35 @@
+import { SlotTimeCalc, TimeSettingsError, createSlotTimeCalc, testnetTimeSettings } from '../../src';
+
+describe('slotTimeCalc', () => {
+  describe('testnet', () => {
+    const slotTimeCalc: SlotTimeCalc = createSlotTimeCalc(testnetTimeSettings);
+
+    it('correctly computes date of the 1st block', () =>
+      expect(slotTimeCalc(1031)).toEqual(new Date(1_564_020_236_000)));
+
+    it('correctly computes date of the genesis block', () =>
+      expect(slotTimeCalc(0)).toEqual(new Date(1_563_999_616_000)));
+
+    it('correctly computes date of some Byron block', () =>
+      expect(slotTimeCalc(1_209_592)).toEqual(new Date(1_588_191_456_000)));
+
+    it('correctly computes date of the last Byron block', () =>
+      expect(slotTimeCalc(1_598_399)).toEqual(new Date(1_595_967_596_000)));
+
+    it('correctly computes date of the 1st Shelley block', () =>
+      expect(slotTimeCalc(1_598_400)).toEqual(new Date(1_595_964_016_000)));
+
+    it('correctly computes date of the 2nd Shelley block', () =>
+      expect(slotTimeCalc(1_598_420)).toEqual(new Date(1_595_964_036_000)));
+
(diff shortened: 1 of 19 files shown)


lgobbi-atix (dismissed):

james-iohk:
@mkazlauskas Missing updated `packages/core/src/index.ts` ?

packages/core/test/util/slotTimeCalc.test.ts
```diff
@@ -0,0 +1,30 @@
+import { SlotTimeCalc, TimeSettingsError, createSlotTimeCalc, testnetTimeSettings } from '../../src';
+
+describe('slotTimeCalc', () => {
+  describe('testnet', () => {
+    const slotTimeCalc: SlotTimeCalc = createSlotTimeCalc(testnetTimeSettings);
+
+    it('correctly computes date of the 1st block', () =>
```
james-iohk, line 7: Isn't first block 
```
expect(slotTimeCalc(1031)).toEqual(new Date(1_564_020_236_000)));
``` 
?

packages/core/test/util/slotTimeCalc.test.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    it('correctly computes date of the 1st Shelley block', () =>
+      expect(slotTimeCalc(1_598_400)).toEqual(new Date(1_595_964_016_000)));
+
+    it('correctly computes date of the 2nd Shelley block', () =>
+      expect(slotTimeCalc(1_598_420)).toEqual(new Date(1_595_964_036_000)));
+
+    it('correctly computes date of some Shelley block', () =>
+      expect(slotTimeCalc(8_078_371)).toEqual(new Date(1_602_443_987_000)));
+  });
```
james-iohk, line 24: Could add
```
it('throws with invalid slot', () =>
  expect(() => slotTimeCalc(-1)).toThrowError(TimeSettingsError));
```

james-iohk (approved):

packages/core/test/util/slotTimeCalc.test.ts
mkazlauskas: You're right, nice catch!

packages/core/test/util/slotTimeCalc.test.ts
```diff
@@ -0,0 +1,30 @@
+import { SlotTimeCalc, TimeSettingsError, createSlotTimeCalc, testnetTimeSettings } from '../../src';
+
+describe('slotTimeCalc', () => {
+  describe('testnet', () => {
+    const slotTimeCalc: SlotTimeCalc = createSlotTimeCalc(testnetTimeSettings);
+
```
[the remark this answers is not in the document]
mkazlauskas: Added the 2 suggested tests

james-iohk (approved):

PR #174, opened 2022-02-02
chore: update ogmios to 5.1.0

rhyslbw:
# Context
I've split this commit out into a separate branch to merge in isolation. Ogmios 5.1.0 is out, and I'm targeting this version in my current WIP. There's a breaking change relating to Plutus scripts, and some numbers are typed as `bigint`

# Proposed Solution
Add the Plutus script type field to the GraphQL query, then use this to determine the Plutus version when converting to core types.


packages/cardano-graphql/src/WalletProvider/queryTransactions/graphqlTransactionsToCore.ts
```diff
@@ -32,7 +32,7 @@ const scriptToCore = (script: GraphqlScript): Cardano.Script => {
     return { native: nativeScriptToCore(script) };
   } else if (script.__typename === 'PlutusScript') {
     return {
-      plutus: script.cborHex
+      'plutus:v1': script.cborHex
```
mkazlauskas, line 35: Should probably do something like:
```
[script.type === 'PlutusScriptV1' ? 'plutus:v1' : 'plutus:v2']: script.cborHex
```

rhyslbw: yeah I was looking for a field that would give me a classification, but there's no `.type` property on `script` or similar. Do you mean refactor to make this possible?

mkazlauskas: No need to refactor, just add it to [query](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-graphql/src/operations/transactions.graphql#L102-L105), then `yarn build && yarn generate` and it should be available

packages/cardano-graphql/src/WalletProvider/queryTransactions/graphqlTransactionsToCore.ts
```diff
@@ -32,7 +32,7 @@ const scriptToCore = (script: GraphqlScript): Cardano.Script => {
     return { native: nativeScriptToCore(script) };
   } else if (script.__typename === 'PlutusScript') {
     return {
-      plutus: script.cborHex
+      [script.type === 'PlutusScriptV1' ? ('plutus:v1' as keyof Cardano.Script) : 'plutus:v2']: script.cborHex
```
mkazlauskas, line 35: I'm wondering if `('plutus:v1' as const)` would do the trick here

mkazlauskas (approved):
Looks good!

packages/cardano-graphql/src/WalletProvider/queryTransactions/graphqlTransactionsToCore.ts
rhyslbw: image

PR #180, opened 2022-02-11
feat: asset metadata schema

mkazlauskas:
# Context

Asset (token/nft) metadata needs to be stored in a database in order to be able to efficiently access it.

# Proposed Solution

Add cardano-graphql schema for cip25 and cip35 data.

There was a number of (mostly) BREAKING changes introduced:

- rename AssetMetadata->TokenMetadata type
- remove deprecated 'logo' and 'image' fields from TokenMetadata: assuming that parsers can check them and set 'icon'
- hoist NftMetadata and accompanying types, as well as `metadatumToCip25` to core package, under `/Asset`
- rename and move core `Cardano.Asset` type to `Asset.AssetInfo`
- rename AssetInfo metadata->tokenMetadata to better reflect it's purpose
- fix `metadatumToCip25` to support either a single value or an array where appropriate. Use only array in our types, as it's more convenient to use (no need to typecheck to access it)

## Follow-up work (out of scope)

- `AssetProvider.getAsset` needs to have an additional argument to specify what optional data to fetch amongst `{nftMetadata, tokenMetadata, history}`
- Once we have a functioning cardano-graphql AssetProvider, we should change the implementation of getting NFT metadata in `wallet` package to use it.


packages/cardano-graphql/src/StakePoolSearchProvider/CardanoGraphQLStakePoolSearchProvider.ts
```diff
@@ -66,7 +66,7 @@ const toCoreStakePool = (responseStakePool: GraphqlStakePool): Cardano.StakePool
     },
     status: stakePool.status,
     transactions: {
-      // TODO: current implementation will only return 1 (latest/active) registration certificate
+      // TODO: current implementation wil l only return 1 (latest/active) registration certificate
```
rhyslbw, line 69:
```suggestion
      // TODO: current implementation will only return 1 (latest/active) registration certificate
```

rhyslbw (changes requested):
Looks good. There's just a minor mistype to correct.

rhyslbw (approved):
Great work @mkazlauskas

PR #182, opened 2022-02-16
refactor(wallet): added second wallet to e2e test

iadmytro:
# Context

added second wallet to e2e test


rhyslbw (changes requested):
Please also grab the workflow file from https://github.com/input-output-hk/cardano-js-sdk/pull/137, so we can trigger the workflow manually when it's in master

rhyslbw (changes requested):
Merging is blocked with that failing test. See if you can resolve it with minimal changes

mkazlauskas:
I think it's just missing `TIME_SETTINGS_PROVIDER` variable in workflow config

PR #186, opened 2022-02-18
fix(core): throw serialization error for invalid metadata fields

james-iohk:
# Context

ADP-1467

# Proposed Solution


mkazlauskas (approved):
Nice work!

PR #187, opened 2022-02-18
fix(core): correct metadata length check

james-iohk:
# Context

Fix a condition from `>=` to `>`

# Proposed Solution


mkazlauskas (changes requested):
Maybe add a test with 64-length? Also, there's another eslint warning in `packages/core/test/CSL/coreToCsl.test.ts` (not from this commit, probably from a previous one)

mkazlauskas (approved):

PR #188, opened 2022-02-23
feat(wallet): data signing

mkazlauskas:
# Context

Wallet needs to be able to sign data/messages, not only transactions.

# Proposed Solution

Add `Wallet.signData`, returning a structure described in [this](https://github.com/cardano-foundation/CIPs/pull/148) CIP-0030 PR.

- Hoist `HexBlob` type under core `Cardano.util` (slightly breaking).
- Add some utils to work with bytes (`bytesToHex`, `castHexBlob`) and some convenience functions to convert from `HexBlob` to more specific hex string types (e.g. tx id).
- Simplify some opaque string type tests by removing the assertion that it's implemented using a specific util. Justification for this is that the code is really simple (1 line of code) and the tests are too complicated for too little value.


rhyslbw (approved):
Looks good. So we're deferring the introduction of a `crypto` package until it's absolutely needed or prior to `1.0`?

mkazlauskas:
> Looks good. So we're deferring the introduction of a `crypto` package until it's absolutely needed or prior to `1.0`?

Just didn't want to bloat the PR with a refactor that can be done separately, I think we should create the package before 1.0.

PR #189, opened 2022-02-23
feat(wallet): Ledger Key Agent

tomislavhoracek:
# Context
SDK needs to be able to support the hardware devices / wallets. LedgerKeyAgent will be responsible to do that.

# Proposed Solution
Implement a KeyAgent to support the Ledger hardware devices

- InMemoryLedgerKeyAgent added for handling device connection (transport).
- ~~`ledger-live-common` lib used to handle communication and exec commands on Ledger devices~~ (Some Cardano methods are not implemented e.g `getAddress` so we will move-to / use `ledgerjs-hw-app-cardano`)
- Web based communication with devices over `@ledgerhq/hw-transport-webhid`
- Node based communication with devices over `@ledgerhq/hw-transport-node-hid`

# TODO:
- [X] Fix lib loading issues
- [X] Consider using different library instead of `ledger-live-common`
- [X] Refactor from `ledger-live-common` to `ledgerjs-hw-app-cardano`
- [X] Web based communication with devices over `@ledgerhq/hw-transport-webhid`
- [x] Add methods to check / recover Transport and DeviceConnection in general
- [x] Allow end apps to use and manipulate with its own Transport
- [ ] ~~Add Bluetooth based communication with devices~~ (it is not in the scope of basic implementation)
- [x] Node based communication with devices over `@ledgerhq/hw-transport-node-hid`
- [x] Add tests - NodeHid based only


packages/wallet/src/KeyManagement/types.ts
```diff
@@ -72,7 +89,7 @@ export type GetPassword = (noCache?: true) => Promise<Uint8Array>;
 export interface KeyAgent {
   get networkId(): Cardano.NetworkId;
   get accountIndex(): number;
-  get serializableData(): SerializableKeyAgentData;
+  get serializableData(): SerializableKeyAgentData | SerializableLedgerKeyAgentData;
```
mkazlauskas, line 92: It would be better to do: `export type SerializableKeyAgentData = SerializableInMemoryKeyAgentData | SerializableLedgerKeyAgentData;`

packages/wallet/src/KeyManagement/util/deviceConnection.ts
```diff
@@ -0,0 +1,51 @@
+import { CommunicationType, DeviceType } from '../types';
```
mkazlauskas, line 1: I think it would be better to convert LedgerKeyAgent to a 'directory module' and put related utils under that directory

packages/wallet/src/KeyManagement/errors/GenericError.ts
```diff
@@ -0,0 +1,19 @@
+import { CustomError } from 'ts-custom-error';
+
+export enum GenericErrorType {
+  NO_METHOD = 'NO_METHOD'
+}
+
+export class GenericError extends CustomError {
+  getMessage(errorType: GenericErrorType): string {
+    if (errorType === GenericErrorType.NO_METHOD) {
+      return 'Method not implemented';
```
mkazlauskas, line 10: There is already a "NotImplementedError" in `core` package, please use that instead

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  async getExtendedAccountPublicKey(): Promise<Cardano.Bip32PublicKey> {
+    await this.checkDeviceConnection();
+    const derivationPath = `1852'/1815'/${this.#accountIndex}'`;
+    const extendedPublicKey = await this.#deviceConnection.getExtendedPublicKey({
+      path: utils.str_to_path(derivationPath) // BIP32Path
+    });
+    const xPubHex = `${extendedPublicKey.publicKeyHex}${extendedPublicKey.chainCodeHex}`;
+    const xPub = Cardano.Bip32PublicKey(xPubHex);
+    this.#extendedAccountPublicKey = xPub;
```
mkazlauskas, line 112: I think it would be better to have an async static function for initially creating/pairing the device key agent and setting `this.#extendedAccountPublicKey` in there.

mkazlauskas (changes requested):
Some missing parts:
- [restoreKeyAgent](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/wallet/src/KeyManagement/restoreKeyAgent.ts#L32) support for LedgerKeyAgent
- Some test coverage: at a bare minimum we need to have a test that would demonstrate basic usage of LedgerKeyAgent. If implementing nodejs support is complicated, creating a small test web app could work too.

packages/cardano-graphql/package.json
```diff
@@ -36,6 +36,7 @@
     "@graphql-codegen/typescript-graphql-request": "~4.3.4",
     "@graphql-codegen/typescript-operations": "~2.2.4",
     "@graphql-tools/utils": "~8.6.1",
+    "@types/validator": "^13.7.1",
```
rhyslbw, line 39: Seems like this was added in the wrong package?

.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 14.17.3
```
rhyslbw, line 1: We're using nvm to manage the Node.js version, so this is a duplication of functionality that needs to be justified, and at least synchronised with the other [config file](https://github.com/input-output-hk/cardano-js-sdk/blob/2d92462a8809ac6deca837a70425a53873130235/.nvmrc#L1)

packages/wallet/src/KeyManagement/types.ts
```diff
@@ -72,7 +89,7 @@ export type GetPassword = (noCache?: true) => Promise<Uint8Array>;
 export interface KeyAgent {
```
rhyslbw, line 89:
```suggestion
export interface KeyAgent<SerializableKeyAgentData> {
```
Re: https://github.com/input-output-hk/cardano-js-sdk/pull/189/files#r825305987
How about keeping the implementation concerns out of this scope, and use a generic type instead?

packages/wallet/src/KeyManagement/util/deviceConnection.ts
rhyslbw: plus a `types.ts` to keep the existing file focused on the abstract base

packages/wallet/test/KeyManagement/InMemoryLedgerKeyAgent.test.ts
```diff
@@ -0,0 +1,27 @@
+// TODO - Used just for fast checks
+/* import { KeyManagement } from '../../src';
+import { TransportType } from '../../src/KeyManagement/InMemoryLedgerKeyAgent';
+
+describe('InMemoryLedgerKeyAgent', () => {
```
rhyslbw, line 5:
```suggestion
describe('LedgerKeyAgent', () => {
```

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -0,0 +1,128 @@
...
+  deviceConnection: AppAda;
+  deviceCommunicationType: DeviceCommunicationType;
+}
+
+export enum TransportType {
+  WebHid = 'webhid',
+  NodeHid = 'nodehid'
+}
+
+// TODO - extend KeyAgentBase
```
rhyslbw, line 21:
```suggestion
```

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -0,0 +1,128 @@
...
+  networkId: Cardano.NetworkId;
+  accountIndex: number;
+  knownAddresses: GroupedAddress[];
+  deviceConnection: AppAda;
+  deviceCommunicationType: DeviceCommunicationType;
+}
+
+export enum TransportType {
+  WebHid = 'webhid',
+  NodeHid = 'nodehid'
```
rhyslbw, line 18:
```suggestion
```

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  get networkId(): Cardano.NetworkId {
+    return this.#networkId;
+  }
+
+  get accountIndex(): number {
+    return this.#accountIndex;
+  }
+
+  get deviceConnection(): AppAda {
```
rhyslbw, line 53: :warning:

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -0,0 +1,128 @@
...
+      accountIndex: this.#accountIndex,
+      extendedAccountPublicKey: this.#extendedAccountPublicKey,
+      knownAddresses: this.#knownAddresses,
+      networkId: this.networkId
+    };
+  }
+
+  // TODO - enable node based apps to establish communication using node-hid
+  // TODO - enable bluetooth communication
+  async checkDeviceConnection(): Promise<AppAda> {
```
rhyslbw, line 81: :warning:

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -0,0 +1,128 @@
...
+      if (error.name === 'DisconnectedDeviceDuringOperation') {
+        const deviceConenction = await establishDeviceConnection(this.#deviceCommunicationType);
+        this.#deviceConnection = deviceConenction;
+        return deviceConenction;
+      }
+      throw error;
+    }
+  }
+
+  async getAppVersion(): Promise<GetVersionResponse> {
```
rhyslbw, line 99: :warning:

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  async signBlob(): Promise<SignBlobResult> {
+    throw new GenericError(GenericErrorType.NO_METHOD);
+  }
+
+  async derivePublicKey(): Promise<Cardano.Ed25519PublicKey> {
+    throw new GenericError(GenericErrorType.NO_METHOD);
+  }
+
+  // TODO - this should be moved / removed from KeyAgentBase. Root private key is stored on device in this case
```
rhyslbw, line 124: Probably better to leave it on the base, and throw a specific error when this operation is not supported. This way the application just has to handle the error condition rather checking for the method.

rhyslbw (changes requested):
The goal of the `KeyAgent` is to unify the interface, so that application logic only requires minimal implementation-specific code, usually related to configuration. Try to refactor `LedgerKeyAgent` class to strictly implement the abstract behaviour of `KeyAgentBase`, with any helper-type methods converted to `static`. The utils seem to have this logic that I expect to see in the key agent class, making it a requirement to return objects such as `adaApp`, and therefore requires very different application logic to orchestrate. If the current base class does not express the behaviour to cover hardware wallets, consider a generic abstract method that can be also resolved by the in-memory implementation. I've marked methods for review as :warning: 

We do need an integration test suite here to run locally, which will need to be separately categorised as `hardware`? and excluded from the other test runs. The npm scripts should be extended to target this suite. 

There should also not be merge commits here from _master_, instead your branch rebased on the latest master state, and a force-push to the remote branch.

packages/wallet/src/KeyManagement/types.ts
mkazlauskas: I've implemented it like that originally, but it just makes it more complicated to use in the end. The goal here is for the user (e.g. the wallet) be able to use the key agent without knowing the specific type of it (and it's serializable data). [restoreKeyAgent](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/wallet/src/KeyManagement/restoreKeyAgent.ts#L32) is a utility that abstracts the specifics of serializable data and wallet initialization. It makes it possible for the user to work with just a `KeyAgent`.

packages/wallet/src/KeyManagement/types.ts
```diff
@@ -32,6 +33,16 @@ export enum AddressType {
   External = 0
 }
 
+export enum DeviceType {
+  Ledger = 'Ledger',
+  Trezor = 'Trezor'
```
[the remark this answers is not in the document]
tomislavhoracek: 🆗

packages/wallet/package.json
```diff
@@ -27,12 +27,16 @@
   },
   "devDependencies": {
     "@cardano-sdk/util-dev": " 0.2.0",
+    "@types/node-hid": "1.3.1",
```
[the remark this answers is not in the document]
tomislavhoracek: ouu. I removed node-hid but not types declaration. Thanks

packages/cardano-graphql/package.json
tomislavhoracek: Actually, I got multiple `validator`errors about missing package while building the app. That is the reason I added / fixed this.

.tool-versions
tomislavhoracek: 🆗

packages/wallet/test/KeyManagement/InMemoryLedgerKeyAgent.test.ts
tomislavhoracek: I will completely remove this file since is not relevant anymore

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
tomislavhoracek: I added this and decided to leave it because we have a guard against unsupported type. But, I can remove it if you think that is a better idea

packages/cardano-graphql/package.json
rhyslbw: Best to then commit this separately as a `fix(cardano-graphql)`

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
rhyslbw: I say drop this until it's implemented, since this is a model of our supported transport types.

packages/cardano-graphql/package.json
tomislavhoracek: @rhyslbw I checked this one more time and without `@types/validator` I'm getting this:
Screenshot 2022-03-14 at 19 37 54

tomislavhoracek:
@rhyslbw @mkazlauskas Thanks guys for the review. 
Code is improved and NodeHid support + tests added!

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -0,0 +1,217 @@
...
+  CommunicationType,
+  GroupedAddress,
+  KeyAgentType,
+  SerializableLedgerKeyAgentData,
+  SignBlobResult,
+  TransportType
+} from './types';
+import { KeyAgentBase } from './KeyAgentBase';
+import { TransportError } from './errors';
+import AppAda, { GetVersionResponse, utils } from '@cardano-foundation/ledgerjs-hw-app-cardano';
```
rhyslbw, line 12:
```suggestion
import DeviceConnection, { GetVersionResponse, utils } from '@cardano-foundation/ledgerjs-hw-app-cardano';
```
All other properties are called `deviceConnection`, so why not use this as the import too?

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  static async createDeviceConnection(activeTransport: Transport): Promise<AppAda> {
+    const deviceConnection = new AppAda(activeTransport);
+    // Perform app check to see if device can respond
+    await deviceConnection.getVersion();
+    return deviceConnection;
+  }
+
+  static async establishDeviceConnection(communicationType: CommunicationType, devicePath?: string): Promise<AppAda> {
```
rhyslbw, line 120: :warning: 
With `KeyAgentBase.initialize()`, establishing the device connection could be be done after creating the device connection non-statically.

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -0,0 +1,217 @@
...
+      // If transport is established we need to close it so we can recover device from previous session
+      if (transport) {
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+        transport.close();
+      }
+      throw error;
+    }
+  }
+
+  static async checkDeviceConnection(communicationType: CommunicationType, deviceConnection?: AppAda): Promise<AppAda> {
```
rhyslbw, line 145: :question: 
1. What triggers the application to use this method?
2. Is it defensive, or as a result on an error?
3. Can disconnection not be resolved by simply re-creating the class instance without this added logic?

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -0,0 +1,217 @@
...
+    } catch (error) {
+      // Device disconnected -> re-establish connection
+      if (error.name === 'DisconnectedDeviceDuringOperation') {
+        return await LedgerKeyAgent.establishDeviceConnection(communicationType);
+      }
+      throw error;
+    }
+  }
+
+  static async getXpub({
```
rhyslbw, line 161: :warning:
`KeyAgentBase` already has an abstract method `getExtendedAccountPublicKey`, so with the other changes to internalise the initialisation, it could be implemented as a non-static method.

packages/wallet/src/KeyManagement/restoreKeyAgent.ts
```diff
@@ -46,8 +49,18 @@ export async function restoreKeyAgent<T extends SerializableKeyAgentData>(
...
+        accountIndex: data.accountIndex,
+        communicationType: data.communicationType,
+        extendedAccountPublicKey: data.extendedAccountPublicKey,
+        knownAddresses: data.knownAddresses,
+        networkId: data.networkId
+      });
+    }
     default:
       throw new InvalidSerializableDataError(
+        // @ts-ignore
```
rhyslbw, line 63: What was `tsc` complaining about here?

packages/wallet/src/KeyManagement/types.ts
```diff
@@ -62,7 +74,16 @@ export interface SerializableInMemoryKeyAgentData extends SerializableKeyAgentDa
   knownAddresses: GroupedAddress[];
 }
 
-export type SerializableKeyAgentData = SerializableInMemoryKeyAgentData;
+export interface SerializableLedgerKeyAgentData extends SerializableKeyAgentDataBase {
+  __typename: KeyAgentType.Ledger;
+  knownAddresses: GroupedAddress[];
```
rhyslbw, line 79: This can be hoisted into `SerializableKeyAgentDataBase`

rhyslbw (changes requested):

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
mkazlauskas: This should be private static then?

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -0,0 +1,217 @@
...
+  }: CreateTransportProps): Promise<TransportType> {
+    if (communicationType === CommunicationType.Node) {
+      return await TransportNodeHid.open(devicePath);
+    }
+    return await (activeTransport && activeTransport instanceof TransportWebHID
+      ? TransportWebHID.open(activeTransport.device)
+      : TransportWebHID.request());
+  }
+
+  static async createDeviceConnection(activeTransport: Transport): Promise<AppAda> {
```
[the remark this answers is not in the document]
mkazlauskas: Key agents would need to have different arguments then. Also we don't necessarily want to create a connection when creating the key agent - a connection is only needed for 1st load (getting xpub key) and signing.

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
mkazlauskas: I think most of those static functions should be private

packages/wallet/src/KeyManagement/types.ts
mkazlauskas: Yeah, I'd say both `extendedAccountPublicKey` and `knownAddresses` should be on base, as they both will be also used by InMemoryKeyAgent.

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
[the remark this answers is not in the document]
rhyslbw: > need to have different arguments

Sure, but this can be typed generically.

> Also we don't necessarily want to create a connection when creating the key agent

Creating the key agent happens in the constructor, `initialize` keeps this separate. Is there another step I'm missing?

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
rhyslbw: It's still exposing internals of hardware device management, and I'd like to understand if we need to do that instead of just using abstract concepts and handling disconnections as an error state.

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
rhyslbw: Isn't this method implementing the intent of `getExtendedAccountPublicKey`?

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  static async getXpub({
+    deviceConnection,
+    communicationType,
+    accountIndex
+  }: GetXpubProps): Promise<Cardano.Bip32PublicKey> {
+    const recoveredDeviceConnection = await LedgerKeyAgent.checkDeviceConnection(communicationType, deviceConnection);
+    const derivationPath = `1852'/1815'/${accountIndex}'`;
+    const extendedPublicKey = await recoveredDeviceConnection.getExtendedPublicKey({
```
mkazlauskas, line 125: Should probably throw AuthenticationError if user refuses/cancels the operation on the device

mkazlauskas (changes requested):

mkazlauskas (approved):
Nice work! :tada: @rhyslbw I'm not sure if you reviewed latest revision so I'll leave this for you to merge

packages/cip2/src/types.ts
```diff
@@ -99,8 +99,6 @@ export type ProtocolParametersForInputSelection = Pick<
   'coinsPerUtxoWord' | 'maxTxSize' | 'maxValueSize' | 'minFeeCoefficient' | 'minFeeConstant'
 >;
 
-export type ProtocolParametersRequiredByInputSelection = Required<
```
rhyslbw, line 102: Just be careful of small stylistic changes in future, possibly as a result of your IDE, as any lint errors should be enforced in CI.

tomislavhoracek:
@rhyslbw actually `yarn lint:fix` changed that.

rhyslbw (approved):
Great work @tomislavhoracek!

packages/cip2/src/types.ts
```diff
@@ -99,8 +99,6 @@ export type ProtocolParametersForInputSelection = Pick<
...
-  {
-    [k in keyof ProtocolParametersForInputSelection]: ProtocolParametersForInputSelection[k];
-  }
->;
+export type ProtocolParametersRequiredByInputSelection = Required<{
```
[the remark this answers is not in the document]
rhyslbw: > actually yarn lint:fix changed that.

Hmm, so looks like it's not totally enforced in CI then.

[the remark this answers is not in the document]
mkazlauskas: It's probably because it doesn't fail on warnings. We could set [--max-warnings 0](https://eslint.org/docs/user-guide/command-line-interface#--max-warnings)

PR #190, opened 2022-02-24
fix: recompute fee after selecting extra utxo due to min utxo value

mkazlauskas:
# Context

Transactions occasionally fail due to fee too low. This is due to a bug within input selection, where it does not recompute the fee after selecting an extra utxo.

# Proposed Solution

Reproduce the bug in a test and fix it.

- Add support for custom random function for round robin random improve
- Update mock selection constraints fee computation to be a function of selected # of inputs instead of a constant

### DIFF
--- a/packages/cip2/src/RoundRobinRandomImprove/change.ts
+++ b/packages/cip2/src/RoundRobinRandomImprove/change.ts
@@ -14,6 +14,7 @@ interface ChangeComputationArgs {
   estimateTxFee: EstimateTxFeeWithOriginalOutputs;
   computeMinimumCoinQuantity: ComputeMinimumCoinQuantity;
   tokenBundleSizeExceedsLimit: TokenBundleSizeExceedsLimit;
+  random: typeof Math.random;
 }
 
 interface ChangeComputationResult {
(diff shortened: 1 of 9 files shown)


packages/cip2/src/RoundRobinRandomImprove/change.ts
```diff
@@ -304,53 +287,69 @@ export const computeChangeAndAdjustForFee = async ({
...
+        estimateTxFee,
+        implicitCoin,
+        outputValues,
+        random,
+        tokenBundleSizeExceedsLimit,
+        uniqueOutputAssetIDs,
+        utxoSelection: pickExtraRandomUtxo(currentUtxoSelection, random)
+      });
+    }
+    // This is not a great error type for this, because the spec says
```
rhyslbw, line 303: So we cannot improve the error type because we're constrained by the spec?

rhyslbw (approved):

packages/cip2/test/RoundRobinRandomImprove.test.ts
```diff
@@ -56,6 +56,26 @@ describe('RoundRobinRandomImprove', () => {
         });
         expect(results.selection.inputs.size).toBe(1);
       });
+      it('Recomputes fee after selecting an extra utxo due to change not meeting minimumCoinQuantity', async () => {
+        const utxo = new Set([
```
james-iohk, line 60: Could include a 4th utxo to check that no additional non-essential inputs are included. E.g.
```typescript
const utxo = new Set([
  TxTestUtil.createUnspentTxOutput({ coins: 1_000_000n }),
  TxTestUtil.createUnspentTxOutput({ coins: 1_000_000n }),
  TxTestUtil.createUnspentTxOutput({ coins: 1_000_000n }),
  TxTestUtil.createUnspentTxOutput({ coins: 975_000_000n })
]);
```
Then you also wouldn't need to worry about the order utxos are considered so you could just use:
`const random = jest.fn().mockReturnValue(0);`

james-iohk: Actually, changing the `random` would result in a slightly different selection process (2 utxos secluded at roundrobin instead of 1), although the result is the same.

mkazlauskas: Added a 4th utxo and a comment explaining how the test works

james-iohk (approved):

PR #192, opened 2022-03-02
Feature/cip30

mchappell:
# Context

Aligning the current sdk cip30 package with the updated cip30 specification

## Important changes introduced

- Implement `cslToCore` transaction conversion
- (BREAKING) Update some core types:
  - Add `Cardano.RedeemerPurpose` and `Cardano.MirCertificatePot` enums
  - Make StakeDelegationCertificate.epoch and PoolRegistrationCertificate.epoch optional
  - Remove PoolRegistrationCertificate.poolId 
  - Made address nullable in txIn model
  - set tx body scriptIntegrityHash to correct length (32 byte)
  - set the correct Ed25519KeyHash length (28 bytes)


packages/e2e/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "description": "e2e tests in other packages",
+  "engines": {
+    "node": "^14"
+  },
+  "main": "dist/index.js",
+  "repository": "https://github.com/input-output-hk/cardano-js-sdk/packages/e2e",
+  "author": "Michael Chappell <martynas.kazlauskas@iohk.io>",
+  "license": "MPL-2.0",
+  "scripts": {
+    "test": "ENVFILE=.env yarn test:chrome",
```
rhyslbw, line 13: In order to run the workspace package.json scripts we need to include:

```
    "build": "shx echo 'build' command not implemented",
    "coverage": "shx echo No coverage report for this package",
    "cleanup": "shx rm -rf node_modules",
    "test:e2e": "yarn test && yarn test:chrome",
    "tscNoEmit": "shx echo typescript --noEmit command not implemented yet",
 ```
 ... plus the devDependency: `"shx": "^0.3.3"`

File: packages/e2e/test/cip30.test.ts [hunk 2]
```diff
@@ -0,0 +1,128 @@
...
+ * @jest-environment jsdom
+ */
+
+//import { testnetTimeSettings } from '@cardano-sdk/core';
+//import { createStubStakePoolSearchProvider, createStubTimeSettingsProvider } from '@cardano-sdk/util-dev';
+//import { createWalletApi, injectWindow, WalletApi, Wallet } from '../../cip30';
+import { handleMessages, injectWindow, Wallet as BigWallet } from '../../cip30/src';
+//import { createWalletPublicApi, WalletPublicApi } from '../../cip30';
+import { RequestAccess, WalletApi, WalletProperties, createUiWallet } from '../../cip30/src';
+import { firstValueFrom } from '../../wallet/dist/util';
```
rhyslbw, line 11:
```suggestion
import { firstValueFrom } from '@cardano-sdk/wallet/util';
```

packages/e2e/test/cip30.test.ts
```diff
@@ -0,0 +1,128 @@
+/**
+ * @jest-environment jsdom
+ */
+
+//import { testnetTimeSettings } from '@cardano-sdk/core';
+//import { createStubStakePoolSearchProvider, createStubTimeSettingsProvider } from '@cardano-sdk/util-dev';
+//import { createWalletApi, injectWindow, WalletApi, Wallet } from '../../cip30';
+import { handleMessages, injectWindow, Wallet as BigWallet } from '../../cip30/src';
```
rhyslbw, line 8:
```suggestion
import { handleMessages, injectWindow, Wallet as BigWallet } from '@cardano-sdk/cip30';
```

Setup a `tsconfig.json` like here: https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/wallet/test/tsconfig.json to establish the TypeScript dependencies. This will build the dependent code when needed.

rhyslbw (changes requested):

packages/cip30/src/Wallet/types.ts
```diff
@@ -1,6 +1,16 @@
...
 import { Cardano } from '@cardano-sdk/core';
 
+/**
+ * Returns the network id of the currently connected account.
+ * 0 is testnet and 1 is mainnet but other networks can possibly be returned by wallets.
+ * Those other network ID values are not governed by this document.
+ *
+ * This result will stay the same unless the connected account has changed.
+ *
+ * Errors: `ApiError`
```
mkazlauskas, line 11: `@throws` would look better in the generated docs

packages/e2e/README.md
```diff
@@ -0,0 +1,9 @@
+# Cardano JS SDK | util-dev
```
mkazlauskas, line 1: `util-dev` -> `e2e`

mkazlauskas (changes requested):
I'm not sure about this new `e2e` package. Maybe it would be better to extract reusable code to `util-dev` and have e2e tests under `cip30` package, @rhyslbw?

rhyslbw:
> have e2e tests under cip30 package

Hmm, yes now you mention it @mkazlauskas, there's no harm in having a dev dependency on `wallet` from within `cip30`, so let's take this approach instead @mchappell.

packages/cip30/src/Wallet/Wallet.ts
```diff
@@ -24,67 +31,88 @@ export type RequestAccess = () => Promise<boolean>;
...
+      icon: this.icon,
+      isEnabled: this.isEnabled.bind(this, hostname),
+      name: this.name
     };
   }
 
-  private getAllowList(): string[] {
-    // JSON.parse(null) seems to be legit
-    return JSON.parse(this.options.storage.getItem(this.name)!) || [];
+  static async getAllowList(_storage?: Storage.LocalStorageArea): Promise<string[]> {
```
mkazlauskas, line 81: I think it's better to keep this non-static: there can be multiple wallets running.

packages/cip30/src/Wallet/Wallet.ts
```diff
@@ -24,67 +31,88 @@ export type RequestAccess = () => Promise<boolean>;
...
+    if (!_storage) return Promise.resolve([]);
+    try {
+      const persistedStorage: Record<string, WalletStorage> = await _storage.get(this.name);
```
mkazlauskas, line 84: `this.name` inside the static function refers to function name (not the name property declared by this class) which is probably not what we want here.

mkazlauskas (changes requested):
Please resolve lint warnings

packages/cip30/src/Wallet/Wallet.ts
```diff
@@ -24,67 +31,88 @@ export type RequestAccess = () => Promise<boolean>;
...
-    this.version = properties.version;
-    this.allowList = this.options.persistAllowList ? this.getAllowList() : [];
+    this.apiVersion = properties.apiVersion;
+    this.icon = properties.icon;
+    this.allowList = allowList;
     this.logger = this.options.logger;
   }
 
-  public getPublicApi(window: WindowMaybeWithCardano) {
+  public getPublicApi(hostname: string) {
```
rhyslbw, line 71: If the private members are designated, this method is probably redundant, as we can just use the class instance as a whole.

rhyslbw (changes requested):

james-iohk:
@mchappell There is no test coverage for error cases. Are they in scope?

james-iohk:
We could include test coverage for optional args such as `GetUtxos = (amount?: Cbor, paginate?: Paginate)`

rhyslbw:
@james-iohk There will be e2e tests in the `wallet` package since there will be a mapping utility sourced there to translate the CIP30 interface to our wallet interface, so this is where we can expect the error cases to be covered. It does raise the issue of not having coverage in this isolated context, so we should probably consider a set of tests using a mocked wallet here, but the best bang for buck assurance we'll get is from the e2e tests to come.

packages/cip30/src/Wallet/types.ts
```diff
@@ -31,14 +42,15 @@ export type GetBalance = () => Promise<Cbor>;
  *
  * The results can be further paginated by `paginate` if it is not `undefined`.
  *
- * Errors: `ApiError` | `PaginateError`
+ * @throws ApiError
+ * @throws PaginateError
```
james-iohk, line 46: The [spec states](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0030#apigetusedaddressespaginate-paginate--undefined-promisecboraddress) that `GetUsedAddresses` doesn't throw `PaginateError`, only `GetUtxos` does. But it is probably wrong and we should correct it.

james-iohk (approved):

packages/cip30/src/Wallet/Wallet.ts
```diff
@@ -24,67 +31,66 @@ export type RequestAccess = () => Promise<boolean>;
...
-  public getPublicApi(window: WindowMaybeWithCardano) {
-    return {
-      enable: this.enable.bind(this, window),
-      isEnabled: this.isEnabled.bind(this, window),
-      name: this.name,
-      version: this.version
-    };
+    this.#api = api;
+    this.#options = { ...defaultOptions, ...options };
+    this.#logger = options?.logger || this.#options.logger;
```
mkazlauskas, line 67: I think this can be simplified to just `this.#logger = this.#options.logger`

packages/cip30/src/Wallet/Wallet.ts
```diff
@@ -24,67 +31,66 @@ export type RequestAccess = () => Promise<boolean>;
...
 
-  private getAllowList(): string[] {
-    // JSON.parse(null) seems to be legit
-    return JSON.parse(this.options.storage.getItem(this.name)!) || [];
+  async #getAllowList(storageKey: string): Promise<string[]> {
+    if (!storageKey) return Promise.resolve([]);
+    try {
+      const persistedStorage: Record<string, WalletStorage> = await this.#options.storage.get(storageKey);
+      return persistedStorage[storageKey].allowList;
+    } catch {
```
mkazlauskas, line 76: Should probably log the error with `this.#logger`

packages/cip30/src/Wallet/Wallet.ts
```diff
@@ -24,67 +31,66 @@ export type RequestAccess = () => Promise<boolean>;
...
-        {
-          allowList: this.getAllowList(),
-          module: 'Wallet',
-          walletName: this.name
-        },
-        'Allow list persisted'
-      );
-    }
+  async #allowApplication(appName: string) {
+    this.#allowList.push(appName);
```
mkazlauskas, line 82: I think `this.#allowList` could potentially be undefined here. Since it's only used in async methods, maybe it would be better to not directly access this, but instead have an async method or a getter to initialize it if undefined and then return it (similar to code in `isEnabled`).

mkazlauskas (changes requested):
Just a few minor issues

packages/wallet/src/util.ts
```diff
@@ -1 +1,136 @@
...
+              (sumCalculation, currentUtxoValue) => currentUtxoValue[1].value.coins + sumCalculation,
+              0n
+            );
+            if (currentSum < BigInt(amount)) return [...utxos, currentUtxo];
+            return utxos;
+          }, [] as Cardano.Utxo[]);
+        logger.info('filteredByAmount', filteredByAmount);
+      }
+
+      // TODO: determine how to paginate effectively
```
mkazlauskas, line 106: You can use rxjs `skip(n)` and `take(n)` operators before converting it to promise, e.g.
```
await firstValueFrom(wallet.utxo.available$.pipe(skip(0), take(5));
```
You should also apply the same technique converting from core types to cbor through `coreToCsl` util

packages/wallet/test/e2e/SingleAddressWallet/cip30mapping.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    xtest('api.getUtxos', async () => {
+      const cip30WindowUtxos = await api.getUtxos();
+      const singleAddressWalletUtxos = await firstValueFrom(wallet.utxo.available$);
+      expect(cip30WindowUtxos).toEqual(singleAddressWalletUtxos);
+    });
+
+    test.skip('api.getBalance', async () => {
+      // const cip30Balance = await api.getBalance();
+      // const walletBalance = wallet.balance.total$.value;
+      // expect(cip30Balance).toEqual(walletBalance);
```
mkazlauskas, line 72: Since cip30 defines all types as cbor, tests should probably use `cardano-serialization-lib` and assert it can parse the returned bytes, e.g.
```typescript
import * as CSL from '@emurgo/cardano-serialization-lib-nodejs'; // or wasm, or load dynamically in beforeAll

const balanceCborBytes = Buffer.from(await api.getBalance(), 'hex');
expect(() => CSL.Value.from_bytes(balanceCborBytes)).not.toThrow(); // I think asserting that it does not throw is sufficient
```

packages/wallet/test/e2e/SingleAddressWallet/delegation.test.ts
```diff
@@ -53,7 +53,7 @@ const waitForNewStakePoolIdAfterTx = (wallet: Wallet) =>
     )
   );
 
-describe('SingleAddressWallet/delegation', () => {
+xdescribe('SingleAddressWallet/delegation', () => {
```
mkazlauskas, line 56: What's `xdescribe`?

mkazlauskas (changes requested):

packages/wallet/src/util.ts
```diff
@@ -1 +1,136 @@
...
+      logger.debug('getting networkId');
+      return Promise.resolve(props.networkId);
+    },
+    /* getRewardAddresses: async (): Promise<Cbor[]> => {
+      logger.info('getting reward addresses');
+
+      const rewardAccounts = await firstValueFrom(wallet.delegation.rewardAccounts$);
+      return rewardAccounts.map((a) => cbor.encode(a.address.toString).toString('hex'));
+    },*/
+    getUnusedAddresses: async (): Promise<Cbor[]> => {
```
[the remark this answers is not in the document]
rhyslbw: The single address could be unused too though, so maybe check if there are transactions and return either empty array or the address.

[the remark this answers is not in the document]
mkazlauskas: I think it ultimately depends on intended use case for this function. Always returning an empty array could give the dapp a hint that it's a single address wallet.

packages/wallet/src/util.ts
mkazlauskas: Actually that won't work, as emitted value is an array. You should instead do something like `wallet.utxo.available$.pipe(map(utxo => utxo.slice(...)));`

packages/wallet/test/e2e/SingleAddressWallet/delegation.test.ts
james-iohk: I think it disables the test

packages/wallet/src/util.ts
```diff
@@ -1 +1,158 @@
...
+      } catch (error) {
+        logger.error(error);
+        throw error;
+      }
+    },
+    getChangeAddress: async (): Promise<Cbor> => {
+      logger.debug('getting changeAddress');
+      try {
+        const [{ address }] = await firstValueFrom(wallet.addresses$);
+        const parsedAddress = parseCslAddress(address as unknown as string);
```
mkazlauskas, line 33: If this is `Cardano.Address` you can do `address.toString()`

packages/wallet/src/util.ts
```diff
@@ -1 +1,158 @@
...
+import { CSL, Cardano, coreToCsl, cslToCore, parseCslAddress } from '@cardano-sdk/core';
+import { KeyAgent } from './KeyManagement';
+import { Logger, dummyLogger } from 'ts-log';
+import { SingleAddressWallet } from '.';
+import { cip30signData } from './KeyManagement/cip8';
+import { firstValueFrom } from 'rxjs';
+import cbor from 'cbor';
+
+type Props = {
+  networkId: Cardano.NetworkId;
```
mkazlauskas, line 11: This is probably redundant: `networkId` is available on KeyAgent

packages/wallet/src/util.ts
```diff
@@ -1 +1,158 @@
...
+    },
+    getUnusedAddresses: async (): Promise<Cbor[]> => {
+      logger.debug('getting unused addresses');
+      return Promise.resolve([]);
+    },
+    getUsedAddresses: async (_paginate?: Paginate): Promise<Cbor[]> => {
+      logger.debug('getting changeAddress');
+
+      const [{ address }] = await firstValueFrom(wallet.addresses$);
+      const parsedAddress = parseCslAddress(address as unknown as string);
```
mkazlauskas, line 74: `toString()`

packages/wallet/src/util.ts
```diff
@@ -1 +1,158 @@
...
+        throw new ApiError(500, 'could not get used addresses');
+      } else {
+        return [Buffer.from(parsedAddress.to_bytes()).toString('hex')];
+      }
+    },
+    getUtxos: async (amount?: Cbor, paginate?: Paginate): Promise<Cardano.Utxo[] | undefined> => {
+      const utxos = await firstValueFrom(wallet.utxo.available$);
+
+      // eslint-disable-next-line unicorn/no-useless-undefined
+      if (!utxos) return Promise.resolve(undefined);
```
mkazlauskas, line 86: This should never be falsy

packages/wallet/src/util.ts
```diff
@@ -1 +1,158 @@
...
+      const cslUtxos = coreToCsl.utxo(utxos);
```
mkazlauskas, line 87: CSL types are difficult to work with. I suggest you do the filtering and pagination before converting to CSL (convert only as you are returning them)

packages/wallet/src/util.ts
```diff
@@ -1 +1,158 @@
...
+
+      if (amount) {
+        try {
+          const filterAmount = CSL.Value.from_bytes(Buffer.from(amount, 'hex'));
+          const filteredUtxosByAmount = cslUtxos.filter(
```
mkazlauskas, line 92: An easy way to get total value of utxo is `Cardano.util.coalesceValueQuantities(utxo.map(([_, txOut]) => txOut.value))`

packages/wallet/src/util.ts
```diff
@@ -1 +1,158 @@
...
+      if (paginate) {
+        return Promise.resolve(
+          cslUtxos.slice(
+            paginate.page * paginate.limit,
+            paginate.page * paginate.limit + paginate.limit
+          ) as unknown as Cardano.Utxo[]
+        );
+      }
+
+      return Promise.resolve(cslUtxos as unknown as Cardano.Utxo[]);
```
mkazlauskas, line 112: This seems off. the return type should probably be `string[]`? `cslUtxos.map(cslUtxo => Buffer.from(cslUtxo.to_bytes()).toString('hex')`

packages/wallet/src/util.ts
```diff
@@ -1 +1,158 @@
...
+    },
+    signData: async (addr: Cbor, sigStructure: string): Promise<Bytes> => {
+      logger.debug('signData', sigStructure);
+      const signWith = cbor.decode(addr, { encoding: 'hex' });
```
mkazlauskas, line 116: Latest spec says that `Address` is bech32 string

packages/wallet/src/util.ts
```diff
@@ -1 +1,158 @@
...
+        throw new ApiError(400, 'could not sign data');
+      }
+
+      return Promise.resolve(signature);
+    },
+    signTx: async (tx: Cbor, _partialSign?: Boolean): Promise<Cbor> => {
+      logger.debug('signTx', tx);
+      try {
+        const signedTx = await wallet.finalizeTx(cbor.decode(tx));
+        return Promise.resolve(Buffer.from(cbor.encode([signedTx.witness.signatures])).toString('hex'));
```
mkazlauskas, line 134: It's probably a better idea to use `KeyAgent.signTransaction` directly as we don't seem to need other witness properties here (only signatures)

packages/wallet/src/util.ts
```diff
@@ -1 +1,158 @@
...
+        throw new TxSignError(1, error);
+      }
+    },
+    submitTx: async (tx: Cbor): Promise<string> => {
+      logger.debug('submitting tx', tx);
+      try {
+        const txDecoded = CSL.Transaction.from_bytes(Buffer.from(tx, 'hex'));
+        const txData: Cardano.NewTxAlonzo = cslToCore.tx(txDecoded);
+        await wallet.submitTx(txData);
+        return Promise.resolve(txData.id as unknown as string);
```
mkazlauskas, line 146: `toString()` here as well. All core `OpaqueString` types have toString()

mkazlauskas (changes requested):

packages/wallet/src/util.ts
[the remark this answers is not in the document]
rhyslbw: Yeah. Small issue either way

packages/cip30/test/Wallet.test.ts
```diff
@@ -26,45 +27,61 @@ describe('Wallet', () => {
...
+    const api = await wallet.enable(windowStub.location.hostname);
     expect(typeof api).toBe('object');
     const methods = Object.keys(api);
     expect(methods).toEqual(apiMethods);
-    expect(await wallet.isEnabled(windowStub)).toBe(true);
+    expect(await wallet.isEnabled(windowStub.location.hostname)).toBe(true);
+  });
+
+  test('prior enabling should persist', async () => {
+    await browser.storage.local.set({ myWallet: { allowList: 'test.invalid' } });
```
rhyslbw, line 61: Where does `test.invalid` come from?

packages/wallet/src/util.ts
```diff
@@ -1 +1,143 @@
...
+        return [Buffer.from(parsedAddress.to_bytes()).toString('hex')];
+      }
+    },
+    getUtxos: async (amount?: Cbor, paginate?: Paginate): Promise<Cardano.Utxo[] | undefined> => {
+      let utxos = await firstValueFrom(wallet.utxo.available$);
+
+      if (amount) {
+        try {
+          const filterAmount = CSL.Value.from_bytes(Buffer.from(amount, 'hex'));
+          const { inputSelection } = await wallet.initializeTx({
```
rhyslbw, line 87: Please add a comment to explain why we're initializing a transaction here, since it's not really self-explanatory.

"Getting UTxOs to meet a required amount is a complex operation, which is handled by input selection capabilities. By initializing a transaction we're able to utilise the internal configuration and algorithm to make this selection, using a wallet address to satisfy the interface only."

rhyslbw (changes requested):
Looking really good @mchappell!

Just need to add more test coverage for the allow list, as it's not clear me if the current implementation is handling the "remember" set appropriately. The scenarios to cover are:

_Following assumes using 2 dApps._
1. Both allowed but not remembered.
2. Both allowed, one remembered
3. One allowed, one disallowed

packages/wallet/customEnvironment.ts
```diff
@@ -0,0 +1,13 @@
+const Environment = require('jest-environment-jsdom');
+
+// Cbor package requires a textEncoder to hav been defined
+module.exports = class CustomTestEnvironment extends Environment {
```
mkazlauskas, line 4: Aren't we using CSL to decode cbor?
If it's really required, I think it's better to put it under `/test` and document required polyfill in README.

packages/wallet/test/e2e/SingleAddressWallet/cip30mapping.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    test('api.getRewardAddresses', async () => {
+      const cipRewardAddresses = await mappedWallet.getRewardAddresses();
+      const walletRewardAddresses = await wallet.addresses$.value?.filter((s) => s.rewardAccount);
+
+      expect(cipRewardAddresses).toEqual(walletRewardAddresses);
+    });
+
+    test('api.signTx', async () => {
+      const txInternals = await createTxInternals(wallet);
+      expect(await mappedWallet.signTx(cbor.encode(txInternals).toString('hex'))).not.toThrow();
```
mkazlauskas, line 118: I don't think this is correct: `Wallet.initializeTx` returns a non-spec-compatible js object. Types defined by cip30 spec should be all supported by CSL and (I think) you can get cbor with (one of) `to_bytes` or `as_bytes`

packages/wallet/test/e2e/SingleAddressWallet/cip30mapping.test.ts
```diff
@@ -0,0 +1,134 @@
...
+      if (!parsedAddress) {
+        throw new Error('No wallet address');
+      }
+
+      expect(cipChangeAddress).toEqual(Buffer.from(parsedAddress.to_bytes()).toString('hex'));
+    });
+
+    test('api.getRewardAddresses', async () => {
+      const cipRewardAddresses = await mappedWallet.getRewardAddresses();
+      const walletRewardAddresses = await wallet.addresses$.value?.filter((s) => s.rewardAccount);
```
mkazlauskas, line 111: Shouldn't this `filter` be `map`?

packages/wallet/test/e2e/SingleAddressWallet/cip30mapping.test.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    test('api.getUnusedAddresses', async () => {
+      const cipUsedAddressess = await mappedWallet.getUnusedAddresses();
+      expect(cipUsedAddressess).toEqual([]);
+    });
+
+    test('api.getChangeAddress', async () => {
+      const cipChangeAddress = await mappedWallet.getChangeAddress();
+      const [{ address }] = await firstValueFrom(wallet.addresses$);
+      const parsedAddress = parseCslAddress(address as unknown as string);
```
mkazlauskas, line 101: Please use `.toString()` to convert all OpaqueString to str

packages/wallet/test/e2e/SingleAddressWallet/cip30mapping.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    test.skip('api.getUtxos', async () => {
+      expect(async () => await mappedWallet.getUtxos()).not.toThrow();
+    });
+
+    test('api.getBalance', async () => {
+      const balanceCborBytes = Buffer.from(await mappedWallet.getBalance(), 'hex');
+      expect(() => CSL.Value.from_bytes(balanceCborBytes)).not.toThrow();
+    });
+
+    test.skip('api.getUsedAddresses', async () => {
```
mkazlauskas, line 84: Why is this skipped? A comment would be nice.
I can see how this test would fail. You could just compare bech32s: `cipUsedAddressess === walletUsedAddresses`

packages/wallet/test/e2e/SingleAddressWallet/cip30mapping.test.ts
```diff
@@ -0,0 +1,134 @@
...
+  });
+
+  describe('API Methods', () => {
+    test('api.getNetworkId', async () => {
+      const cip30NetworkId = await mappedWallet.getNetworkId();
+      expect(cip30NetworkId).toEqual(Number(process.env.NETWORK_ID));
+    });
+
+    test.skip('api.getUtxos', async () => {
+      expect(async () => await mappedWallet.getUtxos()).not.toThrow();
```
mkazlauskas, line 76: Would be great to test if creating CSL objects from mappedWallet response does not throw

mkazlauskas (changes requested):

packages/wallet/customEnvironment.ts
rhyslbw: There may have been early discussion around not using the CSL to stay a bit more generic, but with it we can make better assertions in addition to avoiding this extra test polyfill

mchappell:
> Looking really good @mchappell!
> 
> Just need to add more test coverage for the allow list, as it's not clear me if the current implementation is handling the "remember" set appropriately. The scenarios to cover are:
> 
> _Following assumes using 2 dApps._
> 
> 1. Both allowed but not remembered.
> 2. Both allowed, one remembered
> 3. One allowed, one disallowed

Cases added under injectWindow test

packages/wallet/src/util.ts
```diff
@@ -84,6 +83,11 @@ export const createCip30WalletApiFromWallet = (wallet: SingleAddressWallet, prop
       if (amount) {
         try {
           const filterAmount = CSL.Value.from_bytes(Buffer.from(amount, 'hex'));
+          /**
+           * Getting UTxOs to meet a required amount is a complex operation, which is handled by input selection capabilities.
```
mkazlauskas, line 87: :+1:

packages/wallet/src/util.ts
```diff
@@ -1 +1,151 @@
...
+        }
+      } else if (paginate) {
+        utxos = utxos.slice(paginate.page * paginate.limit, paginate.page * paginate.limit + paginate.limit);
+      }
+
+      return Promise.resolve(utxos);
+    },
+    signData: async (addr: Cardano.Address, payload: Bytes): Promise<Bytes> => {
+      logger.debug('signData');
+      const { signature } = await cip30signData({
```
mkazlauskas, line 108: The return type of this function should be [DataSignature](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0030#datasignature) object, which also includes the key.
```suggestion
      return cip30signData({
```

packages/wallet/src/util.ts
```diff
@@ -1 +1,151 @@
...
+        throw new ApiError(400, 'could not sign data');
+      }
+      return Promise.resolve(Buffer.from(signature).toString('hex'));
+    },
+    signTx: async (tx: Cbor, _partialSign?: Boolean): Promise<Cbor> => {
+      logger.debug('signTx');
+      try {
+        const txDecoded = CSL.Transaction.from_bytes(Buffer.from(tx, 'hex'));
+        const coreTx = cslToCore.tx(txDecoded);
+        const witnessSet = await props.keyAgent.signTransaction({ hash: coreTx.id, body: coreTx.body });
```
mkazlauskas, line 123: A few new lint warnings, one is here

packages/wallet/test/e2e/SingleAddressWallet/cip30mapping.test.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  describe('API Methods', () => {
+    test('api.getNetworkId', async () => {
+      const cip30NetworkId = await mappedWallet.getNetworkId();
+      expect(cip30NetworkId).toEqual(Number(process.env.NETWORK_ID));
+    });
+
+    test('api.getUtxos', async () => {
+      const utxos = await mappedWallet.getUtxos();
+      expect(coreToCsl.utxo(utxos!)).not.toThrow();
```
mkazlauskas, line 76:
```suggestion
      expect(() => coreToCsl.utxo(utxos!)).not.toThrow();
```
Otherwise it's the same as just `coreToCsl.utxo(utxos!)`

packages/wallet/test/e2e/SingleAddressWallet/cip30mapping.test.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    test('api.getChangeAddress', async () => {
+      const cipChangeAddress = await mappedWallet.getChangeAddress();
+      const [{ address }] = await firstValueFrom(wallet.addresses$);
+      const parsedAddress = parseCslAddress(address.toString());
+      if (!parsedAddress) {
+        throw new Error('No wallet address');
+      }
+
+      expect(cipChangeAddress).toEqual(Buffer.from(parsedAddress.to_bytes()).toString('hex'));
```
mkazlauskas, line 103: I think we should go with bech32 format for returning addresses. According to the [CIP](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0030#address), both hex and bech32 are acceptable, but I think hex support is for backwards-compatibility

packages/wallet/test/e2e/SingleAddressWallet/cip30mapping.test.ts
```diff
@@ -0,0 +1,138 @@
...
+
+      expect(cipRewardAddresses).toEqual([Buffer.from(parsedAddress!.to_bytes()).toString('hex')]);
+    });
+
+    test('api.signTx', async () => {
+      const txInternals = await createTxInternals(wallet);
+      const finalizedTx = await wallet.finalizeTx(txInternals);
+
+      const cslTx = coreToCsl.tx(finalizedTx).to_bytes();
+      expect(await mappedWallet.signTx(Buffer.from(cslTx).toString('hex'))).not.toThrow();
```
mkazlauskas, line 119: I think this would be a better test
```suggestion
      const signatures = Buffer.from(await mappedWallet.signTx(Buffer.from(cslTx).toString('hex')), 'hex')
      expect(() => CSLTransactionWitnessSet.from_bytes(signatures)).not.toThrow();
```

mkazlauskas (changes requested):

packages/core/src/CSL/cslToCore.ts
```diff
@@ -81,17 +82,158 @@ export const txInputs = (inputs: CSL.TransactionInputs, address?: Cardano.Addres
...
+  vrfKeyHash: Cardano.util.Hash32ByteBase16(Buffer.from(certificate.vrf_keyhash().to_bytes()).toString())
+});
+
+const mir = (certificate: CSL.MoveInstantaneousRewardsCert): Cardano.MirCertificate => {
+  const potNumber = certificate.move_instantaneous_reward().pot();
+
+  return {
+    __typename: Cardano.CertificateType.MIR,
+    pot: Object.values(Cardano.MirCertificatePot)[potNumber],
+    quantity: '???',
```
mkazlauskas, line 196: Seems this certificate also has `kind()`, so I'm guessing one of those is 'quantity' and the other one is 'rewardAccount'. That would make our type incorrect, as we require both, while CSL suggests they're mutually exclusive. @james-iohk could you please confirm this?

packages/core/src/CSL/cslToCore.ts
```diff
@@ -81,17 +82,158 @@ export const txInputs = (inputs: CSL.TransactionInputs, address?: Cardano.Addres
...
+};
+
+const poolRegistration = (certificate: CSL.PoolRegistration): Cardano.PoolRegistrationCertificate => {
+  const { reward_account, pledge, cost, margin, pool_metadata, pool_owners, relays, vrf_keyhash } =
+    certificate.pool_params();
+  return {
+    __typename: Cardano.CertificateType.PoolRegistration,
+    poolParameters: {
+      cost: BigInt(cost.toString()),
+      id: '???', // Operator?
```
mkazlauskas, line 154: Yes, if I remember correctly, pool id is the operator key hash

packages/core/src/CSL/cslToCore.ts
```diff
@@ -81,17 +82,158 @@ export const txInputs = (inputs: CSL.TransactionInputs, address?: Cardano.Addres
...
+
+const stakeDeregistration = (certificate: CSL.StakeDeregistration): Cardano.StakeAddressCertificate => ({
+  __typename: Cardano.CertificateType.StakeKeyDeregistration,
+  rewardAccount: Cardano.RewardAccount(Buffer.from(certificate.to_bytes()).toString())
+});
+
+const stakeDelegation = (certificate: CSL.StakeDelegation): Cardano.StakeDelegationCertificate => ({
+  __typename: Cardano.CertificateType.StakeDelegation,
+  epoch: certificate.epoch(), // TODO: doesn't exist on this type
+  poolId: Cardano.PoolId(certificate.pool_keyhash().toString()), // TODO: is this correct??
```
mkazlauskas, line 98: Seems to be correct

packages/core/src/CSL/cslToCore.ts
```diff
@@ -81,17 +82,158 @@ export const txInputs = (inputs: CSL.TransactionInputs, address?: Cardano.Addres
...
+});
+
+const stakeDeregistration = (certificate: CSL.StakeDeregistration): Cardano.StakeAddressCertificate => ({
+  __typename: Cardano.CertificateType.StakeKeyDeregistration,
+  rewardAccount: Cardano.RewardAccount(Buffer.from(certificate.to_bytes()).toString())
+});
+
+const stakeDelegation = (certificate: CSL.StakeDelegation): Cardano.StakeDelegationCertificate => ({
+  __typename: Cardano.CertificateType.StakeDelegation,
+  epoch: certificate.epoch(), // TODO: doesn't exist on this type
```
mkazlauskas, line 97: That should be nullable for NewTxAlonzo - need to adjust core types. Perhaps just entirely delete it. I'll look at it and let you know.

mkazlauskas (changes requested):
Looks good! We can scope out the TODOs. We really need to write some tests at this point. `coreToCsl` is somewhat tested and includes some test objects that you can reuse (just put them to `packages/core/test/CSL/util.ts`). You can also rely on coreToCsl working correctly and just assert this: `cslToCore(coreToCsl(obj)).toEqual(obj)`

packages/core/src/CSL/cslToCore.ts
mkazlauskas: Just noticed that `coreToCsl` doesn't support this certificate too. Let's just scope this out to a separate task to implement both conversions.

packages/core/src/CSL/cslToCore.ts
```diff
@@ -143,22 +285,138 @@ export const txBody = (body: CSL.TransactionBody): Cardano.TxBodyAlonzo => {
...
+  const result: Cardano.Redeemer[] = [];
+  for (let j = 0; j < redeemers.len(); j++) {
+    const reedeemer = redeemers.get(j);
+    const index = reedeemer.index();
+    const data = reedeemer.data();
+    const exUnits = reedeemer.ex_units();
+
+    /**
+     * CSL.RedeemerTagKind = Spend, Mint, Cert, Reward
+     * should we modify Cardano.Redeemer.purpose to match or just map reward to withdrawal ??
```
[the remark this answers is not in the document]
rhyslbw: Just map it for now, but make a task to review prior to production release.

packages/core/src/CSL/cslToCore.ts
mchappell: removed, todo added

rhyslbw (changes requested):
I've noticed there's a duplication of the method name array in this [test](https://github.com/input-output-hk/cardano-js-sdk/blob/feature/cip30/packages/cip30/test/Wallet.test.ts). It would be best to hoist [the original](https://github.com/input-output-hk/cardano-js-sdk/blob/f86c345d049b8ee0c7377bd0bc4906f5ba2f95ce/packages/cip30/src/BrowserExtension/uiWallet.ts#L12-L23) to [Wallet](https://github.com/input-output-hk/cardano-js-sdk/blob/feature/cip30/packages/cip30/src/Wallet/Wallet.ts), then import in both contexts.

packages/cip30/src/BrowserExtension/uiWallet.ts
```diff
@@ -1,10 +1,16 @@
...
+import { createMessenger } from './sendMessage';
 
-export const createUiWallet = (logger: Logger = dummyLogger): WalletApi => {
-  const methodNames = [
+export interface CreateUiWalletProps {
+  walletExtensionId?: string;
+  logger?: Logger;
+}
+
+export const createUiWallet = ({ logger = dummyLogger, walletExtensionId }: CreateUiWalletProps = {}): WalletApi => {
```
rhyslbw:
```suggestion
export const createWalletApiMessageClient = ({ logger = dummyLogger, walletExtensionId }: CreateUiWalletProps = {}): WalletApi => {
```

+ rename this file to `walletApiMessageClient.tx`

We need more specificity here to communicate the purpose.

packages/wallet/test/integration/cip30mapping.test.ts
```diff
@@ -0,0 +1,138 @@
...
+      chromeRuntime.onMessage.addListener.mockImplementation((handler: any) => (onMessageHandler = handler));
+      chromeRuntime.sendMessage.mockImplementation((message: any, callback: any) =>
+        onMessageHandler(message, null, callback)
+      );
+      cleanup = cip30.initialize(wallet);
+    });
+    afterAll(() => cleanup());
+
+    it('works using browser runtime messages', async () => {
+      const uiWallet = createUiWallet();
```
rhyslbw:
```suggestion
      const api = createWalletApiMessageClient();
```

packages/wallet/test/integration/cip30mapping.test.ts
```diff
@@ -0,0 +1,138 @@
...
+      const utxos = await uiWallet.getUtxos();
```
rhyslbw, line 124:
```suggestion
      const utxos = await api.getUtxos();
```

rhyslbw (changes requested):

rhyslbw (approved):

rhyslbw (approved):
Excellent work @mchappell and @mkazlauskas :tada:

PR #193, opened 2022-03-03
feat: preparation for SingleAddressWallet persistence implementation

mkazlauskas:
# Context

`SingleAddressWallet` takes a long time to load (fetching all transactions and utxo)

# Proposed Solution

In preparation to implementing wallet persistence, do the following:
- (BREAKING) Remove `address` prop from `SingleAddressWallet` constructor, using `KeyAgent.knownAddresses` for initialization
- Define storage interfaces, implement in-memory storage (intended for development and testing, but could potentially be used to initialize wallets that do not want to store their state, or when shutting down and then restoring the wallet)
- Add `Wallet.syncStatus$` {Syncing, UpToDate}, so that wallet users have some insight on whether their wallet is currently synced or outdated. Implement it by wrapping a `WalletProvider` and tracking calls/responses of each function.

### DIFF
--- a/packages/wallet/src/SingleAddressWallet.ts
+++ b/packages/wallet/src/SingleAddressWallet.ts
@@ -11,7 +11,7 @@ import {
   WalletProvider,
   coreToCsl
 } from '@cardano-sdk/core';
-import { Assets, InitializeTxResult, KeyManagement, SignDataProps } from '.';
+import { Assets, InitializeTxResult, KeyManagement, SignDataProps, SyncStatus } from '.';
 import {
   Balance,
   BehaviorObservable,
@@ -19,6 +19,7 @@ import {
   FailedTx,
   PollingConfig,
   SyncableIntervalTrackerSubject,
+  TrackedWalletProvider,
   TrackerSubject,
   TransactionFailure,
   TransactionalTracker,
@@ -33,7 +34,7 @@ import {
   distinctBlock,
   distinctEpoch
 } from './services';
-import { Cip30DataSignature, cip30signData } from './KeyManagement/cip8/cip30signData';
+import { Cip30DataSignature, cip30signData } from './KeyManagement/cip8';
 import { InitializeTxProps, InitializeTxPropsValidationResult, MinimumCoinQuantity, Wallet } from './types';
 import {
   InputSelector,
@@ -45,6 +46,7 @@ import { Logger, dummyLogger } from 'ts-log';
 import { Observable, Subject, combineLatest, firstValueFrom, lastValueFrom, map, take } from 'rxjs';
 import { RetryBackoffConfig } from 'backoff-rxjs';
 import { TxInternals, createTransactionInternals, ensureValidityInterval } from './Transaction';
+import { createProviderStatusTracker } from './services/ProviderStatusTracker';
 import { isEqual } from 'lodash-es';
 
 export interface SingleAddressWalletProps {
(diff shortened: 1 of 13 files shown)


packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -112,7 +110,7 @@ export class SingleAddressWallet implements Wallet {
     this.#inputSelector = inputSelector;
     this.#walletProvider = walletProvider;
     this.#keyAgent = keyAgent;
-    this.addresses$ = new TrackerSubject<GroupedAddress[]>(this.#initializeAddress(address));
+    this.addresses$ = new TrackerSubject<GroupedAddress[]>(this.#initializeAddress(keyAgent.knownAddresses));
```
rhyslbw, line 120: Great refactor

packages/wallet/src/persistence/inMemoryStores.ts
```diff
@@ -0,0 +1,99 @@
...
+  protected readonly getDocId: GetDocId<T>;
+
+  constructor(getDocId: GetDocId<T>) {
+    this.getDocId = getDocId;
+  }
+
+  get(): Observable<T[]> {
+    return of(this.docs);
+  }
+  upsert(docs: T[]): Observable<void> {
```
rhyslbw, line 17: What about using a `Set` as the class property to reduce the complexity of the code here and in `delete`?

packages/wallet/src/services/ProviderStatusTracker.ts
```diff
@@ -0,0 +1,78 @@
...
+  const syncStatusWithoutTimeout$ = relevantStats.pipe(
+    skipWhile((allStats) => allStats.every(({ numCalls, numResponses }) => numCalls === 0 && numResponses === 0)),
+    map((allStats) =>
+      allStats.some(({ numCalls, numResponses }) => numCalls > numResponses) ? SyncStatus.Syncing : SyncStatus.UpToDate
+    ),
+    distinctUntilChanged(),
+    share()
+  );
+  const upToDate$ = syncStatusWithoutTimeout$.pipe(filter((syncStatus) => syncStatus === SyncStatus.UpToDate));
+  // const syncing$ = syncStatusWithoutTimeout$.pipe(filter((syncStatus) => syncStatus === SyncStatus.Syncing));
```
rhyslbw, line 67: Please explain the purpose of this commented out line

packages/wallet/src/persistence/inMemoryStores.ts
mkazlauskas: `Set` would make the code slightly simpler, but would constrain the user to using the same objects (`===`). Not really a concern for originally intended use case, but current implementation is more resilient.

packages/wallet/test/services/TrackedWalletProvider.test.ts
```diff
@@ -0,0 +1,123 @@
...
+  beforeEach(() => {
+    walletProvider = mockWalletProvider();
+    trackedWalletProvider = new TrackedWalletProvider(walletProvider);
+  });
+
+  test('CLEAN_FN_STATS numResponses is 0', () => {
+    expect(CLEAN_FN_STATS).toEqual({ numCalls: 0, numResponses: 0 });
+  });
+
+  describe('wraps underlying provider functions, tracks # of calls/responses and resets on stats.reset()', () => {
```
rhyslbw, line 18: This will be a useful tool for performance testing as well :heart_eyes:

packages/wallet/src/services/ProviderStatusTracker.ts
mkazlauskas: Nice catch, removed it locally after initially pushing the branch and forgot to re-push. The comment is gone now

packages/wallet/src/persistence/inMemoryStores.ts
mkazlauskas: It's added context for implementing the interface

rhyslbw (approved):
Excellent work @mkazlauskas

PR #197, opened 2022-03-09
Draft: Feat/graphql data projection

jvieiro:
# Context

Writing data to Dgraph from the chain-sync messages and results of Local State Queries can easily get unwieldy and therefore difficult to test if we do not consider a pattern that allows for smaller scoped blocks that still satisfy atomic write requirements.
We need to determine a modular approach to constructing the GraphQL mutations, committing the changes as the final step in the most optimal way.

# Proposed Solution

Modularize the data projection service

# Note

Block handler for assets is gonna be made in this first approach, still `rollForward`, `rollBackward` and `process` methods are in progress.


packages/cardano-graphql/src/DataProjection/RunnableModule.ts
```diff
@@ -0,0 +1,66 @@
+import { InvalidModuleState } from './errors';
+import { Logger, dummyLogger } from 'ts-log';
+import { ModuleState } from './types';
+import { moduleLogger } from '../util';
+
+export type RunnableModuleState = ModuleState | 'starting' | 'running' | 'stopping';
+
+export class RunnableModule {
```
mkazlauskas, line 8: Maybe add something like:
```
// same pattern for 'start' and 'shutdown'
protected abstract initializeImpl(): Promise<void>;
async initialize() {
  this.initializeBefore();
  await this.initializeImpl();
  this.initializeAfter();
}
```

mkazlauskas (commented):
I think you're on the right path. Some notes:
- There are a few lint errors. I suggest setting up eslint extension in your editor.
- It would probably be better to have DataProjection as a separate package (`cardano-graphql-cli`?). Most uses of `cardano-graphql` are only concerned with making the queries (client-side code). What do you think @rhyslbw ?

rhyslbw:
> It would probably be better to have DataProjection as a separate package (cardano-graphql-cli?). Most uses of cardano-graphql are only concerned with making the queries (client-side code). Perhaps leaving only provider implementations in cardano-graphql

Yes you're right @mkazlauskas, that was my mistake with the package scoping. As for naming, maybe `cardano-graphql-services` is a bit more generic? The CLI is one interface, but it's reasonable to think the package could be used as a library too.

jvieiro:
> I think you're on the right path. Some notes:
> 
> * There are a few lint warnings. I suggest setting up eslint extension in your editor.
> * It would probably be better to have DataProjection as a separate package (`cardano-graphql-cli`?). Most uses of `cardano-graphql` are only concerned with making the queries (client-side code). Perhaps leaving only provider implementations in `cardano-graphql`. What do you think @rhyslbw ?

Im actually using the eslint extension on VSCode, can you point me an example of these warnings pls?

mkazlauskas:
> Im actually using the eslint extension on VSCode, can you point me an example of these warnings pls?

GitHub shows the warnings in [Files Changed](https://github.com/input-output-hk/cardano-js-sdk/pull/197/files#diff-e90af146c0fc321c21cd0b7c7c275916dbe4331d1a78f5f4b25626dff2722014R44)

packages/cardano-graphql/src/DataProjection/util.ts
```diff
@@ -21,22 +28,22 @@ export const mergedQuery = async (
...
-//       : { func: noOp(), id: handler.id }
-//   );
+export const mergedProcessingResults = async (
+  blockHandlers: BlockHandler[],
+  ctx: RollForwardContext,
+  mergedQueryResults: any
+): Promise<CombinedProcessingResult[]> =>
+  blockHandlers.map((handler) =>
+    handler.process
+      ? { func: handler.process({ ctx, queryResult: mergedQueryResults }), id: handler.id }
```
mkazlauskas, line 38: How about awaiting processing results here? Would need to make sure handlers are processed in parallel though (Promise.all).

packages/cardano-graphql/src/DataProjection/util.ts
```diff
@@ -21,22 +28,22 @@ export const mergedQuery = async (
...
-// ): Promise<any> =>
-//   blockHandlers.map((handler) =>
-//     handler.process
-//       ? { func: handler.process(ctx, queryResult), handler: handler.id }
-//       : { func: noOp(), id: handler.id }
-//   );
+export const mergedProcessingResults = async (
+  blockHandlers: BlockHandler[],
+  ctx: RollForwardContext,
+  mergedQueryResults: any
```
mkazlauskas, line 34: Are you planning to define query results results type, or keep `any`?

mkazlauskas (commented):
Looking good, just some minor things I noticed

packages/cardano-graphql/src/DataProjection/types.ts
```diff
@@ -32,16 +33,68 @@ export interface Upsert {
...
-  rollForward: (ctx: RollForwardContext, processingResult?: any) => Promise<Upsert>;
+  rollForward: (ctx: RollForwardContext, processingResult: CombinedProcessingResult[]) => Promise<Upsert>;
+}
+
+export interface Signature {
+  signature: string;
+  publicKey: string;
+}
+
+export interface AssetMetadata {
```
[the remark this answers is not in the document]
rhyslbw: Yes, `MetadataClient` isn't exclusively for data projection, so when this code is moved to a new `cardano-graphql-services` package, it can live in the source root.

packages/cardano-graphql/src/DataProjection/RunnableModule.ts
rhyslbw: https://github.com/input-output-hk/cardano-js-sdk/pull/202

rhyslbw:
@jvieiro please drop the following commits, and rebase on `feat/cardano-graphql-services-setup`, to see if you can integrate with this state before we merge.

- e6f69491948ced9ce494122cc29e18d3067a942f
- 7cca593903b8e653d9606808c75c361bb8cd6ac1

packages/cardano-graphql/src/DataProjection/util.ts
jvieiro: It should be typed. I forgot to set it

packages/cardano-graphql/src/DataProjection/MetadataClient.ts
```diff
@@ -0,0 +1,60 @@
...
+        subjects: assetIds,
+        properties: ['decimals', 'description', 'logo', 'name', 'ticker', 'url']
+      });
+      return response.data.subjects;
+    } catch (error) {
+      if (axios.isAxiosError(error) && error?.code === 'ENOTFOUND') {
+        this.logger.error({ err: error });
+        throw new HostDoesNotExist('metadata server');
+      } else {
+        throw error;
```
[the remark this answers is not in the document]
jvieiro: Gotcha. Will include this in the next commit

packages/cardano-graphql/src/DataProjection/blockHandlers/AssetBlockHandler.ts
```diff
@@ -1,9 +1,33 @@
...
+const HANDLER_ID = 'Asset';
+
+export const createAssetBlockHandler = (metadataClient: MetadataClient, logger = dummyLogger): BlockHandler => ({
+  id: HANDLER_ID,
+  process: async ({ queryResult }: ProcessParameters) => {
+    const assetsToInsert = [];
+    if (queryResult) {
+      for (const asset of queryResult.data.assets) {
+        logger.info('About to fetch asset metadata');
+        if (!(await metadataClient.fetch([asset.assetId]))) {
```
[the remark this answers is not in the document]
jvieiro: I will check with @rhyslbw, maybe I misunderstood it.
I've noticed that it can be more performant if I only do one query. The problem I see is how can the asset be mapped to the corresponding metadata? It looks like the response does not contain that information. Also, how would be the insert?  Looks like graphql's asset metadata schema and the response differs in some fields, should I map them?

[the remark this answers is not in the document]
mkazlauskas: > The problem I see is how can the asset be mapped to the corresponding metadata

I would guess it corresponds to index, e.g. `fetch(['id1', 'id2'])` => `[asset1metadata, asset2metadata]`

> Also, how would be the insert? Looks like graphql's asset metadata schema and the response differs in some fields, should I map them?

Does the metadata not cover graphql schema? Please try to map it and leave comments on any missing data.

packages/cardano-graphql-services/src/DataProjection/Service.ts
```diff
@@ -0,0 +1,55 @@
...
+  async initialize() {
+    super.initializeBefore();
+    await this.#dgraphClient.initialize(this.config.dgraph.schema);
+    await this.#chainFollower.initialize(this.config.ogmios?.connection);
+    await this.#metadataClient.initialize();
+    super.initializeAfter();
+  }
+
+  async startImpl() {
+    // Todo: Get most recent point to start sync from
```
mkazlauskas, line 47: We should probably implement a minimal block handler that will write block data so that we can query the tip here

packages/cardano-graphql-services/test/util/stubLogger.ts
```diff
@@ -0,0 +1,12 @@
+import { Logger } from 'ts-log';
```
mkazlauskas, line 1: Wasn't this added by @rhyslbw recently? There might be a git issue to resolve

mkazlauskas (changes requested):
Looks pretty good!

A few topics to discuss/consider:
- Services should probably attempt to retry some of their operations if they fail.
- Is NFT metadata out of scope of this PR?

packages/cardano-graphql-services/src/DataProjection/util.ts
```diff
@@ -0,0 +1,66 @@
...
+  blockHandlers
+    .map((handler) => handler.rollForward(ctx, processingResult))
+    .reduce(async (acc, curr) => lodash.merge(await acc, await curr));
+
+export const mergedRollBackwardUpsert = async (blockHandlers: BlockHandler[], ctx: RollBackwardContext) =>
+  blockHandlers
+    .map((handler) => handler.rollBackward(ctx))
+    .reduce(async (acc, curr) => lodash.merge(await acc, await curr));
+
+export const mapMetadata = (serverAssetMetadata: AssetMetadata): Asset.TokenMetadata => ({
```
mkazlauskas, line 59: I think this file should only have generic utils, this one is specific to AssetBlockHandler - should probably move it there.

packages/cardano-graphql-services/src/DataProjection/types.ts
```diff
@@ -0,0 +1,64 @@
...
+export type ModuleState = null | 'initializing' | 'initialized';
+
+export type DQL = string;
+type Variable = string;
+
+export type QueryVariables = {
+  $assetIds?: string[];
+  $slotNo?: number;
+};
+export interface QueryResult<Variables> {
```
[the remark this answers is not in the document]
jvieiro: @rhyslbw marked me  this as an improve. But maybe I can define it as:

```
 export interface QueryResult {
  query: DQL;
  variables?: QueryVariables;
}
```

rhyslbw (changes requested):

packages/cardano-graphql-services/src/DataProjection/Service.ts
rhyslbw: Yes, see my comment here from a pending review I forgot to submit. 
https://github.com/input-output-hk/cardano-js-sdk/pull/197#discussion_r830489411

packages/cardano-graphql-services/src/DataProjection/blockHandlers/AssetBlockHandler.ts
```diff
@@ -0,0 +1,124 @@
...
+      tokenMetadata: null,
+      totalQuantity: null,
+      uid: 'uid(Asset)'
+    };
+    return {
+      mutations: { ...deleteMutation },
+      variables: { defines: ['Asset'], dql: query }
+    };
+  },
+  rollForward: async (ctx: RollForwardContext, processingResult: CombinedProcessingResult[]) => {
```
mkazlauskas, line 113: Shouldn't this receive `CombinedQueryResult`?

packages/cardano-graphql-services/src/DataProjection/blockHandlers/helpers.ts
```diff
@@ -0,0 +1,79 @@
...
+};
+
+export const mapMetadata = (serverAssetMetadata: AssetMetadata): Asset.TokenMetadata => ({
+  decimals: serverAssetMetadata.decimals.value,
+  desc: serverAssetMetadata.description?.value,
+  icon: serverAssetMetadata.logo?.value,
+  name: serverAssetMetadata.name?.value,
+  ticker: serverAssetMetadata.ticker?.value,
+  url: serverAssetMetadata.url?.value
+  // Missing fields: sizedIcons
```
mkazlauskas, line 22: This is fine for now. Could open an issue in metadata server repo. It should come as ["icon-\<size>"](https://github.com/cardano-foundation/CIPs/pull/137/files#diff-f478f61a4b67bc7fee1d41bbb427028a7615beaacf726efc9289104e64f26ffaR48). Let's keep the comment.

packages/cardano-graphql-services/src/DataProjection/blockHandlers/helpers.ts
```diff
@@ -0,0 +1,79 @@
...
+      blockNo: block.header?.blockHeight,
+      hash: block.header?.blockHash,
+      opCert: block.header?.opCert.sigma,
+      previousBlock: { hash: block.header?.prevHash },
+      size: block.header?.blockSize,
+      slot: block.header?.slot,
+      totalFees: BigIntMath.sum(block.body.map((body) => body.body.fee)),
+      totalOutput: BigIntMath.sum(outputsAmount),
+      transactions: block?.body?.map((tx) => mapTransaction(tx, block?.header?.blockHeight))
+      // Missing: nextBlock, nextBlockProtocolVersion, confirmations, totalLiveStake, epoch, stake pool issuer
```
mkazlauskas, line 76: Let's keep this comment for now and implement all those as a separate PR. Might need to temporarily make some required fields optional: dgraph will most likely complain about not setting them.

packages/cardano-graphql-services/src/DataProjection/blockHandlers/helpers.ts
```diff
@@ -0,0 +1,79 @@
...
+  let b: Schema.BlockMary | Schema.BlockAlonzo | undefined;
+  if (isAlonzoBlock(block)) {
+    b = block.alonzo as Schema.BlockAlonzo;
+  } else if (isMaryBlock(block)) {
+    b = block.mary as Schema.BlockMary;
+  }
+  return b;
+};
+
+export const mapMetadata = (serverAssetMetadata: AssetMetadata): Asset.TokenMetadata => ({
```
mkazlauskas, line 15: Since `mapBlock` is used for mutations, we should be mapping to our [graphql schema types](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-graphql-services/src/Schema/types/Asset/Asset.ts) instead of core types

packages/cardano-graphql-services/src/DataProjection/blockHandlers/helpers.ts
```diff
@@ -0,0 +1,79 @@
...
+  desc: serverAssetMetadata.description?.value,
+  icon: serverAssetMetadata.logo?.value,
+  name: serverAssetMetadata.name?.value,
+  ticker: serverAssetMetadata.ticker?.value,
+  url: serverAssetMetadata.url?.value
+  // Missing fields: sizedIcons
+});
+
+const mapAssets = (assets: { [k: string]: Schema.AssetQuantity } | undefined) =>
+  // Missing: transactionOutput, asset
```
mkazlauskas, line 26: You're probably looking at the wrong type:
- `Token` represents value on utxo
- `Asset` is what's minted - it's what we want here

packages/cardano-graphql-services/src/DataProjection/blockHandlers/helpers.ts
```diff
@@ -0,0 +1,79 @@
...
+  assets ? Object.entries(assets).map(([name, quantity]) => ({ asset: { name }, quantity })) : [];
+
+const mapOutput = (output: Schema.TxOut) =>
+  // Missing: index, transaction, datumHash
```
mkazlauskas, line 30: - index: `outputs.indexOf(output)`
- transaction: graph edge to Transaction object
- datumHash: [available in Ogmios schema](https://github.com/CardanoSolutions/ogmios/blob/master/clients/TypeScript/packages/schema/src/index.ts#L1113)

packages/cardano-graphql-services/src/DataProjection/blockHandlers/helpers.ts
```diff
@@ -0,0 +1,79 @@
...
+  ({
+    address: output.address,
+    value: { assets: mapAssets(output.value.assets), coin: output.value.coins }
+  });
+
+const mapInput = (input: Schema.TxIn) =>
+  // Missing: address:, redeemer, sourceTransaction, transaction, value
```
mkazlauskas, line 37: - address and value: it should be usually available in our database, this is data from previous outputs. Should query those during our "query" step. Let's change our schema so that they're nullable (it's probably the correct model anyway, as there might not be any address for 1st transaction?) and leave this out of scope for this PR. Might end up just removing those fields.
- redeemer: let's leave this out of scope of this PR
- sourceTransaction: graph edge to [this](https://github.com/CardanoSolutions/ogmios/blob/master/clients/TypeScript/packages/schema/src/index.ts#L1104)
- transaction: graph edge to current Transaction object

packages/cardano-graphql-services/src/DataProjection/blockHandlers/helpers.ts
```diff
@@ -0,0 +1,79 @@
...
+  ({ index: input.index });
+
+const mapAuxiliaryData = (metadata: Schema.AuxiliaryData) => ({
```
mkazlauskas, line 40: This is a complex nested object (see [Metadatum](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-graphql-services/src/Schema/types/Transaction/AuxiliaryData/Metadatum.ts)). Let's leave it out of scope of this PR.

packages/cardano-graphql-services/src/DataProjection/blockHandlers/helpers.ts
```diff
@@ -0,0 +1,79 @@
...
+  auxiliaryData: tx.metadata ? mapAuxiliaryData(tx.metadata) : tx.metadata,
+  blockHeight,
+  fee: tx.body.fee,
+  hash: tx.id,
+  inputs: tx.body.inputs.map(mapInput),
+  invalidBefore: tx.body.validityInterval.invalidBefore,
+  invalidHereafter: tx.body.validityInterval.invalidHereafter,
+  mint: mapAssets(tx.body.mint.assets),
+  outputs: tx.body.outputs.map(mapOutput)
+  // Missing: size, certificates, scriptIntegrityHash, requiredExtraSignatures, deposit, index, collaterals
```
mkazlauskas, line 60: Let's leave those out of scope of this PR

mkazlauskas (changes requested):
Let's keep Transaction data to the minimum in this PR. You can set any transaction properties (that you don't necessarily need right now) in schema to `{nullable: true}` (just please leave comments so we can gradually roll back to non-nullable as we implement more functionality).

I think we're at the point where we need to run some basic e2e tests - it should flesh out any remaining issues.

packages/cardano-graphql-services/src/DataProjection/blockHandlers/helpers.ts
jvieiro: Am I ? I intended to map the [Token](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-graphql-services/src/Schema/types/Transaction/Token.ts) type. I will rename this fc to `mapToken` to be more clear.
I think the best is to  just leave `tx.mint` and `output.value.assets`  empty for now with a comment that `assetId` is missing (to reference the asset edge).

packages/cardano-graphql-services/src/DataProjection/blockHandlers/AssetBlockHandler.ts
jvieiro: Isnt the `CombinedProccessingResult` the one that adds the metadata to the asset?

packages/cardano-graphql-services/src/DataProjection/blockHandlers/helpers.ts
jvieiro: Ok, so the idea is in the first instance to insert the Asset object into this BlockHandler and then at AssetHandler update each of these asset objects with its metadata? In this case be must be sure that the block handler runs first.
So, the mint property would be necessary there, I will add this.
Not sure how the block handler interface will help

packages/cardano-graphql-services/src/DataProjection/blockHandlers/AssetBlockHandler.ts
mkazlauskas: Yeah you're right, I thought maybe it could receive both if there is a handler that doesn't have a processing step. Currrent approach is good too: nothing to do.

packages/cardano-graphql-services/src/DataProjection/blockHandlers/helpers.ts
mkazlauskas: > Ok, so the idea is in the first instance to insert the Asset object into this BlockHandler and then at AssetHandler update each of these asset objects with its metadata?

> So, the mint property would be necessary there, I will add this.

Something like that, but no, it doesn't sound right. The purpose of BlockHandlers is to decompose BlockBlockHandler and make it possible to create a database that has only a subset of data by using a subset of BlockHandlers. So I think the BlockBlockHandler is not supposed to write Assets at all.

> In this case be must be sure that the block handler runs first.

Original idea was to use [variables](https://github.com/input-output-hk/cardano-js-sdk/pull/197/files#diff-e9fbda3cd7bac9880ca294f5c1de0b2e102ac4a15a111f8799da73a1e2d66182R36-R37) to build a dependency tree: it should be possible to compute the correct order of handlers. Let's keep it out of scope of this PR and just put the handlers in order that we need.

> So, the mint property would be necessary there, I will add this.

I think BlockBlockHandler shouldn't insert assets. It's what AssetBlockHandler.

mkazlauskas: Also I'm thinking that since we need to query the tip in order to start ChainFollower, maybe it would be best to write partial blocks as "core functionality" (like a hidden BlockHandler that always runs first). Thoughts @rhyslbw ?

rhyslbw: > Also I'm thinking that since we need to query the tip in order to start ChainFollower, maybe it would be best to write partial blocks as "core functionality" (like a hidden BlockHandler that always runs first).

Yes, this is the first handler to implement. The asset handler was being nominated first since it requires both the query and processing step, but it does need to come after a pure chain data handler to allow the sync state to be understood at startup.

rhyslbw:
Parking this PR for now

PR #201, opened 2022-03-15
Feat/ogmios tx submit provider

rhyslbw:
# Context
- Ogmios provides a number of useful interfaces we need to integrate with, including transaction submission.
- The core package contains a dependency on Ogmios, given it forms the basis of our integration with `cardano-node`.

# Proposed Solution
I've chosen to source the provider in the core package for now given the existing dependency, with a mock service to facilitate testing. Only a single error path is tested, as the provider is only catching and translating non-domain errors to the [error type](https://github.com/input-output-hk/cardano-js-sdk/blob/7e4a058abd404aad6b96cd66dd1a80948c6abee4/packages/core/src/Cardano/types/TxSubmissionErrors.ts#L69-L73) defined by the SDK, so it would be a lot of mocking and test code for little benefit. The two core package tests demonstrate the two top-level responses, with further test coverage to be handled within the wallet's e2e tests.
 
:pushpin: **feat(core): ogmiosTxSubmitProvider**
- Uses a mock Ogmios server, that we can also use to facilitate chain-sync testing.
- random free port selected within the test

:pushpin: **feat(wallet): ogmiosTxSubmitProvider in e2e tests**
Configuration to select the provider in wallet e2e tests, using envalid for improved
env validation.

:pushpin: r**efactor(wallet): refactor remaining e2e config.ts using envalid**
Also positions the options for easy review when determining config.
Existence checks are not needed since envalid enforces values.

### DIFF
--- a/packages/ogmios/src/ogmiosTxSubmitProvider.ts
+++ b/packages/ogmios/src/ogmiosTxSubmitProvider.ts
@@ -1,0 +1,54 @@
+import { Buffer } from 'buffer';
+import { Cardano, TxSubmitProvider } from '@cardano-sdk/core';
+import {
+  ConnectionConfig,
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  TxSubmission,
+  UnknownResultError,
+  createInteractionContext,
+  createTxSubmissionClient
+} from '@cardano-ogmios/client';
+import { Logger, dummyLogger } from 'ts-log';
+
+/**
+ * Connect to an [Ogmios](https://ogmios.dev/) instance
+ *
+ * @param {ConnectionConfig} connectionConfig Ogmios connection configuration
+ * @param {Logger} logger object implementing the Logger abstract class
+ * @returns {TxSubmitProvider} TxSubmitProvider
+ * @throws {TxSubmission.errors}
+ */
+export const ogmiosTxSubmitProvider = (
+  connectionConfig: ConnectionConfig,
+  logger: Logger = dummyLogger
+): TxSubmitProvider => ({
+  submitTx: async (signedTransaction) => {
+    // The Ogmios client supports opening a long-running ws connection,
+    // however as the provider interface doesn't include shutdown handling,
+    // we're using the one time interaction type for now.
+    try {
+      const interactionContext = await createInteractionContext(
+        (error) => {
+          logger.error({ error: error.name, module: 'ogmiosTxSubmitProvider' }, error.message);
+        },
+        logger.info,
+        {
+          connection: connectionConfig,
+          interactionType: 'OneTime'
+        }
+      );
+      const txSubmissionClient = await createTxSubmissionClient(interactionContext);
+      const txHex = Buffer.from(signedTransaction).toString('hex');
+      return await txSubmissionClient.submitTx(txHex);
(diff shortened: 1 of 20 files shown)


packages/wallet/test/e2e/config.ts
```diff
@@ -31,10 +32,22 @@ export const assetProvider = (() => {
 })();
 
 export const txSubmitProvider = (() => {
-  const projectId = process.env.BLOCKFROST_API_KEY;
-  if (!projectId) throw new Error('BLOCKFROST_API_KEY not set (for txSubmitProvider)');
-  const blockfrost = new BlockFrostAPI({ isTestnet, projectId });
-  return blockfrostTxSubmitProvider(blockfrost);
+  const env = envalid.cleanEnv(process.env, {
```
mkazlauskas, line 35: Nice

packages/wallet/.env.example
```diff
@@ -6,4 +7,7 @@ NETWORK_ID=0
 MNEMONIC_WORDS="actor scout worth mansion thumb device mass pave gospel secret height document merge text broom kind lesson invest across estate erase interest end century"
 WALLET_PASSWORD=some_password
 POOL_ID_1=pool1euf2nh92ehqfw7rpd4s9qgq34z8dg4pvfqhjmhggmzk95gcd402
-POOL_ID_2=pool1fghrkl620rl3g54ezv56weeuwlyce2tdannm2hphs62syf3vyyh
\ No newline at end of file
+POOL_ID_2=pool1fghrkl620rl3g54ezv56weeuwlyce2tdannm2hphs62syf3vyyh
+OGMIOS_HOST=localhost
```
mkazlauskas, line 11: Since there are now 2 submit providers available, we should probably document the options either as comments in `.env.example` or README.

mkazlauskas (changes requested):
I think it would be better to move this to a new `ogmios` package. Even though `core` already had ogmios client as a dependency, it's not a "hard dependency" - more of an implementation detail for convenience. We could remove it without breaking too much if we wanted.

packages/wallet/.env.example
rhyslbw: Done in c079f6668a0ede6787ec546c13296a4a04e9ef33 and 2aba96af2329c4a28f078325fe9d8fc7ff3893d4

packages/wallet/test/e2e/config.ts
rhyslbw: Completed the upgrade in 2aba96af2329c4a28f078325fe9d8fc7ff3893d4

packages/ogmios/src/ogmiosTxSubmitProvider.ts
```diff
@@ -0,0 +1,54 @@
+import { Buffer } from 'buffer';
+import { Cardano, TxSubmitProvider } from '@cardano-sdk/core';
+import {
+  ConnectionConfig,
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
mkazlauskas, line 5: We could just disable this globally. Speaking of global configurations, it would be nice to have separate config for tests allowing explicit any and long lines. Can be addressed separately.

mkazlauskas (approved):
Nice work!

mkazlauskas (approved):

PR #203, opened 2022-03-18
feat: add Provider interface, use as base for TxSubmitProvider

rhyslbw:
# Context
Checking the health of a service is an important aspect for application logic and system monitoring. All providers have this common model, but we don't currently have a base interface. The service of interest is tx submission due to a dependent chunk of work now on hold awaiting this feature.

# Proposed Solution
Implement a base interface with minimal model for a boolean response, but room for an extension to include other metrics at a later date. I've purposely left the other provider refactors out of this scope to avoid overloading this PR, as it can be completed separately.


mkazlauskas (approved):
Looks good!

PR #205, opened 2022-03-21
refactor: move Dgraph schema generation to services package

rhyslbw:
# Context
Generating Dgraph schema is better suited to be sourced in `cardano-graphql-services`, since this is where the projection code is.

# Proposed Solution
Move the code, scripts, and documentation


mkazlauskas (changes requested):
I'm not sure about moving `yarn generate`. It generates `sdk.ts` which is part of client-side scope that belongs in `cardano-graphql`. Also, one of the inputs to codegen is `operations/` graphql files for queries, which definitely belong in `cardano-graphql`.

At the same time, it would be cleaner if `cardano-graphql` didn't need to have a docker-compose.yml.

Perhaps we should both:
- Keep codegen in `cardano-graphql`
- Keep docker-compose under `cardano-graphql-services`, documenting in `cardano-graphql` that you need to have it running for codegen

packages/cardano-graphql/codegen.yml
```diff
@@ -1,4 +1,4 @@
-schema: ./dist/dgraph.graphql
+schema: http://localhost:8080/
```
mkazlauskas, line 1: Nice, have you tested this?

mkazlauskas (approved):
Please confirm that this is the new workflow after updating schema (might be worth documenting too?):
- `cd packages/cardano-graphql-services && yarn build && yarn build:schema:dgraph`
- `cd packages/cardano-graphql` && yarn build && yarn generate`

rhyslbw:
> Please confirm that this is the new workflow after updating schema (might be worth documenting too?):

eb5aea45ed6556a115b057a4b9e2972470472253 includes those docs, plus starting and stopping the Docker Compose stack.

jvieiro (approved):

PR #208, opened 2022-03-22
fix(ogmios): fix ogmios ts errors

tomislavhoracek:
# Context
Fix `Ogmios` TS errors

- `ogmiosTxSubmitProvider` - add type to try/catch to fix ts error


mkazlauskas (approved):
Would be nicer to check if error is instance of ogmios error, but let's just trust it won't throw stuff like `null` or `numbers`

PR #211, opened 2022-03-30
refactor(wallet): added second wallet to e2e test

iadmytro:
Added second wallet to e2e test


mkazlauskas (approved):
Looks good! I think that since the changes are only for tests, a more suitable commit message would be `test(wallet):`.

Follow-up work could be asserting `destWallet` balance after `sourceWallet` sends coin to it. At this point we should probably be renaming the test.

PR #212, opened 2022-03-30
ci: added pre-release workflow

iadmytro:
Added CI pre-release workflow

### DIFF
--- a/.github/workflows/pre-release.yaml
+++ b/.github/workflows/pre-release.yaml
@@ -1,0 +1,47 @@
+name: Pre-release
+
+on:
+  pull_request:
+    branches: [ "master" ]
+  workflow_dispatch:
+
+jobs:
+  wallet-e2e-tests:
+    strategy:
+      matrix:
+        os: [ ubuntu-20.04 ]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: 📥 Checkout repository
+        uses: actions/checkout@v2.3.3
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,44 @@
...
+        run: |
+          yarn install --frozen-lockfile --non-interactive --logevel=error
+          yarn build
+
+      - name: 🔬 Test
+        working-directory: packages/wallet
+        env:
+          BLOCKFROST_API_KEY: ${{ secrets.BLOCKFROST_API_KEY }}
+          MNEMONIC_WORDS_ONE: ${{ secrets.MNEMONIC_WORDS_ONE }}
+          MNEMONIC_WORDS_TWO: ${{ secrets.MNEMONIC_WORDS_TWO }}
```
james-iohk, line 35: We only want `MNEMONIC_WORDS`

james-iohk (changes requested):

.github/workflows/pre-release.yaml
mkazlauskas: Also seems like we're missing some variable(s) in env because the check on this PR failed

Newman1515 (commented):
Submit the approve
And
Submit the review

rhyslbw:
We now have a _release_ workflow, so this one is no longer relevant.

PR #213, opened 2022-03-30
Feat/Tx Submit Prometheus

rhyslbw:
# Context
We need to expose Prometheus metrics for service monitoring

# Proposed Solution
Attaches express middleware in `HttpServer` when enabled, using `express-prom-bundle`. `prom-client` options are exposed at the class level, but have not yet been exposed at the entrypoints since it's complex and will need either a decision on using configuration files or inline JSON. Until then the default is likely suitable for current needs.
 
Enabling this feature exposes a `/metrics` endpoint


packages/cardano-graphql-services/src/Http/HttpServer.ts
```diff
@@ -1,27 +1,66 @@
...
+  metrics?: {
+    enabled: boolean;
+    options?: expressPromBundle.Opts;
+  };
+  name?: string;
   listen: net.ListenOptions;
 };
 
+export interface HttpServerDependencies {
+  healthCheckCallback: () => Promise<{ ok: boolean }>;
```
mkazlauskas, line 22: Exporting `{ok: boolean}` from core Provider.ts would be nice. It's likely to change in the future.
Also, I think a nicer name here is just 'healthCheck' or 'checkHealth' or 'healthCheckImpl'.

I would make it optional, having default just resolve {ok: true}

packages/cardano-graphql-services/src/Http/HttpServer.ts
```diff
@@ -1,27 +1,66 @@
...
+  protected constructor(public config: HttpServerConfig, { logger = dummyLogger, ...rest }: HttpServerDependencies) {
+    super(config.name || 'HttpServer', logger);
+    this.#dependencies = { logger, ...rest };
   }
 
   protected async initializeImpl(): Promise<void> {
+    this.#dependencies.router.use(bodyParser.json());
+    this.#dependencies.router.get('/health', async (req, res) => {
+      this.logger.debug('/health', { ip: req.ip });
+      let body: { ok: boolean } | Error['message'];
```
mkazlauskas, line 41:
```suggestion
      let body: HealthCheckResult | Error['message'];
```

packages/cardano-graphql-services/src/Http/HttpServer.ts
```diff
@@ -1,27 +1,66 @@
...
-    this.app.use(this.config.router);
+    if (this.config.metrics?.enabled) {
+      this.app.use(
+        expressPromBundle({
+          includeMethod: true,
+          promRegistry: new promClient.Registry(),
+          ...this.config.metrics.options
+        })
+      );
+      this.logger.info(`Prometheus metrics configured at ${this.config.metrics.options?.metricsPath || '/metrics'}`);
```
mkazlauskas, line 60: I'm wondering maybe the configured path is available on what `expressPromBundle` returns. Super minor, feel free to ignore this comment

packages/cardano-graphql-services/src/TxSubmit/TxSubmitHttpServer.ts
```diff
@@ -67,7 +58,11 @@ export class TxSubmitHttpServer extends HttpServer {
       }
       res.send(body);
     });
-    return new TxSubmitHttpServer(config, router, { logger, txSubmitProvider });
+    return new TxSubmitHttpServer(
+      { logger, txSubmitProvider },
+      { router, healthCheckCallback: () => txSubmitProvider.healthCheck(), logger },
```
mkazlauskas, line 63: Lint warning here. Let's configure it to fail on warnings. appending `--max-warnings 0` to root package.json lint script should do it

packages/cardano-graphql-services/test/Http/HttpServer.test.ts
```diff
@@ -26,18 +28,18 @@ describe('HttpServer', () => {
...
 
   beforeEach(async () => {
     port = await getRandomPort();
     apiUrlBase = `http://localhost:${port}`;
-    httpServer = SomeHttpServer.create({ host: 'localhost', port });
   });
 
   describe('initialize', () => {
     it('initializes the express application', async () => {
+      httpServer = SomeHttpServer.create({ listen: { host: 'localhost', port } });
```
mkazlauskas, line 42: This has many occurrences. I think wrapping it in a fn would be nice

```suggestion
const createHttpServer = () => SomeHttpServer.create({ listen: { host: 'localhost', port } });
...
      httpServer = createHttpServer();
```

mkazlauskas (changes requested):

packages/cardano-graphql-services/src/Http/HttpServer.ts
rhyslbw: I did look for that before resorting to this clumsy solution, but will double-check.

packages/cardano-graphql-services/src/Http/HttpServer.ts
rhyslbw: Thanks, I was meaning to come back and define this type in the core. Also agree with the function rename

packages/cardano-graphql-services/test/Http/HttpServer.test.ts
rhyslbw: The repetition is a bit awkward, but we don't gain much by lifting to a function since `port` is being set in the `before` block, and it would create some disconnect when reading the test. I'll leave it as-is for now, but we can revisit it should the interface get more complicated.

mkazlauskas: Sounds reasonable

mkazlauskas (approved):
Nice work!

PR #216, opened 2022-03-31
feature(wallet): Ledger Hardware Wallet Transaction Signing

tomislavhoracek:
# Context
SDK needs to be able to support transaction signing using hardware devices / wallets. LedgerKeyAgent will be responsible to do that.

# Proposed Solution
As a hardware wallet user who has constructed a transaction (or had a transaction constructed for them) I want to sign a transaction so that I can submit a transaction to the network & spend the assets “stored” in my hardware wallet.

- Transaction signing with Ledger devices

# TODO:
- TBD


packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -164,11 +170,68 @@ export class LedgerKeyAgent extends KeyAgentBase {
...
+
   async signBlob(): Promise<SignBlobResult> {
     throw new NotImplementedError('signBlob');
   }
 
   async exportRootPrivateKey(): Promise<Cardano.Bip32PrivateKey> {
     throw new NotImplementedError('Operation not supported!');
   }
+
+  async #prepareDataForSigning({ body, hash, inputSelection }: InitializeTxResult) {
```
mkazlauskas, line 198: Should probably explicitly specify the return type from ledger library. Same when converting each transaction property.

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -164,11 +170,68 @@ export class LedgerKeyAgent extends KeyAgentBase {
...
+    // TODO - Isolate change outputs
+    // TODO - Prepare token bundle
+    const inputs = body.inputs.map((input) => ({
+      txHashHex: input.txId,
+      outputIndex: input.txIndex,
+      // TODO - Derivation Path - get from `ownSignatureKeyPaths`
```
mkazlauskas, line 204: I think this is an odd API and I suggest setting all paths on inputs to `null`, using `ownSignatureKeyPaths` to specify `additionalWitnessPaths` in `SignTransactionRequest`.

Note that ledger will probably be able to compute that it needs stake key (`role: KeyType.Stake`) by itself, so you may have to omit it from `additionalWitnessPaths`

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -164,11 +170,68 @@ export class LedgerKeyAgent extends KeyAgentBase {
...
+    }));
+
+    const outputs = body.outputs.map((output) => ({
+      amount: output.value.coins,
+      tokenBundle: output.value.assets || [], // TODO - Check & Calculate
+      destination: {
+        // TODO - remove address type declaration comments
+        // DEVICE_OWNED - for change address
+        // THIRD_PARTY - for regular output
+        type: TxOutputDestinationType.THIRD_PARTY,
```
mkazlauskas, line 215: `if this.addresses.some(({address}) => address === output.address) then DEVICE_OWNED`

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -164,11 +170,68 @@ export class LedgerKeyAgent extends KeyAgentBase {
...
+          addressHex: utils.buf_to_hex(utils.bech32_decodeAddress(output.address))
+        },
+      },
+    }));
+
+    return {
+      tx: {
+        network: {
+          networkId: this.serializableData.networkId,
+          protocolMagic: 42, // TODO - replace with real magic - Faked to Testnet
```
mkazlauskas, line 226: This can be a prop when creating key agent

mkazlauskas (changes requested):

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
tomislavhoracek: @mkazlauskas aha, I know. I have that in my TODO list.

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
tomislavhoracek: I will double check but as I can see from integration tests it is required.

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
tomislavhoracek: Yes, I need to calculate outputs but also add additional parameters if address is change address ( + spendingPath
& stakingPath). I have TODO for that `// TODO - Isolate change outputs`

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
tomislavhoracek: cool, I was thinking about constant because protocolMagic depends on `networdId` and automatically return it regarding passed network. What you think about that @rhyslbw @mkazlauskas?

mkazlauskas: That sounds ok to me but don't really know what it the magic is

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
tomislavhoracek: @mkazlauskas here is the definition and maybe we should construct path. Additional witnesses are used for e.g. minting
```
path: [BIP32Path](https://vacuumlabs.github.io/ledgerjs-cardano-shelley/4.0.0/index.html#bip32path) | null
Describes path used for witnessing this UTxO. The API will sign transaction with this path.

Note: null indicates we don't want to sign this utxo. This is highly non-standard and the only usecase so far is pool registration as owner. We therefore don't mark it as optional so that people won't forget specifying it

Note: Device has no ability to really check whether path is correct witnessing path for this UTxO.
```

mkazlauskas: Hm no I think it would be easier to set all paths as "additional". Mapping inputs to addresses would make it more complicated.

tomislavhoracek: ok :)

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
mkazlauskas: Our staking path is always the same (2, 0) and spending path is available on objects in `this.addresses`

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
tomislavhoracek: It could be e.g.
```
// Cardano Byron Testnet network magic
  [TESTNET]: [1097911063, 0],
  // Cardano Alonzo Purple network magic
  [ALONZO_PURPLE]: [8, 0],
  // Cardano Shelley QA network magic
  [SHELLEY_QA]: [3, 0],
```

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
tomislavhoracek: aha, staking path is same `1852'/1815'/{accountIndex}'/0/0`. 
ouu yeah, spending path is in address obj

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
rhyslbw: Yeah network ID is a network _type_ identifier, telling us if it's the mainnet or **a** testnet. Network magic is the network identifier. This should really come from a provider, so we can use a remote service to provision the key agent during a handshake if that's an option.

mkazlauskas (changes requested):
Looking good so far! Besides stuff in comments, there are some lint errors.

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -36,6 +39,8 @@ export interface CreateTransportProps {
   devicePath?: string;
 }
 
+const transportTypedError = (error?: any) => new TransportError('Transport failed', error);
```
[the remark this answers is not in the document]
tomislavhoracek: @mkazlauskas oke, I was in dilemma about that. To use AuthenticationError or Transport error. Now is cleaner

packages/wallet/src/KeyManagement/util/mapHardwareSigningData.ts
```diff
@@ -0,0 +1,454 @@
...
+    ipv6 = ipv6.slice(0, -1);
+    return ipv6;
+  }
+  return null;
+};
+
+export const txToLedger = async ({
+  tx,
+  networkId,
+  keys,
```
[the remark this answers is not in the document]
tomislavhoracek: 👌

packages/wallet/src/KeyManagement/errors/HwMappingError.ts
```diff
@@ -0,0 +1,12 @@
+import { CustomError } from 'ts-custom-error';
+
+const formatMessage = (detail?: string) => {
+  const messageDetail = detail ? `: ${detail}` : '';
+  return `Hardware date mapping failure${messageDetail}`;
+};
+
+export class HwMappingError extends CustomError {
```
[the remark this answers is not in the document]
tomislavhoracek: Yeah, I was thinking about that but wanted to isolate that. But in the end, it is some kind of serialization and you are right. I will replace that

packages/wallet/test/hardware/LedgerKeyAgent.test.ts
```diff
@@ -32,6 +59,42 @@ describe('LedgerKeyAgent', () => {
...
+    };
+    const txInternals = await wallet.initializeTx(props);
+    const tx = await wallet.finalizeTx(txInternals);
+    const txSubmitting = firstValueFrom(wallet.transactions.outgoing.submitting$);
+    const txPending = firstValueFrom(wallet.transactions.outgoing.pending$);
+    const txInFlight = firstValueFrom(wallet.transactions.outgoing.inFlight$.pipe(skip(1)));
+    expect(tx.body).toBe(txInternals.body);
+    expect(tx.id).toBe(txInternals.hash);
+    expect(tx.witness.signatures.size).toBe(1);
+    await wallet.submitTx(tx);
```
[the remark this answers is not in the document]
tomislavhoracek: 👌

packages/wallet/src/KeyManagement/errors/HwMappingError.ts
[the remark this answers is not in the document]
mkazlauskas: I'm actually not sure about this one, so feel free to keep it if you feel it deserves a separate error type

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -145,6 +183,10 @@ export class LedgerKeyAgent extends KeyAgentBase {
     return await recoveredDeviceConnection.getVersion();
   }
 
+  /**
+   * @throws AuthenticationError
+   * @throws TransportError
```
mkazlauskas, line 188: :+1:

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -164,6 +206,46 @@ export class LedgerKeyAgent extends KeyAgentBase {
     });
   }
 
+  async signTransaction(
+    { body }: TxInternals,
+    { inputAddressResolver }: SignTransactionOptions
+  ): Promise<Cardano.Signatures> {
+    try {
+      const { accountIndex, networkId, communicationType, knownAddresses } = this
+        .serializableData as SerializableLedgerKeyAgentData;
```
mkazlauskas, line 215: These are already available as `this.member`. You should probably create a new `this.#communicationType` setting it in the constructor.

packages/wallet/src/KeyManagement/LedgerKeyAgent.ts
```diff
@@ -164,6 +206,46 @@ export class LedgerKeyAgent extends KeyAgentBase {
...
+      const cslTxBody = coreToCsl.txBody(body);
+      const addressHex = utils.buf_to_hex(utils.bech32_decodeAddress(this.knownAddresses[0].address.toString()));
+      const ledgerTxData = await txToLedger({
+        accountIndex,
+        addressHex,
```
mkazlauskas, line 220: This works when used with SingleAddessWallet, but in general it might have multiple addresses. You should probably remove this, so the mapper can check all knownAddresses to find if it's a change output.

packages/wallet/src/KeyManagement/errors/HwMappingError.ts
```diff
@@ -0,0 +1,12 @@
+import { CustomError } from 'ts-custom-error';
+
+const formatMessage = (detail?: string) => {
+  const messageDetail = detail ? `: ${detail}` : '';
+  return `Hardware date mapping failure${messageDetail}`;
```
mkazlauskas, line 5: I'm having trouble understanding what 'date' means here

packages/wallet/src/KeyManagement/util/mapHardwareSigningData.ts
```diff
@@ -0,0 +1,547 @@
...
+  const ledgerWithdrawals = [];
+  for (let i = 0; i < withdrawals.keys().len(); i++) {
+    const withdrawal = { stakeCredential: {} } as Withdrawal;
+    const rewardAddress = withdrawals.keys().get(i);
+    const paymentCredentials = rewardAddress.payment_cred();
+    const paymentCredentialsScriptHash = paymentCredentials.to_scripthash();
+    if (rewardAddress.payment_cred().kind() === 0) {
+      withdrawal.stakeCredential.type = StakeCredentialParamsType.KEY_PATH;
+      // Ignore! KeyStakeCredentialParams / ScriptStakeCredentialParams are not exported
+      // @ts-ignore
```
mkazlauskas, line 401: What are the errors for these @ts-ignores and are we sure the code is correct? Maybe just coerce type with `as any` instead?

packages/wallet/src/KeyManagement/util/mapHardwareSigningData.ts
```diff
@@ -0,0 +1,547 @@
...
+      for (let k = 0; k < assets.keys().len(); k++) {
+        const assetName = assets.keys().get(k);
+        const amount = assets.get(assetName);
+        const positiveAmount = amount?.as_positive()?.to_str();
+        const negativeAmount = amount?.as_negative()?.to_str();
+        if (!amount || !positiveAmount || !negativeAmount) {
+          throw new HwMappingError('Missing token amount.');
+        }
+        tokens.push({
+          amount: amount.is_positive() ? positiveAmount : negativeAmount,
```
mkazlauskas, line 442: I'm like 80% sure this is a bug
```suggestion
          amount: amount.is_positive() ? positiveAmount : ('-' + negativeAmount),
```

mkazlauskas (changes requested):
Looks great @tomislavhoracek! Let's resolve the new comments and add this key agent as an option for e2e tests. To do that:
1. Add `KEY_AGENT` variable to [.env](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/wallet/.env.example), with options being 'InMemory' or 'Ledger'.
2. Add a switch statement [here](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/wallet/test/e2e/config.ts#L117) to initialize the correct key agent depending on the new variable

packages/wallet/src/KeyManagement/util/mapHardwareSigningData.ts
tomislavhoracek: yep, you are right. It should be like here:
https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/core/src/CSL/cslToCore/cslToCore.ts#L113

mkazlauskas (changes requested):
Looks great! Still need to add a LedgerKeyAgent option for e2e tests

tomislavhoracek:
> Looks great! Still need to add a LedgerKeyAgent option for e2e tests

Aha, I'm working on e2e tests

mkazlauskas (dismissed):
Nice work @tomislavhoracek :tada: Please confirm e2e tests pass when used with ledger key agent

packages/wallet/src/KeyManagement/util/mapHardwareSigningData.ts
```diff
@@ -0,0 +1,567 @@
+// Based on Nami implementation
+// https://github.com/Berry-Pool/nami-wallet/blob/main/src/api/util.js
```
rhyslbw, line 2: Link to specific commit

rhyslbw (changes requested):
Looks good @tomislavhoracek. One minor issue, but otherwise it's approved assuming e2e tests are passing

packages/wallet/src/KeyManagement/util/mapHardwareSigningData.ts
tomislavhoracek: Thanks @rhyslbw . Comment updated and e2e with ledger are green.

mkazlauskas (approved):
:ok_hand:

PR #217, opened 2022-04-01
feat(cip30): implement ApiError{InternalError}

mkazlauskas:
# Context

Underlying wallet in cip30 throwing for any reason unspecified in the spec should result in ApiError{InternalError}

# Proposed Solution

try-catch in handleMessages


rhyslbw (approved):
Nice

PR #219, opened 2022-04-06
Feature/cip30 callbacks

mchappell:
# Context
https://input-output.atlassian.net/browse/ADP-1611

Introduces callback methods for both signData, and submitTx methods for cip30 api


packages/wallet/test/integration/cip30mapping.test.ts
```diff
@@ -35,7 +35,7 @@ describe('cip30', () => {
     let mappedWallet: WalletApi;
 
     beforeAll(async () => {
-      mappedWallet = cip30.createWalletApi(wallet);
+      mappedWallet = cip30.createWalletApi(wallet, () => Promise.resolve(true));
```
iadmytro, line 38: We also need case where user rejects the tx and verification that callback was called.

packages/wallet/src/cip30.ts
```diff
@@ -122,10 +148,22 @@ export const createWalletApi = (
...
   signData: async (addr: Cardano.Address, payload: Bytes): Promise<Cip30DataSignature> => {
     logger.debug('signData');
-    return cip30signData({
-      keyAgent: wallet.keyAgent,
-      payload: Cardano.util.HexBlob(payload),
-      signWith: addr
+    return confirmationCallback({
+      data: {
+        addr,
+        payload: Cardano.util.HexBlob(payload)
```
mkazlauskas, line 154: Could store payload in a variable to not run validation twice

packages/wallet/src/cip30.ts
```diff
@@ -122,10 +148,22 @@ export const createWalletApi = (
...
+      },
+      type: Cip30ConfirmationCallbackType.SignData
+    }).then(async (proceed: boolean) => {
```
mkazlauskas, line 157: Since this is already an async function, let's convert this (and signTx confirmation) to `await` syntax

packages/wallet/src/cip30.ts
```diff
@@ -154,15 +192,24 @@ export const createWalletApi = (
...
+    const txDecoded = CSL.Transaction.from_bytes(Buffer.from(tx, 'hex'));
+    const txData: Cardano.NewTxAlonzo = cslToCore.newTx(txDecoded);
+    return confirmationCallback({ data: txData, type: Cip30ConfirmationCallbackType.SubmitTx }).then(
+      async (proceed: boolean) => {
+        if (proceed) {
+          try {
+            await wallet.submitTx(txData);
+            return txData.id.toString();
+          } catch (error) {
+            logger.eror(error);
```
mkazlauskas, line 204: `eror`?

mkazlauskas (changes requested):
Looking good, just need some additional test coverage

packages/wallet/test/integration/cip30mapping.test.ts
```diff
@@ -106,6 +106,64 @@ describe('cip30', () => {
...
+    beforeAll(() => {
+      let onMessageHandler: any;
+      const chromeRuntime = (global as any).chrome.runtime;
+      chromeRuntime.onMessage.addListener.mockImplementation((handler: any) => (onMessageHandler = handler));
+      chromeRuntime.sendMessage.mockImplementation((_: any, message: any, callback: any) =>
+        onMessageHandler(message, null, callback)
+      );
+      cip30.initialize(
+        wallet,
+        jest
```
mkazlauskas, line 121: The tests would be clearer and less brittle if you set it to a variable, e.g.
```ts
describe('confirmation callbacks', () => {
  let confirmationCallback: jest.Mock;
  ...
```
Then you can set `mockResolvedValue` within the tests instead of beforeAll

packages/wallet/src/cip30.ts
```diff
@@ -154,14 +197,20 @@ export const createWalletApi = (
...
-      const txDecoded = CSL.Transaction.from_bytes(Buffer.from(tx, 'hex'));
-      const txData: Cardano.NewTxAlonzo = cslToCore.newTx(txDecoded);
-      await wallet.submitTx(txData);
-      return Promise.resolve(txData.id.toString());
-    } catch (error) {
-      logger.error(error);
-      throw error;
+    const txDecoded = CSL.Transaction.from_bytes(Buffer.from(tx, 'hex'));
+    const txData: Cardano.NewTxAlonzo = cslToCore.newTx(txDecoded);
+    const shouldProceed = await confirmationCallback({ data: txData, type: Cip30ConfirmationCallbackType.SubmitTx });
```
mkazlauskas, line 202: Actually this callback should be called in signTx. Don't think we need to guard against submitting. This will also change the data type from NewTxAlonzo to either NewTxBodyAlonzo or TxInternals

packages/wallet/test/integration/cip30mapping.test.ts
```diff
@@ -106,6 +106,64 @@ describe('cip30', () => {
...
+      test('signData confirm callback resolves', async () => {
+        await expect(() => api.signData(signDataParams.addr, signDataParams.payload)).resolves.not.toThrow();
+      });
+
+      test('signData confirm callback rejects', async () => {
+        await expect(() => api.signData(signDataParams.addr, signDataParams.payload)).rejects.toThrow();
+      });
+    });
+
+    describe('submitTx', () => {
```
mkazlauskas, line 148:
```suggestion
    describe('signTx', () => {
```

packages/wallet/test/integration/cip30mapping.test.ts
```diff
@@ -106,6 +106,64 @@ describe('cip30', () => {
...
+
+        cslTx = Buffer.from(coreToCsl.tx(finalizedTx).to_bytes()).toString('hex');
+      });
+
+      test('submitTx confirm callback resolves', async () => {
+        await expect(() => api.submitTx(cslTx)).resolves.not.toThrow();
+      });
+
+      test('submitTx confirm callback rejects', async () => {
+        await expect(() => api.submitTx(cslTx)).rejects.toThrow();
```
mkazlauskas, line 162:
```suggestion
        await expect(() => api.submitTx(cslTx)).rejects.toThrowError(TxSignError);
```

mkazlauskas (changes requested):

packages/wallet/src/cip30.ts
mchappell: Don't we need both?
**SubmitTx**
> If the wallet accepts the transaction and tries to send it,
**SignTx**
> The wallet should ask the user for permission,

Sounds like `signData`, `signTx`, and `submitTx` should all have confirmation callbacks for them as they are user-led actions, what do you think to me adding another instead of simply switching?

mkazlauskas: Yeah adding a another sounds right

packages/wallet/src/cip30.ts
```diff
@@ -16,13 +19,49 @@ import { Logger, dummyLogger } from 'ts-log';
...
+  data: Cardano.NewTxBodyAlonzo;
+};
+
+type SubmitTxCallbackParams = {
+  type: Cip30ConfirmationCallbackType.SubmitTx;
+  // Sending the entire tx allows for future extensibility of FE without API modification
+  data: Cardano.NewTxAlonzo;
+};
+
+type CallbackConfirmation = (
```
mkazlauskas, line 53: Params types would be good to export too
```suggestion
export type CallbackConfirmation = (
```

packages/wallet/test/integration/cip30mapping.test.ts
```diff
@@ -99,71 +101,112 @@ describe('cip30', () => {
...
-      let cslTx: string;
+    describe('signTx', () => {
+      let hexTxBody: string;
       beforeAll(async () => {
         const txInternals = await wallet.initializeTx(simpleTxProps);
         const finalizedTx = await wallet.finalizeTx(txInternals);
+        hexTxBody = Buffer.from(coreToCsl.tx(finalizedTx).body().to_bytes()).toString('hex');
+      });
+
+      /* eslint-disable sonarjs/no-duplicate-string */
```
mkazlauskas, line 158: Could just disable for entire file

mkazlauskas (changes requested):
Nice tests!

PR #215, opened 2022-03-31
feat(cardano-services): stake pool search http server

jvieiro:
# Context
Implement stake pool search microservice. setup http server. add mainly queries to implement stake pool search

# Proposed Solution

- [x] Setup new service at `src/StakePoolSearch` , similar to `src/TxSubmit` . These are separate microservices, so must remain independent.

- [x] Implement `DbSyncStakePoolSearchProvider` to resolve from PostgreSQL

- [x] Implement  `StakePoolSearchHttpServer` , using the `StakePoolSearchProvider` interface as a dependency.

- [x] Add db snapshot in order to tests queries

# Important Changes Introduced
n/a

PR #220, opened 2022-04-07
feat: generic http provider client

mkazlauskas:
# Context

Need to implement a http provider client for every type of provider

# Proposed Solution

- [x] feat(cardano-graphql-services): add generic http provider client that uses Proxy
- [x] refactor(cardano-services-client)!: reimplement txSubmitHttpProvider using createHttpProvider
  - fix TxSubmissionError deserialization into correct error type
  - wrap submitTx errors in ProviderError, simplify to always return 1 obj
  - move serialize-error to devDependencies
- [x] feat(cardano-services-client): add stakePoolSearchHttpProvider
- [x] refactor TxSubmitHttpServer to be compatible with the new generic HttpServer
  - feat(cardano-services): add HttpServer.sendJSON as a generic util that all http provider servers can use. 
    - Update base HttpServer to use core fromSerializableObject when deserializing req body.
  - refactor(cardano-services)!: make TxSubmitHttpServer compatible with createHttpProvider<T>
    - move body size validation to HttpServer
    - add providerHandler util to validate args presence in req body

- [x] chore: rm unused .gitmodules
- [x] chore: disable 'jsdoc/require-returns-type' eslint rule
- [x] refactor: hoist pouchdb obj processing to core as serializableObject
- [x] refactor(core): rm UnknownTxSubmissionError from TxSubmissionErrors
  - add ProviderFailure.ConnectionFailure and ProviderFailure.BadRequest
  - refactor: update UnknownTxSubmissionError usages

### DIFF
--- a/packages/cardano-services-client/src/StakePoolSearchProvider/stakePoolSearchHttpProvider.ts
+++ b/packages/cardano-services-client/src/StakePoolSearchProvider/stakePoolSearchHttpProvider.ts
@@ -1,0 +1,20 @@
+import { HttpProviderConfigPaths, createHttpProvider } from '../HttpProvider';
+import { StakePoolSearchProvider } from '@cardano-sdk/core';
+
+export const defaultStakePoolSearchProviderPaths: HttpProviderConfigPaths<StakePoolSearchProvider> = {
+  queryStakePools: '/search'
+};
+
+/**
+ * Connect to a StakePoolSearchHttpServer instance
+ *
+ * @param {string} baseUrl server root url, w/o trailing /
+ */
+export const stakePoolSearchHttpProvider = (
+  baseUrl: string,
+  paths = defaultStakePoolSearchProviderPaths
+): StakePoolSearchProvider =>
+  createHttpProvider<StakePoolSearchProvider>({
+    baseUrl,
+    paths
+  });
--- a/packages/cardano-services/src/Http/HttpServer.ts
+++ b/packages/cardano-services/src/Http/HttpServer.ts
@@ -2,7 +2,7 @@ import { HealthCheckResponse, ProviderError, util } from '@cardano-sdk/core';
 import { Logger, dummyLogger } from 'ts-log';
 import { RunnableModule } from '../RunnableModule';
 import { listenPromise, serverClosePromise } from '../util';
-import bodyParser from 'body-parser';
+import bodyParser, { Options } from 'body-parser';
 import express from 'express';
 import expressPromBundle from 'express-prom-bundle';
 import http from 'http';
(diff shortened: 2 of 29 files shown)


packages/cardano-services-client/src/StakePoolSearchProvider/stakePoolSearchHttpProvider.ts
```diff
@@ -0,0 +1,20 @@
+import { HttpProviderConfigPaths, createHttpProvider } from '../HttpProvider';
+import { StakePoolSearchProvider } from '@cardano-sdk/core';
+
+export const defaultStakePoolSearchProviderPaths: HttpProviderConfigPaths<StakePoolSearchProvider> = {
+  queryStakePools: '/search'
+};
+
+/**
+ * Connect to a TxSubmitHttpServer instance
```
jvieiro, line 9:
```suggestion
 * Connect to a StakePoolSearchHttpServer instance
```

jvieiro:
Looks really good!

packages/cardano-services/src/Http/HttpServer.ts
```diff
@@ -41,7 +44,10 @@ export abstract class HttpServer extends RunnableModule {
     this.app = express();
     // must use this before router
     this.app.use(
-      bodyParser.json({ reviver: (key, value) => (key === '' ? util.fromSerializableObject(value) : value) })
+      bodyParser.json({
+        limit: this.config.bodyParser?.limit || '500kB',
+        reviver: (key, value) => (key === '' ? util.fromSerializableObject(value) : value)
```
jvieiro, line 49: This means that only the last attribute of the rquest will be parsed by `fromSerializableObject`?

jvieiro (approved):

packages/cardano-services/src/Http/HttpServer.ts
mkazlauskas: I don't think so, tested this:
```js
JSON.parse('{"1": 1, "2": 2, "3": {"4": 4, "5": {"6": 6}}}', (key, value) => {
  if (key === '') console.log(`"${key}"`, value); // log the current property name, the last is "".
  return value;     // return the unchanged property value.
});
```
prints `"" {1: 1, 2: 2, 3: {…}}`

It's the last call, but as I understand it's for mapping the entire object.

iadmytro (approved):

rhyslbw (approved):
Great refactor @mkazlauskas

mkazlauskas:
@rhyslbw should we just close this and merge the changes together with #215 ?

rhyslbw:
@mkazlauskas Yep, should be fine

mkazlauskas:
Will be merged with #215

PR #223, opened 2022-04-12
refactor(cip30): remove storage.local dependency

mchappell:
# Context

The lightwallet application breaks if the `webextension-polyfill` is imported in anything other than the popup

# Proposed Solution
Remove the default import of this package and instead pass in as a required prop


mkazlauskas (changes requested):
Looks good, but maybe let's apply the same changes and inject `browser.runtime` to `handleMessages`

mchappell:
> Looks good, but maybe let's apply the same changes and inject `browser.runtime` to `handleMessages`

@mkazlauskas Applied this through the cip30 flow now

mkazlauskas (approved):
LGTM :+1:

PR #224, opened 2022-04-12
feat(core): add tx inspector

lgobbi-atix:
# Context

Added a transaction inspection feature to differentiate between transaction types

# Proposed Solution

Create util function `createTxInspector` in `core` package that receives inspector functions to identify different operations in a transaction.
Create inspector functions for sending, receiving, delegation, stake key registration, stake key registration transactions and withdrawals.

### DIFF
--- a/packages/core/src/util/txInspector.ts
+++ b/packages/core/src/util/txInspector.ts
@@ -1,0 +1,113 @@
+import {
+  Address,
+  CertificateType,
+  Lovelace,
+  StakeAddressCertificate,
+  StakeDelegationCertificate,
+  TxAlonzo,
+  Value
+} from '../Cardano';
+import { BigIntMath } from './BigIntMath';
+import { coalesceValueQuantities } from '../Cardano/util';
+import { isAddressWithin, isOutgoing } from '../Address/util';
+
+type Inspector<Inspection> = (tx: TxAlonzo) => Inspection;
+type Inspectors = { [k: string]: Inspector<unknown> };
+type TxInspector<T extends Inspectors> = (tx: TxAlonzo) => {
+  [k in keyof T]: ReturnType<T[k]>;
+};
+
+// Inspectors result types
+export type SendReceiveValueInspection = Value;
+export type DelegationInspection = StakeDelegationCertificate[];
+export type StakeKeyRegistrationInspection = StakeAddressCertificate[];
+export type WithdrawalInspection = Lovelace;
+
+/**
+ * Inspects a transaction for value (coins + assets) sent by the provided addresses.
+ *
+ * @param {Address[]} ownAddresses own wallet's addresses
+ * @returns {Value} total value sent
+ */
+export const valueSentInspector =
+  (ownAddresses: Address[]): Inspector<Value> =>
+  (tx: TxAlonzo): SendReceiveValueInspection => {
+    if (!isOutgoing(tx, ownAddresses)) return { coins: 0n };
+    const sentOutputs = tx.body.outputs.filter((out) => !isAddressWithin(ownAddresses)(out));
+    return coalesceValueQuantities(sentOutputs.map((output) => output.value));
+  };
+
+/**
+ * Inspects a transaction for value (coins + assets) received by the provided addresses.
+ *
+ * @param {Address[]} ownAddresses own wallet's addresses
+ * @returns {Value} total value received
+ */
+export const valueReceivedInspector =
+  (ownAddresses: Address[]): Inspector<Value> =>
+  (tx: TxAlonzo): SendReceiveValueInspection => {
+    if (isOutgoing(tx, ownAddresses)) return { coins: 0n };
+    const receivedOutputs = tx.body.outputs.filter((out) => isAddressWithin(ownAddresses)(out));
+    return coalesceValueQuantities(receivedOutputs.map((output) => output.value));
+  };
+
+/**
+ * Inspects a transaction for a stake delegation certificate.
+ *
+ * @param {TxAlonzo} tx transaction to inspect
+ * @returns {DelegationInspection} array of delegation certificates
+ */
+export const delegationInspector: Inspector<DelegationInspection> = (tx: TxAlonzo) =>
+  (tx.body.certificates?.filter(
+    (cert) => cert.__typename === CertificateType.StakeDelegation
+  ) as StakeDelegationCertificate[]) ?? [];
+
+/**
+ * Inspects a transaction for a stake key registration certificate.
+ *
+ * @param {TxAlonzo} tx transaction to inspect
+ * @returns {StakeKeyRegistrationInspection} array of stake key registration certificates
+ */
+export const stakeKeyRegistrationInspector: Inspector<StakeKeyRegistrationInspection> = (tx: TxAlonzo) =>
+  (tx.body.certificates?.filter(
+    (cert) => cert.__typename === CertificateType.StakeKeyRegistration
+  ) as StakeAddressCertificate[]) ?? [];
+
+/**
+ * Inspects a transaction for a stake key deregistration certificate.
+ *
+ * @param {TxAlonzo} tx transaction to inspect
+ * @returns {StakeKeyRegistrationInspection} array of stake key deregistration certificates
+ */
+export const stakeKeyDeregistrationInspector: Inspector<StakeKeyRegistrationInspection> = (tx: TxAlonzo) =>
+  (tx.body.certificates?.filter(
+    (cert) => cert.__typename === CertificateType.StakeKeyDeregistration
+  ) as StakeAddressCertificate[]) ?? [];
+
+/**
+ * Inspects a transaction for withdrawals.
+ *
+ * @param {TxAlonzo} tx transaction to inspect
+ * @returns {WithdrawalInspection} accumulated withdrawal quantities
+ */
+export const withdrawalInspector: Inspector<WithdrawalInspection> = (tx: TxAlonzo) =>
+  tx.body.withdrawals?.length ? BigIntMath.sum(tx.body.withdrawals.map(({ quantity }) => quantity)) : 0n;
+
+/**
+ * Returns a function to convert lower level transaction data to a higher level object, using the provided inspectors.
+ *
+ * @param {Inspectors} inspectors inspector functions scoped to a domain concept.
+ */
+export const createTxInspector =
+  <T extends Inspectors>(inspectors: T): TxInspector<T> =>
+  (tx: TxAlonzo) =>
+    Object.keys(inspectors).reduce(
+      (result, key) => {
+        const inspector = inspectors[key];
+        result[key as keyof T] = inspector(tx) as ReturnType<T[keyof T]>;
+        return result;
+      },
+      {} as {
+        [k in keyof T]: ReturnType<T[k]>;
+      }
+    );
--- a/packages/core/test/Address/util.test.ts
+++ b/packages/core/test/Address/util.test.ts
@@ -1,18 +1,69 @@
-import { Address } from '../../src';
-
-jest.mock('../../src/CSL/parseCslAddress');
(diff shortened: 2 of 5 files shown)


packages/core/src/util/txInspector.ts
```diff
@@ -0,0 +1,145 @@
...
+      for (const [asset, amount] of output.value.assets) {
+        const total = totalAssets.get(asset) ?? 0n;
+        totalAssets.set(asset, total + amount);
+      }
+    }
+  }
+
+  if (totalAssets.size > 0) result.assets = totalAssets;
+  return result;
+};
```
lgobbi-atix, line 38: Could these methods be moved somewhere else? `isMyAddress` and `isOutgoing` were pretty much copied from `TransactionsTracker`

packages/core/src/util/txInspector.ts
```diff
@@ -0,0 +1,145 @@
...
+ * Inspects a transaction for a stake delegation certificate.
+ * Returns `null` if the transaction doesn't have one.
+ *
+ * @param {TxAlonzo} tx transaction to inspect
+ */
+export const delegationInspector = (tx: TxAlonzo): DelegateInspector => {
+  const delegation = tx.body.certificates?.find((cert) => cert.__typename === CertificateType.StakeDelegation) as
+    | StakeDelegationCertificate
+    | undefined;
+  return delegation ? { stakeKeyHash: delegation.stakeKeyHash, stakePool: delegation.poolId } : null;
```
lgobbi-atix, line 99: Could a tx have more than one certificate of the same type? If so, should we return an array instead?

packages/core/src/util/txInspector.ts
```diff
@@ -0,0 +1,145 @@
...
+  PoolId,
+  StakeAddressCertificate,
+  StakeDelegationCertificate,
+  TokenMap,
+  TxAlonzo,
+  TxOut,
+  Value
+} from '../Cardano';
+
+const isMyAddress =
```
rhyslbw, line 15:
```suggestion
const isAddressWithin =
```
The function isn't concerned with ownership, all it's doing is checking for presence in the given array.

packages/core/src/util/txInspector.ts
```diff
@@ -0,0 +1,145 @@
...
+  (addresses: Address[]) =>
+  ({ address }: { address: Address }) =>
+    addresses.includes(address!);
+
+const isOutgoing = (tx: TxAlonzo, ownAddresses: Address[]): boolean => tx.body.inputs.some(isMyAddress(ownAddresses));
```
rhyslbw, line 20:
```suggestion
const isOutgoing = (tx: TxAlonzo, ownAddresses: Address[]): boolean => tx.body.inputs.some(isAddressWithin(ownAddresses));
```
... an _address within_ the wallet's _own addresses_  can be interpreted as "my address"

packages/core/src/util/txInspector.ts
rhyslbw: `Address/util.ts`

packages/core/src/util/txInspector.ts
rhyslbw: Yes, return type should be an array

rhyslbw (changes requested):

packages/core/src/util/txInspector.ts
```diff
@@ -0,0 +1,145 @@
...
+};
+
+type InspectionResult = { [k: string]: unknown } | null;
+type InspectTx = (tx: TxAlonzo) => InspectionResult;
+type Inspectors = { [k: string]: InspectTx };
+type TxInspector<T extends Inspectors> = (tx: TxAlonzo) => {
+  [k in keyof T]: ReturnType<T[k]>;
+};
+
+export type SendReceiveCoinInspector = {
```
mkazlauskas, line 47: It's not just coin, but also assets. Also think the 'inspector' is a function, and this is a result?
I think returning `{coins: 0n}` is nice than null for this inspector.
```suggestion
export type SendReceiveValue = Cardano.Value;
```

packages/core/src/util/txInspector.ts
```diff
@@ -0,0 +1,145 @@
...
+  stakeKeyHash: Ed25519KeyHash;
+} | null;
+
+/**
+ * Inspects a transaction for coins and assets sent from the provided addresses.
+ * Returns `null` if it's not sending anything.
+ *
+ * @param {Address[]} ownAddresses own wallet's addresses
+ */
+export const sendCoinInspector =
```
mkazlauskas, line 67:
```suggestion
export const sendValueInspector =
```

packages/core/src/util/txInspector.ts
```diff
@@ -0,0 +1,145 @@
...
+  (ownAddresses: Address[]) =>
+  (tx: TxAlonzo): SendReceiveCoinInspector => {
+    if (isOutgoing(tx, ownAddresses)) return null;
+    const receivedOutputs = tx.body.outputs.filter((out) => isMyAddress(ownAddresses)(out));
+    return getTotalValue(receivedOutputs);
+  };
+
+/**
+ * Inspects a transaction for a stake delegation certificate.
+ * Returns `null` if the transaction doesn't have one.
```
mkazlauskas, line 91: This looks nicer in generated docs
```suggestion
 * @returns `null` if the transaction doesn't have one.
```

mkazlauskas (changes requested):

packages/core/src/util/txInspector.ts
rhyslbw: I vote for just typing the return value from the inspector as `Cardano.Value`

packages/core/test/util/txInspector.test.ts
```diff
@@ -0,0 +1,186 @@
...
+      ])
+    );
+  });
+
+  test('a delegation transaction should return the pool id and key hash', () => {
+    const tx = buildMockTx({
+      certificates: [delegationCert]
+    });
+    const inspectTx = createTxInspector({
+      delegate: delegationInspector,
```
rhyslbw, line 141:
```suggestion
      delegation: delegationInspector,
```

packages/core/src/util/txInspector.ts
```diff
@@ -0,0 +1,145 @@
...
+  const registration = tx.body.certificates?.find(
+    (cert) => cert.__typename === CertificateType.StakeKeyDeregistration
+  ) as StakeAddressCertificate | undefined;
+  return registration ? { stakeKeyHash: registration.stakeKeyHash } : null;
+};
+
+/**
+ * Returns a function to inspect a transaction depending on the inspectors provided on creation
+ *
+ * @param {Inspectors} inspectors functions with the logic to process and inspect a transaction
```
rhyslbw, line 131:
```suggestion
 * @param {Inspectors} inspector functions scoped to a domain concept. 
```

packages/core/src/util/txInspector.ts
```diff
@@ -0,0 +1,145 @@
...
+  stakePool: PoolId;
+  stakeKeyHash: Ed25519KeyHash;
+} | null;
+
+export type StakeKeyRegistrationInspector = {
+  stakeKeyHash: Ed25519KeyHash;
+} | null;
+
+/**
+ * Inspects a transaction for coins and assets sent from the provided addresses.
```
rhyslbw, line 62:
```suggestion
 * Inspects a transaction for values sent from the provided addresses.
```

packages/core/src/util/txInspector.ts
rhyslbw: Please also add docblocks after hoisting

rhyslbw (changes requested):
Fundamentally this looks good @lgobbi-atix, just some tweaks for polish.

It would be good to also add nested `describe` blocks in the test for each inspector, so we can extend the test suite with multiple scenarios without getting lost. It will also help to keep the test case names shorter.

e.g

``` TypeScript
describe('delegationInspector', () => {
    test('returns the pool id and key hash when a delegation certificate is found', () => {
    
    test('returns null when no delegation certificates are found', () => {
    
```

packages/core/src/util/txInspector.ts
```diff
@@ -0,0 +1,145 @@
...
+        totalAssets.set(asset, total + amount);
+      }
+    }
+  }
+
+  if (totalAssets.size > 0) result.assets = totalAssets;
+  return result;
+};
+
+type InspectionResult = { [k: string]: unknown } | null;
```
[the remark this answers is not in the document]
mkazlauskas: Also, is the `| null` part really needed here?

packages/core/src/util/txInspector.ts
lgobbi-atix: @rhyslbw @mkazlauskas In case there are no certificates and considering we are removing `null` for the send/receive ones, what do you think it's best to return: `null` or just an empty array?

mkazlauskas: I think both options are ok, just go with whichever you think would be easier or more intuitive to consume in the wallet

packages/core/src/util/txInspector.ts
lgobbi-atix: Done in 19ae735f102d9a785ed2ab7a84f055380eb1ce77, ended up removing `getTotalValue` because I realized that `coalesceValueQuantities` exists

packages/core/src/util/txInspector.ts
lgobbi-atix: Done in 19ae735f102d9a785ed2ab7a84f055380eb1ce77, removed all nulls

packages/core/src/util/txInspector.ts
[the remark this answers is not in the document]
lgobbi-atix: Ended up removing this type in 19ae735f102d9a785ed2ab7a84f055380eb1ce77

packages/core/test/Address/util.test.ts
```diff
@@ -1,19 +1,70 @@
-import { Address } from '../../src';
-
-jest.mock('../../src/CSL/parseCslAddress');
-const { parseCslAddress } = jest.requireMock('../../src/CSL/parseCslAddress');
+import * as parseCslAddress from '../../src/CSL/parseCslAddress';
+import { Address, Cardano } from '../../src';
+import { CSL as SerializationLib } from '../../src/CSL';
 
 describe('Address', () => {
+  const parseCslAddressSpy = jest.spyOn(parseCslAddress, 'parseCslAddress');
```
lgobbi-atix, line 6: I had to modify the way it's mocked so I could use the real implementation in the new tests

packages/core/src/util/txInspector.ts
```diff
@@ -0,0 +1,113 @@
...
+
+type Inspector<Inspection> = (tx: TxAlonzo) => Inspection;
+type Inspectors = { [k: string]: Inspector<unknown> };
+type TxInspector<T extends Inspectors> = (tx: TxAlonzo) => {
+  [k in keyof T]: ReturnType<T[k]>;
+};
+
+// Inspectors results types
+export type SendReceiveValueInspection = Value;
+export type DelegateInspection = StakeDelegationCertificate[];
```
rhyslbw, line 22:
```suggestion
export type DelegationInspection = StakeDelegationCertificate[];
```
Sorry, I missed this one on the first review

rhyslbw (changes requested):

packages/core/src/util/txInspector.ts
rhyslbw: done in [c1a8d4f](https://github.com/input-output-hk/cardano-js-sdk/pull/224/commits/c1a8d4fb1567f0971a5291cefa0eccf3c2a125bd)

mkazlauskas (dismissed):
Nice work! :tada:

mkazlauskas (approved):

rhyslbw (approved):
Great work @lgobbi-atix

PR #228, opened 2022-04-18
feat(wallet): add unspendable utxo observable and store

lgobbi-atix:
# Context

Allow wallet to set unspendable UTxO set, which won't be present in available UTxO or balance for input selection.

Jira Issue: [ADP-1647](https://input-output.atlassian.net/browse/ADP-1647)

# Proposed Solution

- Add `unspendable$` at `UtxoTracker` and `BalanceTracker`
- Add `setUnspendable(utxo: Cardano.Utxo[])` at `UtxoTracker` to update store values
- Filter out unspendable utxo in `available$` utxo in `UtxoTracker`

- `UtxoTracker` now has `unspendableStore` and `unspendableUtxoSource$` arguments
- `PersistentCollectionTrackerSubject.next` now also sets all store values

### DIFF
--- a/packages/wallet/src/SingleAddressWallet.ts
+++ b/packages/wallet/src/SingleAddressWallet.ts
@@ -188,7 +188,7 @@ export class SingleAddressWallet implements Wallet {
     this.utxo = createUtxoTracker({
       addresses$,
       retryBackoffConfig,
-      store: stores.utxo,
+      stores,
       tipBlockHeight$,
       transactionsInFlight$: this.transactions.outgoing.inFlight$,
       walletProvider: this.walletProvider
--- a/packages/wallet/src/services/UtxoTracker.ts
+++ b/packages/wallet/src/services/UtxoTracker.ts
@@ -1,14 +1,14 @@
 import { Cardano, WalletProvider } from '@cardano-sdk/core';
-import { CollectionStore } from '../persistence';
-import { Observable, combineLatest, map, switchMap } from 'rxjs';
+import { NEVER, Observable, combineLatest, concat, map, of, switchMap } from 'rxjs';
 import { PersistentCollectionTrackerSubject, TrackerSubject, coldObservableProvider, utxoEquals } from './util';
 import { RetryBackoffConfig } from 'backoff-rxjs';
-import { TransactionalTracker } from './types';
+import { UtxoTracker } from './types';
+import { WalletStores } from '../persistence';
 
 export interface UtxoTrackerProps {
   walletProvider: WalletProvider;
   addresses$: Observable<Cardano.Address[]>;
-  store: CollectionStore<Cardano.Utxo>;
+  stores: Pick<WalletStores, 'utxo' | 'unspendableUtxo'>;
   transactionsInFlight$: Observable<Cardano.NewTxAlonzo[]>;
   tipBlockHeight$: Observable<number>;
   retryBackoffConfig: RetryBackoffConfig;
@@ -16,6 +16,7 @@ export interface UtxoTrackerProps {
 
 export interface UtxoTrackerInternals {
   utxoSource$?: PersistentCollectionTrackerSubject<Cardano.Utxo>;
+  unspendableUtxoSource$?: PersistentCollectionTrackerSubject<Cardano.Utxo>;
 }
 
 export const createUtxoProvider = (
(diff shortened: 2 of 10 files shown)


packages/wallet/src/services/types.ts
```diff
@@ -18,10 +18,12 @@ export interface Balance extends Cardano.Value {
 export interface TransactionalObservables<T> {
   total$: BehaviorObservable<T>;
   available$: BehaviorObservable<T>;
+  reserved$: BehaviorObservable<T>;
 }
 
 export interface TransactionalTracker<T> extends TransactionalObservables<T> {
   shutdown(): void;
+  setReserved(utxo: Cardano.Utxo[]): void;
 }
```
lgobbi-atix, line 27: Changes in `TransactionalTracker` and `TransactionalObservables` also affect `BalanceTracker`, but I don't think we need `reserved$` and `setReserved` there, right? Should I make a new interface for `UtxoTracker` that extends `TransactionalTracker` instead?

rhyslbw: The concept of reserved UTxOs is relevant, as it reveals the spendable balance

mkazlauskas: I think we should have the `TransactionalObservables.unspendable$` since it's relevant for both balance and utxo.

Since unspendable balance is computed from unspendable utxo, `setUnspendable` should only be a method on UtxoTracker.

packages/wallet/src/services/UtxoTracker.ts
```diff
@@ -8,14 +8,15 @@ import { TransactionalTracker } from './types';
 export interface UtxoTrackerProps {
   walletProvider: WalletProvider;
   addresses$: Observable<Cardano.Address[]>;
-  store: CollectionStore<Cardano.Utxo>;
+  store: { utxo: CollectionStore<Cardano.Utxo>; reservedUtxo: CollectionStore<Cardano.Utxo> };
```
mkazlauskas, line 11: How about
```suggestion
  stores: Pick<WalletStores, 'utxo' | 'reservedUtxo'>;
```

packages/wallet/src/services/UtxoTracker.ts
```diff
@@ -40,28 +41,34 @@ export const createUtxoTracker = (
   {
     utxoSource$ = new PersistentCollectionTrackerSubject<Cardano.Utxo>(
       () => createUtxoProvider(walletProvider, addresses$, tipBlockHeight$, retryBackoffConfig),
-      store
-    )
+      store.utxo
+    ),
+    reservedUtxoSource$ = new TrackerSubject(store.reservedUtxo.getAll())
```
mkazlauskas, line 46: `getAll()` will emit 0 or 1 times and complete. You need to `concat` it with something that emits when you `setReserved()`

mkazlauskas (changes requested):

packages/wallet/src/services/UtxoTracker.ts
lgobbi-atix: I'm a bit lost here, I don't understand what I need to do to stop it from completing. 
> Concat it with another Subject, calling .next() on setReserved

If I go with this option, what can I concat it with? 

and if I use `PersistentCollectionTrackerSubject`, what should I use as the `source` if I only have the store? You mentioned a refactor, but I don't know what I should do besides adding
```ts
next(value: T[]): void {
  this.store.setAll(value);
}
```

packages/wallet/test/services/util/persistentTrackerSubjects.test.ts
```diff
@@ -44,7 +44,7 @@ describe('PersistentCollectionTrackerSubject', () => {
       const tracker$ = new PersistentCollectionTrackerSubject<string>(() => source$, store);
       expectObservable(tracker$.asObservable()).toBe('a-b--c', { a, b, c });
     });
-    expect(store.setAll).toBeCalledTimes(2);
+    expect(store.setAll).toBeCalledTimes(4);
```
lgobbi-atix, line 47: This is because `next` now also calls `store.setAll`

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -205,7 +205,7 @@ export class SingleAddressWallet implements Wallet {
     this.utxo = createUtxoTracker({
       addresses$,
       retryBackoffConfig,
-      store: stores.utxo,
+      store: { unspendableUtxo: stores.unspendableUtxo, utxo: stores.utxo },
```
mkazlauskas, line 208: Let's rename this since it's multiple stores. Could as well just use the existing `stores` object instead of creating a new one then.
```suggestion
      stores,
```

packages/wallet/src/services/util/persistentTrackerSubjects.ts
```diff
@@ -28,6 +29,11 @@ export class PersistentCollectionTrackerSubject<T> extends TrackerSubject<T[]> {
         })
       )
     );
+    this.store = store;
+  }
+  next(value: T[]): void {
+    this.store && this.store.setAll(value);
```
mkazlauskas, line 35: If we're doing this in .next(), maybe we can:
- remove `tap(store.setAll.bind(store))` in the constructor
- get rid of [unspendableUtxoSubject$](https://github.com/input-output-hk/cardano-js-sdk/pull/228/files#diff-b2b947493b30368269c8ae9c56fb7723a623e8c10f22c17a4d8da41bda21e33bR49) (use [NEVER](https://rxjs.dev/api/index/const/NEVER) instead) and in `setUnspendable` just call `.next` on persistent subject directly

packages/wallet/test/services/UtxoTracker.test.ts
```diff
@@ -30,18 +30,90 @@ describe('createUtxoTracker', () => {
...
       expectObservable(utxoTracker.total$).toBe('a---|', { a: utxo });
+      expectObservable(utxoTracker.unspendable$).toBe('-a-b|', { a: [], b: [utxo[0]] });
       expectObservable(utxoTracker.available$).toBe('-a-b|', {
         a: utxo,
         b: utxo.slice(1)
       });
     });
   });
+
+  it('sets unspendable utxos and locks them', () => {
```
mkazlauskas, line 91: :+1:

mkazlauskas (changes requested):

packages/wallet/test/services/util/persistentTrackerSubjects.test.ts
mkazlauskas: Suggestion in [this comment](https://github.com/input-output-hk/cardano-js-sdk/pull/228#discussion_r854838987) might help

packages/wallet/src/services/UtxoTracker.ts
```diff
@@ -36,33 +37,44 @@ export const createUtxoProvider = (
...
-  { walletProvider, addresses$, store, transactionsInFlight$, retryBackoffConfig, tipBlockHeight$ }: UtxoTrackerProps,
+  { walletProvider, addresses$, stores, transactionsInFlight$, retryBackoffConfig, tipBlockHeight$ }: UtxoTrackerProps,
   {
     utxoSource$ = new PersistentCollectionTrackerSubject<Cardano.Utxo>(
       () => createUtxoProvider(walletProvider, addresses$, tipBlockHeight$, retryBackoffConfig),
-      store
+      stores.utxo
+    ),
+    unspendableUtxoSource$ = new PersistentCollectionTrackerSubject(
+      (stored) => (stored.length > 0 ? NEVER : concat(of([]), NEVER)),
```
mkazlauskas, line 47: I had this in mind:
```suggestion
      () => NEVER
```
But I guess you'd run into the same null value problem?

mkazlauskas (approved):
LGTM! Could you please also verify that `yarn test:e2e` from `wallet` package is still passing? You might need to create a `.env` with your own secrets to do that

packages/wallet/src/services/UtxoTracker.ts
lgobbi-atix: Yes, I'm getting the same errors as I had before when running the tests

lgobbi-atix:
> LGTM! Could you please also verify that `yarn test:e2e` from `wallet` package is still passing? You might need to create a `.env` with your own secrets to do that

They pass 
```
Test Suites: 5 passed, 5 total
Tests:       7 passed, 7 total
```

rhyslbw (approved):
Looks great, well done @lgobbi-atix and @mkazlauskas

PR #229, opened 2022-04-20
feat: additional output validation

mkazlauskas:
# Context

Wallets might want to validate some transaction properties before building it. We currently have `Wallet.validateInitializeTxProps`, which was originally designed as a general validation function that could be called prior to calling `Wallet.initializeTx` to ensure that `initializeTx` will succeed, as well as providing additional context about any validation errors. Problem with this approach is that one of the potential errors is the overall transaction size being too large. To perform this validation, you have to run input selection algorithm, which has randomness involved and could therefore select a different number of inputs, resulting in a different transaction size when you later call `initializeTx`. Therefore, it is way more practical to keep validation of transaction size as an exception of `initializeTx` (currently as `MaximumInputCountExceeded` error).

# Proposed Solution

- Since it's not practical to implement assurance of valid InitializeTxProps prior to actually calling `initializeTx`, get rid of `Wallet.validateInitializeTxProps` by splitting it into more atomic validation functions, moving it under `Wallet.util`:
  - Wallet.util.validateOutput
  - Wallet.util.validateOutputs
- Validate token bundle (value) size of an output
- In addition, hoist `SingleAddressWallet.resolveInputAddress` to `Wallet.util.resolveInputAddress` since it fits nicely under this newly created scope, simplifying the root Wallet interface a little bit.

- rename type of validation result to OutputValidation
- rename InputAddressResolver to ResolveInputAddress since it's a function, not an object
- [fix(cip2): computeSelectionLimit constraint logic error](https://github.com/input-output-hk/cardano-js-sdk/commit/cbcf9b5ee23964389f671cdab835c47a72427e1b)

### DIFF
--- a/packages/cip2/src/selectionConstraints.ts
+++ b/packages/cip2/src/selectionConstraints.ts
@@ -67,7 +67,7 @@ const getTxSize = (tx: CSL.Transaction) => tx.to_bytes().length;
  * that adjust selection based on selection limit. RRRI implementation uses this after selecting all the inputs
  * and throws MaximumInputCountExceeded if the constraint returns a limit higher than number of selected utxo.
  *
- * @returns {ComputeSelectionLimit} constraint that returns txSize <= maxTxSize ? utxo[].length : utxo[].length+1
+ * @returns {ComputeSelectionLimit} constraint that returns txSize <= maxTxSize ? utxo[].length : utxo[].length-1
  */
 export const computeSelectionLimit =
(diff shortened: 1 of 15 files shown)


rhyslbw (changes requested):
Just a couple of JSDoc warnings. Otherwise is a good refactor :+1:

mkazlauskas:
> Just a couple of JSDoc warnings.

That was sort of intentional, the rule was disabled globally in #215 so I thought eslint-disable here would quickly become trash. Should pass the checks now

rhyslbw (approved):
Well done

PR #230, opened 2022-04-20
refactor(wallet): change transactions history all type to TxAlonzo

lgobbi-atix:
# Context
Simplify `SingleAddressWallet.transactions.history.all` by changing it’s type to `Cardano.TxAlonzo[]`

# Proposed Solution
- Use the transaction inspector to identify an outgoing or incoming transaction and get rid of `DirectionalTransaction` type

- Changed `transactions.history.all$` type from `DirectionalTransaction` to `TxAlonzo`
- Removed `DirectionalTransaction` and `TransactionDireciton`

### DIFF
--- a/packages/core/src/util/txInspector.ts
+++ b/packages/core/src/util/txInspector.ts
@@ -17,11 +17,17 @@ type TxInspector<T extends Inspectors> = (tx: TxAlonzo) => {
   [k in keyof T]: ReturnType<T[k]>;
 };
 
-// Inspectors result types
+// Inspection result types
 export type SendReceiveValueInspection = Value;
 export type DelegationInspection = StakeDelegationCertificate[];
 export type StakeKeyRegistrationInspection = StakeAddressCertificate[];
 export type WithdrawalInspection = Lovelace;
+
+// Inspector types
+export type SendReceiveValueInspector = (ownAddresses: Address[]) => Inspector<SendReceiveValueInspection>;
+export type DelegationInspector = Inspector<DelegationInspection>;
+export type StakeKeyRegistrationInspector = Inspector<StakeKeyRegistrationInspection>;
+export type WithdrawalInspector = Inspector<WithdrawalInspection>;
 
 /**
  * Inspects a transaction for value (coins + assets) sent by the provided addresses.
@@ -29,13 +35,11 @@ export type WithdrawalInspection = Lovelace;
  * @param {Address[]} ownAddresses own wallet's addresses
  * @returns {Value} total value sent
  */
-export const valueSentInspector =
-  (ownAddresses: Address[]): Inspector<Value> =>
-  (tx: TxAlonzo): SendReceiveValueInspection => {
-    if (!isOutgoing(tx, ownAddresses)) return { coins: 0n };
-    const sentOutputs = tx.body.outputs.filter((out) => !isAddressWithin(ownAddresses)(out));
-    return coalesceValueQuantities(sentOutputs.map((output) => output.value));
-  };
+export const valueSentInspector: SendReceiveValueInspector = (ownAddresses: Address[]) => (tx: TxAlonzo) => {
+  if (!isOutgoing(tx, ownAddresses)) return { coins: 0n };
+  const sentOutputs = tx.body.outputs.filter((out) => !isAddressWithin(ownAddresses)(out));
+  return coalesceValueQuantities(sentOutputs.map((output) => output.value));
+};
 
 /**
  * Inspects a transaction for value (coins + assets) received by the provided addresses.
@@ -43,13 +47,11 @@ export const valueSentInspector =
  * @param {Address[]} ownAddresses own wallet's addresses
  * @returns {Value} total value received
  */
-export const valueReceivedInspector =
-  (ownAddresses: Address[]): Inspector<Value> =>
-  (tx: TxAlonzo): SendReceiveValueInspection => {
-    if (isOutgoing(tx, ownAddresses)) return { coins: 0n };
-    const receivedOutputs = tx.body.outputs.filter((out) => isAddressWithin(ownAddresses)(out));
-    return coalesceValueQuantities(receivedOutputs.map((output) => output.value));
-  };
+export const valueReceivedInspector: SendReceiveValueInspector = (ownAddresses: Address[]) => (tx: TxAlonzo) => {
+  if (isOutgoing(tx, ownAddresses)) return { coins: 0n };
+  const receivedOutputs = tx.body.outputs.filter((out) => isAddressWithin(ownAddresses)(out));
+  return coalesceValueQuantities(receivedOutputs.map((output) => output.value));
+};
 
 /**
  * Inspects a transaction for a stake delegation certificate.
@@ -57,7 +59,7 @@ export const valueReceivedInspector =
  * @param {TxAlonzo} tx transaction to inspect
  * @returns {DelegationInspection} array of delegation certificates
  */
-export const delegationInspector: Inspector<DelegationInspection> = (tx: TxAlonzo) =>
+export const delegationInspector: DelegationInspector = (tx: TxAlonzo) =>
   (tx.body.certificates?.filter(
     (cert) => cert.__typename === CertificateType.StakeDelegation
   ) as StakeDelegationCertificate[]) ?? [];
@@ -68,7 +70,7 @@ export const delegationInspector: Inspector<DelegationInspection> = (tx: TxAlonz
  * @param {TxAlonzo} tx transaction to inspect
  * @returns {StakeKeyRegistrationInspection} array of stake key registration certificates
  */
-export const stakeKeyRegistrationInspector: Inspector<StakeKeyRegistrationInspection> = (tx: TxAlonzo) =>
+export const stakeKeyRegistrationInspector: StakeKeyRegistrationInspector = (tx: TxAlonzo) =>
   (tx.body.certificates?.filter(
     (cert) => cert.__typename === CertificateType.StakeKeyRegistration
   ) as StakeAddressCertificate[]) ?? [];
@@ -79,7 +81,7 @@ export const stakeKeyRegistrationInspector: Inspector<StakeKeyRegistrationInspec
  * @param {TxAlonzo} tx transaction to inspect
  * @returns {StakeKeyRegistrationInspection} array of stake key deregistration certificates
  */
-export const stakeKeyDeregistrationInspector: Inspector<StakeKeyRegistrationInspection> = (tx: TxAlonzo) =>
+export const stakeKeyDeregistrationInspector: StakeKeyRegistrationInspector = (tx: TxAlonzo) =>
   (tx.body.certificates?.filter(
     (cert) => cert.__typename === CertificateType.StakeKeyDeregistration
   ) as StakeAddressCertificate[]) ?? [];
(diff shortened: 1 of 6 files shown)


packages/core/src/util/txInspector.ts
```diff
@@ -17,47 +17,49 @@ type TxInspector<T extends Inspectors> = (tx: TxAlonzo) => {
   [k in keyof T]: ReturnType<T[k]>;
 };
 
-// Inspectors result types
+// Inspection result types
```
rhyslbw, line 20: Please separate the isolated refactor within this util into a separate commit

rhyslbw (changes requested):

packages/wallet/src/services/TransactionsTracker.ts
```diff
@@ -157,17 +147,23 @@ export const createTransactionsTracker = (
...
+          value: inspector(addresses)
+        });
+        return transactions.filter((tx) => inspectTx(tx).value.coins > 0n);
+      }),
       distinctUntilChanged(transactionsEquals)
     );
+
   const incomingTransactionHistory$ = new TrackerSubject<Cardano.TxAlonzo[]>(
-    providerTransactionsByDirection$(TransactionDirection.Incoming)
+    providerTransactionsByDirection$(valueReceivedInspector)
```
mkazlauskas, line 163: This will classify transactions that have any change outputs as incoming, which is not what we want here. This observable should emit transactions where no coin was sent from the wallet

packages/wallet/test/integration/transactionTime.test.ts
```diff
@@ -11,7 +11,7 @@ describe('integration/transactionTime', () => {
   });
 
   it('provides utils necessary for computing transaction time', async () => {
-    const transactions = await firstValueFrom(wallet.transactions.history.incoming$);
+    const transactions = await firstValueFrom(wallet.transactions.history.all$);
```
mkazlauskas, line 14: This is fine, but why was this change needed?

packages/wallet/test/services/TransactionsTracker.test.ts
```diff
@@ -183,7 +175,7 @@ describe('TransactionsTracker', () => {
         const tip$ = hot<Cardano.Tip>('----a|', { a: tip });
         const submitting$ = hot('-a---|', { a: tx });
         const pending$ = hot('---a-|', { a: tx });
-        const transactionsSource$ = hot<Cardano.TxAlonzo[]>('-----|');
+        const transactionsSource$ = hot<Cardano.TxAlonzo[]>('-----|') as unknown as TrackerSubject<Cardano.TxAlonzo[]>;
```
mkazlauskas, line 178: Just wondering why this was needed, doesn't seem like type of this changed :thinking:

packages/core/src/util/txInspector.ts
mkazlauskas: Also, changes in `wallet` package are breaking. We're using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) to indicate it

mkazlauskas (changes requested):

packages/core/src/util/txInspector.ts
```diff
@@ -17,47 +17,49 @@ type TxInspector<T extends Inspectors> = (tx: TxAlonzo) => {
...
  */
-export const valueReceivedInspector =
-  (ownAddresses: Address[]): Inspector<Value> =>
-  (tx: TxAlonzo): SendReceiveValueInspection => {
-    if (isOutgoing(tx, ownAddresses)) return { coins: 0n };
-    const receivedOutputs = tx.body.outputs.filter((out) => isAddressWithin(ownAddresses)(out));
-    return coalesceValueQuantities(receivedOutputs.map((output) => output.value));
-  };
+export const valueReceivedInspector: SendReceiveValueInspector = (ownAddresses: Address[]) => (tx: TxAlonzo) => {
+  if (isOutgoing(tx, ownAddresses)) return { coins: 0n };
```
mkazlauskas, line 51: Not a change in this PR, but this line is not entirely correct. You could have a tx that both sends and receives:
- from plutus script address
- with a tx signed by multiple parties (yourself and someone else)

Note: if we change this then TransactionsTracker.incoming$ will also be affected and we could have transactions that are both incoming and outgoing. Not sure if that would break anything :thinking: 

@rhyslbw @james-iohk thoughts?

rhyslbw: Right, this catch should be lower in the logic stack, as a fall-back value instread

packages/wallet/test/integration/transactionTime.test.ts
lgobbi-atix: `valueReceivedInspector` checks that the wallet addresses are within the transaction inputs to see if it's incoming.

Now `incoming$` is always empty as the inspector cannot match the addresses and inputs because the `createWallet` used for testing generates a new key agent with different address each run and the mocked wallet provider always returns the same transactions with hardcoded inputs and addresses.

packages/wallet/test/services/TransactionsTracker.test.ts
lgobbi-atix: My bad, was something I tried but forgot to change it back

packages/core/src/util/txInspector.ts
lgobbi-atix: I see, but in that case how do you know which are change utxo and what are you actually receiving?

I could rename `isOutgoing` to `isSending` where "is sending" means "my addresses are included in the tx inputs" (I guess we could also rename it to `isWithinInputs` or something like that)

* `valueSentInspector`:

```ts
export const valueSentInspector: SendReceiveValueInspector = (ownAddresses: Address[]) => (tx: TxAlonzo) => {
  if (!isSending(tx, ownAddresses)) return { coins: 0n };
  // Filter out change utxo
  const sentOutputs = tx.body.outputs.filter((out) => !isAddressWithin(ownAddresses)(out));
  return coalesceValueQuantities(sentOutputs.map((output) => output.value));
};
```
If I'm not wrong that check in the `valueSentInspector` should still be fine, if your address is not in the tx inputs then you are not sending anything. And if it is included in some input, then all outputs that don't include your address is what you are actually sending.

But what about this case: 
> with a tx signed by multiple parties (yourself and someone else)

could someone else be sending something in that same tx? Could there be change utxo for that other wallet?

For example, could something like this happen in the same transaction?
- `MyWallet` sends to `WalletA` -> here you would have sending outputs to `WalletA` and change outputs for `MyWallet`  
- `WalletA` sends to `WalletB` -> here you would have sending outputs to `WalletB` and change outputs for `WalletA`

How do you differentiate between `sending outputs to WalletA` and `change outputs for WalletA`? Do we even care about that or should we just filter out `change outputs for MyWallet` and consider everything else as sending as we are doing now?

lgobbi-atix: - `valueReceivedInspector`:

```ts
export const valueReceivedInspector: SendReceiveValueInspector = (ownAddresses: Address[]) => (tx: TxAlonzo) => {
  if (isSending(tx, ownAddresses)) return { coins: 0n };
  const receivedOutputs = tx.body.outputs.filter((out) => isAddressWithin(ownAddresses)(out));
  return coalesceValueQuantities(receivedOutputs.map((output) => output.value));
};
```

For this function, as you mentioned, the `isSending` shouldn't be here. But could something like this happen?

- `MyWallet` sends to `WalletA` -> here you would have sending outputs to `WalletA` and change outputs for `MyWallet`
- `WalletA` sends to `MyWallet` -> here you would have sending outputs to `MyWallet` and change outputs for `WalletA`

In that case, we should only consider `sending outputs to MyWallet` from `WalletA` as the received value, which `tx.body.outputs.filter((out) => isAddressWithin(ownAddresses)(out))` would get, but it would also include `change outputs for MyWallet`, how do we identify them and filter them out?

packages/wallet/test/integration/transactionTime.test.ts
mkazlauskas: I see, but still not sure why it worked before. Wondering if we've changed how it works :thinking:

packages/core/src/util/txInspector.ts
mkazlauskas: > But could something like this happen

I guess it's technically possible but would be unpractical.

I'm now thinking that the concept of incoming$ and outgoing$ observables is just too abstract and difficult to define. Maybe we should just leave it up to the SDK user to apply the filters he wants. How about refactoring the interface to have just `transactions.history$` (the current `all$`), intended to be filtered with the inspectors?

If we do that then we should definitely remove logic like this: `if (isSending(tx, ownAddresses)) return { coins: 0n };` and have well named inspectors for:
1. sum value of all inputs from own addr
2. sum value of all outputs to own addr
3. net value spent = 1 - 2
4. net value received = 2 - 1

lgobbi-atix: Ok, I could take a look at this. Should I do it in this PR or should we merge this one as it is and create a new branch for this changes?

Some questions:
> 1. sum value of all inputs from own addr
Inputs do not have value, how can I get this without using the wallet provider?

> 5. could also do a "did I have to sign the transaction" inspector that would check 1. + stuff in certificates. I think that's the typical "send transaction" classification that a user might expect
I don't think I understand this one

lgobbi-atix: > I'm now thinking that the concept of incoming$ and outgoing$ observables is just too abstract and difficult to define. Maybe we should just leave it up to the SDK user to apply the filters he wants. How about refactoring the interface to have just transactions.history$ (the current all$), intended to be filtered with the inspectors?

About this, we can remove `history.incoming$` and `history.outgoing$`, but what about the others:
```ts
history: {
  all$: historicalTransactions$,
  incoming$: incomingTransactionHistory$,
  outgoing$: outgoingTransactionHistory$
},
incoming$: newTransactions$(incomingTransactionHistory$),
outgoing: {
  confirmed$,
  failed$,
  inFlight$,
  pending$,
  submitting$
},
```
Most depend on `outgoingTransactionHistory$` and `incomingTransactionHistory$` in one way or another.
The light wallet only uses `outgoing.inFlight$` though, which needs `txConfirmed$` and that one needs `outgoingTransactionHistory$` which I think could be replaced by `historicalTransactions$`

Should we maybe leave it like this?
```ts
{
  history$: historicalTransactions$,
  confirmed$,
  failed$,
  inFlight$,
  pending$,
  submitting$
},
```
after making `txConfirmed$` use `historicalTransactions$` instead of `outgoingTransactionHistory$` if it makes sense

mkazlauskas: > Should I do it in this PR or should we merge this one as it is and create a new branch for these changes?

I think merging this is better, but let's do that tomorrow, so if you can get some progress in the mean time then great.

> Inputs do not have value, how can I get this without using the wallet provider?

Right, seems like this will require resolving it with the provider. We should probably change the return type of transactions fetched via provider. Let's scope this out of this PR.

> I don't think I understand this one

I think most users would consider any transactions that they sign with their wallet as "sent", so we probably want to have an inspector for that. My understanding is that a way to recognize such transaction is either:

- it has any inputs from own addr
- it has any certificates referencing own reward account. I'm actually not 100% sure about the rules of signature validation for all certificate types, but an example would be a delegation certificate - you need to sign such transactions with your stake key

It is unlikely that a tx would have such certificates without any inputs from own address, but I think it's technically possible - tx fee could be paid with an input owned by plutus script address for example, or just a different wallet.

> Should we maybe leave it like this?

Sorry if I wasn't clear on my suggestion here. I wanted to suggest that we should probably replace this:

```
history: {
  all$: historicalTransactions$,
  incoming$: incomingTransactionHistory$,
  outgoing$: outgoingTransactionHistory$
},
```
with `history$: historicalTransactions$,`

mkazlauskas (approved):
Since all tests pass I think this is good enough to merge, but I'd like to get @rhyslbw approval before we do that. @lgobbi-atix if you make some clean commits today based on the discussions in comments please merge into this branch and we'll review on Monday.

lgobbi-atix:
> Since all tests pass I think this is good enough to merge, but I'd like to get @rhyslbw approval before we do that. @lgobbi-atix if you make some clean commits today based on the discussions in comments please merge into this branch and we'll review on Monday.

I created this draft PR with those changes: https://github.com/input-output-hk/cardano-js-sdk/pull/233

rhyslbw (approved):
Well done guys

PR #231, opened 2022-04-20
WIP: Feat/add pool stats

jvieiro:
# Context

Add StakePoolMetrics to current stake pools response at DbSyncStakePoolProvider

# Proposed Solution

- [x] Create corresponding queries

- [x] Create data mappers

- [x] Adapt tests / add new ones

n/a


mkazlauskas:
@jvieiro please rebase on master

packages/cardano-services/src/StakePoolSearch/DbSyncStakePoolSearchProvider/mappers.ts
```diff
@@ -60,7 +64,8 @@ export const toCoreStakePool = ({
       hexId: poolData.hexId,
       id: poolData.id,
       margin: poolData.margin,
-      metrics: {} as Cardano.StakePoolMetrics, // TODO: fetch metrics
+      metrics:
+        poolMetrics.find((metrics) => metrics.hashId === poolData.hashId)?.metrics || ({} as Cardano.StakePoolMetrics),
```
mkazlauskas, line 68: Aren't these supposed to correspond to the same index?

mkazlauskas (commented):
Didn't look into queries, but LGTM in general!

packages/cardano-services/src/StakePoolSearch/DbSyncStakePoolSearchProvider/queries.ts
```diff
@@ -91,15 +328,16 @@ FROM pool_relay
...
 
-// TODO: probably this should be filtered by registration tx_id = last_update.registration_tx_id
 export const findPoolsOwners = `
 SELECT 
-  DISTINCT stake.view AS address,
+  stake.view AS address,
   owner.pool_hash_id AS hash_id
 FROM pool_owner owner
+JOIN pool_update 
+    ON pool_update.registered_tx_id = owner.registered_tx_id
```
jvieiro, line 337: @rhyslbw @mkazlauskas Can you please double check this.

> Pools can be registered, then retired and then registered again. What happens if the owners of the first pool registration are different from the last registration? Should we filter them? Thats why I put this comment. We currently have the tx where the owner was registered and this should match with the pool's registration.

packages/cardano-services/src/StakePoolSearch/DbSyncStakePoolSearchProvider/mappers.ts
jvieiro: poolMetrics contains the metrics of all the fetched pools in order to only make one query to fetch all pools metrics. Then, they are matched through hashId

packages/cardano-services/src/StakePoolSearch/DbSyncStakePoolSearchProvider/queries.ts
mkazlauskas: Hm I'm not sure. May be best to ask db-sync devs how this scenario is handled so we know how to query it.

rhyslbw: > What happens if the owners of the first pool registration are different from the last registration?

@james-iohk is this a legal state change?

packages/cardano-services/src/StakePoolSearch/DbSyncStakePoolSearchProvider/queries.ts
```diff
@@ -10,47 +10,284 @@ export const findLastEpoch = `
  LIMIT 1
 `;
 
+export const findTotalAda = `
```
iadmytro, line 13: we can use simpler query from the dbsync doc:
select sum (value)
from tx_out as tx_outer where
    not exists
        ( select tx_out.id
          from tx_out
          inner join tx_in on
              tx_out.tx_id = tx_in.tx_out_id and
              tx_out.index = tx_in.tx_out_index
          where tx_outer.id = tx_out.id
        );

packages/cardano-services/src/StakePoolSearch/DbSyncStakePoolSearchProvider/queries.ts
```diff
@@ -10,47 +10,284 @@ export const findLastEpoch = `
...
+    GROUP BY pd.pool_hash_id
+  ),
+  owners_total_utxos AS (
+    SELECT
+      sum(tx_out.value) AS amount,
+      o.pool_hash_id 
+    FROM tx_out
+    JOIN pool_owner o on 
+      o.addr_id = tx_out.stake_address_id and 
+      o.pool_hash_id = 16
```
iadmytro, line 101: hardcoded value

packages/cardano-services/src/StakePoolSearch/DbSyncStakePoolSearchProvider/queries.ts
jvieiro: This query performs better than the current one. But Ive found that the results doesnt match. Will take a deeper look at it

rhyslbw (changes requested):
The HTTP server can be configured to run one or more services, so we'll need to instead define the OpenAPI schema at the service level (one per service), attaching the validation middleware to the service's router instead.

packages/cardano-services/src/StakePoolSearch/DbSyncStakePoolSearchProvider/queries.ts
rhyslbw: > is this a legal state change?

Having thought about this more, we don't need to be concerned with this. As @mkazlauskas suggested, the last pool update owner is all we care about.

jvieiro: Yeah, that is what this change does. So I think we are good :)

jvieiro:
> The HTTP server can be configured to run one or more services, so we'll need to instead define the OpenAPI schema at the service level (one per service), attaching the validation middleware to the service's router instead.

Done

packages/cardano-services/src/StakePoolSearch/openApi.json
```diff
@@ -0,0 +1,421 @@
+{
+  "openapi": "3.0.0",
+  "info": {
+    "title": "Cardano Services API",
```
rhyslbw, line 4:
```suggestion
    "title": "Stake Pool Search",
```

packages/cardano-services/src/TxSubmit/openApi.json
```diff
@@ -0,0 +1,75 @@
+{
+  "openapi": "3.0.0",
+  "info": {
+    "title": "Cardano Services API",
```
rhyslbw, line 4:
```suggestion
    "title": "Tx Submit",
```

rhyslbw (changes requested):
16128839240a626f9d20871970465153add56514 should really be deferred until `stakePoolStats` is hoisted from `WalletProvider`.

jvieiro:
> [1612883](https://github.com/input-output-hk/cardano-js-sdk/commit/16128839240a626f9d20871970465153add56514) should really be deferred until `stakePoolStats` is hoisted from `WalletProvider`.

Ive dropped the commit to then be addded all the changes together

rhyslbw (changes requested):
Looks good. Please squash to minimal commits (looks like 1 would be ideal).

packages/cardano-services/src/TxSubmit/openApi.json
```diff
@@ -0,0 +1,75 @@
...
+          }
+        }
+      },
+      "txsubmit_submit_body": {
+        "type": "object",
+        "properties": {
+          "args": {
+            "type": "array",
+            "items": {
+              "type": "object"
```
mkazlauskas, line 68: UInt8Array gets serialized to hex string via `toSerializableObject`
```suggestion
              "type": "string"
```

mkazlauskas (changes requested):

packages/cardano-services/src/TxSubmit/openApi.json
jvieiro: If I do so, tests will fail. I think that it expects an object with this format `{ __type: 'Buffer', value: string }`

mkazlauskas: Ah yes, you're right - that's what it is. Perhaps specify this type in openApi.json too then?

jvieiro: For some reason if one field starts with double underscore the other fields get ignore , looks like a bug on open api validator. I let the fields of the object as optional for now. Ive tried with this format and works perfectly `{ _type: string, value: string }` but I dont like much the idea to adapt the code to the library.

packages/cardano-services/src/StakePoolSearch/DbSyncStakePoolSearchProvider/queries.ts
```diff
@@ -10,47 +10,294 @@ export const findLastEpoch = `
...
-    COALESCE((r_e.total_amount - ((r_e.total_amount * pu.margin) + pu.fixed_cost))
-      /total_stake.active_stake, 0) AS member_roi
-  from total_rewards_per_epoch r_e
-  join epochs e on 
+    COALESCE(r_e.total_amount,0) AS total_rewards,
+    CASE
+    WHEN pu.fixed_cost >= r_e.total_amount
+    THEN r_e.total_amount
+    ELSE (
+      COALESCE (FLOOR((r_e.total_amount-pu.fixed_cost) * pu.margin + pu.fixed_cost),0)
```
jvieiro, line 283: @rhyslbw @mkazlauskas Do you think we can use this formula or has to strictly be  the `5.5.4.1` of the ledger? What differs from that is a sum factor on the margin value which is not straightforward to calculate.

packages/cardano-services/src/TxSubmit/openApi.json
mkazlauskas: Changing field name is still an option at this point. Maybe we could come up with something less conflict-prone than `_type` :thinking: This field is not supposed to be used in code directly, only by toSerializableObject and fromSerializableObject, so we can even use special characters if we want.

jvieiro: Maybe something like `_objType` will be less conflictibe

packages/cardano-services/src/StakePoolSearch/DbSyncStakePoolSearchProvider/queries.ts
```diff
@@ -10,47 +10,294 @@ export const findLastEpoch = `
...
-    COALESCE((r_e.total_amount - ((r_e.total_amount * pu.margin) + pu.fixed_cost))
-      /total_stake.active_stake, 0) AS member_roi
-  from total_rewards_per_epoch r_e
-  join epochs e on 
+    COALESCE(r_e.total_amount,0) AS total_rewards,
+    CASE
+    WHEN pu.fixed_cost >= r_e.total_amount
+    THEN r_e.total_amount
+    ELSE (
+      COALESCE (FLOOR((r_e.total_amount-pu.fixed_cost) * pu.margin) + pu.fixed_cost,0)
```
rhyslbw, line 283: What is  `total_amount-pu`, and where is it defined?

packages/cardano-services/src/TxSubmit/openApi.json
jvieiro: Ok, I see what is going on. It wasnt the naming being used. It is actually the bodyParser middleware that maps the request when it invokes `fromSerializableObject`. So, in this case we have a Uint8Array object and the initially schema is the one which works. This leads the openApi doc a bit confusing.

packages/cardano-services/src/StakePoolSearch/DbSyncStakePoolSearchProvider/queries.ts
rhyslbw: Cont https://github.com/input-output-hk/cardano-js-sdk/pull/231#discussion_r863495064

packages/cardano-services/src/StakePoolSearch/DbSyncStakePoolSearchProvider/queries.ts
rhyslbw: Looks good to me

mkazlauskas (approved):
Looks good to me, nice work!

packages/cardano-services/src/Http/HttpServer.ts
```diff
@@ -64,6 +63,13 @@ export class HttpServer extends RunnableModule {
       this.app.use(`/${service.slug}`, service.router);
       this.logger.debug(`Using /${service.slug}`);
     }
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    this.app.use((err: any, _req: express.Request, res: express.Response, _n: express.NextFunction) => {
```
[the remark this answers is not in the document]
jvieiro: I think it will be easier to resolve in #240

iadmytro:
I think we are missing unit tests for DbSyncStakePoolSearchProvider,
namely functions:
 queryPoolData
 queryPoolRelays
 queryPoolOwners
 queryRegistrations
 queryRetirements
 queryPoolRewards
 getLastEpoch
 queryPoolMetrics
 
 so far I have found corner case that live stake is not calculated for pool with no withdrawals.
 Unit test for queryPoolMetrics would be ideal place for that but the method is private.
 Thoughts?

rhyslbw:
They are mostly covered though @iadmytro 
image

> Thoughts

If we're not able to test that scenario via the public method, we could lift those private methods to a separate module, rather than making them public here.

iadmytro:
> If we're not able to test that scenario via the public method, we could lift those private methods to a separate module, rather than making them public here.

We could test it via public method, but those tests more about filters and conditions and we have complicated queries with logic which better be tested separately.

packages/cardano-services/src/DbSyncProvider/index.ts
```diff
@@ -0,0 +1,17 @@
+import { HealthCheckResponse, Provider } from '@cardano-sdk/core';
+import { Pool } from 'pg';
+
+const HEALTH_CHECK_QUERY = 'SELECT 1';
+
+export abstract class DbSyncProvider implements Provider {
```
rhyslbw, line 6: Nice, but I think this should be at the root in a file named `DbSyncProvider.ts` given there's nothing else in the dir.

rhyslbw (changes requested):

rhyslbw (approved):
Well done @jvieiro, great work :tada:

PR #233, opened 2022-04-22
refactor: transactions tracker

lgobbi-atix:
# Context

`outgoing$` and `incoming$` transactions definitions are not clear. It's better if we removed them and instead let the sdk user use the transaction inspector to classify transactions as needed.

Thread discussing the issue: https://github.com/input-output-hk/cardano-js-sdk/pull/230#discussion_r855336487

# Proposed Solution
- Remove `history.outgoing$` and `history.incoming$`, replace `history.all$` with just `history$`.
- Improve inspectors to properly tell if a transaction was sent by an address, calculate inputs and outputs total value sent and received and calculate total net value sent and received

**TransactionsTracker:**
- `TransactionsTracker.history.outgoing$` and `TransactionsTracker.history.incoming$` were removed 
- `TransactionsTracker.incoming$` was removed
- `TransactionsTracker.history.all$` was removed
- `TransactionsTracker.history` was changed to `TransactionsTracker.history$` and now has what `TransactionsTracker.history.all$` used to have

**TxInspector:**
- `valueSentInspector` and `valueSentInspector were modified to calculate sent and received net values
- `totalAddressInputsValueInspector` and `totalAddressOutputsValueInspector` were added
- `sentInspector` was added
- utility functions needed for the new inspectors were added

### DIFF
--- a/packages/core/src/Address/util.ts
+++ b/packages/core/src/Address/util.ts
@@ -1,4 +1,4 @@
-import { Address, TxAlonzo } from '../Cardano';
+import { Address, TxAlonzo, TxIn } from '../Cardano';
 import { parseCslAddress } from '../CSL';
 
 /**
@@ -15,8 +15,10 @@ export const isAddressWithin =
     addresses.includes(address!);
 
 /**
- * Receives a transaction and a set of addresses to check if the transaction is outgoing,
- * i.e., some of the addresses are included in the transaction inputs
+ * Receives a transaction and a set of addresses to check if
+ * some of them are included in the transaction inputs
+ *
+ * @returns {TxIn[]} array of inputs that contain any of the addresses
  */
-export const isOutgoing = (tx: TxAlonzo, ownAddresses: Address[]): boolean =>
-  tx.body.inputs.some(isAddressWithin(ownAddresses));
+export const inputsWithAddresses = (tx: TxAlonzo, ownAddresses: Address[]): TxIn[] =>
+  tx.body.inputs.filter(isAddressWithin(ownAddresses));
--- a/packages/core/src/Asset/util/subtractTokenMaps.ts
+++ b/packages/core/src/Asset/util/subtractTokenMaps.ts
@@ -1,0 +1,22 @@
+import { TokenMap } from '../../Cardano';
+import { util } from '../../util';
+
+/**
+ * Subtract asset quantities in order
+ */
+export const subtractTokenMaps = (assets: (TokenMap | undefined)[]): TokenMap | undefined => {
+  if (assets.length <= 0 || !util.isNotNil(assets[0])) return undefined;
+  const result: TokenMap = assets[0];
+  const rest: TokenMap[] = assets.slice(1).filter(util.isNotNil);
+  for (const assetTotals of rest) {
+    for (const [assetId, assetQuantity] of assetTotals.entries()) {
+      const total = result.get(assetId) ?? 0n;
+      const diff = total - assetQuantity;
+      diff === 0n ? result.delete(assetId) : result.set(assetId, diff);
+    }
+  }
+  if (result.size === 0) {
+    return undefined;
+  }
+  return result;
+};
--- a/packages/core/src/Cardano/util/resolveInputValue.ts
+++ b/packages/core/src/Cardano/util/resolveInputValue.ts
@@ -1,0 +1,13 @@
+import { TxAlonzo, TxIn, Value } from '../types';
+
+/**
+ * Resolves the value of an input by looking for the matching output in a list of transactions
+ *
+ * @param {TxIn} input input to resolve value for
+ * @param {TxAlonzo[]} transactions list of transactions to find the matching output
+ * @returns {Value | undefined} input value or undefined if not found
+ */
+export const resolveInputValue = (input: TxIn, transactions: TxAlonzo[]): Value | undefined => {
+  const tx = transactions.find((transaction) => transaction.id === input.txId);
+  return tx?.body.outputs[input.index]?.value;
+};
--- a/packages/core/src/Cardano/util/subtractValueQuantities.ts
+++ b/packages/core/src/Cardano/util/subtractValueQuantities.ts
@@ -1,0 +1,10 @@
+import { Asset, BigIntMath } from '../..';
+import { Value } from '../types';
+
+/**
+ * Subtract all quantities
+ */
+export const subtractValueQuantities = (quantities: Value[]) => ({
+  assets: Asset.util.subtractTokenMaps(quantities.map(({ assets }) => assets)),
+  coins: BigIntMath.subtract(quantities.map(({ coins }) => coins))
+});
--- a/packages/core/src/util/txInspector.ts
+++ b/packages/core/src/util/txInspector.ts
@@ -1,15 +1,21 @@
 import {
   Address,
+  Certificate,
   CertificateType,
+  Ed25519KeyHash,
   Lovelace,
+  RewardAccount,
   StakeAddressCertificate,
   StakeDelegationCertificate,
+  TokenMap,
   TxAlonzo,
+  TxIn,
   Value
 } from '../Cardano';
 import { BigIntMath } from './BigIntMath';
-import { coalesceValueQuantities } from '../Cardano/util';
-import { isAddressWithin, isOutgoing } from '../Address/util';
+import { coalesceValueQuantities, resolveInputValue, subtractValueQuantities } from '../Cardano/util';
+import { inputsWithAddresses, isAddressWithin } from '../Address/util';
+import { removeNegativesFromTokenMap } from '../Asset/util';
 
 type Inspector<Inspection> = (tx: TxAlonzo) => Inspection;
 type Inspectors = { [k: string]: Inspector<unknown> };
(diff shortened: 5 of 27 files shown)


packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
```diff
@@ -53,10 +53,7 @@ export const certificateTransactionsWithEpochs = (
   slotEpochCalc$: Observable<SlotEpochCalc>,
   certificateTypes: Cardano.CertificateType[]
 ): Observable<TxWithEpoch[]> =>
-  combineLatest([
-    transactionsWithCertificates(transactionsTracker.history.outgoing$, certificateTypes),
-    slotEpochCalc$
-  ]).pipe(
+  combineLatest([transactionsWithCertificates(transactionsTracker.history$, certificateTypes), slotEpochCalc$]).pipe(
```
lgobbi-atix, line 56: Is this ok? We don't have `transactionsTracker.history.outgoing$` anymore

packages/wallet/src/services/TransactionsTracker.ts
```diff
@@ -148,26 +140,8 @@ export const createTransactionsTracker = (
...
-  const incomingTransactionHistory$ = new TrackerSubject<Cardano.TxAlonzo[]>(
-    providerTransactionsByDirection$(valueReceivedInspector)
-  );
-  const outgoingTransactionHistory$ = new TrackerSubject<Cardano.TxAlonzo[]>(
-    providerTransactionsByDirection$(valueSentInspector)
-  );
-
   const txConfirmed$ = (tx: Cardano.NewTxAlonzo) =>
-    newTransactions$(outgoingTransactionHistory$).pipe(
+    newTransactions$(historicalTransactions$).pipe(
```
lgobbi-atix, line 144: Does this change make sense?

`txConfirmed$` is being used for most things here:
```ts
outgoing: {
  confirmed$,
  failed$,
  inFlight$,
  pending$,
  submitting$
},
```

packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
mkazlauskas: Looking at this now, it wasn't 100% correct to start with. This change made it a little bit worse, but let's address it separately in ADP-1719

The issue is that this is then used [here](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/wallet/src/services/DelegationTracker/RewardsHistory.ts#L37) (and possibly a few other places), which does not check `rewardAccount` in the certificate, so if someone sends you some coin and delegates in the same tx, it will be a bug.

packages/wallet/src/services/TransactionsTracker.ts
mkazlauskas: Yep, looks good

packages/core/src/Address/util.ts
```diff
@@ -15,8 +15,8 @@ export const isAddressWithin =
     addresses.includes(address!);
 
 /**
- * Receives a transaction and a set of addresses to check if the transaction is outgoing,
- * i.e., some of the addresses are included in the transaction inputs
+ * Receives a transaction and a set of addresses to check if
+ * some of them are included in the transaction inputs
  */
-export const isOutgoing = (tx: TxAlonzo, ownAddresses: Address[]): boolean =>
+export const isSending = (tx: TxAlonzo, ownAddresses: Address[]): boolean =>
```
mkazlauskas, line 21: Your rename makes it better, but I'm thinking we should move it under inspectors as it's inspecting a tx object. Also, how about this? Even more explicit:

```suggestion
export const anyInputFromAddress = (tx: TxAlonzo, ownAddresses: Address[]): boolean =>
```

packages/core/src/Asset/util/subtractTokenMaps.ts
```diff
@@ -0,0 +1,24 @@
+import { TokenMap } from '../../Cardano';
+import { util } from '../../util';
+
+/**
+ * Subtract asset quantities in order
+ */
+export const subtractTokenMaps = (totals: (TokenMap | undefined)[]): TokenMap | undefined => {
```
mkazlauskas, line 7: This needs some tests

packages/core/src/Cardano/util/subtractValueQuantities.ts
```diff
@@ -0,0 +1,10 @@
+import { Asset, BigIntMath } from '../..';
+import { Value } from '../types';
+
+/**
+ * Subtract all quantities
+ */
+export const subtractValueQuantities = (quantities: Value[]) => ({
```
mkazlauskas, line 7: This needs some tests

packages/core/src/Asset/util/subtractTokenMaps.ts
```diff
@@ -0,0 +1,24 @@
...
+  if (totals.length <= 0 || !util.isNotNil(totals[0])) return undefined;
+  const result: TokenMap = totals[0];
+  const rest: TokenMap[] = totals.slice(1).filter(util.isNotNil);
+  for (const assetTotals of rest) {
+    for (const [assetId, assetQuantity] of assetTotals.entries()) {
+      const total = result.get(assetId);
+      if (total) {
+        const diff = total - assetQuantity;
+        diff > 0 ? result.set(assetId, total) : result.delete(assetId);
```
mkazlauskas, line 16: Given function name, I would expect this to return negative quantities too. Maybe let's create a separate function to check for negative quantities or to remove them.

```suggestion
        diff === 0n ? result.delete(assetId) : result.set(assetId, total);
```

Btw `subtractValueQuantities` can already return negative quantity of coin

packages/core/src/Cardano/util/resolveInputValue.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Resolves the value of an input by looking for the matching output in a list of transactions
+ *
+ * @param {TxIn} input input to resolve value for
+ * @param {TxAlonzo[]} transactions list of transactions to find the matching output
+ * @returns {Value | undefined} input value or undefined if not found
+ */
+export const resolveInputValue = (input: TxIn, transactions: TxAlonzo[]) => {
+  let value: Value | undefined;
+  transactions.some((transaction) => {
```
mkazlauskas, line 12: This would be cleaner:
```suggestion
  return transactions.some((transaction) => {
```

packages/core/src/Cardano/util/resolveInputValue.ts
```diff
@@ -0,0 +1,17 @@
...
+    value = transaction.body.outputs.find((_, index) => input.txId === transaction.id && input.index === index)?.value;
```
mkazlauskas, line 13:
```suggestion
    return transaction.body.outputs.find((_, index) => input.txId === transaction.id && input.index === index)?.value || undefined
```

packages/core/src/util/txInspector.ts
```diff
@@ -24,35 +28,96 @@ export type StakeKeyRegistrationInspection = StakeAddressCertificate[];
 export type WithdrawalInspection = Lovelace;
 
 // Inspector types
+interface SentInspectorArgs {
+  addresses?: Address[];
```
mkazlauskas, line 32: Is there a reason to have these args as nullable?

packages/core/src/util/txInspector.ts
```diff
@@ -24,35 +28,96 @@ export type StakeKeyRegistrationInspection = StakeAddressCertificate[];
...
  * @param {Address[]} ownAddresses own wallet's addresses
- * @returns {Value} total value sent
+ * @param {TxAlonzo[]} historicalTxs wallet's historical transactions
+ * @returns {Value} total value in inputs
  */
-export const valueSentInspector: SendReceiveValueInspector = (ownAddresses: Address[]) => (tx: TxAlonzo) => {
-  if (!isOutgoing(tx, ownAddresses)) return { coins: 0n };
-  const sentOutputs = tx.body.outputs.filter((out) => !isAddressWithin(ownAddresses)(out));
-  return coalesceValueQuantities(sentOutputs.map((output) => output.value));
+export const totalInputValuesInspector: TotalInputValuesInspector = (ownAddresses, historicalTxs) => (tx) => {
```
mkazlauskas, line 54: How about this? Suggestions @rhyslbw ?
```suggestion
export const totalAddressInputsValueInspector: TotalAddressInputsValueInspector = (ownAddresses, historicalTxs) => (tx) => {
```

packages/core/src/util/txInspector.ts
```diff
@@ -24,35 +28,96 @@ export type StakeKeyRegistrationInspection = StakeAddressCertificate[];
...
 // Inspector types
+interface SentInspectorArgs {
+  addresses?: Address[];
+  stakeKeyHashes?: Ed25519KeyHash[];
+  rewardAccounts?: RewardAccount[];
+}
+export type SentInspector = (args: SentInspectorArgs) => Inspector<boolean>;
+export type TotalInputValuesInspector = (
+  ownAddresses: Address[],
+  historicalTxs: TxAlonzo[]
```
mkazlauskas, line 39: This would allow using the same inspector object for a longer period of time
```suggestion
  getHistoricalTxs: () => TxAlonzo[]
```

packages/core/src/util/txInspector.ts
```diff
@@ -24,35 +28,96 @@ export type StakeKeyRegistrationInspection = StakeAddressCertificate[];
...
+ * Inspects a transaction for values (coins + assets) in outputs
+ * containing any of the provided addresses.
  *
  * @param {Address[]} ownAddresses own wallet's addresses
- * @returns {Value} total value received
+ * @returns {Value} total value in outputs
  */
-export const valueReceivedInspector: SendReceiveValueInspector = (ownAddresses: Address[]) => (tx: TxAlonzo) => {
-  if (isOutgoing(tx, ownAddresses)) return { coins: 0n };
+export const totalOutputValuesInspector: SendReceiveValueInspector = (ownAddresses) => (tx) => {
```
mkazlauskas, line 70: Let's rename this too to correspond to a new name of `totalInputValuesInspector`

packages/core/src/util/txInspector.ts
```diff
@@ -24,35 +28,96 @@ export type StakeKeyRegistrationInspection = StakeAddressCertificate[];
...
+export const sentInspector: SentInspector =
+  ({ addresses, rewardAccounts, stakeKeyHashes }) =>
+  (tx: TxAlonzo) => {
+    const isAddressInInputs = addresses?.length && isSending(tx, addresses);
+    const isStakeKeyInCertificate =
+      stakeKeyHashes?.length &&
+      tx.body.certificates?.some((cert) => 'stakeKeyHash' in cert && stakeKeyHashes.includes(cert.stakeKeyHash));
+    const isRewardAccountInCertificate =
+      rewardAccounts?.length &&
+      tx.body.certificates?.some((cert) => 'rewardAccount' in cert && rewardAccounts.includes(cert.rewardAccount));
```
mkazlauskas, line 91: I'm wondering if we should refactor `MirCertificate` to also use `stakeKeyHash` instead, like other certificates. The original refactor of `rewardAccount`->`stakeKeyHash` was needed to not require `networkId` in `coreToCsl`. MIR was left behind because it's conversion was not implemented. What do you think @rhyslbw ?

Also, we probably want to check `poolParameters.rewardAccount` - I assume you have to sign such certificate with your stake key too?

packages/core/src/util/txInspector.ts
```diff
@@ -24,35 +28,96 @@ export type StakeKeyRegistrationInspection = StakeAddressCertificate[];
...
 };
 
+/**
+ * Inspects a transaction to see if any of the addresses provided are included in a transaction input
+ * or if any of the rewards accounts or stake keys are included in a certificate
+ *
+ * @param {SentInspectorArgs} args array of addresses, stake key hashes and/or reward accounts
+ * @returns {boolean} true if any of the arguments provided were found in the transaction, false otherwise
+ */
+export const sentInspector: SentInspector =
```
mkazlauskas, line 82: Need some tests for the inspectors too

mkazlauskas (changes requested):

packages/core/src/Cardano/util/resolveInputValue.ts
lgobbi-atix: Related to https://github.com/input-output-hk/cardano-js-sdk/pull/233#discussion_r857611544

packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
lgobbi-atix: > let's address it separately in ADP-1719

Should I leave it like this then?

packages/core/src/Cardano/util/resolveInputValue.ts
mkazlauskas: Ah yes, it would have to be `transactions.find`

packages/core/src/util/txInspector.ts
mkazlauskas: Makes sense, just make sure different combinations are covered by tests

lgobbi-atix: Question, does it make sense if I remove the `stakeKeyHashes` argument and ask only for `rewardAccounts`? And then do something like this:

```ts
const stakeKeyHashes = rewardAccounts?.map((account) => Ed25519KeyHash.fromRewardAccount(account));
```

I think it'd make it easier to use as with the wallet you can just use `wallet.delegation.rewardAccounts$`

mkazlauskas: Agree, good idea :+1:

packages/core/src/util/txInspector.ts
rhyslbw: So `stakeKeyHash` instead of `rewardAccount`? It's probably best to make it consistent and simplify this logic.

mkazlauskas (approved):
Nice work @lgobbi-atix! :rocket:

PR #235, opened 2022-04-26
Refactor/single cardano services  entrypoint

rhyslbw:
# Context
We currently have multiple CLIs and run entrypoints, which are remnants of the [previous refactor](https://github.com/input-output-hk/cardano-js-sdk/pull/234). Now that we only run a single HTTP server, we can easily compress down to a single set of entrypoints.

# Proposed Solutiom
- Formalises the program elements into a new module, reducing duplicated code in each entrypoint.
- Service names are now a runtime argument for the top-level entrypoints.
- While we don't need to specify defaults in the CLI definition for the execution, we do need them defined to be shown within the help screens. The Ogmios URL default is derived from the newly re-exported function from `@cardano-ogmios/client` via the `@cardano-sdk/ogmios` package.

- I've exported a single function from `@cardano-ogmios/client` in 92af5472ceff52b747428c37c953ffd3c940d950. I don't recall if this is something we want to avoid, so let's discuss. 
- There was a missing Jest snapshot, so I've committed that in d81a4057af7c03ff29515f6fd61bf369410579ef.
- The test Docker container should now never bomb when the test suite is terminated without the `after` hooks firing to remove the container. b09bcd9f1c93dc0a01b31e63c4bed4548f3ab8f7

# Todo
- [x] Add tests for the missing configuration based on service nomination


packages/cardano-services/src/cli.ts
```diff
@@ -0,0 +1,79 @@
...
+console.log('Cardano Services CLI');
+
+const program = new Command('cardano-services');
+
+program.version(packageJson.version);
+
+program
+  .command('start-server')
+  .description('Start the HTTP server')
+  .argument('<serviceNames...>', 'List of services to attach')
```
mkazlauskas, line 25: A list of options would be nice here

packages/ogmios/src/index.ts
```diff
@@ -1 +1,2 @@
 export * from './ogmiosTxSubmitProvider';
+export { createConnectionObject } from '@cardano-ogmios/client';
```
mkazlauskas, line 2: I think this is fine. Re-exports to avoid are for `core` package types, as we want them to be stable and they tend to benefit from some changes compared to ogmios types (opaque strings with validation etc.).

mkazlauskas (commented):
Looking good! I'm wondering if we really need `run.ts` - aren't use cases covered by the cli interface?

rhyslbw:
> Looking good! I'm wondering if we really need run.ts - aren't use cases covered by the cli interface?

It's mainly there to support Docker Compose, which requires awkward workarounds dealing with variable command line arguments.

packages/cardano-services/src/cli.ts
rhyslbw: Done in 416c09fd72a97dfe9ad74fb71dbb7db94bf0b377

mkazlauskas (dismissed):
Looks good!

mkazlauskas (approved):

PR #236, opened 2022-04-27
refactor: asset provider extra data

lgobbi-atix:
# Context
Assets can have a huge number of mint/burn transactions.
For most use cases this information is not crucial and it can significantly increase load time.

# Proposed Solution
- Add `AssetInfo.mintOrBurnCount` and update blockfrost `AssetProvider.getAsset` implementation
- Add `extraData = {nftMetadata?: boolean, tokenMetadata?: boolean, history?: boolean}` argument to `getAsset` and implement it for blockfrost. Fetch metadata of last mint tx and use `metadatumToCip25` for `nftMetadata`
- Update wallet package to fetch all extra data and remove `nftMetadataProvider`

- Optional `extraData` argument added to `AssetProvider.getAsset`
- Fetch nft metadata in `AssetProvider` instead of `NftMetadataProvider`
- Remove `NftMetadataProvider`


rhyslbw (changes requested):
Looks good @lgobbi-atix , just need to remove the `blockfrost` and `wallet` scopes in each of the commit messages since they're touching multiple packages.

e.g 2bf57b462ccd1d348796227a7fb866ea6408ef0d
image

packages/wallet/src/services/AssetsTracker.ts
```diff
@@ -8,46 +8,31 @@ import { distinct, from, map, mergeMap, of, scan, tap } from 'rxjs';
 export const createAssetService =
   (assetProvider: TrackedAssetProvider, retryBackoffConfig: RetryBackoffConfig) => (assetId: Cardano.AssetId) =>
     coldObservableProvider(
-      () => assetProvider.getAsset(assetId),
+      () => assetProvider.getAsset(assetId, { history: true, nftMetadata: true, tokenMetadata: true }),
```
mkazlauskas, line 11: Do we really need to fetch history here?

packages/blockfrost/src/blockfrostAssetProvider.ts
```diff
@@ -41,26 +41,64 @@ export const blockfrostAssetProvider = (blockfrost: BlockFrostAPI): AssetProvide
...
-                quantity,
-                transactionId: Cardano.TransactionId(response.initial_mint_tx_hash)
-              }
-            ]
-          : await getAssetHistory(assetId),
+      history: extraData?.history ? await history() : undefined,
+      mintOrBurnCount: response.mint_or_burn_count,
       name,
-      policyId: Cardano.PolicyId(response.policy_id),
+      nftMetadata: extraData?.nftMetadata ? await nftMetadata() : undefined,
```
mkazlauskas, line 98: Let's change AssetInfo type's `nftMetadata` and `tokenMetadata` slightly: `nftMetadata?: NftMetadata | null` and add a doc comment like: "`undefined` if not loaded, `null` if no metadata found"

mkazlauskas (changes requested):

packages/wallet/src/services/AssetsTracker.ts
lgobbi-atix: I don't think so, I did it like this because the ticket said to "Update wallet package to fetch all extra data" but I might have misunderstood. I'll remove it

packages/blockfrost/src/blockfrostAssetProvider.ts
lgobbi-atix: Done

rhyslbw (approved):
Great work @lgobbi-atix

mkazlauskas (approved):
Nice work!

PR #237, opened 2022-04-27
fix(wallet): check reward account in rewards history tracker

lgobbi-atix:
# Context

If someone sends you some coin and delegates in the same tx, it will be a bug, because code will assume it’s your certificate as it doesn’t check `rewardAccount`

# Proposed Solution
Check if certificate includes the `rewardAccount` in `RewardsHistoryTracker`


rhyslbw (changes requested):
With single commit changes like this it's best to embed the whole context and solution in the commit message, so we're not relying on GitHub PRs for important information. Also, try to improve the solution so it's more clear, as: 

> Check if certificate includes the rewardAccount in RewardsHistoryTracker

... is a little vague. Something like this:

**fix(wallet): filter to own delegation certs when detecting new delegations**

If someone sends you coin and issues a delegation in the same tx a bug will be triggered. The wallet blindly assumes any delegation certificate present in transactions of relevance is it's own, since it doesn’t check the `rewardAccount` for ownership. The outcome of this bug is invalid wallet delegation state.

This change inspects the delegation certificate to check it originated from the owned reward account.

packages/wallet/src/services/DelegationTracker/RewardsHistory.ts
```diff
@@ -34,11 +34,20 @@ export const createRewardsHistoryProvider =
...
 export type RewardsHistoryProvider = ReturnType<typeof createRewardsHistoryProvider>;
 
-const firstDelegationEpoch$ = (transactions$: Observable<TxWithEpoch[]>) =>
+const firstDelegationEpoch$ = (transactions$: Observable<TxWithEpoch[]>, rewardAccounts: Cardano.RewardAccount[]) =>
   transactions$.pipe(
     map((transactions) =>
       first(
-        transactions.filter(({ tx }) => transactionHasAnyCertificate(tx, [Cardano.CertificateType.StakeDelegation]))
+        transactions.filter(({ tx }) => {
+          const hasCertificate = transactionHasAnyCertificate(tx, [Cardano.CertificateType.StakeDelegation]);
```
mkazlauskas, line 42: I think this is redundant, as `sentInspector` (almost) does the same job. We might be able to delete `transactionHasAnyCertificate` because it's not that useful without checking rewardAccounts. It should probably be replaced with an inspector. Also, deleting `transactionHasAnyCertificate` will probably flesh out some other similar issues to fix.

packages/wallet/src/services/DelegationTracker/RewardsHistory.ts
```diff
@@ -34,11 +34,20 @@ export const createRewardsHistoryProvider =
...
+          if (hasCertificate) {
+            const inspectTx = createTxInspector({
+              sentInspection: sentInspector({ rewardAccounts })
```
mkazlauskas, line 45: Maybe let's hoist the sentInspector part that finds certificates into a separate inspector

packages/wallet/src/services/DelegationTracker/RewardsHistory.ts
```diff
@@ -34,11 +34,20 @@ export const createRewardsHistoryProvider =
 
 export type RewardsHistoryProvider = ReturnType<typeof createRewardsHistoryProvider>;
 
-const firstDelegationEpoch$ = (transactions$: Observable<TxWithEpoch[]>) =>
+const firstDelegationEpoch$ = (transactions$: Observable<TxWithEpoch[]>, rewardAccounts: Cardano.RewardAccount[]) =>
   transactions$.pipe(
     map((transactions) =>
       first(
-        transactions.filter(({ tx }) => transactionHasAnyCertificate(tx, [Cardano.CertificateType.StakeDelegation]))
+        transactions.filter(({ tx }) => {
```
mkazlauskas, line 41: Let's map transactions to either 'null' or `{epoch}` here instead. This function has to return epoch of the first tx with delegation from own reward account.

```suggestion
        transactions.map(({ tx }) => {
```

mkazlauskas (changes requested):
Let's cover this fix with a test in RewardsHistory: use transactions that have certificates of both own and some other reward account and assert that only own is considered.

packages/wallet/src/services/DelegationTracker/RewardsHistory.ts
lgobbi-atix: I added the inspector: 
```ts
SignedCertificatesInspector = (
  rewardAccounts: RewardAccount[],
  certificateTypes?: CertificateType[]
) => Inspector<Certificate[]>;
```

but `transactionHasAnyCertificate` is used by `transactionsWithCertificates` and that one by `DelegationTracker.certificateTransactionsWithEpochs`.

I could delete `transactionHasAnyCertificate` and change `transactionsWithCertificates` to receive reward accounts and use the inspector but that would only return the transactions with a certificate signed by those accounts, is that ok?

mkazlauskas: Yes, I think in wallet context we always only care about own certificates

packages/wallet/src/services/DelegationTracker/RewardsHistory.ts
lgobbi-atix: Do you mean like this?
```ts
  transactions$.pipe(
    map((transactions) =>
      first(
        transactions.map(({ tx, epoch }) => {
          const inspectTx = createTxInspector({
            signedCertificates: signedCertificatesInspector(rewardAccounts, [Cardano.CertificateType.StakeDelegation])
          });
          return inspectTx(tx).signedCertificates.length > 0 ? { epoch } : null;
        })
      )
    ),
    map((tx) => (util.isNotNil(tx) ? tx.epoch + 3 : null)),
    distinctUntilChanged()
```
with this if the first transaction in `transactions$` is not signed or is not a delegation it'll return null, before it used to filter those out and get the first transaction that was a delegation and was signed

mkazlauskas: Ah yes, you need to do `map(...).filter(util.isNotNil) then`

rhyslbw (approved):

PR #238, opened 2022-04-28
feat(wallet): Trezor Key Agent

tomislavhoracek:
# Context
SDK needs to be able to support the hardware devices / wallets. TrezorKeyAgent is responsible for communication with the Trezor device model.

# Proposed Solution
Implement a KeyAgent to support the Trezor hardware devices

- TrezorKeyAgent added for handling Trezor device connection / communication.
- Web and Node based communication with Trezor devices via `trezor-connect` package

### DIFF
--- a/packages/wallet/src/KeyManagement/TrezorKeyAgent.ts
+++ b/packages/wallet/src/KeyManagement/TrezorKeyAgent.ts
@@ -1,0 +1,134 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { AuthenticationError, TransportError } from './errors';
+import { Cardano, NotImplementedError } from '@cardano-sdk/core';
+import { CommunicationType, KeyAgentType, SerializableTrezorKeyAgentData, SignBlobResult, TrezorConfig } from './types';
+import { KeyAgentBase } from './KeyAgentBase';
+import TrezorConnect, { Features } from 'trezor-connect';
+
+export interface TrezorKeyAgentProps extends Omit<SerializableTrezorKeyAgentData, '__typename'> {
+  isTrezorInitialized?: boolean;
+}
+
+export interface GetTrezorXpubProps {
+  accountIndex: number;
+}
+
+export interface CreateTrezorKeyAgentProps {
+  networkId: Cardano.NetworkId;
+  accountIndex?: number;
+  trezorConfig: TrezorConfig;
+}
+
+const transportTypedError = (error?: any) =>
+  new AuthenticationError('Trezor transport failed', new TransportError('Trezor transport failed', error));
+
+export class TrezorKeyAgent extends KeyAgentBase {
+  readonly isTrezorInitialized: Promise<boolean>;
+  constructor({ isTrezorInitialized, ...serializableData }: TrezorKeyAgentProps) {
+    super({ ...serializableData, __typename: KeyAgentType.Trezor });
+    if (!isTrezorInitialized) {
+      this.isTrezorInitialized = TrezorKeyAgent.initializeTrezorTransport(serializableData.trezorConfig);
+    }
+  }
+
+  /**
+   * @throws AuthenticationError
+   */
+  static async initializeTrezorTransport({
+    manifest,
+    communicationType,
+    silentMode = false,
+    lazyLoad = false
+  }: TrezorConfig): Promise<boolean> {
+    try {
+      await TrezorConnect.init({
+        // eslint-disable-next-line max-len
+        // Set to "false" (default) if you want to start communication with bridge on application start (and detect connected device right away)
+        // Set it to "true", then trezor-connect will not be initialized until you call some TrezorConnect.method()
+        // This is useful when you don't know if you are dealing with Trezor user
+        lazyLoad: communicationType !== CommunicationType.Node && lazyLoad,
+        // Manifest is required from Trezor Connect 7:
+        // https://github.com/trezor/connect/blob/develop/docs/index.md#trezor-connect-manifest
+        manifest,
+        // Show Trezor Suite popup. Disabled for node based apps
+        popup: communicationType !== CommunicationType.Node && !silentMode
+      });
+      return true;
+    } catch (error: any) {
+      if (error.code === 'Init_AlreadyInitialized') return true;
+      throw transportTypedError(error);
+    }
+  }
+
+  /**
+   * @throws AuthenticationError
+   */
+  static async checkDeviceConnection(): Promise<Features> {
+    try {
+      const deviceFeatures = await TrezorConnect.getFeatures();
+      if (!deviceFeatures.success) {
+        throw new TransportError('Failed to get device', deviceFeatures.payload);
+      }
+      if (deviceFeatures.payload.model !== 'T') {
+        throw new TransportError(`Trezor device model "${deviceFeatures.payload.model}" is not supported.`);
+      }
+      return deviceFeatures.payload;
+    } catch (error) {
+      throw transportTypedError(error);
+    }
+  }
+
+  /**
+   * @throws AuthenticationError
+   */
+  static async getXpub({ accountIndex }: GetTrezorXpubProps): Promise<Cardano.Bip32PublicKey> {
+    try {
+      await TrezorKeyAgent.checkDeviceConnection();
+      const derivationPath = `m/1852'/1815'/${accountIndex}'`;
+      const extendedPublicKey = await TrezorConnect.cardanoGetPublicKey({
+        path: derivationPath,
+        showOnTrezor: true
+      });
+      if (!extendedPublicKey.success) {
+        throw new TransportError('Failed to export extended account public key', extendedPublicKey.payload);
+      }
+      return Cardano.Bip32PublicKey(extendedPublicKey.payload.publicKey);
+    } catch (error: any) {
+      throw transportTypedError(error);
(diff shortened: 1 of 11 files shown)


packages/wallet/src/KeyManagement/TrezorKeyAgent.ts
```diff
@@ -0,0 +1,129 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { AuthenticationError, TransportError } from './errors';
+import { Cardano, NotImplementedError } from '@cardano-sdk/core';
+import { CommunicationType, KeyAgentType, SerializableTrezorKeyAgentData, SignBlobResult } from './types';
+import { KeyAgentBase } from './KeyAgentBase';
+import TrezorConnect, { Features } from 'trezor-connect';
+
+export type TrezorKeyAgentProps = Omit<SerializableTrezorKeyAgentData, '__typename'>;
+
+export interface CreateWithDevice {
```
mkazlauskas, line 10: Let's rename all 'Device' to 'Trezor' since there are now 2 hw key agent device types. Maybe review LedgerKeyAgent type names too to make them more specific, etc. Device->Ledger, Transport->LedgerTransport
```suggestion
export interface CreateTrezorKeyAgentProps {
```

packages/wallet/src/KeyManagement/types.ts
```diff
@@ -80,7 +81,14 @@ export interface SerializableLedgerKeyAgentData extends SerializableKeyAgentData
   communicationType: CommunicationType;
 }
 
-export type SerializableKeyAgentData = SerializableInMemoryKeyAgentData | SerializableLedgerKeyAgentData;
+export interface SerializableTrezorKeyAgentData extends SerializableKeyAgentDataBase {
+  __typename: KeyAgentType.Trezor;
```
mkazlauskas, line 85: Don't we need to save manifest and connection type?

mkazlauskas (changes requested):
So far so good!

packages/wallet/src/KeyManagement/types.ts
tomislavhoracek: @mkazlauskas there is no need for that. We need that only for `initializeTrezorTransport` which should be called on app start or before interacting with a device but just once per session otherwise, it will throw an error. So, we don't need it after that. Trezor connect is a little bit different and simpler than Ledger communication

packages/wallet/src/KeyManagement/TrezorKeyAgent.ts
tomislavhoracek: @mkazlauskas I wanted to unify naming to prevent breaking changes in LW but I agree that it should be better :) 
I will do that

tomislavhoracek: Done!

rhyslbw (changes requested):
Looks good @tomislavhoracek. Please squash the last commit with the first

packages/wallet/src/KeyManagement/types.ts
mkazlauskas: But if we save it we can then call it within key agent (on restoration or before signing etc.) :thinking: Otherwise user (wallet) has to treat this key agent differently from other key agents, right?

tomislavhoracek: @mkazlauskas It is a little bit different than Ledger. Even if you have e.g deviceConnection you can't call e.g. deviceConnection.getFeatures(). All operations you want to perform come from TrezorConnect e.g `TrezorConnect.getFeatures()`. In the Ledger key agent, we need a device connection to prove and to keep an open channel between the end app and Cardano App on the device. Everything is simpler in the scope of Trezor because you don't have Carano App installed on it and everything works out of the box. So, you will always communicate with the last physically connected device.

tomislavhoracek:
> squash

Thanks @rhyslbw. 
Squashed!

packages/wallet/src/KeyManagement/types.ts
mkazlauskas: From user's point of view this is not simpler, because as I understand you have to call `initializeTrezorTransport` separately, as in [this test](https://github.com/input-output-hk/cardano-js-sdk/pull/238/files#diff-7f5c8ec02d8adf78f37dfc3c9f2ef1161ae277a22ea00a9370d9209954fc4057R31) - for both creating the key agent, and restoring the key agent. It would be great to encapsulate this to be called within `LedgerKeyAgent`. I think it can be done if you add required props to `createWithDevice` and serializableData :thinking:

tomislavhoracek: @mkazlauskas but `initializeTrezorTransport` must be called only once and it won't return deviceConnection but just prepare the app to accept signals from "transport". It should be called ideally on the app start.

mkazlauskas: I understand, but why can't it be called internally?

rhyslbw: > initializeTrezorTransport must be called only once

Doesn't `createWithDevice` only get called once too? If so, is there a reason the transport cannot be initialised there?

tomislavhoracek: @mkazlauskas it could be called internally but then we need to know if is already initialized or not since it will throw an error if is called multiple times. From my perspective that is more complicated for users because they will need to take care of whether TrezorConnect has been already initialized or not. That is why I decided to do that in this way 🤔 . So, one initialization responsible for multiple devices and multiple trezor key agents and you don't user doesn't need to care if some key agent already exists to check that the connection is already initialized and to pass a prop to say to new key agent that is already initialized

@rhyslbw `Doesn't createWithDevice only get called once too? `:
yeah, it is called once per keyAgent but TrezorConnect should be initialized once on app start for all agents not per device

Maybe I'm wrong and don't see something 🤔  Don't get me wrong please, I just want to explain why I did that on this way :)

mkazlauskas: > then we need to know if is already initialized or not since it will throw an error if is called multiple times

Don't see this as a problem - we can keep track if it was already initialized with a module-scoped variable, and throw if it's already initialized but we receive different props (different communication type or different manifest).

rhyslbw: Given the goal is to make the key agent implementations as common as possible, it would be best to try and keep the concerns of initialising the global Trezor transport internal.
 
> then we need to know if is already initialised or not since it will throw an error if is called multiple times

How about converting this logic to being an idempotent operation. i.e catch and inspect the error, returning `void` if it's the expected error, or re-throw if not.

tomislavhoracek: ok, I will change that

mkazlauskas (changes requested):
Looks good but I think you still need to initializeTrezorTransport upon restoreKeyAgent

packages/wallet/src/KeyManagement/TrezorKeyAgent.ts
```diff
@@ -1,38 +1,34 @@
...
-  trezorConfig: InitializeTrezorTransportProps;
+  trezorConfig: TrezorConfig;
 }
 
 const transportTypedError = (error?: any) =>
   new AuthenticationError('Trezor transport failed', new TransportError('Trezor transport failed', error));
 
 export class TrezorKeyAgent extends KeyAgentBase {
-  constructor({ ...serializableData }: TrezorKeyAgentProps) {
+  readonly isTrezorInitialized: Promise<boolean>;
```
mkazlauskas, line 26: Since we're doing try-catch, do we really need this member variable?

It seems like it's redundant as it will always be true (we're throwing if `TrezorConnect.init` throws)

mkazlauskas (changes requested):

packages/wallet/src/KeyManagement/TrezorKeyAgent.ts
tomislavhoracek: @mkazlauskas it is more like a guard if e.g key agent is restored with serialized data and e.g signing triggered immediately we will wait for this to be resolved. I added a comment here
https://github.com/input-output-hk/cardano-js-sdk/blob/bd154c06cc81c6eee6bc5443270132fec4005141/packages/wallet/src/KeyManagement/TrezorKeyAgent.ts#L119-L125

mkazlauskas: Ah I see, it's because it's async :+1: `restoreKeyAgent` can have some logic specific to key agents so it could be awaited in there, but this works too!

mkazlauskas (approved):
Nice work! Let's just squash first and last commits again before merging

packages/wallet/src/KeyManagement/TrezorKeyAgent.ts
tomislavhoracek: 👌

tomislavhoracek:
> st commits again before

Guys @rhyslbw , @mkazlauskas Thanks for the review and all the help. Really appreciate that.

mkazlauskas (approved):

tomislavhoracek:
@mkazlauskas Squashed! Can you approve again please 🙏

rhyslbw (approved):

PR #243, opened 2022-05-09
Chain history data

lgobbi-atix:
# Context

As a developer wishing to access a full chain history
I want to have the option of using a remote service without depending on a specific one,
so that I can write application logic without being coupled to a particular service provider

# Proposed Solution

- Implement `ChainHistoryProvider`
- Implement `ChainHistoryHttpService`
- Implement `ChainHistoryHttpProvider`
- Implement `ChainHistoryDbSyncProvider`
- Implement `ChainHistoryBuilder`

- Added ChainHistory provider, queries builder and http service
- Added DBSyncChainHistory
- query transaction by addresses or hashes
- query blocks by hashes


packages/cardano-services/src/ChainHistory/openApi.json
```diff
@@ -0,0 +1,406 @@
...
+        "type": "object",
+        "properties": {
+          "args": {
+            "type": "array",
+            "items": {
+              "oneOf": [{ "type": "array", "items": { "$ref": "#/components/schemas/Address" } }, { "type": "number" }]
+            }
+          }
+        }
+      },
```
lgobbi-atix, line 397: Not sure about this

packages/cardano-services/src/ChainHistory/ChainHistoryHttpService.ts
```diff
@@ -0,0 +1,91 @@
...
+    router.use(
+      OpenApiValidator.middleware({
+        apiSpec,
+        ignoreUndocumented: true, // otherwhise /metrics endpoint should be included in spec
+        validateRequests: true,
+        validateResponses: true
+      })
+    );
+    // TODO: refactor providerHandler to get Args and Response types on its own?
+    //       and use it as providerHandler<typeof chainHistoryProvider.blocksByHashes>
```
lgobbi-atix, line 46: Is there any reason why we should pass args and response types manually? Does it make sense to change `providerHandler` to be able to use it like this instead?

```ts
providerHandler<typeof chainHistoryProvider.blocksByHashes>(async (args, _, res, handler) => {
``` 

Or even something like this: 
```ts
providerHandler(chainHistoryProvider.blocksByHashes)(async (args, _, res, handler) => {
        try {
          return HttpServer.sendJSON(res, await handler(...args));
        } catch (error) {
          logger.error(error);
          return HttpServer.sendJSON(res, new ProviderError(ProviderFailure.Unhealthy, error), 500);
        }
      }
```

mkazlauskas: No reason, I agree, inferring the types would be better

packages/core/src/Asset/util/assetId.ts
```diff
@@ -10,6 +10,15 @@ export const assetNameFromAssetId = (assetId: AssetId): AssetName => AssetName(a
 export const createAssetId = (scriptHash: CSL.ScriptHash, assetName: CSL.AssetName): AssetId =>
   AssetId(util.bytesToHex(scriptHash.to_bytes()) + util.bytesToHex(assetName.name()).toString());
 
+/**
+ * @returns {AssetId} concatenated policy id and asset name
+ */
+export const assetIdFromPolicyAndName = (policyId: PolicyId, assetName: AssetName): AssetId =>
+  createAssetId(
```
mkazlauskas, line 17: Can't we do just `policyId.toString() + assetName.toString()`? Looking at the test for this it seems to be the case

mkazlauskas (changes requested):

packages/cardano-services/src/ChainHistory/openApi.json
rhyslbw: Specifically?:
```
              "oneOf": [{ "type": "array", "items": { "$ref": "#/components/schemas/Address" } }, { "type": "number" }]
```

Yeah, why is there a `number` expected?

lgobbi-atix: The number is because `transactionsByAddresses` has a `sinceBlock` optional argument. https://github.com/input-output-hk/cardano-js-sdk/pull/243/files#diff-cc6fbcd6f5f3a6dee2c12a90216ba559ea451cb187699b838f8bed6b600d65c3R339

If we base this on that method arguments like the others the request body should be something like `args: [[addresses], blockNo]`. But I don't think we can validate that with just open api, what I did here would allow `[array, number]`, `[number,array]`, `[array,array]`, `[number, number]`, etc. We could still validate them manually in the provider but the endpoint documentation would be misleading.

Should I change it to an object instead: `args: [{ addresses, blockNo? }]`?
It will be different than the others though as the request body and the provider method arguments don't match. And if you omit `blockNo` the request would look like `args: [{ addresses: Address[] }]`, unlike the others where it's just `args: [[hashes]]`, which is a bit inconsistent I guess

rhyslbw: > We could still validate them manually in the provider handler but the endpoint documentation would be misleading

Which is an expression of the API :-)  Let's aim for self-describing interfaces at all times 👍 
 
> Should I change it to an object instead: args: [{ addresses, blockNo? }]?

With a slight consistency correction: 
`args: [{ addresses, sinceBlock? }]`

> It will be different than the others though as the request body and the provider method arguments don't match.
...
 which is a bit inconsistent I guess
 
 Please log it as a SPIKE for @mkazlauskas to review, assigned to the production-readiness epic, and we'll press on here.

packages/cardano-services/test/ChainHistory/DbSyncChainHistoryProvider/ChainHistoryBuilder.test.ts
```diff
@@ -0,0 +1,191 @@
...
+  });
+
+  describe('queryProtocolParams', () => {
+    test('query protocol parameters from last epoch', async () => {
+      const result = await builder.queryProtocolParams();
+      expect(result).toMatchSnapshot();
+    });
+  });
+
+  describe('queryCertificatesByHashes', () => {
```
iadmytro, line 169: missing cases for:
-    MirCertificate
-    GenesisKeyDelegationCertificate  
-    StakeKeyDeregistrationCertificate

iadmytro (changes requested):

packages/cardano-services/test/ChainHistory/DbSyncChainHistoryProvider/ChainHistoryBuilder.test.ts
lgobbi-atix: I'm not sure if we can get `GenesisKeyDelegationCertificate` from db-sync, I didn't make any queries to get them and they are not being fetched by the blockfrost provider either.

Added the others in https://github.com/input-output-hk/cardano-js-sdk/pull/243/commits/5372e875ce7ce0f93277b75e1f5b2bdbaa008274

packages/cardano-services/src/ChainHistory/openApi.json
lgobbi-atix: I changed the open api specification to an object in this commit https://github.com/input-output-hk/cardano-js-sdk/pull/243/commits/50827879ec95916cb72e72952de9214a3a35de68, but had to delete these tests here https://github.com/input-output-hk/cardano-js-sdk/pull/243/commits/50827879ec95916cb72e72952de9214a3a35de68#diff-cc6fbcd6f5f3a6dee2c12a90216ba559ea451cb187699b838f8bed6b600d65c3L276 because the open api validation fails when the http provider method makes a request to the http service as the request and method arguments are different now (request: `({ addresses, sinceBlock })`, method: `(addresses, sinceBlock)`). 

Besides that, I don't really see the point of those kinds of tests, the provider already has `chainHistoryHttpProvider.test.ts` and the HttpService is already covered by the tests that use `doChainHistoryRequest`

So I'm not sure what to do, should we revert this commit and leave the request body as an array? Should we change the provider method arguments? Should we leave it like this?

rhyslbw: >  Should we change the provider method arguments to match the http service?

Yes. I'll make a ticket to review all our Provider interfaces for consistency

packages/cardano-services/test/ChainHistory/DbSyncChainHistoryProvider/ChainHistoryBuilder.test.ts
rhyslbw: Please leave a comment in the test to explain why there are no tests for these cert types @lgobbi-atix

iadmytro: It's blocked by https://github.com/input-output-hk/cardano-db-sync/issues/290

rhyslbw: https://github.com/input-output-hk/cardano-js-sdk/commit/07a978ee39e38f0f8a62a662895af610b22db10c#diff-3e16e6fa3a0c30bc7dc482538653e7edca22402ca67614ba46757d93f533b0e8R150

packages/cardano-services/src/ChainHistory/DbSyncChainHistoryProvider/types.ts
```diff
@@ -0,0 +1,153 @@
...
+export type WithCertIndex<T extends Cardano.Certificate> = T & { cert_index: number };
+export type WithCertType<T extends CertificateModel> = T & {
+  type: 'retire' | 'update' | 'mir' | 'stake' | 'delegation';
+};
+
+export interface PoolRetireCertModel extends CertificateModel {
+  retiring_epoch: number;
+  pool_id: string;
+}
+export const isPoolRetireCertModel = (
```
rhyslbw, line 107: Only types or interfaces should be in this file. Please move all functions to a new `./util.ts`

packages/cardano-services/src/ChainHistory/DbSyncChainHistoryProvider/types.ts
```diff
@@ -0,0 +1,153 @@
...
+export interface TxInOutModel {
+  address: string;
+  coin_value: string;
+  datum?: Buffer | null;
+  id: string;
+  index: number;
+  tx_id: Buffer;
+}
+
+export interface TxOutput extends Cardano.TxOut {
```
rhyslbw, line 56:
```suggestion
export interface TxOutputModel extends Cardano.TxOut {
```
?

packages/cardano-services/src/Program/loadHttpServer.ts
```diff
@@ -53,6 +54,19 @@ export const loadHttpServer = async (args: ProgramArgs): Promise<HttpServer> =>
           })
         );
         break;
+      case ServiceNames.ChainHistory:
+        if (args.options?.dbConnectionString === undefined)
+          throw new MissingProgramOption(ServiceNames.ChainHistory, ProgramOptionDescriptions.DbConnection);
+        services.push(
+          await ChainHistoryHttpService.create({
+            chainHistoryProvider: new DbSyncChainHistoryProvider(
+              new Pool({ connectionString: args.options.dbConnectionString }),
```
rhyslbw, line 63: Should we hoist this before the switch to share the instance across services?

packages/cardano-services/src/types.ts
```diff
@@ -1 +1,2 @@
 export type ModuleState = null | 'initializing' | 'initialized';
+export type Awaited<T> = T extends PromiseLike<infer U> ? U : T;
```
rhyslbw, line 2: This is generic, but probably not worth moving until we have implemented the `util` package.

packages/cardano-services/src/util/hexString.ts
```diff
@@ -0,0 +1 @@
+export const hexStringToBuffer = (hex: string) => Buffer.from(hex, 'hex');
```
rhyslbw, line 1: Also generic, but also probably not worth moving until we have implemented the util package.

packages/cardano-services/test/StakePoolSeach/DbSyncStakePoolSearchProvider/__snapshots__/StakePoolSearchBuilder.test.ts.snap
```diff
@@ -140,8 +140,8 @@ Object {
 exports[`StakePoolSearchBuilder buildAndQuery buildAndQuery, queryPoolHashes & queryTotalCount 2`] = `
 Array [
   Object {
-    "id": "15",
-    "updateId": "309",
+    "id": "16",
```
rhyslbw, line 143: What changed to invalidate the snapshot?

rhyslbw (changes requested):
Need to resolve the lint warning:
image

Looking at the code I'd be fine with that rule being disbled inline

packages/cardano-services/src/ChainHistory/DbSyncChainHistoryProvider/ChainHistoryBuilder.ts
```diff
@@ -0,0 +1,187 @@
...
+  public async queryCertificatesByHashes(
+    hashes: Cardano.TransactionId[]
+  ): Promise<TransactionDataMap<Cardano.Certificate[]>> {
+    const byteHashes = hashes.map((hash) => hexStringToBuffer(hash.toString()));
+    this.#logger.debug('About to find certificates for txs:', byteHashes);
+    const poolRetireCerts: QueryResult<PoolRetireCertModel> = await this.#db.query(Queries.findPoolRetireCerts, [
+      byteHashes
+    ]);
+
+    const poolUpdateCerts: QueryResult<PoolUpdateCertModel> = await this.#db.query(Queries.findPoolUpdateCerts, [
```
iadmytro, line 141: better to rename it to poolRegisterCerts

packages/cardano-services-client/src/ChainHistoryProvider/chainHistoryHttpProvider.ts
```diff
@@ -0,0 +1,30 @@
+import { ChainHistoryProvider, ProviderError, ProviderFailure } from '@cardano-sdk/core';
+import { HttpProviderConfigPaths, createHttpProvider } from '../HttpProvider';
+
+export const defaultChainHistoryProviderPaths: HttpProviderConfigPaths<ChainHistoryProvider> = {
+  blocksByHashes: '/blocks/hash',
```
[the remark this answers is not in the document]
mkazlauskas: I agree, it is cleaner and could make a potential future refactor of hoisting this to a new BlocksProvider easier. I'm fine with a temporary divergence, changing the existing paths is a small task anyway. `blocks/by-hashes` :+1:

packages/cardano-services/src/util/hexString.ts
mkazlauskas: Yep, similar to [bytesToHex](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/core/src/util/misc/bytesToHex.ts) that's currently in `core`

mkazlauskas (changes requested):

packages/cardano-services/test/StakePoolSeach/DbSyncStakePoolSearchProvider/__snapshots__/StakePoolSearchBuilder.test.ts.snap
mkazlauskas: Snapshots failed [here](https://github.com/input-output-hk/cardano-js-sdk/runs/6507593263?check_suite_focus=true) too after merge, but the PR checks were green :thinking:

packages/cardano-services/src/ChainHistory/DbSyncChainHistoryProvider/mappers.ts
```diff
@@ -0,0 +1,239 @@
...
+    certificates,
+    withdrawals
+  }),
+  index: txModel.index,
+  txSize: txModel.size,
+  witness: {
+    // TODO: fetch bootstrap witnesses, datums and scripts
+    redeemers,
+    // TODO: fetch signatures
+    signatures: new Map()
```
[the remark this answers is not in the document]
rhyslbw: or make it an optional property?

packages/cardano-services/test/StakePoolSeach/DbSyncStakePoolSearchProvider/__snapshots__/StakePoolSearchBuilder.test.ts.snap
lgobbi-atix: I needed to add some tables to the dump and the database used before was a db that Juan Cruz had in his computer. The one I used had different ids (but same data) for some records, so I had to update the snapshots. We are now sharing the same one

packages/cardano-services/src/ChainHistory/DbSyncChainHistoryProvider/mappers.ts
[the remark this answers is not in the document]
rhyslbw: > I'd prefer not to mix the 2 styles (optional vs no property)

Agree. Let's drop it

packages/cardano-services/test/StakePoolSeach/DbSyncStakePoolSearchProvider/__snapshots__/StakePoolSearchBuilder.test.ts.snap
rhyslbw: Can the original one be used instead @lgobbi-atix, or would that just shift the impact?

rhyslbw: Can we generate a snapshot from the original DB, or would that just shift the impact?

lgobbi-atix: Thinking about it with Juan Cruz maybe it's better to use a db with the [snapshots](https://updates-cardano-testnet.s3.amazonaws.com/cardano-db-sync/index.html#12/) that db-sync provides but without running db-sync.
That way all devs could have their own frozen local database (using the same snapshot version of course) and then run the dump script on it if they need to add more tables/blocks to the fixture.

There shouldn't be any issues with ids anymore if everyone is using the same snapshot and db-sync is not running.

packages/cardano-services/src/ChainHistory/DbSyncChainHistoryProvider/mappers.ts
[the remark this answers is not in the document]
lgobbi-atix: Should we do this in another task?

packages/core/src/Provider/ChainHistoryProvider/types.ts
```diff
@@ -0,0 +1,32 @@
...
+   * @returns {Cardano.TxAlonzo[]} an array of transactions involving the addresses
+   */
+  transactionsByAddresses: (args: TransactionsByAddressesArgs) => Promise<Cardano.TxAlonzo[]>;
+  /**
+   * Gets the transactions matching the provided hashes.
+   *
+   * @param {Cardano.TransactionId[]} hashes array of transaction ids
+   * @returns {Cardano.TxAlonzo[]} an array of transactions
+   */
+  transactionsByHashes: (hashes: Cardano.TransactionId[]) => Promise<Cardano.TxAlonzo[]>;
```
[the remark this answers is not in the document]
lgobbi-atix: Done in 70dd1e23cbb08257e051305e215dc03c001d0d99

packages/cardano-services-client/src/ChainHistoryProvider/chainHistoryHttpProvider.ts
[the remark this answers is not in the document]
lgobbi-atix: Done in 70dd1e23cbb08257e051305e215dc03c001d0d99

packages/cardano-services-client/test/ChainHistoryProvider/chainHistoryProvider.test.ts
```diff
@@ -0,0 +1,93 @@
+/* eslint-disable sonarjs/no-duplicate-string */
+import { ProviderError, ProviderFailure, util } from '@cardano-sdk/core';
+import { chainHistoryHttpProvider } from '../../src';
+import got, { HTTPError, Response } from 'got';
```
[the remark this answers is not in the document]
lgobbi-atix: Done in 70dd1e23cbb08257e051305e215dc03c001d0d99

packages/cardano-services/src/ChainHistory/DbSyncChainHistoryProvider/types.ts
lgobbi-atix: Done in 70dd1e23cbb08257e051305e215dc03c001d0d99

packages/cardano-services/src/ChainHistory/DbSyncChainHistoryProvider/types.ts
lgobbi-atix: Done in 70dd1e23cbb08257e051305e215dc03c001d0d99

packages/cardano-services/src/Program/loadHttpServer.ts
lgobbi-atix: Done in 70dd1e23cbb08257e051305e215dc03c001d0d99

packages/cardano-services/test/StakePoolSeach/DbSyncStakePoolSearchProvider/__snapshots__/StakePoolSearchBuilder.test.ts.snap
lgobbi-atix: We also found another issue with the order of the queries' results, apparently it was not always the same. For example, running `yarn test chainhistoryhttpservice` would pass, but running `yarn test` would fail because it was expecting the same data but in a different order in the snapshot. I added `order by` to the queries and now it's working

packages/cardano-services/src/ChainHistory/DbSyncChainHistoryProvider/ChainHistoryBuilder.ts
lgobbi-atix: Done in 70dd1e23cbb08257e051305e215dc03c001d0d99

packages/cardano-services/src/ChainHistory/ChainHistoryHttpService.ts
```diff
@@ -0,0 +1,68 @@
...
+    router.use(
+      OpenApiValidator.middleware({
+        apiSpec,
+        ignoreUndocumented: true, // otherwhise /metrics endpoint should be included in spec
+        validateRequests: true,
+        validateResponses: true
+      })
+    );
+
+    const routeHandler: ProviderHandler = async (args, _r, res, _n, handler) => {
```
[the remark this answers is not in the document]
lgobbi-atix: Done in 70dd1e23cbb08257e051305e215dc03c001d0d99

packages/cardano-services/package.json
```diff
@@ -17,7 +17,7 @@
   "license": "MPL-2.0",
   "scripts": {
     "build": "tsc --build ./src && yarn copy-openApi",
-    "copy-openApi": "cp ./src/StakePoolSearch/openApi.json ./dist/StakePoolSearch/openApi.json && cp ./src/TxSubmit/openApi.json ./dist/TxSubmit/openApi.json",
+    "copy-openApi": "cp ./src/StakePoolSearch/openApi.json ./dist/StakePoolSearch/openApi.json && cp ./src/TxSubmit/openApi.json ./dist/TxSubmit/openApi.json && cp ./src/ChainHistory/openApi.json ./dist/ChainHistory/openApi.json",
```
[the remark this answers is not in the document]
lgobbi-atix: Done in 70dd1e23cbb08257e051305e215dc03c001d0d99

packages/cardano-services/test/StakePoolSeach/DbSyncStakePoolSearchProvider/__snapshots__/StakePoolSearchBuilder.test.ts.snap
lgobbi-atix: That should be the reason why this was happening

> Snapshots failed [here](https://github.com/input-output-hk/cardano-js-sdk/runs/6507593263?check_suite_focus=true) too after merge, but the PR checks were green

lgobbi-atix: CI now is failing because of dates, it looks like an issue with timezones. There is a 3 hour difference between expected and the snapshot, which I guess makes sense because I'm in UTC-3

rhyslbw (changes requested):
@lgobbi-atix the changes in 70dd1e23cbb08257e051305e215dc03c001d0d99 look good, but you won't be able to squash this into the relevant commits without keeping them separate here for the review. Remember, the end goal is to merge atomic changes of the `master` branch state, and the PR branch is your staging area to do so.

packages/cardano-services/src/ChainHistory/DbSyncChainHistoryProvider/mappers.ts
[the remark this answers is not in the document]
rhyslbw: Yes, a separate task. I've made a ticket ADP-1792

packages/cardano-services/test/StakePoolSeach/DbSyncStakePoolSearchProvider/__snapshots__/StakePoolSearchBuilder.test.ts.snap
rhyslbw: Do we have relative dates in the dataset? The DB just stores UTC. We could strip some irrelevant data like DB IDs and dates?

lgobbi-atix: It looks like the issue is that db-sync stores timestamps in UTC but without timezone (`timestamp` instead of `timestamptz`), so when making a query it assumes it's stored in the computer's timezone and tries to convert it to UTC, in my case by adding 3 extra hours.

For example, for [block 1598507](https://explorer.cardano-testnet.iohkdev.io/en/block?id=7a48b034645f51743550bbaf81f8a14771e58856e031eb63844738ca8ad72298), with time `2020/07/29 03:58:16 UTC`

If I just do `SELECT time FROM block WHERE block_no = 1598507;` I get as the `db.query` result: `time:  2020-07-29T06:58:16.000Z` (which would actually be UTC+3).
But if I specify in which timezone the `time` is by doing `SELECT time AT TIME ZONE 'UTC' FROM block WHERE block_no = 1598507;` it should always return it as `time:  2020-07-29T03:58:16.000Z`, which is correct

ssong-iohk (changes requested):
1. I've commented with missing test cases in `ChainHistoryBuilder.test.ts`. 
2. Unit test for `DbSyncChainHistoryProvider` class is missing. If relevant tests are existed, please point them.

packages/cardano-services/test/ChainHistory/DbSyncChainHistoryProvider/ChainHistoryBuilder.test.ts
```diff
@@ -0,0 +1,217 @@
...
+      const result = await builder.queryTransactionInputsByHashes([]);
+      expect(result).toHaveLength(0);
+    });
+    test('query transaction inputs when tx hashes not found', async () => {
+      const result = await builder.queryTransactionInputsByHashes([
+        Cardano.TransactionId('0000000000000000000000000000000000000000000000000000000000000000')
+      ]);
+      expect(result).toHaveLength(0);
+    });
+
```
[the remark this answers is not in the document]
rhyslbw: This would fail in the `Cardano.TransactionId` function, which is tested [here](https://github.com/input-output-hk/cardano-js-sdk/blob/dc33f153288255b256453209433f08dae1a22291/packages/core/test/Cardano/types/Transaction.test.ts). It would be best to branch off _master_, and include it there.

packages/cardano-services/test/ChainHistory/DbSyncChainHistoryProvider/ChainHistoryBuilder.test.ts
```diff
@@ -0,0 +1,217 @@
...
+      expect(result).toMatchSnapshot();
+    });
+    test('query multi assets with empty array', async () => {
+      const result = await builder.queryMultiAssetsByTxOut([]);
+      expect(result.size).toEqual(0);
+    });
+    test('query multi assets when tx output not found or has no multi-assets', async () => {
+      const result = await builder.queryMultiAssetsByTxOut(['0', '5396852']);
+      expect(result.size).toEqual(0);
+    });
```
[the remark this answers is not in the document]
lgobbi-atix: You are right, changed it to `BigInt` https://github.com/input-output-hk/cardano-js-sdk/pull/243/commits/6fc21e8ed6f3d5f7a7b13c644a379001f15c9568#diff-3e16e6fa3a0c30bc7dc482538653e7edca22402ca67614ba46757d93f533b0e8L62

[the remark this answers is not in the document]
rhyslbw: Great, so with that change it's covered by type safety.

packages/cardano-services/test/ChainHistory/DbSyncChainHistoryProvider/ChainHistoryBuilder.test.ts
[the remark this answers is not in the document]
ssong-iohk: LGTM

ssong-iohk (approved):

rhyslbw (changes requested):
For consistency, class filenames should match the class name. Also, the structure of the module could be improved by dropping `Provider` from the dir name:

```
packages/cardano-services/src/ChainHistory/DbSyncChainHistoryProvider/DbSyncChainHistory.ts
```

```
packages/cardano-services/src/ChainHistory/DbSyncChainHistory/DbSyncChainHistoryProvider.ts
```

lgobbi-atix:
> 2. Unit test for `DbSyncChainHistoryProvider` class is missing. If relevant tests are existed, please point them.

I could add a new `DbSyncChainHistoryProvider.test.ts` file but the provider should already be covered with these kind of tests https://github.com/input-output-hk/cardano-js-sdk/pull/243/files#diff-cc6fbcd6f5f3a6dee2c12a90216ba559ea451cb187699b838f8bed6b600d65c3R101

lgobbi-atix:
> For consistency, class filenames should match the class name. Also, the structure of the module could be improved by dropping `Provider` from the dir name:
> 
> ```
> packages/cardano-services/src/ChainHistory/DbSyncChainHistoryProvider/DbSyncChainHistory.ts
> ```
> 
> ```
> packages/cardano-services/src/ChainHistory/DbSyncChainHistory/DbSyncChainHistoryProvider.ts
> ```

@rhyslbw Done in https://github.com/input-output-hk/cardano-js-sdk/pull/243/commits/790b52abc46b7b238a27ea16ed17c50f810fea38

rhyslbw:
> but the provider should already be covered with these kind of tests

Yes, given we're mapping 1:1, the coverage will be sufficient.

ssong-iohk:
> > 2. Unit test for `DbSyncChainHistoryProvider` class is missing. If relevant tests are existed, please point them.
> 
> I could add a new `DbSyncChainHistoryProvider.test.ts` file but the provider should already be covered with these kind of tests https://github.com/input-output-hk/cardano-js-sdk/pull/243/files#diff-cc6fbcd6f5f3a6dee2c12a90216ba559ea451cb187699b838f8bed6b600d65c3R101

Thanks. I can see generic happy/negative cases in `chainHistoryByProvider.test.ts`. However, some negative cases could be caused by given arguments are missing or exception occurred inside the method (`transactionsByHashes()`). I will create a ticket to cover these cases later.

rhyslbw (approved):
Approved. Ready for final squashing, including the last into 0146f7c29148fd6430531efc196aea3f568c72d2 if that can be done cleanly.

rhyslbw (approved):
Great work @lgobbi-atix 🎉

PR #242, opened 2022-05-09
Refactor/modularize stake pool queries

jvieiro:
# Context

In order to improve testing quality, modularize `DbSyncStakePoolSearch` queries

# Proposed Solution

Create a new dependency into `DbSyncStakePoolSearch` called `StakePoolSearchBuilder` with the previous `DbSyncStakePoolSearch's` private methods (query builders and query calls).

# Important Changes Introduced

n/a

PR #246, opened 2022-05-11
Feat/expose unpaginated result count in stake pool search

jvieiro:
# Context

Let the consumer of StakePoolSearch service know the total amount of stake pools that satisfies the given search.

# Proposed Solution

- [x] Modify StakePoolProvider interface
- [x] Add total count query, mapper and types
- [x] Modify corresponding returning types 
- [x] Modify openApi.json schema

n/a

Depends on #242

### DIFF
--- a/packages/wallet/src/services/DelegationTracker/RewardAccounts.ts
+++ b/packages/wallet/src/services/DelegationTracker/RewardAccounts.ts
@@ -22,13 +22,15 @@ export const createQueryStakePoolsProvider =
       store.getValues(poolIds),
       coldObservableProvider(
         () =>
-          stakePoolSearchProvider.queryStakePools({
-            filters: { identifier: { values: poolIds.map((poolId) => ({ id: poolId })) } }
-          }),
+          stakePoolSearchProvider
+            .queryStakePools({
+              filters: { identifier: { values: poolIds.map((poolId) => ({ id: poolId })) } }
+            })
+            .then(({ pageResults }) => pageResults),
         retryBackoffConfig
       ).pipe(
-        tap((stakePools) => {
-          for (const stakePool of stakePools) {
+        tap((pageResults) => {
+          for (const stakePool of pageResults) {
             store.setValue(stakePool.id, stakePool);
           }
         })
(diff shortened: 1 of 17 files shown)


packages/wallet/src/services/DelegationTracker/RewardAccounts.ts
```diff
@@ -22,13 +22,15 @@ export const createQueryStakePoolsProvider =
...
       coldObservableProvider(
         () =>
-          stakePoolSearchProvider.queryStakePools({
-            filters: { identifier: { values: poolIds.map((poolId) => ({ id: poolId })) } }
-          }),
+          stakePoolSearchProvider
+            .queryStakePools({
+              filters: { identifier: { values: poolIds.map((poolId) => ({ id: poolId })) } }
+            })
+            .then(({ pageResults }) => pageResults),
```
rhyslbw, line 29: We need to handle the `totalCount` too, which is needed for pagination. Please review the application implementation to guide us 👍

rhyslbw (changes requested):
The last three commits can be squashed, as they're all part of the same atomic change of
> add totalResultCount to StakePoolSearch

- 26d082dbb18905afaef94b4fb93fb70ae893fe72
- e32717250a29967a6f71809c2a4ddfa24e361208
- 2ac693104da6895efe3fb43f39d946049287c771

packages/wallet/src/services/DelegationTracker/RewardAccounts.ts
jvieiro: Maybe we can return an object with this structure:
```
{
   totalCount: number,
   stakePools: KeyValueStore<Cardano.PoolId, Cardano.StakePool>
}
```

rhyslbw: Actually, ignore this feedback :smile:. `totalCount` is not required in this context, as it's here to simply cache the stake pool the wallet is delegated to. The app interfaces directly with the provider for search results, so can handle the new return type as it sees fit.

rhyslbw (approved):
Great work @jvieiro :tada:

PR #248, opened 2022-05-13
Rewards data

jvieiro:
# Context

Create independent service to access rewards information.

# Proposed Solution

- [x] Implement RewardsProvider
- [x] Implement RewardsHttpService
- [x] Implement RewardsHttpProvider
- [x] Implement RewardsDbSyncProvider
- [x] Move rewards methods from WalletProvider to Rewards Provider

Move rewards methods from WalletProvider to Rewards Provider


packages/cardano-services/src/Reward/RewardHttpService.ts
```diff
@@ -0,0 +1,61 @@
...
+        apiSpec,
+        ignoreUndocumented: true,
+        validateRequests: true,
+        validateResponses: true
+      })
+    );
+    // Add initial healthCheck of the provider when implemented
+    router.post(
+      '/account-balance',
+      providerHandler<[Cardano.RewardAccount], bigint>(async ([rewardAccount], _, res) => {
```
mkazlauskas, line 39: Catch block here is exactly the same for both handlers (and I suspect most other provider http services too). Maybe we could refactor it to hoist try-catch code to `providerHandler`, adding an optional argument like `errorHandler?: (error: unknown) => void`

Then maybe we could simplify it to something like this, as the success status code is always the same for all providers:
```suggestion
      providerHandler<[Cardano.RewardAccount], bigint>(async ([rewardAccount], _, res) => rewardProvider.rewardAccountBalance(rewardAccount))
```

mkazlauskas (changes requested):
Looking good, just need to:
- Resolve CI error. Not sure how it got in there as it's not from changes in this PR :thinking: 
- Update `wallet/test/e2e` tests and make sure they still pass
- After approvals when squashing: faa34a2baf631c6cb467c27e8bd9b3dcd4d0b28f probably deserves a separate commit, so it could use a better commit message

packages/cardano-services/src/Reward/RewardHttpService.ts
jvieiro: > Catch block here is exactly the same for both handlers (and I suspect most other provider http services too). Maybe we could refactor it to hoist try-catch code to `providerHandler`, adding an optional argument like `errorHandler?: (error: unknown) => void`
> 
> Then maybe we could simplify it to something like this, as the success status code is always the same for all providers:
> 
> This is actually duplicate with what @lgobbi-atix did [here](https://github.com/input-output-hk/cardano-js-sdk/pull/248#discussion_r877859900) so it can be resolved when merging

Sorry but this link redirects me to my current page. So, this suggestion may be already resolved in Leo's PR?

mkazlauskas: Sorry must have pasted in the wrong link. [This](https://github.com/input-output-hk/cardano-js-sdk/pull/243/files#diff-d403c7615a0c015170e3bc860f71607b78c5a840ca10bfc832991ba1760f7d65R34) is the correct one. Yes, it resolves this suggestion

packages/cardano-services/src/Program/ServiceNames.ts
```diff
@@ -4,5 +4,6 @@
  */
 export enum ServiceNames {
   StakePoolSearch = 'stake-pool-search',
-  TxSubmit = 'tx-submit'
+  TxSubmit = 'tx-submit',
+  Reward = 'rewards'
```
mkazlauskas, line 8:
```suggestion
  Rewards = 'rewards'
```

mkazlauskas (changes requested):
Let's make the names consistent: since it's called Reward**s**Provider, names in code and filenames should also be plural (e.g. TrackedRewardProvider -> TrackedRewardsProvider).

packages/cardano-services/src/Program/ServiceNames.ts
jvieiro: Yes, sorry. I changed it in a new commit to then be squashed once approved, sounds good?

rhyslbw: Yes, that's the correct way to resolve.

packages/cardano-services/test/Reward/RewardsHttpService.test.ts
```diff
@@ -0,0 +1,168 @@
...
+import { Pool } from 'pg';
+import { getPort } from 'get-port-please';
+import { rewardsHttpProvider } from '@cardano-sdk/cardano-services-client';
+import got from 'got';
+
+const APPLICATION_JSON = 'application/json';
+const UNSUPPORTED_MEDIA_STRING = 'Response code 415 (Unsupported Media Type)';
+const APPLICATION_CBOR = 'application/cbor';
+
+describe('RewardsHttpService', () => {
```
iadmytro, line 14: missing cases for malformed json

packages/cardano-services/test/Reward/RewardsHttpService.test.ts
```diff
@@ -0,0 +1,168 @@
...
+        });
+      });
+      describe('rewardsHistory', () => {
+        it('returns rewards address history', async () => {
+          const response = await provider.rewardsHistory({
+            rewardAccounts: [rewardAcc]
+          });
+          expect(response).toMatchSnapshot();
+        });
+        it('returns address rewards history with epochs ', async () => {
```
iadmytro, line 137: empty response

iadmytro (changes requested):

packages/wallet/test/SingleAddressWallet/load.test.ts
```diff
@@ -130,10 +138,18 @@ const assertWalletProperties2 = async (wallet: ObservableWallet) => {
...
     await assertWalletProperties(wallet1, somePartialStakePools[0].id);
     wallet1.shutdown();
-    const wallet2 = await createWallet(stores, mocks.mockWalletProvider2(100), mocks.mockUtxoProvider2(100));
+    const wallet2 = await createWallet(stores, {
+      rewardsProvider: mocks.mockRewardsProvider2(100),
+      utxoProvider: mocks.mockUtxoProvider2(100),
+      walletProvider: mocks.mockWalletProvider2(100)
+    });
     await assertWalletProperties(wallet2, somePartialStakePools[0].id);
     await waitForWalletStateSettle(wallet2);
```
jvieiro, line 154: This method is what makes the tests timeout. But not sure why

packages/core/src/Provider/WalletProvider/types.ts
```diff
@@ -16,26 +16,13 @@ export type ProtocolParametersRequiredByWallet = Required<
...
-export interface RewardHistoryProps {
-  rewardAccounts: Cardano.RewardAccount[];
-  epochs?: EpochRange;
-}
-
-export interface EpochRewards {
-  epoch: Cardano.Epoch;
-  rewards: Cardano.Lovelace;
-}
+export type StakePoolStats = {
```
rhyslbw: Maybe a merge resolution mistake?

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -337,6 +337,8 @@ export class SingleAddressWallet implements ObservableWallet {
     this.txSubmitProvider.stats.shutdown();
     this.networkInfoProvider.stats.shutdown();
     this.stakePoolProvider.stats.shutdown();
+    this.utxoProvider.stats.shutdown();
```
rhyslbw, line 340: The inclusion of `this.utxoProvider.stats.shutdown();` is a separate fix commit we'll merge to master. The other changes in this PR can be squashed into the original commit.

packages/wallet/test/SingleAddressWallet/load.test.ts
rhyslbw: Was it fixed in ad3209143fee1180ed4f3c0793331f1c6ba5628c?

packages/wallet/src/SingleAddressWallet.ts
mkazlauskas: #268 has this fix too, so you could just remove it from this PR

packages/wallet/test/SingleAddressWallet/load.test.ts
jvieiro: That's correct

rhyslbw (dismissed):

mkazlauskas (changes requested):
Need to also update:
- `web-extension/e2e/.env.example`
- `web-extension/e2e/extension/background/config.ts`
- `web-extension/e2e/extension/background/wallet.ts`

rhyslbw (approved):

mkazlauskas (approved):

iadmytro (approved):
Testcase "RewardsHttpService healthy state with rewardsHttpProvider rewardsHistory returns address rewards history with epochs  1" will be updated in separate PR

PR #249, opened 2022-05-16
refactor: use `axios` in `HttpProvider` instead of `got` to support browser usage

lgobbi-atix:
# Context

`got` does not support browser environment so we need to replace it with something else.

# Proposed Solution

Replace `got` in `HttpProvider` with `axios`, which can be used in both node and browser environments

- refactor `HttpProvider`
- replace all uses of `got` with `axios` in tests


rhyslbw (approved):
Nice work @lgobbi-atix

mkazlauskas (approved):
`cardano-services` tests were fine with `got`, but I can see how it's now better for consistency within SDK. Nice work!

PR #250, opened 2022-05-16
feat(cardano-services): add ordering control to the stake pool search interface

IvayloAndonov:
# Context

Add ordering control to the stake pool search interface

# Proposed Solution

## Note
Sort by `saturation` needs to be discussed before this PR is finalized. ( WIP ) [comments](https://input-output.atlassian.net/browse/ADP-1755)

# Important Changes Introduced

- extend stake pool query options interface with `sort`
- apply sort by order and field ( name ) on SQL query
- apply default sort if not provided - ASC by NAME
- add debug config for HTTP service if needed

Most of the DB Stake pools [snapshots]([StakePoolSearchHttpService.test.ts.snap](https://github.com/input-output-hk/cardano-js-sdk/pull/250/files#diff-7b52e133bc30c5d563eb2f5850d621e7e4165310759d49e913e7599f07970264)) are modified due to the default sort of stake pools. (by name ascending) - [StakePoolSearchHttpService.test.ts.snap](https://github.com/input-output-hk/cardano-js-sdk/pull/250/files#diff-7b52e133bc30c5d563eb2f5850d621e7e4165310759d49e913e7599f07970264)

PR #252, opened 2022-05-17
feat: add ordering to stake pool search interface

IvayloAndonov:
# Context

Add ordering control to the stake pool search interface

# Proposed Solution

Apply a sort by given field and order at the database level before the actual pagination. This ensures that in case we have passed pagination with sort, the entire result set will be sorted A-z, not only each individual page.

- extend stake pool query options interface with a sort
- apply sort by order and field (order by name) on queryPoolData query
- apply default sort if not provided (order by name then by pool id ascending)
- apply the pagination after sorting
- add debug config for local HTTP service (VSCode)

Most of the DB Stake pools snapshots are modified due to the default sort of stake pools. (by name ascending) - [StakePoolSearchHttpService.test.ts.snap](https://github.com/input-output-hk/cardano-js-sdk/pull/250/files#diff-7b52e133bc30c5d563eb2f5850d621e7e4165310759d49e913e7599f07970264)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,18 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Cardano http server",
+            "type": "node",
+            "request": "attach",
+            "port": 9229,
+            "address": "localhost",
+            "protocol": "inspector",
+            "localRoot": "${workspaceRoot}",
+            "remoteRoot": "/usr/app"
+        }
+    ]
+}
--- a/packages/cardano-services/package.json
+++ b/packages/cardano-services/package.json
@@ -28,6 +28,7 @@
     "prepack": "yarn build",
     "test:debug": "DEBUG=true yarn test",
     "run:http-server": "ts-node --transpile-only src/run.ts",
+    "run:http-server:debug": "npx nodemon --legacy-watch --exec 'node -r ts-node/register --inspect=0.0.0.0:9229 src/run.ts'",
     "cli": "ts-node --transpile-only src/cli.ts"
   },
   "devDependencies": {
--- a/packages/cardano-services/test/StakePoolSeach/StakePoolSearchHttpService.test.ts
+++ b/packages/cardano-services/test/StakePoolSeach/StakePoolSearchHttpService.test.ts
@@ -11,9 +11,24 @@ import got from 'got';
 const UNSUPPORTED_MEDIA_STRING = 'Response code 415 (Unsupported Media Type)';
 const APPLICATION_CBOR = 'application/cbor';
 const APPLICATION_JSON = 'application/json';
+const STAKE_POOL_NAME = 'THE AMSTERDAM NODE';
 
 const setFilterCondition = (options: StakePoolQueryOptions, condition: 'and' | 'or'): StakePoolQueryOptions => ({
   filters: { ...options.filters, _condition: condition }
+});
+
+const setSortCondition = (
+  options: StakePoolQueryOptions,
+  order: 'asc' | 'desc',
+  field: 'name'
+): StakePoolQueryOptions => ({
+  ...options,
+  sort: { ...options.sort, field, order }
+});
+
+const setPagination = (options: StakePoolQueryOptions, startAt: number, limit: number): StakePoolQueryOptions => ({
+  ...options,
+  pagination: { ...options.pagination, limit, startAt }
 });
 
 const addStatusFilter = (options: StakePoolQueryOptions, status: Cardano.StakePoolStatus): StakePoolQueryOptions => ({
@@ -167,7 +182,7 @@ describe('StakePoolSearchHttpService', () => {
               identifier: {
                 _condition: 'or',
                 values: [
-                  { name: 'THE AMSTERDAM NODE' },
+                  { name: STAKE_POOL_NAME },
                   { name: 'banderini' },
                   { ticker: 'TEST' },
                   { id: '98a6rqr7a7yakanm5sz6asx9gfjsr343mus0tsye23wmg70' as unknown as Cardano.PoolId }
@@ -290,7 +305,7 @@ describe('StakePoolSearchHttpService', () => {
             identifier: {
               _condition: 'or',
               values: [
-                { name: 'THE AMSTERDAM NODE' },
+                { name: STAKE_POOL_NAME },
                 { name: 'banderini' },
                 { ticker: 'TEST' },
                 { id: Cardano.PoolId('pool1jcwn98a6rqr7a7yakanm5sz6asx9gfjsr343mus0tsye23wmg70') }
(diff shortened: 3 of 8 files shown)


packages/cardano-services/test/StakePoolSeach/StakePoolSearchHttpService.test.ts
```diff
@@ -468,6 +483,75 @@ describe('StakePoolSearchHttpService', () => {
...
+        });
+
+        it('sort by name asc order', async () => {
+          const response = await doServerRequest(setSortCondition({}, 'asc', 'name'));
+          expect(response).toMatchSnapshot();
+        });
+
+        it('if sort not provided defaults by name asc order', async () => {
+          const response = await doServerRequest({});
+          expect(response).toMatchSnapshot();
```
rhyslbw, line 513: Maybe a better test here is to query explicitly and compare the result, rather than writing a snapshot that needs to be verified

packages/cardano-services/test/StakePoolSeach/__snapshots__/StakePoolSearchHttpService.test.ts.snap
```diff
@@ -3091,6 +3001,21 @@ Object {
...
       "vrf": "3409a1bebeaa47e6d99e0748a99f65dee60b7f7e9a64dc865d52b4fb445b98ab",
     },
+  ],
+  "totalResultCount": 6,
+}
+`;
+
+exports[`StakePoolSearchHttpService healthy state /search search pools by multiple filters identifier & status  & pledgeMet filters pledgeMet true, status activating, and condition 1`] = `
+Object {
+  "pageResults": Array [],
```
rhyslbw, line 3011: DB snapshot likely doesn't include results, and it's not practical to try and capture another snapshot with them included. Ensuring test coverage here is not part of this PR scope though, as it's present in _master_

rhyslbw (changes requested):
Looking good @IvayloAndonov 

I think we need to add a few tests that stub the DB response, as we're missing coverage demonstrating the ordering when there's no bech32 encoded `poolId`, and also the inline comment pointing out the empty response in our current snapshot.

Also, small typo in your commit message in d362432b254ff7e9ce015b2b10f2d05984dbdea7 (extent -> extend)

packages/cardano-services/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@cardano-sdk/cardano-services",
-  "version": "0.2.0",
+  "version": "0.3.0",
```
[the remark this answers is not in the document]
IvayloAndonov: I saw that we use `"npx standard-version"` to bump versions but also noticed the `package.json` was updated automatically once I added the debug script to it, fixed now.

packages/cardano-services/test/StakePoolSeach/__snapshots__/StakePoolSearchHttpService.test.ts.snap
IvayloAndonov: Yes, I saw there are multiple places with empty results related to StakePoolSearchHttpService in _master_, no actions need to be taken as part of this PR correct?

rhyslbw: Correct. I'll make a ticket for this to be looked at separately

packages/cardano-services/test/StakePoolSeach/StakePoolSearchHttpService.test.ts
IvayloAndonov: Addressed [79ff5bc](https://github.com/input-output-hk/cardano-js-sdk/pull/252/commits/79ff5bcc52709801b3afd7d2ddb9b65950738d58)

packages/cardano-services/package.json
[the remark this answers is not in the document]
IvayloAndonov: Addressed here [0c1b623](https://github.com/input-output-hk/cardano-js-sdk/pull/252/commits/0c1b623ba927b5f3e57d0a9f1af95bb9ca8e569b)

packages/cardano-services/test/StakePoolSeach/StakePoolSearchHttpService.test.ts
```diff
@@ -508,11 +517,35 @@ describe('StakePoolSearchHttpService', () => {
           expect(response).toMatchSnapshot();
         });
 
-        it('if sort not provided defaults by name asc order', async () => {
+        it('if sort not provided defaults order by name and then by pool id asc', async () => {
           const response = await doServerRequest({});
+
+          const resultSortedCopy = [...response.pageResults].sort(sortByNameThenByPoolId);
+
+          expect(response.pageResults).toEqual(resultSortedCopy);
```
rhyslbw, line 525: Nice. I was thinking of just making two queries, one with the explicit sort and the other without, then compare, but this better communicates the expectation.

packages/cardano-services/test/StakePoolSeach/StakePoolSearchHttpService.test.ts
```diff
@@ -468,6 +483,108 @@ describe('StakePoolSearchHttpService', () => {
...
+        it('if sort not provided defaults order by name and then by pool id asc', async () => {
+          const response = await doServerRequest({});
+
+          const resultSortedCopy = [...response.pageResults].sort(sortByNameThenByPoolId);
+
+          expect(response.pageResults).toEqual(resultSortedCopy);
+          expect(response).toMatchSnapshot();
+        });
+
+        it('if stake pool name is not presented, order should be then by pool id', async () => {
```
rhyslbw, line 529:
```suggestion
        it('positions stake pools with no name registered after named pools, sorted by poolId', async () => {
```

packages/cardano-services/test/StakePoolSeach/StakePoolSearchHttpService.test.ts
```diff
@@ -468,6 +483,108 @@ describe('StakePoolSearchHttpService', () => {
...
+          const fistNoMetadataPoolId = Cardano.PoolId('pool126zlx7728y7xs08s8epg9qp393kyafy9rzr89g4qkvv4cv93zem');
+          const secondNoMetadataPoolId = Cardano.PoolId('pool1y25deq9kldy9y9gfvrpw8zt05zsrfx84zjhugaxrx9ftvwdpua2');
```
rhyslbw, line 531:
```suggestion
          const secondNoMetadataPoolId = Cardano.PoolId('pool1y25deq9kldy9y9gfvrpw8zt05zsrfx84zjhugaxrx9ftvwdpua2');
          const firstNamedPoolId = Cardano.PoolId('...');
          const secondNamedPoolId = Cardano.PoolId('...');
```

rhyslbw: Let's cover this scenario where there's a mix of named pools and unnamed pools, as that's the real-world scenario

rhyslbw (changes requested):

rhyslbw (approved):
Great work @IvayloAndonov :tada: 
Please squash to minimal commits, for QA review

rhyslbw (approved):

.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
+{
+    // Use IntelliSense to learn about possible attributes.
```
mkazlauskas, line 2: Technically JSON doesn't support comments
```suggestion
    "__comment": "Use IntelliSense to learn about possible attributes."
```

mkazlauskas (approved):
Nice work @IvayloAndonov! :tada:

PR #240, opened 2022-04-29
UTxO data

jvieiro:
# Context

Move `utxosByAddresses` from `WalletProvider` to a new http service

# Proposed Solution

- [x] Implement UtxoProvider
- [x] Implement UtxoHttpProvider
- [x] Implement UtxoHttpService
- [x] Implement UtxoDbSyncProvider
- [x] move utxosByAddresses method from WalletProvider to UtxoProvider
- [x] Add Tests

# Important Changes Introduced
n/a

Depends on #242

PR #253, opened 2022-05-17
refactor!: rename `StakePoolSearchProvider` to `StakePoolProvider`

lgobbi-atix:
# Context

Generalize `StakePoolSearchProvider` as part of the refactor to remove `WalletProvider`

# Proposed Solution

Rename `StakePoolSearchProvider` to `StakePoolProvider` and move `stakePoolStats` from `WalletProvider`

- Renamed all occurrences of `StakePoolSearchProvider` to `StakePoolProvider`
- Removed `stakePoolStats` from `WalletProvider`
- Added `stakePoolStats` to `StakePoolProvider` with its corresponding query and endpoint


rhyslbw (changes requested):
I found a few more occurrences of "StakePoolSearch":

- [This](https://github.com/input-output-hk/cardano-js-sdk/blob/e8142e90997ae2e1a31ea2337d527dbe3fde85b5/packages/wallet/test/services/ProviderTracker/TrackedUtxoProvider.test.ts#L6) is an error, but please fix in this PR too.
- https://github.com/input-output-hk/cardano-js-sdk/blob/773c926b78ee9bfbad7e25e54288401298a3f548/packages/cardano-services/test/StakePool/StakePoolHttpService.test.ts#L85
- https://github.com/input-output-hk/cardano-js-sdk/blob/773c926b78ee9bfbad7e25e54288401298a3f548/packages/cardano-services/test/StakePool/StakePoolHttpService.test.ts#L110

rhyslbw (approved):
Thanks @lgobbi-atix

packages/blockfrost/src/blockfrostWalletProvider.ts
```diff
@@ -50,26 +50,18 @@ export const blockfrostWalletProvider = (blockfrost: BlockFrostAPI, logger = dum
...
-      return newCount;
-    };
-    return {
-      qty: {
-        active: await tallyPools('pools'),
-        retired: await tallyPools('poolsRetired'),
-        retiring: await tallyPools('poolsRetiring')
-      }
-    };
+  const utxoByAddresses: WalletProvider['utxoByAddresses'] = async (addresses) => {
```
rhyslbw, line 53: Why is this being reintroduced? 
https://github.com/input-output-hk/cardano-js-sdk/pull/240/commits/e8142e90997ae2e1a31ea2337d527dbe3fde85b5

lgobbi-atix: Removed

mkazlauskas (approved):
Looks good to me!

rhyslbw (dismissed):
``` ts
export const mockWalletProvider = (): WalletProvider => ({
  ...
});
```

Looks like we're missing the typing in both ` mockAssetProvider` and `mockNetworkInfoProvider` too. They can all be committed along with the fix.

mkazlauskas:
> Looks like we're missing the typing in both mockAssetProvider and mockNetworkInfoProvider too. They can all be committed along with the fix.

That is actually intentionally not typed, so that tests can do stuff like `provider.method.mockResolvedValue(...)` instead of `(provider as jest.Mock).method.mockResolvedValue(...)`.

It's losing type-checks for compatibility with provider interface, but is more convenient to use. I was thinking we can rely on tests to fail if some method is missing or is incompatible.

rhyslbw:
> I was thinking we can rely on tests to fail if some method is missing or is incompatible.

The only gap here is a superset of methods, but I don't think it's worth getting hung up in this now.

rhyslbw (approved):
Thanks @lgobbi-atix

PR #245, opened 2022-05-11
Flakify

jonringer:
# Context

Attempt to package this with nix.

depends on #256

To build the cardano services cli run `nix build .#@cardano-sdk/cardano-services` from in the repo.

PR #256, opened 2022-05-19
refactor: remove circular devDependencies

mkazlauskas:
# Context

`nix` builds are not working due to circular devDependencies (see #245)

# Proposed Solution

- Remove dependencies:
  - `core` -> `util-dev` 4df6958
  - `util-dev` -> `cip2` 33fc14d
  - `cip30` -> `web-extension` 1c4ab46
- revert 7076fc2 in c911692 (hoist ObservableWallet to cip30 WalletApi mapping back to 'wallet' package)
- rm `cardano-services-client` dependency to itself 2a70267

See [updated dependency tree](https://drive.google.com/file/d/176Yc-cGEUJbodJqZ-Om5tlW9saadHL5_/view?usp=sharing):
\- - - is a `devDependency`
--- is a `dependency`


packages/web-extension/package.json
```diff
@@ -60,6 +59,7 @@
   },
   "dependencies": {
     "@cardano-sdk/core": " 0.2.0",
+    "@cardano-sdk/cip30": " 0.2.0",
```
rhyslbw, line 62: I can see why this is being done to untangle the deps, but this is really the inverse of what we want right? Maybe log a spike to review how we can achieve the goal of a more generic `web-extension` package that `cip30` depends on. Maybe we can consider a pattern of `{PACKAGE_NAME}-dev` for smaller aligned scoping.

rhyslbw (commented):
The dependency tree is pretty helpful, and actually revealed the missing dep between `rabbitmq` and `core` (logged as ADP-1790). I've added another ticket (ADP-1791) to generate and host as part of our documentation. What tool are you using to generate it?

mkazlauskas:
> The dependency tree is pretty helpful, and actually revealed the missing dep between rabbitmq and core (logged as ADP-1790). I've added another ticket (ADP-1791) to generate and host as part of our documentation. What tool are you using to generate it?

Agree, I thought about adding it in this PR, but the tool I used ([monorepo-mapper](https://npm.io/package/monorepo-mapper))  has an external dependency (graphviz). There is [something](https://github.com/marketplace/actions/setup-graphviz) we could try to make it work, but there are some alternatives too: [Nx](https://nx.dev/getting-started/intro) seems like it could simplify our setup and dev experience when it comes to building/rebuilding, it can generate dependency graph and it [seems like](https://dev.to/ipreda/how-to-easily-create-js-libraries-compatible-with-esamdumdcjs-module-systems-1kfd) it can also help us build to multiple module formats without too much setup work (cjs + esm etc.).

packages/core/test/AssetId.ts
```diff
@@ -0,0 +1,5 @@
+import { Cardano } from '../src';
+
+export const TSLA = Cardano.AssetId('659f2917fb63f12b33667463ee575eeac1845bbc736b9c0bbc40ba8254534c41');
```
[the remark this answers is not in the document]
mkazlauskas: This is super small. We do duplicate other strings in tests too (hashes etc.)

packages/wallet/test/Transaction/createTransactionInternals.test.ts
```diff
@@ -1,6 +1,6 @@
 import { Cardano, WalletProvider } from '@cardano-sdk/core';
 import { CreateTxInternalsProps, createTransactionInternals } from '../../src/Transaction';
-import { SelectionConstraints } from '@cardano-sdk/util-dev';
+import { SelectionConstraints } from '../../../cip2/test/util';
```
mkazlauskas, line 3: @rhyslbw this is what I'm most unsure about in this PR. I added `cip2/test/util` under tsconfig {includes} for this to work and it's the only occurrence of such pattern in entire sdk.

packages/util-dev/test/selectionConstraints.test.ts
```diff
@@ -1,19 +0,0 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
-import { mockConstraintsToConstraints } from '../src/selectionConstraints';
-
-describe('selectionConstraints', () => {
-  it('mockConstraintsToConstraints ', async () => {
```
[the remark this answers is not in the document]
mkazlauskas: Yes, it was moved under `cip2/test/util` and we don't have any other tests for such code

packages/wallet/test/Transaction/createTransactionInternals.test.ts
rhyslbw: Can you not just add `../../cip2/test` as a project reference?

packages/web-extension/package.json
mkazlauskas: Depends how you look at it. I think current dependency tree is reasonable. web-extension is currently the highest level package and will have highest level e2e tests. It's only purpose right now is to set up web extension messaging and it exports some cip30-specific setup code for that as well.

Consider this scenario: cip30 is extended for some other environment: (android, desktop wallet etc.), utilizing a different communication interface (like named pipes). You'd still be able to use most of the stuff that currently resides in cip30 package, but you wouldn't want to have a dependency on `web-extension`.

packages/wallet/test/Transaction/createTransactionInternals.test.ts
mkazlauskas: Might be able to do that, but I'm not sure it's any better. `cip2-dev` package as you suggested would be the cleanest solution, but also feels like an overkill for what it would do.

rhyslbw: > Might be able to do that, but I'm not sure it's any better.

Project references establish the build requirements for the current tsconfig scope.

packages/web-extension/package.json
rhyslbw: OK, I'm with you on that.

packages/wallet/test/Transaction/createTransactionInternals.test.ts
mkazlauskas: Converted to project reference in 0e58c2f4064381de144033abbcc7fed341f7f3e4

rhyslbw (dismissed):
Great work @mkazlauskas, ready for squashing

rhyslbw (approved):

PR #257, opened 2022-05-20
Refactor StakePoolBuilder tests

jvieiro:
# Context

Since Db Snapshot can be taken from different databases some autoincremental IDs can vary from db to db due to cases such rollbacks. So, instead of checking for every ID number, we would check only object's data by replacing every registry ID per a generic one

# Proposed Solution

- [x] Strip id field to corresponding StakePoolBuilder queries
- [x] Update Snapshot

### DIFF
--- a/packages/cardano-services/test/StakePoolSeach/DbSyncStakePoolSearchProvider/StakePoolSearchBuilder.test.ts
+++ b/packages/cardano-services/test/StakePoolSeach/DbSyncStakePoolSearchProvider/StakePoolSearchBuilder.test.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
 import { Cardano, StakePoolQueryOptions } from '@cardano-sdk/core';
 import { Pool } from 'pg';
 import { StakePoolSearchBuilder } from '../../../src';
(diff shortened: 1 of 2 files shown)


rhyslbw (changes requested):
Can you instead just strip these fields from the object altogether? Also, retaining the prettiness of the snapshot is important, as they're verified manually.

jvieiro:
> Can you instead just strip these fields from the object altogether? Also, retaining the prettiness of the snapshot is important, as they're verified manually.

I've stripped them in most of tests. Except poolHashes query, where ids are the only result it expects.

packages/cardano-services/test/StakePoolSeach/DbSyncStakePoolSearchProvider/StakePoolSearchBuilder.test.ts
```diff
@@ -27,18 +30,27 @@ describe('StakePoolSearchBuilder', () => {
...
   });
   describe('queryRegistrations', () => {
     it('queryRegistrations', async () => {
       const registrations = await builder.queryRegistrations([1, 2, 3]);
-      expect(registrations).toMatchSnapshot();
+      const striptedIds = registrations.map((r) => {
+        const { hashId, ...rest } = r;
+        return rest;
+      });
+      expect(striptedIds).toMatchSnapshot();
```
rhyslbw, line 47: We're not expecting the stripped IDs to match the snapshot :smile: Always think of the person reading this for the first time :+1: 
```suggestion
      expect(registrations).toMatchSnapshot();
```

rhyslbw (changes requested):

rhyslbw (approved):
Thanks @jvieiro

Newman1515 (commented):
Submit the approve
And 
Submit the review

PR #258, opened 2022-05-24
Chore: Correct and cleanup tsconfig project references

rhyslbw:
# Context
I noticed there were a few issues with the tsconfig project references.

# Proposed Solution
See individual commits.

n.a.


mkazlauskas (approved):
Is eslint linting json files? We should probably have some check to avoid future formatting issues

rhyslbw:
> Is eslint linting json files? We should probably have some check to avoid future formatting issues

It's now in the backlog as ADP-1803

PR #262, opened 2022-05-27
feat: ADP-1772 util package

mirceahasegan:
# Context

1. ADP-1772 Create Util package

Originally this item was a way of decoupling web-extension from wallet. In the meantime, it was decided that it is ok for web-extension to depend on wallet.
Even so, some common code, like pure RxJS extensions or code that does not have any cardano specifics/dependencies, should be in a common package, to minimize coupling.

2. Some refactoring was needed:
  - not all eslintrc.js files are needed
  - linting eslintrc.js files (yarn lint-staged invoked in pre-commit hook) will fail because they are not part of the typescript project configuration

# Proposed Solution

1. Moved some utilities in packages/util
2. Removed unnecessary .eslintrc files. Checked that linting still works (yarn lint + vscode extension)
  - Added tsconfig.eslint.json file referenced by .eslintrc files for proper linting (yarn lint-staged)


packages/util/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "lint": "eslint --ignore-path ../../.eslintignore \"**/*.ts\"",
+    "lint:fix": "eslint --fix --ignore-path ../../.eslintignore \"**/*.ts\"",
+    "test": "jest -c ./jest.config.js",
+    "test:e2e": "shx echo 'test:e2e' command not implemented yet",
+    "coverage": "yarn test --coverage",
+    "prepack": "yarn build",
+    "test:debug": "DEBUG=true yarn test"
+  },
+  "devDependencies": {
+    "@cardano-sdk/util-dev": " 0.2.0"
```
mkazlauskas, line 25: This will be a circular dependency since `util-dev` depends on `core`. Let's get rid of this dependency.

I'm also not a big fan of `core` to depending on `util` as it is now, as it adds an indirect `rxjs` dependency to it :thinking: We should consider `util-rxjs` package. What do you think @rhyslbw @mirceahasegan ?

mkazlauskas (changes requested):
- A better scope for first 3 commits with eslint changes would be `chore`
- Since this repo is public and jira is not, let's not use links to it in PR description. Mentioning issue ID is ok though.
- There's a conflict with `master` - need to rebase
- I've made some changes to serializableObj in #263 so this will need some manual merging

packages/util/NOTICE
```diff
@@ -0,0 +1,5 @@
+Copyright 2021 IOHK
```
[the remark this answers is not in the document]
rhyslbw: Yes

packages/util/package.json
rhyslbw: > I'm also not a big fan of core to depending on util as it is now, as it adds an indirect rxjs dependency to it

I can't see this. Please link

mkazlauskas: Ah core is missing a dependency in package json right now, but it's [using](https://github.com/input-output-hk/cardano-js-sdk/pull/262/files#diff-faa5b269997a2961d92093073919a379bf50ef93fb2e947bbb9d87ee7343f987R2) `util`

rhyslbw: > We should consider `util-rxjs` package

Yes, I agree. `util` will then have no external dependencies.

mirceahasegan: I'm not sure what is the action item here.

rhyslbw: Two packages instead of one. `util` and `util-rxjs`

packages/util/NOTICE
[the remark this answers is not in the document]
mirceahasegan: @mkazlauskas , I don't think manual changes will be needed to PR#263. The file is marked as moved (renamed), so a rebase should detect that and apply your changes in the moved file.

mirceahasegan:
Force pushed the rebase with conflicts resolution and 'refactoring'->'chore' rename

packages/util/NOTICE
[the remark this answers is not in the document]
mirceahasegan: Update year in [f6f3acb](https://github.com/input-output-hk/cardano-js-sdk/pull/262/commits/f6f3acba076afdf8411f083d31a7a061378f8d1d)

packages/wallet/src/tsconfig.json
```diff
@@ -5,14 +5,9 @@
...
-    {
-      "path": "../../cip30/src"
-    },
-    {
-      "path": "../../core/src"
-    }
+    { "path": "../../cip2/src" },
+    { "path": "../../cip30/src" },
+    { "path": "../../core/src" },
+    {"path": "../../util/src"}
```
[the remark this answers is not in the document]
mirceahasegan: Done in [5486867](https://github.com/input-output-hk/cardano-js-sdk/pull/262/commits/5486867d759f39961e69c24c0c99221d489c38c1)

mirceahasegan:
Rebased and resolved conflicts (round 2)

packages/util/package.json
mirceahasegan: I cannot find a way to do this cleanly with fixup commits that I can squash later. I will start rewriting the history of this branch to add the util-rxjs package first, then the util package. I think this is better than keeping the history where TrackerSubject was first part of util, then moved to util-rxjs.
This means that you will not see the new changes separately.

rhyslbw: > This means that you will not see the new changes separately.

No problem, thanks for the heads up

mirceahasegan:
Rebased on master and fixed conflicts.
Added util-rxjs package to break core-util-'util-dev'-core dependency.
Added util and util-rxjs dependencies in all package/*/package.json files that require it and the tsconfig.json files

PR #265, opened 2022-05-30
test(client): fixed 'address already in use' issue

iadmytro:
Fixed test that may cause 'address already in use' error

### DIFF
--- a/packages/cardano-services-client/test/HttpProvider.test.ts
+++ b/packages/cardano-services-client/test/HttpProvider.test.ts
@@ -60,8 +60,8 @@ describe('createHttpServer', () => {
         res.send();
       });
       const response = await provider.noArgsEmptyReturn();
-      expect(response).toBe(undefined);
       await closeServer();
+      expect(response).toBe(undefined);
     });
   });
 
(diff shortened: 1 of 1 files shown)


rhyslbw (changes requested):
Thanks, @iadmytro, well done pinpointing the test bug. I am curious to know why the server isn't being closed in the original version, as while your change ensures the server is closed before making the assertion, there could be a test timeout if the promise never resolves (assuming the server close isn't working as expected).

Normally we would use lifecycle hooks to open and close mock servers since these blocks always fire, even if an error is thrown in the `it` block, plus it keeps the test logic clear. Maybe consider a small refactor to use both it and a `beforeAll` on line 56 to start the server, and an `afterAll` to close it.

Check [this](https://github.com/input-output-hk/cardano-js-sdk/blob/cba8871570af197490755d65ed47afe4eee0779f/packages/cardano-services/test/Http/HttpServer.test.ts#L95-L115) out for a reference

rhyslbw:
> since there is an assertion in closeServer it should be part of the test in my opinion.

Great point, yes I agree.

PR #267, opened 2022-05-31
refactor!: remove transactions and blocks methods from blockfrost wallet provider

lgobbi-atix:
# Context

Hoist `transactionsByAddresses`, `transactionsByHashes`, `blocksByHashes` from `WalletProvider`

# Proposed Solution

Remove `transactionsByAddresses`, `transactionsByHashes`, `blocksByHashes` from `WalletProvider` and create a new `blockfrostChainHistoryProvider`

- `transactionsByAddresses`, `transactionsByHashes`, `blocksByHashes` removed from `WalletProvider`
- Added new  `blockfrostChainHistoryProvider`
- Added new `chainHistoryProvider` dependency to `SingleAddressWallet`


packages/wallet/test/e2e/config.ts
```diff
@@ -68,7 +71,8 @@ const blockfrostApi = [
   env.WALLET_PASSWORD,
   env.TX_SUBMIT_PROVIDER,
   env.ASSET_PROVIDER,
-  env.NETWORK_INFO_PROVIDER
+  env.NETWORK_INFO_PROVIDER,
+  env.CHAIN_HISTORY_PROVIDER
```
lgobbi-atix, line 75: Isn't this missing `env.UTXO_PROVIDER`?

And shouldn't it be `env.WALLET_PROVIDER` instead of `env.WALLET_PASSWORD`?

The same is happening here https://github.com/input-output-hk/cardano-js-sdk/pull/267/files#diff-b5bf86cc0c1c2281ead2068f3647944bb80c64c141fe77e6a2aebff890cb84b4R58

packages/blockfrost/src/blockfrostChainHistoryProvider.ts
```diff
@@ -0,0 +1,305 @@
...
+      index: response.index,
+      txSize: response.size,
+      witness: {
+        redeemers: await fetchRedeemers(response),
+        signatures: new Map() // not available in blockfrost
+      }
+    };
+  };
+
+  const currentWalletProtocolParameters = async (): Promise<ProtocolParametersRequiredByWallet> => {
```
lgobbi-atix, line 229: This one is also in `blockfrostWalletProvider` but we need it here too. This one is very simple but I'm thinking that in the future we could have more functions like this needed by several providers, maybe we could have some sort of shared base provider where you can do something like this instead:

```ts
export const blockfrostChainHistoryProvider = (
  blockfrost: BlockFrostAPI,
  logger = dummyLogger
): ChainHistoryProvider => {
  const { currentWalletProtocolParameters } = baseBlockfrostProvider(blockfrost);

  const transactionsByHashes: ChainHistoryProvider['transactionsByHashes'] = async (hashes) => {
    const protocolParameters = await currentWalletProtocolParameters();
    return Promise.all(hashes.map((hash) => fetchTransaction(hash, protocolParameters)));
  }

  return { transactionsByHashes };
}
```

Does it make sense?

Newman1515 (commented):
Submit the approve

packages/blockfrost/src/blockfrostChainHistoryProvider.ts
rhyslbw: If we moved `currentWalletProtocolParameters` to `NetworkInfoProvider`, then it could be:

``` ts
export const blockfrostChainHistoryProvider = (
  blockfrost: BlockFrostAPI,
  networkInfoProvider: NetworkInfoProvider,
  logger = dummyLogger
): ChainHistoryProvider => {
  const currentWalletProtocolParameters = await networkInfoProvider.currentWalletProtocolParameters();

  const transactionsByHashes: ChainHistoryProvider['transactionsByHashes'] = async (hashes) => {
    return Promise.all(hashes.map((hash) => fetchTransaction(hash, currentWalletProtocolParameters)));
  }

  return { transactionsByHashes };
}
```

@mkazlauskas thoughts?

packages/wallet/test/e2e/config.ts
mkazlauskas: > And shouldn't it be env.WALLET_PROVIDER instead of env.WALLET_PASSWORD

Both are used. WALLET_PASSWORD is used for creating `InMemoryKeyAgent`

packages/blockfrost/src/blockfrostChainHistoryProvider.ts
mkazlauskas: Agree, providers can depend on other providers :+1:

packages/wallet/test/e2e/config.ts
rhyslbw: `WALLET_PASSWORD` is being used to conditionally instantiate `BlockFrostAPI`: 

https://github.com/input-output-hk/cardano-js-sdk/blob/3a202ea407b9260b8598a3e1097d535cb3ca7f34/packages/wallet/test/e2e/config.ts#L69-L84

It'd be nice to isolate that as a named function, to better describe it :sunglasses:

mkazlauskas: Oh this doesn't look right, yeah should be `WALLET_PROVIDER` in here

packages/blockfrost/src/blockfrostChainHistoryProvider.ts
lgobbi-atix: For now I added a `TODO` comment to move `currentWalletProtocolParameters` to `NetworkInfoProvider`

rhyslbw: ADP-1856

mkazlauskas (approved):
Looks good!

PR #270, opened 2022-06-03
feat: add sort stake pools by saturation

lgobbi-atix:
# Context

The current sort order for results returned from the `DbSyncStakePoolProvider.queryStakePools` method is ordered by name. We require the interface to also include `saturation`.

# Proposed Solution

- Add `saturation` sort field with `desc` order as default
- Adapt `withSort` function to be able to sort queries other than `findPoolData`
- Refactor `DbSyncStakePoolProvider.queryStakePools` so it can identify which query to use to order the response

- `DbSyncStakePoolProvider.queryStakePools` refactor
- Added `getQueryBySortType` private method to `DbSyncStakePoolProvider`
- Added sort by saturation to pool metrics query
- Fix a bug where sorting by name was case sensitive, leading to results like: `['CLIO1','July 2021','banderini']`
- Fix a bug where sorting by name `desc` would return pools with no name first instead of last
- Fixed tests related to those bugs

### DIFF
--- a/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePool.ts
+++ b/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePool.ts
@@ -1,9 +1,10 @@
-/* eslint-disable sonarjs/no-nested-template-literals */
+import { CommonPoolInfo, PoolData, PoolMetrics, PoolSortType } from './types';
 import { DbSyncProvider } from '../../DbSyncProvider';
 import { Logger, dummyLogger } from 'ts-log';
 import { Pool } from 'pg';
 import { StakePoolBuilder } from './StakePoolBuilder';
 import { StakePoolProvider, StakePoolQueryOptions, StakePoolSearchResults, StakePoolStats } from '@cardano-sdk/core';
+import { getStakePoolSortType } from './util';
 import { isNotNil } from '@cardano-sdk/util';
 import { toCoreStakePool } from './mappers';
 
@@ -17,6 +18,23 @@ export class DbSyncStakePoolProvider extends DbSyncProvider implements StakePool
     this.#logger = logger;
   }
 
+  private getQueryBySortType(
+    sortType: PoolSortType,
+    queryArgs: { hashesIds: number[]; updatesIds: number[]; totalAdaAmount: string }
+  ) {
+    const { hashesIds, updatesIds, totalAdaAmount } = queryArgs;
+    // Identify which query to use to order and paginate the result
+    // Should be the only one to get the sort options, rest should be ordered by their own defaults
+    switch (sortType) {
+      // Add more cases as more sort types are supported
+      case 'metrics':
+        return (options?: StakePoolQueryOptions) => this.#builder.queryPoolMetrics(hashesIds, totalAdaAmount, options);
+      case 'data':
+      default:
+        return (options?: StakePoolQueryOptions) => this.#builder.queryPoolData(updatesIds, options);
+    }
+  }
+
   public async queryStakePools(options?: StakePoolQueryOptions): Promise<StakePoolSearchResults> {
     const { params, query } =
       options?.filters?._condition === 'or'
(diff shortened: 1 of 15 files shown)


mkazlauskas (commented):
Looks good overall. Query building is getting complicated (and therefore difficult to read), but I don't think there's an easy solution to that.

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePool.ts
```diff
@@ -25,31 +43,60 @@ export class DbSyncStakePoolProvider extends DbSyncProvider implements StakePool
...
+        this.#builder.queryRetirements(orderedResultHashIds),
+        this.#builder.queryPoolRewards(orderedResultHashIds, options?.rewardsHistoryLimit),
+        sortType === 'metrics'
+          ? (orderedResult as PoolMetrics[])
+          : this.#builder.queryPoolMetrics(orderedResultHashIds, totalAdaAmount),
+        this.#builder.queryTotalCount(query, params)
+      ]);
+
+    this.#logger.debug('About to get last epoch');
+    const lastEpoch = await this.#builder.getLastEpoch();
```
IvayloAndonov, line 97: What is the reason we move `getLastEpoch()` out of `Promise.all()`?

IvayloAndonov, line 97: What is the reason we move `getLastEpoch()` out of `Promise.all()`?

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePool.ts
```diff
@@ -25,31 +43,60 @@ export class DbSyncStakePoolProvider extends DbSyncProvider implements StakePool
...
+        });
+        poolDatas.push(...restOfPoolData);
+        orderedResultUpdateIds.push(...restOfPoolData.map(({ updateId }) => updateId));
+        orderedResultHashIds.push(...restOfPoolData.map(({ hashId }) => hashId));
+      }
+    } else {
+      poolDatas = orderedResult as PoolData[];
+    }
+
+    this.#logger.debug('About to stake pool extra information');
```
IvayloAndonov, line 80:
```suggestion
    this.#logger.debug('About to stake pool extra information');
```
```suggestion
    this.#logger.debug('About to query stake pool extra information');
```

IvayloAndonov, line 80:
```suggestion
    this.#logger.debug('About to stake pool extra information');
```
```suggestion
    this.#logger.debug('About to query stake pool extra information');
```

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/mappers.ts
```diff
@@ -47,43 +47,51 @@ interface ToCoreStakePoolInput {
...
+          registration: registrations.map((r) => r.transactionId),
+          retirement: retirements.map((r) => r.transactionId)
+        },
+        vrf: poolData.vrfKeyHash
+      };
+      if (poolData.metadata) toReturn.metadata = poolData.metadata;
+      if (poolData.metadataJson) toReturn.metadataJson = poolData.metadataJson;
+      return toReturn;
+    })
+    .filter((result) => result !== undefined) as Cardano.StakePool[],
```
IvayloAndonov, line 94: What is the case where `result` is _undefined_ here? At first look it seems we iterate over the collection `poolHashIds` (passed as `orderedResultHashIds`  which are previously filled with additional [restOfPoolData](https://github.com/input-output-hk/cardano-js-sdk/pull/270/files#diff-5ad8b26faabe50605214aec922ed96428a741e5e07da541676062cec7a693a17R74) ) which is bigger than `poolDatas`

IvayloAndonov, line 94: What is the case where `result` is _undefined_ here? At first look it seems we iterate over the collection `poolHashIds` (passed as `orderedResultHashIds`  which are previously filled with additional [restOfPoolData](https://github.com/input-output-hk/cardano-js-sdk/pull/270/files#diff-5ad8b26faabe50605214aec922ed96428a741e5e07da541676062cec7a693a17R74) ) which is bigger than `poolDatas`

IvayloAndonov:
More of a question, considering the intense increase in complexity and theoretically more sorting fields requested in the future, do we think that maintaining modular DB queries + additional [logic](https://github.com/input-output-hk/cardano-js-sdk/pull/270/files#diff-5ad8b26faabe50605214aec922ed96428a741e5e07da541676062cec7a693a17R59-R74) populating minimum requested data for a simple operation like sort is still the better option than having one composed DB query (with joined required tables) providing [stake pool data](https://github.com/input-output-hk/cardano-js-sdk/blob/52d71a70700b05902cca6205fe01a63f811ba5af/packages/core/src/Provider/StakePoolProvider/types/StakePoolProvider.ts#L53) at once  (sorted, filtered and paginated)? 🤔

rhyslbw:
@IvayloAndonov Please create a SPIKE for further exploration of this technical debt reducer in a later sprint.

rhyslbw (commented):
I expect `fix: make stake pool order by name case insensitive and move nulls to last` to be a separate commit before either a single commit with the refactor + feature, or two extra commits.

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/util.ts
```diff
@@ -0,0 +1,8 @@
+import { PoolSortType } from './types';
+import { isPoolDataSortField, isPoolMetricsSortField } from '@cardano-sdk/core';
+
+export const getStakePoolSortType = (field: string): PoolSortType => {
+  if (isPoolDataSortField(field)) return 'data';
+  if (isPoolMetricsSortField(field)) return 'metrics';
+  return 'data';
```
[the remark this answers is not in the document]
lgobbi-atix: Done

lgobbi-atix:
> I expect `fix: make stake pool order by name case insensitive and move nulls to last` to be a separate commit before either a single commit with the refactor + feature, or two extra commits.

I left the first commit with the refactor + feature as it's kinda hard to separate them and added a new commit after with the fix

rhyslbw (approved):
Great work @lgobbi-atix.

mkazlauskas (approved):
Looks good, nice work! :tada:

IvayloAndonov (dismissed):

ssong-iohk (changes requested):
Can you add the test case (i.e. happy path - the request with filters/sort query parameters)  to [stakePoolHttpProvider.test.ts](https://github.com/input-output-hk/cardano-js-sdk/blob/feature/ADP-1793-Add-stake-pool-order-by-saturation/packages/cardano-services-client/test/StakePoolProvider/stakePoolHttpProvider.test.ts)

lgobbi-atix:
> Can you add the test case (i.e. happy path - the request with filters/sort query parameters) to [stakePoolHttpProvider.test.ts](https://github.com/input-output-hk/cardano-js-sdk/blob/feature/ADP-1793-Add-stake-pool-order-by-saturation/packages/cardano-services-client/test/StakePoolProvider/stakePoolHttpProvider.test.ts)

The provider interface didn't change. The response for these tests are mocked, it doesn't make sense to add the filters and everything there. Those cases should already be covered with https://github.com/input-output-hk/cardano-js-sdk/pull/270/files#diff-7bf95288ddba72cc435fbe56613884016d332bdba413afd23c5e1e7bcc69ce42R684

rhyslbw:
> it doesn't make sense to add the filters and everything there

Right, the `HTTPService` implementations are just plumbing, and we could even look at refactoring the base class to be usable without needing to extend.

ssong-iohk (commented):
LGTM for me. 
I will run regression test against `testnet` then approve unless any issues are found. 
And @lgobbi-atix I think you can resolve conflict.

PR #273, opened 2022-06-07
ADP-1821 remove sync properties from ObservableWallet interface

mirceahasegan:
# Context
Since ObservableWallet is intended to be used asynchronously (wallet running in background process), any sync properties will not work.

# Proposed Solution
Convert all BehaviorObervable into Observable (including in all nested objects). Update tests (and maybe code too) by replacing observable$.value with firstValueFrom(observable$)


mirceahasegan:
I will update the commit messages shortly, to reflect the breaking changes

mkazlauskas (changes requested):
Looking good, just a few notes for commits (please wait for other review(s) before changing the commits):

- 71bb08478179e7e6697228aab699395306e78b57 should have no scope as it affects multiple packages
- a more suitable commit type for 2af87ee97821ae8e2f2883af3a1c6c9f94fa56d8 would be `test`. However since other commits (except for the 1st one) are broken without this, I think we should just squash them to 1.
- ResolveInputAddress breaking change is also an important one, let's point this out in the commit message when squashing

Also, please confirm that `wallet` e2e tests are still passing

mirceahasegan:
> Looking good, just a few notes for commits (please wait for other review(s) before changing the commits):
> 
> * [71bb084](https://github.com/input-output-hk/cardano-js-sdk/commit/71bb08478179e7e6697228aab699395306e78b57) should have no scope as it affects multiple packages
> * a more suitable commit type for [2af87ee](https://github.com/input-output-hk/cardano-js-sdk/commit/2af87ee97821ae8e2f2883af3a1c6c9f94fa56d8) would be `test`. However since other commits (except for the 1st one) are broken without this, I think we should just squash them to 1.
> * ResolveInputAddress breaking change is also an important one, let's point this out in the commit message when squashing
> 
> Also, please confirm that `wallet` e2e tests are still passing

i confirm wallet e2e tests are still passing.

rhyslbw (approved):
> I think we should just squash them to 1

Agree. It's been good for the review, but would be best squashed into one:
image

rhyslbw (dismissed):

rhyslbw (approved):

mkazlauskas (approved):
Nice work! :rocket:

PR #274, opened 2022-06-08
fix: Core TokenMap to CSL multiasset insert de-dupe  

rhyslbw:
# Context
ADP-1763

# Proposed Solution
remove duplicated insert when assets from the same policy are present in the token map

# Important Info
Conversion back to core types test is failing. Currently investigating

### DIFF
--- a/packages/core/src/CSL/coreToCsl/coreToCsl.ts
+++ b/packages/core/src/CSL/coreToCsl/coreToCsl.ts
@@ -44,14 +44,28 @@ import { SerializationError } from '../../errors';
 import { SerializationFailure } from '../..';
 import { parseCslAddress } from '../parseCslAddress';
 
-export const tokenMap = (assets: Cardano.TokenMap) => {
+export const tokenMap = (map: Cardano.TokenMap) => {
   const multiasset = MultiAsset.new();
-  for (const assetId of assets.keys()) {
-    const { scriptHash, assetName } = Asset.util.parseAssetId(assetId);
-    const assetsObj = Assets.new();
-    const amount = BigNum.from_str(assets.get(assetId)!.toString());
-    assetsObj.insert(assetName, amount);
-    multiasset.insert(scriptHash, assetsObj);
+  const policyMap: Map<string, { assetsMap: Map<AssetName, BigNum>; scriptHash: ScriptHash }> = new Map();
+
+  for (const assetId of map.keys()) {
+    const { assetName, scriptHash } = Asset.util.parseAssetId(assetId);
+    const policyId = Asset.util.policyIdFromAssetId(assetId).toString();
+    const amount = BigNum.from_str(map.get(assetId)!.toString());
+    if (!policyMap.has(policyId)) {
+      policyMap.set(policyId, { assetsMap: new Map([[assetName, amount]]), scriptHash });
+    } else {
+      const { assetsMap } = policyMap.get(policyId)!;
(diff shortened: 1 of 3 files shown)


packages/core/test/CSL/testData.ts
```diff
@@ -30,6 +30,7 @@ export const poolParameters: Cardano.PoolParameters = {
   vrf
 };
 
+// REVIEW: Also add these additional assets here?
```
mkazlauskas, line 33: Good idea :+1:

packages/core/src/CSL/coreToCsl/coreToCsl.ts
```diff
@@ -44,14 +44,29 @@ import { SerializationError } from '../../errors';
...
-    const amount = BigNum.from_str(assets.get(assetId)!.toString());
-    assetsObj.insert(assetName, amount);
-    multiasset.insert(scriptHash, assetsObj);
+  const policyObj: {
+    [policyId: string]: { assetsMap: Map<AssetName, BigNum>; scriptHash: ScriptHash };
+  } = {};
+
+  for (const assetId of map.keys()) {
+    const { assetName, scriptHash } = Asset.util.parseAssetId(assetId);
+    // REVIEW: Could also return the policyId from Asset.util.parseAssetId
```
mkazlauskas, line 55: It already returns it as `scriptHash` object. Perhaps it should be exported as `coreToCsl.asset` though :thinking:

packages/core/test/CSL/coreToCsl.test.ts
```diff
@@ -40,15 +40,17 @@ describe('coreToCsl', () => {
...
-    for (const [assetId, expectedAssetQuantity] of txOut.value.assets!.entries()) {
-      const { scriptHash, assetName } = Asset.util.parseAssetId(assetId);
-      const assetQuantity = BigInt(multiasset.get(scriptHash)!.get(assetName)!.to_str());
-      expect(assetQuantity).toBe(expectedAssetQuantity);
-    }
+  describe('tokenMap', () => {
+    it('inserts a single multiasset per asset policy', () => {
+      const multiasset = coreToCsl.tokenMap(txOut.value.assets!);
+      expect(multiasset).toBeInstanceOf(CSL.MultiAsset);
+      expect(multiasset.len()).toBe(3);
```
mkazlauskas, line 47: That would be a little clearer as keys of MultiAsset is ScriptHashes which is what we want to test here. Not sure if there's any difference functionally, but I think more explicit is better:

```suggestion
      expect(multiasset.keys().len()).toBe(3);
```

mkazlauskas (changes requested):

packages/core/test/CSL/coreToCsl.test.ts
rhyslbw: Done in 37e9d008328ca1f17efc8e93bb8eade91317762f

packages/core/test/CSL/testData.ts
mkazlauskas: Done in [9cb225b](https://github.com/input-output-hk/cardano-js-sdk/pull/274/commits/9cb225be0fea0ec2da65f393f7120be986f0929d)

packages/core/src/CSL/coreToCsl/coreToCsl.ts
mkazlauskas: Captured in ADP-1876

iadmytro (approved):

PR #275, opened 2022-06-08
chore: esm builds

mkazlauskas:
# Context

The only way to consume SDK packages right now is through bundler, because they are currently built to CommonJS but has `lodash-es` dependency which doesn't work in non-`type="module"` nodejs apps. We need to make it more accessible.

# Proposed Solution

- Set up build to produce both CJS and ESM modules
  - There's now `dist/cjs` and `dist/esm`
    - exception is `cardano-services` package, which is only built to `cjs` because I couldn't find an easy way to make `__dirname` compatible with both formats. To be addressed in ADP-1887
  - Set up `package.json` to support consuming the same npm package in both module formats
- Replace `lodash-es` with `lodash`
  - Update lodash imports to individual functions `{stuff} from 'lodash-es'` -> `stuff from 'lodash/stuff'` and update eslint config to enforce that.

## Testing

Verified that SDK package:

- works in node.js app that uses CJS
- works in node.js app that uses ESM (`type="module"`)
  - requires `--experimental-specifier-resolution=node` because `tsc` does not append extensions to relative imports/exports (keeps `import stuff from './stuff'`, while ES modules are expected to do `import stuff from './stuff.js'`). We could address this in the future by doing some post-processing after the build or replacing `tsc` with `babel` for ESM builds. 
- webpack finds ESM files and tree shaking works

## Other changes

- [test(web-extension): add required protocolMagic to LedgerKeyAgent init](https://github.com/input-output-hk/cardano-js-sdk/pull/275/commits/0d447378364690b321b10937c251127396f3d123)
- [chore: bump typescript-eslint version](https://github.com/input-output-hk/cardano-js-sdk/pull/275/commits/65648e0fd1ca787ffb861af778ec570f6ce147bd)


packages/blockfrost/src/tsconfig.json
```diff
@@ -1,11 +1,15 @@
...
   "extends": "../../../tsconfig.json",
   "compilerOptions": {
-    "outDir": "../dist/",
+    "outDir": "../dist/cjs",
     "rootDir": "."
   },
   "references": [
-    { "path": "../../util/src" },
-    { "path": "../../core/src" }
+    {
```
mirceahasegan, line 8: Any ideea why this is being formatted like this?

mkazlauskas: My vscode autoformats to this. I don't seem to have any extra formatting plugins so it must be from eslint :thinking:

mirceahasegan (dismissed):

packages/blockfrost/src/tsconfig.json
mkazlauskas: Agree, formatting should differ based on who makes the changes, but we have ADP-1803 to work on this

packages/cardano-services-client/package.json
```diff
@@ -5,18 +5,29 @@
...
     "Martynas Kazlauskas <martynas.kazlauskas@iohk.io> (https://iohk.io)"
   ],
   "license": "MPL-2.0",
   "scripts": {
-    "build": "tsc --build ./src",
-    "generate": "graphql-codegen",
+    "build:esm": "tsc -p src/tsconfig.json --outDir ./dist/esm --module es2020",
+    "build:cjs": "tsc --build src",
+    "build": "run-s build:cjs build:esm module-fixup",
+    "module-fixup": "shx cp ../../build/cjs-package.json ./dist/cjs/package.json && cp ../../build/esm-package.json ./dist/esm/package.json",
```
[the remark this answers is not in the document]
mkazlauskas: It's needed for node.js es module resolution. Main `package.json` files of packages do not specify `type` because we want them to work in both commonjs and esm setups. A separate `package.json` under `dist/esm` is required to treat files under it as es modules. The `cjs-package.json` one might actually be redundant because commonjs is the default `type`, but I put it in there for consistency and future-proofing.

packages/cardano-services/package.json
```diff
@@ -33,7 +33,7 @@
     "cleanup": "run-s cleanup:dist cleanup:nm",
     "lint": "eslint --ignore-path ../../.eslintignore \"**/*.ts\"",
     "lint:fix": "eslint --fix --ignore-path ../../.eslintignore \"**/*.ts\"",
-    "test": "jest -c ./jest.config.js",
+    "test": "shx echo typescript --noEmit command not implemented yet",
```
rhyslbw, line 36: :raised_eyebrow:

rhyslbw (changes requested):
Nice work @mkazlauskas, great to see the `patchRequire` workaround and the eslint warning gone :sunflower: 

Just have the one issue I've highlighted inline

mkazlauskas:
Finally got CI to pass. Updated PR description:
> exception is cardano-services package, which is only built to cjs because I couldn't find an easy way to make __dirname compatible with both formats. To be addressed in ADP-1887

Had to rebase on master. See last 4 commits for new changes.

mirceahasegan (approved):

mirceahasegan:
Nice work 👍 . Good to go imho

PR #276, opened 2022-06-08
feat: Cache heavy NetworkInfo db-sync queries

IvayloAndonov:
# Context

We need to cache the following values in the context of [DbSyncNetworkInfoProvider](https://github.com/input-output-hk/cardano-js-sdk/blob/b70ae5222a54634ba50bba0c9cc73b03832afe00/packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts#L28-L31) to limit impact on the DB:

`this.#builder.queryTotalSupply(maxLovelaceSupply)` -  Constant for the epoch
`this.#builder.queryCirculatingSupply()` - Changes every block
`this.#builder.queryActiveStake()` - Constant for the epoch
`this.#builder.queryLiveStake()` -  Changes every block

The two constant values can be cached using knowledge of the current epoch, triggered at the point of epoch rollover, and then set a configurable value in minutes with a sensible default to cover the other two.

# Proposed Solution
**Capturing the epoch rollover ( cache invalidation for epoch values )**
- The implemented approach here is to just start up an interval on boot with a light query, and it run indefinitely.
- The optimized approach is to start a Chain-Sync Ogmios client with an epoch check in the `rollForward` handler, start polling the DB once we observe a rollover.

In order to know when there is epoch rollover, we query the latest one from `epoch` table and store it in the cache as current. The validation query is set to run every 10 seconds change, check if there is a new epoch with data created greater than the cached one, if so - override it and invalidate the cache.

**Validation requirements**
- Default caching duration of 120 minutes for queryCirculatingSupply  and queryLiveStake
- Duration must be an integer from 1 to 2880 (2 days). This does not impact queryTotalSupply or queryActiveStake, as the cache invalidation is defined by the epoch rollover.
- Polling interval is 10 seconds

- add `DB_QUERIES_CACHE_TTL` env in `run` with custom range validation
- add `DB_POLL_INTERVAL` env in `run` 
- add `--db-queries-cache-ttl` option in `cli`with custom range validation
- add `--db-poll-interval` option in `cli`
- add [CacheService](https://github.com/input-output-hk/cardano-js-sdk/pull/276/files#diff-9e8c39a0834f189fcd5eb822e9a84eac51cbf6ab7b401e8fdfb160625a78044eR7) to cache heavy db queries (used `node-cache` package )
- add db sync polling on [NetworkInfoProvider](https://github.com/input-output-hk/cardano-js-sdk/blob/12ce8b4c18d1df087fb5d1a7295ebe569aa6d188/packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts#L12) provider start (10 secs polling interval)
- create test utils [wrapWithTransaction](https://github.com/input-output-hk/cardano-js-sdk/blob/4f2694e56028ba76cf96bd4e93e9c7883d5850fd/packages/cardano-services/test/util.ts#L58) and [ingestDbData](https://github.com/input-output-hk/cardano-js-sdk/blob/4f2694e56028ba76cf96bd4e93e9c7883d5850fd/packages/cardano-services/test/util.ts#L40) helping where needs to be reproduced specific test cases but finally db snapshot and state shouldn't be mutated for other tests
- extend `Provider` and `HttpService` in order to keep services builder pure [(7ceca49):](https://github.com/input-output-hk/cardano-js-sdk/pull/276/commits/7ceca494f698905e7e7492ad09bae5fbb250bea6)
-- extend `Provider` interface with `start` optional method
-- extend `DbSyncProvider` and `HttpSrvice` with `start` method
-- invoke services `start()` within `HttpServer.startImpl()`
-- remove db `healthCheck` during services creation, keep them pure
-- align tests and enhance `loadHttpServer.test.ts`

### DIFF
--- a/packages/cardano-services/.env.test
+++ b/packages/cardano-services/.env.test
@@ -1,2 +1,3 @@
 DB_CONNECTION_STRING=postgresql://postgres:doNoUseThisSecret!@127.0.0.1:5433/testnet
 CARDANO_NODE_CONFIG_PATH=./config/network/testnet/cardano-node/config.json
+DB_QUERIES_CACHE_TTL=120
--- a/packages/cardano-services/src/InMemoryCache/InMemoryCache.ts
+++ b/packages/cardano-services/src/InMemoryCache/InMemoryCache.ts
@@ -1,0 +1,87 @@
+import { CACHE_TTL_DEFAULT } from './defaults';
+import NodeCache from 'node-cache';
+
+export type Key = string | number;
+export type DbSyncQuery<T> = () => Promise<T>;
+
+export class InMemoryCache {
+  #cache: NodeCache;
+  #ttlDefault: number;
+
+  constructor(cacheTtlInMins: number = CACHE_TTL_DEFAULT, cache: NodeCache = new NodeCache()) {
+    this.#ttlDefault = cacheTtlInMins * 60;
+    this.#cache = cache;
+  }
+
+  /**
+   * Get a cached key value, if not found execute db query and cache the result with that key
+   *
+   * @param key cache key
+   * @param dbSyncQuery cache key
+   * @param ttl cache duration in seconds
+   * @returns The value stored with the key
+   */
+  public async get<T>(key: Key, dbSyncQuery: DbSyncQuery<T>, ttl = this.#ttlDefault): Promise<T> {
+    const cachedValue: T | undefined = this.#cache.get(key);
+    if (cachedValue) {
+      return cachedValue;
+    }
+
+    const result = await dbSyncQuery();
+    this.#cache.set(key, result, ttl);
+    return result;
+  }
+
+  /**
+   * Get a cached key
+   *
+   * @param key cache key
+   * @returns The value stored in the key
+   */
+  public getVal<T>(key: Key) {
+    return this.#cache.get<T>(key);
+  }
+
+  /**
+   * Set a cached key
+   *
+   * @param key Cache key
+   * @param value A value to cache
+   * @param ttl The time to live in seconds.
+   */
+  public set<T>(key: Key, value: T, ttl: number) {
+    return this.#cache.set<T>(key, value, ttl);
+  }
+
+  /**
+   * Invalidate cached values
+   *
+   * @param keys cache key to delete or a array of cache keys
+   */
+  public invalidate(keys: Key | Key[]) {
+    this.#cache.del(keys);
+  }
+
+  /**
+   * List all keys within this cache
+   *
+   * @returns An array of all keys
+   */
+  public keys() {
+    return this.#cache.keys();
+  }
+
+  /**
+   * Clear the interval timeout which is set on check period option. Default: 600
+   */
+  public shutdown() {
+    this.#cache.close();
+  }
+
+  /**
+   * Clear the whole data and reset the stats
+   */
+  public clear() {
+    this.#cache.flushAll();
+  }
+}
--- a/packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
+++ b/packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
@@ -1,21 +1,42 @@
 import { DbSyncProvider } from '../../DbSyncProvider';
 import { GenesisData } from './types';
+import { InMemoryCache, UNLIMITED_CACHE_TTL } from '../../InMemoryCache';
 import { Logger, dummyLogger } from 'ts-log';
 import { NetworkInfo, NetworkInfoProvider, timeSettingsConfig } from '@cardano-sdk/core';
 import { NetworkInfoBuilder } from './NetworkInfoBuilder';
+import { NetworkInfoCacheKey } from '.';
 import { Pool } from 'pg';
+import { Shutdown } from '@cardano-sdk/util';
 import { loadGenesisData, toNetworkInfo } from './mappers';
+import { pollDbSync } from './utils';
 
+export interface NetworkInfoProviderProps {
+  cardanoNodeConfigPath: string;
+  dbPollInterval: number;
+}
+export interface NetworkInfoProviderDependencies {
+  db: Pool;
+  cache: InMemoryCache;
+  logger?: Logger;
+}
 export class DbSyncNetworkInfoProvider extends DbSyncProvider implements NetworkInfoProvider {
   #logger: Logger;
+  #cache: InMemoryCache;
   #builder: NetworkInfoBuilder;
   #genesisDataReady: Promise<GenesisData>;
+  #dbPollInterval: number;
+  #pollService: Shutdown | null;
 
-  constructor(cardanoNodeConfigPath: string, db: Pool, logger = dummyLogger) {
+  constructor(
+    { cardanoNodeConfigPath, dbPollInterval }: NetworkInfoProviderProps,
+    { db, cache, logger = dummyLogger }: NetworkInfoProviderDependencies
+  ) {
     super(db);
     this.#logger = logger;
+    this.#cache = cache;
     this.#builder = new NetworkInfoBuilder(db, logger);
     this.#genesisDataReady = loadGenesisData(cardanoNodeConfigPath);
+    this.#dbPollInterval = dbPollInterval;
   }
 
   public async networkInfo(): Promise<NetworkInfo> {
@@ -24,11 +45,15 @@ export class DbSyncNetworkInfoProvider extends DbSyncProvider implements Network
 
     this.#logger.debug('About to query network info data');
 
-    const [totalSupply, circulatingSupply, activeStake, liveStake] = await Promise.all([
-      this.#builder.queryTotalSupply(maxLovelaceSupply),
-      this.#builder.queryCirculatingSupply(),
-      this.#builder.queryActiveStake(),
-      this.#builder.queryLiveStake()
+    const [liveStake, circulatingSupply, activeStake, totalSupply] = await Promise.all([
+      this.#cache.get(NetworkInfoCacheKey.LIVE_STAKE, () => this.#builder.queryLiveStake()),
+      this.#cache.get(NetworkInfoCacheKey.CIRCULATING_SUPPLY, () => this.#builder.queryCirculatingSupply()),
+      this.#cache.get(NetworkInfoCacheKey.ACTIVE_STAKE, () => this.#builder.queryActiveStake(), UNLIMITED_CACHE_TTL),
+      this.#cache.get(
+        NetworkInfoCacheKey.TOTAL_SUPPLY,
+        () => this.#builder.queryTotalSupply(maxLovelaceSupply),
+        UNLIMITED_CACHE_TTL
+      )
     ]);
 
     return toNetworkInfo({
@@ -41,5 +66,16 @@ export class DbSyncNetworkInfoProvider extends DbSyncProvider implements Network
       timeSettings,
       totalSupply
     });
+  }
+
+  async start(): Promise<void> {
+    if (!this.#pollService)
+      this.#pollService = pollDbSync(this.#cache, () => this.#builder.queryLatestEpoch(), this.#dbPollInterval);
+  }
+
+  async close(): Promise<void> {
+    this.#pollService?.shutdown();
+    this.#pollService = null;
+    this.#cache.shutdown();
   }
 }
--- a/packages/cardano-services/src/NetworkInfo/NetworkInfoHttpService.ts
+++ b/packages/cardano-services/src/NetworkInfo/NetworkInfoHttpService.ts
@@ -29,6 +29,7 @@ export class NetworkInfoHttpService extends HttpService {
 
   static create({ logger = dummyLogger, networkInfoProvider }: NetworkInfoServiceDependencies) {
     const router = express.Router();
+
     const apiSpec = path.join(__dirname, 'openApi.json');
     router.use(
       OpenApiValidator.middleware({
@@ -46,5 +47,13 @@ export class NetworkInfoHttpService extends HttpService {
       )
     );
     return new NetworkInfoHttpService({ logger, networkInfoProvider }, router);
+  }
+
+  async start(): Promise<void> {
+    await this.#networkInfoProvider.start();
+  }
+
+  async close(): Promise<void> {
+    await this.#networkInfoProvider.close();
   }
 }
--- a/packages/cardano-services/src/Program/ProgramOptionDescriptions.ts
+++ b/packages/cardano-services/src/Program/ProgramOptionDescriptions.ts
@@ -7,7 +7,9 @@ enum HttpServerOptionDescriptions {
   OgmiosUrl = 'Ogmios URL',
   RabbitMQUrl = 'RabbitMQ URL',
   UseQueue = 'Enables RabbitMQ',
-  CardanoNodeConfigPath = 'Cardano node config path'
+  CardanoNodeConfigPath = 'Cardano node config path',
+  DbQueriesCacheTtl = 'Db queries cache TTL in minutes between 1 and 2880',
+  DbPollInterval = 'Db poll interval'
 }
 
 export type ProgramOptionDescriptions = CommonOptionDescriptions | HttpServerOptionDescriptions;
--- a/packages/cardano-services/src/Program/loadHttpServer.ts
+++ b/packages/cardano-services/src/Program/loadHttpServer.ts
@@ -7,6 +7,7 @@ import { DbSyncRewardsProvider, RewardsHttpService } from '../Rewards';
 import { DbSyncStakePoolProvider, StakePoolHttpService } from '../StakePool';
 import { DbSyncUtxoProvider, UtxoHttpService } from '../Utxo';
 import { HttpServer, HttpServerConfig, HttpService } from '../Http';
+import { InMemoryCache } from '../InMemoryCache';
 import { MissingProgramOption, UnknownServiceName } from './errors';
 import { ProgramOptionDescriptions } from './ProgramOptionDescriptions';
 import { RabbitMqTxSubmitProvider } from '@cardano-sdk/rabbitmq';
@@ -18,6 +19,8 @@ import pg from 'pg';
 
 export interface HttpServerOptions extends CommonProgramOptions {
   dbConnectionString?: string;
+  dbQueriesCacheTtl: number;
+  dbPollInterval: number;
   cardanoNodeConfigPath?: string;
   metricsEnabled?: boolean;
   useQueue?: boolean;
@@ -36,7 +39,7 @@ export interface ProgramArgs {
   options?: HttpServerOptions;
 }
 
-const serviceMapFactory = (args: ProgramArgs, logger: Logger, db?: pg.Pool) => ({
+const serviceMapFactory = (args: ProgramArgs, logger: Logger, cache: InMemoryCache, db?: pg.Pool) => ({
   [ServiceNames.StakePool]: async () => {
     if (!db) throw new MissingProgramOption(ServiceNames.StakePool, ProgramOptionDescriptions.DbConnection);
     return await StakePoolHttpService.create({
@@ -67,7 +70,10 @@ const serviceMapFactory = (args: ProgramArgs, logger: Logger, db?: pg.Pool) => (
 
     return await NetworkInfoHttpService.create({
       logger,
-      networkInfoProvider: new DbSyncNetworkInfoProvider(args.options?.cardanoNodeConfigPath, db, logger)
+      networkInfoProvider: new DbSyncNetworkInfoProvider(
+        { cardanoNodeConfigPath: args.options?.cardanoNodeConfigPath, dbPollInterval: args.options?.dbPollInterval },
+        { cache, db, logger }
+      )
     });
   },
   [ServiceNames.TxSubmit]: async () =>
@@ -98,7 +104,9 @@ export const loadHttpServer = async (args: ProgramArgs): Promise<HttpServer> => 
     ? new pg.Pool({ connectionString: args.options.dbConnectionString })
     : undefined;
 
-  const serviceMap = serviceMapFactory(args, logger, db);
+  const cache = new InMemoryCache(args.options?.dbQueriesCacheTtl);
+
+  const serviceMap = serviceMapFactory(args, logger, cache, db);
 
   for (const serviceName of args.serviceNames) {
     if (serviceMap[serviceName]) {
--- a/packages/cardano-services/src/cli.ts
+++ b/packages/cardano-services/src/cli.ts
@@ -145,5 +145,6 @@ if (process.argv.slice(2).length === 0) {
   program.parseAsync(process.argv).catch((error) => {
     // eslint-disable-next-line no-console
     console.error(error);
+    process.exit(1);
   });
 }
--- a/packages/cardano-services/src/cli.ts
+++ b/packages/cardano-services/src/cli.ts
@@ -8,8 +8,10 @@ import {
   ServiceNames,
   loadHttpServer
 } from './Program';
+import { CACHE_TTL_DEFAULT } from './InMemoryCache';
 import { Command } from 'commander';
 import { CommonOptionDescriptions } from './ProgramsCommon';
+import { DB_POLL_INTERVAL_DEFAULT } from './NetworkInfo';
 import { InvalidLoggerLevel } from './errors';
 import {
   PARALLEL_MODE_DEFAULT,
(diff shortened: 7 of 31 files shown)


packages/cardano-services/src/CacheService/keys.ts
```diff
@@ -0,0 +1,6 @@
+// Network Info namespace
```
mkazlauskas, line 1: Maybe putting this under `/NetworkInfo` would be better. Also you can 'namespace' with an object, like `export const NetworkInfoCacheKey = {TOTAL_SUPPLY, ...}`

mkazlauskas (changes requested):

packages/cardano-services/test/Program/loadHttpServer.test.ts
```diff
@@ -45,9 +47,10 @@ describe('loadHttpServer', () => {
         options: {
           cardanoNodeConfigPath,
           dbConnectionString,
+          dbQueriesCacheDuration,
           ogmiosUrl: new URL(ogmiosConnection.address.webSocket)
         },
-        serviceNames: [ServiceNames.StakePool, ServiceNames.TxSubmit, ServiceNames.NetworkInfo]
+        serviceNames: [ServiceNames.StakePool, ServiceNames.TxSubmit]
```
IvayloAndonov, line 53: This one is the problematic test where `getaddrinfo ENOTFOUND database.server.com` is thrown, caused by service [healthCheck](https://github.com/input-output-hk/cardano-js-sdk/blob/12ce8b4c18d1df087fb5d1a7295ebe569aa6d188/packages/cardano-services/src/NetworkInfo/NetworkInfoHttpService.ts#L34-L36). Same behavior if we add [ChainHistoryHttpService](https://github.com/input-output-hk/cardano-js-sdk/blob/64aa7aeff061aa2cf9bc6196347f6cf5b9c7f6be/packages/cardano-services/src/ChainHistory/ChainHistoryHttpService.ts#L16) in `serviceNames` - the containerised db is not accesssible during the service creation when there is `db.query` health check prior inisialisation. Looks like a raise condition, this is the only failing test if all available services are added here.

mkazlauskas: cc @iadmytro

packages/cardano-services/src/Program/loadHttpServer.ts
```diff
@@ -65,7 +67,13 @@ const serviceMapFactory = (args: ProgramArgs, logger: Logger, db?: Pool) => ({
 
     return await NetworkInfoHttpService.create({
       logger,
-      networkInfoProvider: new DbSyncNetworkInfoProvider(args.options?.cardanoNodeConfigPath, db, logger)
+      networkInfoProvider: new DbSyncNetworkInfoProvider(
+        args.options?.cardanoNodeConfigPath,
+        POLL_INTERVAL,
```
IvayloAndonov, line 72: This one could be passed as env var as well, any thoughts?

packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
```diff
@@ -6,16 +6,26 @@ import { NetworkInfoBuilder } from './NetworkInfoBuilder';
...
+
 export class DbSyncNetworkInfoProvider extends DbSyncProvider implements NetworkInfoProvider {
   #logger: Logger;
   #builder: NetworkInfoBuilder;
   #genesisDataReady: Promise<GenesisData>;
 
-  constructor(cardanoNodeConfigPath: string, db: Pool, logger = dummyLogger) {
+  constructor(
+    { cardanoNodeConfigPath, dbQueriesCacheDuration }: NetworkInfoProviderParams,
+    db: Pool,
```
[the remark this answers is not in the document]
IvayloAndonov: Addressed [f0d1480](https://github.com/input-output-hk/cardano-js-sdk/pull/276/commits/f0d148025fb2491bad63bde71a338d2be6f5869f)

packages/cardano-services/src/CacheService/keys.ts
IvayloAndonov: Addressed [e470d15](https://github.com/input-output-hk/cardano-js-sdk/pull/276/commits/e470d156e91e9a5cbf074bf9a725d0b334256850)

packages/cardano-services/src/CacheService/CacheService.ts
```diff
@@ -0,0 +1,87 @@
...
+   * Clear the interval timeout which is set on check period option. Default: 600
+   */
+  public close() {
+    this.#cache.close();
+  }
+
+  /**
+   * Flush the whole data and reset the stats
+   */
+  public flush() {
```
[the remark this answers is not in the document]
IvayloAndonov: [e470d15](https://github.com/input-output-hk/cardano-js-sdk/pull/276/commits/e470d156e91e9a5cbf074bf9a725d0b334256850)

packages/cardano-services/src/Program/loadHttpServer.ts
```diff
@@ -65,7 +67,10 @@ const serviceMapFactory = (args: ProgramArgs, logger: Logger, db?: Pool) => ({
 
     return await NetworkInfoHttpService.create({
       logger,
-      networkInfoProvider: new DbSyncNetworkInfoProvider(args.options?.cardanoNodeConfigPath, db, logger)
+      networkInfoProvider: new DbSyncNetworkInfoProvider(
+        { cardanoNodeConfigPath: args.options?.cardanoNodeConfigPath, pollInterval: POLL_INTERVAL },
```
IvayloAndonov, line 71: `POLL_INTERVAL` could be passed as an env var as well, any thoughts?

mkazlauskas: Agree, maybe `CACHE_POLL_INTERVAL`?

IvayloAndonov: Actual the poll interval is not associated with the cache, rather than with the DB sync provider. The cache service shouldn't know about the provider's custom polling, no? 🤔

mkazlauskas: Right. You don't have to use this exact name, just wanted to point out that POLL_INTERVAL is too generic for an env variable

rhyslbw: `DB_POLL_INTERVAL` ?

packages/cardano-services/test/NetworkInfo/NetworkInfoHttpService.test.ts
```diff
@@ -58,6 +83,10 @@ describe('NetworkInfoHttpService', () => {
     });
 
     describe('/network', () => {
+      const path = '/network';
+      const dbSyncPollQueriesCount = 1;
```
rhyslbw, line 87:
```suggestion
      const DB_POLL_QUERIES_COUNT = 1;
```
1. We've been using this format for `number`, `bigint`, or `string` constants, although it would be good to have a lint rule to enforce it, so I'm ok with it either way.
2. We can just refer to this as DB polling since we're interfacing with PostgreSQL. `cardano-db-sync` is the projector.

packages/cardano-services/test/NetworkInfo/NetworkInfoHttpService.test.ts
```diff
@@ -71,9 +100,74 @@ describe('NetworkInfoHttpService', () => {
         }
       });
 
+      describe('cached', () => {
+        it('should call db-sync queries only once when the response is cached', async () => {
```
rhyslbw, line 104:
```suggestion
        it('should query the DB only once when the response is cached', async () => {
```

packages/cardano-services/test/entrypoints.test.ts
```diff
@@ -161,6 +166,34 @@ describe('entrypoints', () => {
           });
         });
 
+        it('cli:start-server network-info exits with code 0 when cache duration is out of range', (done) => {
+          const cacheDurationOutOfRange = '3000';
```
rhyslbw, line 170:
```suggestion
          expect.assertions(2);
          const cacheDurationOutOfRange = '3000';
```

packages/cardano-services/test/util.ts
```diff
@@ -31,3 +36,30 @@ export const doServerRequest =
...
+};
+
+export const deleteDbData = async (db: Pool, tableName: string, field: string, value: any) => {
+  const sqlString = `DELETE FROM ${tableName} WHERE ${field} = ${value};`;
+
+  await db.query(sqlString);
+};
+
+/**
+ * Wraps unit test within a transaction and once test has finished, rollback is executed
```
rhyslbw, line 55:
```suggestion
 * Wraps integration test within a transaction and once test has finished, rollback is executed
```

IvayloAndonov: [6e5ee7c](https://github.com/input-output-hk/cardano-js-sdk/pull/276/commits/6e5ee7c71babfe32d15e36f92ea25177a42f75a5)

packages/cardano-services/test/entrypoints.test.ts
IvayloAndonov: [6e5ee7c](https://github.com/input-output-hk/cardano-js-sdk/pull/276/commits/6e5ee7c71babfe32d15e36f92ea25177a42f75a5)

packages/cardano-services/test/NetworkInfo/NetworkInfoHttpService.test.ts
IvayloAndonov: [6e5ee7c](https://github.com/input-output-hk/cardano-js-sdk/pull/276/commits/6e5ee7c71babfe32d15e36f92ea25177a42f75a5)

packages/cardano-services/test/NetworkInfo/NetworkInfoHttpService.test.ts
IvayloAndonov: [6e5ee7c](https://github.com/input-output-hk/cardano-js-sdk/pull/276/commits/6e5ee7c71babfe32d15e36f92ea25177a42f75a5)

packages/cardano-services/src/CacheService/CacheService.ts
```diff
@@ -0,0 +1,87 @@
+import { CACHE_DURATION_DEFAULT } from './defaults';
+import NodeCache from 'node-cache';
+
+export type Key = string | number;
+export type DbSyncQuery<T> = () => Promise<T>;
+
+export class CacheService {
```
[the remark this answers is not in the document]
IvayloAndonov: [6e5ee7c](https://github.com/input-output-hk/cardano-js-sdk/pull/276/commits/6e5ee7c71babfe32d15e36f92ea25177a42f75a5)

packages/cardano-services/src/NetworkInfo/NetworkInfoHttpService.ts
```diff
@@ -27,8 +27,9 @@ export class NetworkInfoHttpService extends HttpService {
     return this.#networkInfoProvider.healthCheck();
   }
 
-  static create({ logger = dummyLogger, networkInfoProvider }: NetworkInfoServiceDependencies) {
+  static async create({ logger = dummyLogger, networkInfoProvider }: NetworkInfoServiceDependencies) {
```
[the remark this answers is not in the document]
IvayloAndonov: [6e5ee7c](https://github.com/input-output-hk/cardano-js-sdk/pull/276/commits/6e5ee7c71babfe32d15e36f92ea25177a42f75a5)

packages/cardano-services/src/Program/loadHttpServer.ts
IvayloAndonov: [6e5ee7c](https://github.com/input-output-hk/cardano-js-sdk/pull/276/commits/6e5ee7c71babfe32d15e36f92ea25177a42f75a5)

packages/cardano-services/test/entrypoints.test.ts
```diff
@@ -220,6 +254,25 @@ describe('entrypoints', () => {
           });
         });
 
+        it('run network-info exits with code 1 when cache duration is out of range', (done) => {
+          const cacheDurationOutOfRange = '3000';
```
rhyslbw, line 258:
```suggestion
          expect.assertions(2);
          const cacheDurationOutOfRange = '3000';
```

Sorry, I should have been more explicit. Just ensure any of the callback-style tests with assertions in handlers follow this pattern (eg. line 277 as well)

packages/cardano-services/test/entrypoints.test.ts
```diff
@@ -220,6 +253,25 @@ describe('entrypoints', () => {
           });
         });
 
+        it('run network-info exits with code 1 when cache duration is out of range', (done) => {
```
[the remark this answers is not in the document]
rhyslbw: Actually, this one is appropriate, it's the CLI that's wrong. Can you see if [this](https://github.com/input-output-hk/cardano-js-sdk/blob/b003d499b6fad289b8d9656c6293a4f23856b5ed/packages/cardano-services/src/cli.ts#L131) is being executed?

rhyslbw (changes requested):

packages/cardano-services/src/InMemoryCache/InMemoryCache.ts
```diff
@@ -0,0 +1,87 @@
...
+import NodeCache from 'node-cache';
+
+export type Key = string | number;
+export type DbSyncQuery<T> = () => Promise<T>;
+
+export class InMemoryCache {
+  #cache: NodeCache;
+  #ttlDefault: number;
+
+  constructor(cacheDurationInMins: number = CACHE_DURATION_DEFAULT) {
```
rhyslbw, line 11:
```suggestion
  constructor(cacheTtlInMins: number = CACHE_TTL_DEFAULT) {
```
Let's rename all variables for consistency: cache duration -> cache TTL

rhyslbw (changes requested):

packages/cardano-services/test/entrypoints.test.ts
IvayloAndonov: Does it mean to add assertions in all other tests that are missing?

rhyslbw: Yes. It's to ensure future regression causing the assertions to not be executed do not produce a false-positive.

packages/cardano-services/test/entrypoints.test.ts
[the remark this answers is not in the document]
IvayloAndonov: This test is about `run.ts` only, not `cli` ( it exits with 0 properly ). It's failing [here](https://github.com/input-output-hk/cardano-js-sdk/blob/6e5ee7c71babfe32d15e36f92ea25177a42f75a5/packages/cardano-services/src/run.ts#L20-L22) where we do not have try-catch block.

packages/cardano-services/src/InMemoryCache/InMemoryCache.ts
IvayloAndonov: [00d4bba](https://github.com/input-output-hk/cardano-js-sdk/pull/276/commits/00d4bbae9dd517b46acd8abc7125bcff480b513c)

packages/cardano-services/test/entrypoints.test.ts
IvayloAndonov: [00d4bba](https://github.com/input-output-hk/cardano-js-sdk/pull/276/commits/00d4bbae9dd517b46acd8abc7125bcff480b513c)

packages/cardano-services/src/CacheService/CacheService.ts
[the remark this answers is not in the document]
IvayloAndonov: [00d4bba](https://github.com/input-output-hk/cardano-js-sdk/pull/276/commits/00d4bbae9dd517b46acd8abc7125bcff480b513c)

packages/cardano-services/.env.test
```diff
@@ -1,2 +1,3 @@
 DB_CONNECTION_STRING=postgresql://postgres:doNoUseThisSecret!@127.0.0.1:5433/testnet
-CARDANO_NODE_CONFIG_PATH=./config/network/testnet/cardano-node/config.json
\ No newline at end of file
+CARDANO_NODE_CONFIG_PATH=./config/network/testnet/cardano-node/config.json
+DB_QUERIES_CACHE_DURATION=120
```
rhyslbw, line 3:
```suggestion
DB_QUERIES_CACHE_TTL=120
```

packages/cardano-services/src/Program/ProgramOptionDescriptions.ts
```diff
@@ -7,7 +7,9 @@ enum HttpServerOptionDescriptions {
   OgmiosUrl = 'Ogmios URL',
   RabbitMQUrl = 'RabbitMQ URL',
   UseQueue = 'Enables RabbitMQ',
-  CardanoNodeConfigPath = 'Cardano node config path'
+  CardanoNodeConfigPath = 'Cardano node config path',
+  DbQueriesCacheDuration = 'Db queries cache duration in minutes between 1 and 2880',
```
rhyslbw, line 11:
```suggestion
  DbQueriesCacheTtl = 'Db queries cache TTL in minutes between 1 and 2880',
```

packages/cardano-services/src/Program/loadHttpServer.ts
```diff
@@ -17,6 +18,8 @@ import Logger, { createLogger } from 'bunyan';
 
 export interface HttpServerOptions extends CommonProgramOptions {
   dbConnectionString?: string;
+  dbQueriesCacheDuration: number;
```
rhyslbw, line 21:
```suggestion
  dbQueriesCacheTtl: number;
```

packages/cardano-services/src/cli.ts
```diff
@@ -23,6 +25,7 @@ import {
   loadTxWorker
 } from './TxWorker';
 import { URL } from 'url';
+import { cacheDurationValidator } from './util/validators';
```
rhyslbw, line 28:
```suggestion
import { cacheTtlValidator } from './util/validators';
```

packages/cardano-services/src/cli.ts
```diff
@@ -74,6 +77,18 @@ commonOptions(
     new URL(url).toString()
   )
   .option('--cardano-node-config-path <cardanoNodeConfigPath>', ProgramOptionDescriptions.CardanoNodeConfigPath)
+  .option(
+    '--db-queries-cache-duration <dbQueriesCacheDuration>',
```
rhyslbw, line 81:
```suggestion
    '--db-queries-cache-ttl <dbQueriesCacheTtl>',
```

packages/cardano-services/test/InMemoryCache/InMemoryCache.test.ts
```diff
@@ -0,0 +1,76 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { CACHE_TTL_DEFAULT, InMemoryCache } from '../../src/InMemoryCache';
+import NodeCache from 'node-cache';
+
+jest.mock('node-cache');
+
+describe('In-Memory Cache', () => {
```
rhyslbw, line 7: Best to just use the module name here
```suggestion
describe('InMemoryCache', () => {
```

packages/cardano-services/test/entrypoints.test.ts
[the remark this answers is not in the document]
rhyslbw: We just should have consistent behaviour between the run script and the CLI, so if `envalid` is doing the right thing, let's correct the CLI to produce the same exit code.

rhyslbw (changes requested):
There are still some references to "duration", which I've highlighted some, but not all.

rhyslbw (approved):
Great work @IvayloAndonov. Just waiting for follow-up review from @mkazlauskas, then we're ready for squashing prior to QA review.

packages/cardano-services/src/CacheService/keys.ts
mkazlauskas: @IvayloAndonov can we move `packages/cardano-services/src/InMemoryCache/keys.ts` file to `packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/keys.ts`?

packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
```diff
@@ -42,4 +65,13 @@ export class DbSyncNetworkInfoProvider extends DbSyncProvider implements Network
       totalSupply
     });
   }
+
+  async start(): Promise<void> {
+    this.#pollId = pollDbSync(this.#cache, () => this.#builder.queryLatestEpoch(), this.#dbPollInterval);
```
mkazlauskas, line 70: This would be better, wouldn't leak implementation details (setInterval):
```suggestion
    this.#pollService = pollDbSync(this.#cache, () => this.#builder.queryLatestEpoch(), this.#dbPollInterval);
    ...
    async close(): Promise<void> {
      this.#pollService?.shutdown();
      this.#cache.shutdown();
    }
```

mkazlauskas (changes requested):

mkazlauskas (dismissed):
Nice work @IvayloAndonov! :tada:

packages/core/src/Provider/Provider.ts
```diff
@@ -3,6 +3,10 @@ export type HealthCheckResponse = {
 };
 
 export interface Provider {
+  start?(): Promise<void>;
```
mkazlauskas, line 6: Looking at this again, I'm actually not liking core Provider having `{start, close}`. Can we move those to cardano-services? I suggest `DbSyncProvider` or a new interface that extends core Provider. What do you think @rhyslbw?

IvayloAndonov:
The commits are being squashed. Ready for QA :)

packages/core/src/Provider/Provider.ts
mkazlauskas: Makes perfect sense: `type RunnableProvider = Provider & RunnableModule`?

IvayloAndonov: I like the suggestion of `RunnableProvider`. So in the end we agree on: 
1. Leave only `healthCheck` method within `Provider` and get rid of [{start, close}](https://github.com/input-output-hk/cardano-js-sdk/blob/2c61c6eb2a7a7de81ce37c5d524e526bc17767de/packages/core/src/Provider/Provider.ts#L6-L13)?
2. Rename `RunnableModule`'s [start](https://github.com/input-output-hk/cardano-js-sdk/blob/71a939b874296d86183d89fce7877c565630e921/packages/cardano-services/src/RunnableModule.ts#L28-L32) to `run` and all inners methods (before/after run)?
3. Create `type RunnableProvider = Provider & Pick<RunnableModule, 'run' | 'shutdown'>` within `cardano-services` package
4. These two changes should be a separate commit as part of this PR?

mkazlauskas: Yes to all questions, unless @rhyslbw objects

rhyslbw: Just make sure to mark the method name change as breaking

mkazlauskas: I think `runAfter` setting running state is confusing. Perhaps not such a great refactor idea :thinking:

IvayloAndonov: Yeah, since we have `'starting' | 'running' | 'stopping'` states, I guess we shouldn't ignore `start` as a runnable module lifecycle method?

IvayloAndonov: It turned out there is another issue with the refactoring - as discussed we remove `{start/stop}` methods from `core`'s [Provider](https://github.com/input-output-hk/cardano-js-sdk/blob/a2bfcc62c25e71d78d07b961267d7ce9679b6cf4/packages/core/src/Provider/Provider.ts#L5) interface and create `type RunnableProvider<T extends Provider> = T & Pick<RunnableModule, 'start' | 'shutdown'>` inside `cardano-services` and start using it. But we have another packages such as [rabbitmqTxSubmitProvider](https://github.com/input-output-hk/cardano-js-sdk/blob/3c0f6048c5cfa04654f0a5463dfccefd24c9054e/packages/rabbitmq/src/rabbitmqTxSubmitProvider.ts#L11) implementing [TxSubmitProvider](https://github.com/input-output-hk/cardano-js-sdk/blob/e155ed4efcd1338a54099d1a9034ccbeddeef1cc/packages/core/src/Provider/TxSubmitProvider/types.ts#L5) extending Provider. So the issue is we want to align all providers with `RunnableProvider` (located within `cardano-services`) where some of them depends only on the `core` package.
I'm happy to discuss the approach here in a quick huddle.

mkazlauskas (approved):

rhyslbw (approved):

mkazlauskas:
Looks good to merge @rhyslbw ?

PR #280, opened 2022-06-13
fix: division by zero error at pool rewards query

lgobbi-atix:
# Context

When stake-pool server is started with testnet-db-snapshot DB then called POST ”stake-pool/search”   with the request - 
```
{ "args":[] }
```
It returns 500 and divide by zero error. 

# Proposed Solution
Add `WHEN COALESCE(stake.active_stake, 0) = 0 THEN 0` check to epoch rewards query

### DIFF
--- a/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/StakePoolBuilder.ts
+++ b/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/StakePoolBuilder.ts
@@ -82,7 +82,7 @@ export class StakePoolBuilder {
     return Promise.all(
       hashesIds.map(async (hashId) => {
         const result: QueryResult<EpochRewardModel> = await this.#db.query(Queries.findPoolEpochRewards(limit), [
-          hashId
+          [hashId]
         ]);
         return result.rows.length > 0 ? mapEpochReward(result.rows[0], hashId) : undefined;
       })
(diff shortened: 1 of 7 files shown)


ssong-iohk (approved):
It works now.
See the successful response. 
[280-response.json.txt](https://github.com/input-output-hk/cardano-js-sdk/files/8899224/280-response.json.txt)

mkazlauskas (changes requested):
Would be great to have a test that fails without  changes in 1db1b37a4a2e101a07cccd973170c0a3b494b3c5

lgobbi-atix:
> Would be great to have a test that fails without changes in [1db1b37](https://github.com/input-output-hk/cardano-js-sdk/commit/1db1b37a4a2e101a07cccd973170c0a3b494b3c5)

The test should be this one, which returns all stake pools https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services/test/StakePool/StakePoolHttpService.test.ts#L96

Or even this one https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services/test/StakePool/DbSyncStakePoolProvider/StakePoolBuilder.test.ts#L48 but with a stake pool id with 0 total stake.

Right now we don't have any cases like that in the fixture data, that's why it was not failing before. 
With the way we are currently handling the fixture data adding a case like that it's not easy without hardcoding it

rhyslbw (approved):

PR #281, opened 2022-06-13
feat: add stake pool apy field and sort by cost

lgobbi-atix:
# Context

The current sort order for results returned from the DbSyncStakePoolProvider queryStakePools method is ordered by name. We require the interface to also include `APY` and `cost`.

# Proposed Solution

Add APY calculation to epoch rewards query and corresponding sort field
Add cost sort field, which orders by cost first and margin second.

- Add new `queryPoolAPY` and `findPoolAPY` functions to calculate pool APY
- Add `apy` property to `Cardano.StakePool` interface
- Modify `findPoolEpochRewards` to work for several stake pools and be able to use the response to calculate the APY and order by it
- Add stake pool `cost` sort field


packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/queries.ts
```diff
@@ -244,72 +244,123 @@ LEFT JOIN owners_total_utxos otu on
...
-   order by rewards.epoch_no desc
+)`;
+
+export const findPoolEpochRewards = (limit?: number) => `
+  ${epochRewardsSubqueries(limit)}
+  SELECT * FROM epoch_rewards
+  ORDER BY epoch_no desc
+`;
+
+export const findPoolAPY = (limit?: number) => `
```
lgobbi-atix, line 334: I'm getting some very big APY values for some stake pools, this was based on `packages/core/src/Cardano/util/estimateStakePoolAPY.ts`.
Are we sure the formulas for this and `memberROI` are correct?

packages/core/src/Cardano/types/StakePool/StakePool.ts
```diff
@@ -86,4 +86,8 @@ export interface StakePool extends PoolParameters {
    * Sorted by epoch in ascending order.
    */
   epochRewards: StakePoolEpochRewards[];
+  /**
+   * Annual Percentage Yield (APY)
+   */
+  apy?: number;
```
lgobbi-atix, line 92: Not sure if this is the correct place for this

mkazlauskas (changes requested):

packages/core/src/Cardano/types/StakePool/StakePool.ts
lgobbi-atix: I thought about it but I need the pool rewards to calculate this and the metrics are retrieved by another query. I would need to adjust the types but it's doable

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/queries.ts
lgobbi-atix: In my local db with testnet data there are 67 stake pools with APY >= 1. The biggest one is
```
hash_id|apy                  |
-------|---------------------|
    226|   871048215088976000|
```

Epoch rewards for that stake pool are:
```
epoch_no|epoch_length|hash_id|total_rewards|active_stake|operator_fees|member_roi        |
--------|------------|-------|-------------|------------|-------------|------------------|
      89|   431989000|    226|            0|   693587479|          0.0|               0.0|
      88|   431983000|    226|    500000000|    97247840|    352000000|1.5218847020149753|
```

Avg. daily ROI:
```
hash_id|avg_roi            |
-------|-------------------|
    226|0.15219445933531395|
```

Avg. daily ROI calculation in the code:
```ts
const roisPerDay = rewardsHistory.map(
    ({ epochLength, memberROI }) => memberROI / (epochLength / MILLISECONDS_PER_DAY)
  );
const averageDailyROI = sum(roisPerDay) / roisPerDay.length;
```
which would be
``` 
(0 + (1.5218847020149753 / (431983000/86400000)) / 2 = 
0.30438891867 / 2 = 
0.15219445933
```

And then to calculate the final number:
```ts
const epochLengthInDays = rewardsHistory[rewardsHistory.length - 1].epochLength / MILLISECONDS_PER_DAY;
// epochLengthInDays = 431989000 / 86400000 = 4.99987268519

const roiPerEpoch = averageDailyROI * epochLengthInDays;
// roiPerEpoch = 0.15219445933 * 4.99987268519 = 0.76095292004

const numEpochs = 365 / epochLengthInDays;
// numEpochs = 365 / 4.99987268519 = 73.0018588436

return Math.pow(1 + roiPerEpoch, numEpochs) - 1;
// APY = ((1 + 0.76095292004) ^ 73.0018588436) - 1 = 8.7104821e+17
```

which is what I'm getting from the query

lgobbi-atix: For the member roi calculation, in epoch 88 the total rewards (500000000) are greater than fixed cost (340000000) so it does:
```
(rewards - (FLOOR((rewards - cost) * margin) + cost)) / stake =

(500000000 - (((500000000 - 340000000) * 0.075) + 340000000)) / 97247840 =

1.52188470201
```

in epoch 89 fixed cost is greater than the rewards (0), and it does:
```
(rewards - rewards) / stake
```
which will always return 0

lgobbi-atix: Another things is that rewards can be type `[leader, member, reserves, treasury, refund]`, is it ok that we take all of them into account when calculating the roi and stuff?

mkazlauskas: `1.5218847020149753` for `member_roi` is way too much, it would be 152% interest per epoch :laughing: Could be possible with very small stake and incredible luck though? :thinking: 

> Another thing is that rewards can be type [leader, member, reserves, treasury, refund], is it ok that we take all of them into account when calculating the roi and stuff?

No, we're only interested in member rewards here when computing APY, so something like:
```
epoch_member_rewards = SUM(SELECT ... WHERE epoch=... AND type="member")
member_roi = epoch_member_rewards / active_stake
```

lgobbi-atix: With `member` rewards only the numbers make more sense, biggest member roi being `0.0002945421377973` and APY `0.012573814334454259`

mkazlauskas: Hmm that's much better, but now it seems too small for the biggest member roi. Although I guess it could be correct for the testnet data :thinking Are you using somewhat fully synced data? We might want to test this on an up-to-date one.

lgobbi-atix: I stopped db-sync after restoring the snapshot at block 3590461

lgobbi-atix: I'm getting 0 roi and apy for this one https://testnet.cexplorer.io/pool/pool167u07rzwu6dr40hx2pr4vh592vxp4zen9ct2p3h84wzqzv6fkgv

packages/core/src/Cardano/types/StakePool/StakePool.ts
lgobbi-atix: Yes, you are right, it was easier than I thought. Fixed it

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/queries.ts
mkazlauskas: If you look at https://testnet.cexplorer.io/pool/pool167u07rzwu6dr40hx2pr4vh592vxp4zen9ct2p3h84wzqzv6fkgv/rewards#data delegators get negligible rewards. I don't think we should be counting in operator (`leader`) rewards when computing APY

mkazlauskas: Maybe missing a cast to float somewhere so it gets to 0 :thinking:

lgobbi-atix: No, it's 0 because of this case at the member roi calculation: 
```sql
   WHEN pool.fixed_cost >= rewards.total_amount
    THEN (
      --  (rewards - rewards) / stake will always be 0
      COALESCE(
      	(rewards.total_amount - COALESCE(rewards.total_amount, 0)) / stake.active_stake
      , 0)
    )
```
This stake pool rewards are less than the fixed_cost
```
fixed_cost|total_rewards|
----------|-------------|
 340000000|     33436585|
 340000000|     31308365|
 340000000|     29652223|
 340000000|     29274580|
 340000000|     28246202|
 340000000|     27212254|
```

mkazlauskas: https://testnet.cardanoscan.io/pool/d7b8ff0c4ee69a3abee65047565e85530c1a8b332e16a0c6e7ab8401 also shows lifetime rewards as 0. Seems like we're good then?

packages/cardano-services/test/Reward/DbSyncRewardsProvider/__snapshots__/RewardsBuilder.test.ts.snap
```diff
@@ -8,16 +8,62 @@ Object {
...
 Object {
-  "balance": "48642468465",
+  "balance": "97675046319",
 }
 `;
 
 exports[`RewardsBuilder getRewardsHistory returns RewardEpochModel when there is epochs field 1`] = `Array []`;
 
-exports[`RewardsBuilder getRewardsHistory returns RewardEpochModel when there is no epochs field 1`] = `Array []`;
+exports[`RewardsBuilder getRewardsHistory returns RewardEpochModel when there is no epochs field 1`] = `
```
rhyslbw, line 17: This snapshot will fail in the next epoch. The result we snapshot needs to be consistent

packages/cardano-services/test/Reward/DbSyncRewardsProvider/__snapshots__/RewardsBuilder.test.ts.snap
```diff
@@ -8,16 +8,62 @@ Object {
...
+    "epoch": "161",
+    "quantity": "16214946715",
+  },
+]
+`;
 
 exports[`RewardsBuilder getRewardsHistory returns RewardEpochModel when there is no epochs field and empty accounts 1`] = `Array []`;
 
-exports[`RewardsBuilder getRewardsHistory returns RewardEpochModel when there is partially epochs field with lowerBound 1`] = `Array []`;
+exports[`RewardsBuilder getRewardsHistory returns RewardEpochModel when there is partially epochs field with lowerBound 1`] = `
```
rhyslbw, line 44: Same with this. Need to make the result segment consistent

packages/cardano-services/test/Reward/DbSyncRewardsProvider/__snapshots__/RewardsBuilder.test.ts.snap
```diff
@@ -8,16 +8,62 @@ Object {
...
+  },
+  Object {
+    "address": "stake_test1uqfu74w3wh4gfzu8m6e7j987h4lq9r3t7ef5gaw497uu85qsqfy27",
+    "epoch": "161",
+    "quantity": "16214946715",
+  },
+]
+`;
 
 exports[`RewardsBuilder getRewardsHistory returns RewardEpochModel when there is no epochs field and empty accounts 1`] = `Array []`;
```
rhyslbw, line 42: Let's change this to assert the expected response is an empty array

packages/cardano-services/test/StakePool/DbSyncStakePoolProvider/StakePoolBuilder.test.ts
```diff
@@ -115,6 +117,26 @@ describe('StakePoolBuilder', () => {
         expect(pools).toHaveLength(3);
         expect(pools).toMatchSnapshot();
       });
+      it('by cost and margin desc', async () => {
```
rhyslbw, line 120:
```suggestion
      it('orders results by real-world cost by considering fixed cost and margin when specifying sort by cost desc', async () => {
```

packages/cardano-services/test/StakePool/DbSyncStakePoolProvider/StakePoolBuilder.test.ts
```diff
@@ -115,6 +117,26 @@ describe('StakePoolBuilder', () => {
...
+        const pools = (await builder.queryPoolData([14, 15, 20], { sort: { field: 'cost', order: 'desc' } })).map(
+          (qR) => {
+            const { hashId: _1, updateId: _2, ...poolData } = qR;
+            return poolData;
+          }
+        );
+        expect(pools).toHaveLength(3);
+        expect(pools).toMatchSnapshot();
+      });
+      it('by cost and margin asc', async () => {
```
rhyslbw, line 130:
```suggestion
      it('orders results by real-world cost by considering fixed cost and margin when specifying sort by cost asc', async () => {
```

packages/cardano-services/test/StakePool/DbSyncStakePoolProvider/StakePoolBuilder.test.ts
```diff
@@ -279,4 +301,38 @@ describe('StakePoolBuilder', () => {
...
+    describe('sort', () => {
+      it('by default sort (APY desc)', async () => {
+        const result = (await builder.queryPoolAPY([1, 15])).map(({ apy }) => ({ apy }));
+        expect(result).toHaveLength(2);
+        expect(result[0].apy).toBeGreaterThan(result[1].apy);
+        expect(result).toMatchSnapshot();
+      });
+      it('by APY asc', async () => {
+        const result = (await builder.queryPoolAPY([1, 15], { sort: { field: 'apy', order: 'asc' } })).map(
+          ({ apy }) => ({ apy })
```
rhyslbw, line 314:
```suggestion
          ({ apy }) => ({ apy })
```
Should we include a pool identifier to improve the snapshots?

rhyslbw (changes requested):

packages/cardano-services/test/StakePool/DbSyncStakePoolProvider/StakePoolBuilder.test.ts
lgobbi-atix: This query has the `hashId` as the identifier to match with the rest of the pool information but that could change as it's generated by postgres. We could add a couple of `JOIN`s to the query to get the bech32 pool id but it won't be used anywhere else than in tests.

packages/cardano-services/test/NetworkInfo/__snapshots__/NetworkInfoHttpService.test.ts.snap
```diff
@@ -3,7 +3,7 @@
 exports[`NetworkInfoHttpService healthy state /network with NetworkInfoHttpProvider response is an object of network info 1`] = `
 Object {
   "lovelaceSupply": Object {
-    "circulating": 69163989759711432n,
+    "circulating": 69173735562813828n,
```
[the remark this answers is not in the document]
lgobbi-atix: I'm getting that `circulating` is greater than both `total` and `max`

I could do something like this for now but I don't think the `circulating` calculation is correct if it should be less than `max` and `total`
```ts
expect(response.lovelaceSupply.circulating).toBeGreaterThan(0);
expect(response.lovelaceSupply.total).toBeGreaterThan(0);
expect(response.lovelaceSupply.max).toBeGreaterThan(response.lovelaceSupply.total);
```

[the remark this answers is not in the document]
rhyslbw: Yes, let's treat the circulating supply issue separately then. I'll log a ticket

packages/cardano-services/test/StakePool/DbSyncStakePoolProvider/StakePoolBuilder.test.ts
rhyslbw: @mkazlauskas thoughts on the presence of the pool ID in the query to facilitate verification? I'm a little concerned since it's `JOIN`ing :thinking:

mkazlauskas: Don't think we should be adding JOINs just for tests. We're already querying by `hashesIds: number[]` as first argument in `builder.queryPoolAPY([1, 15]` so adding it in the snapshot won't make it worse.

mkazlauskas (changes requested):
Looks good, just need to rebase on `master` to resolve snapshot conflicts

mkazlauskas (approved):
Looks good! Again a few conflicts though :sweat_smile:

packages/cardano-services/test/NetworkInfo/__snapshots__/NetworkInfoHttpService.test.ts.snap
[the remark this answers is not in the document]
rhyslbw: ADP-1973

rhyslbw (approved):
Great work @lgobbi-atix

mkazlauskas (approved):

ssong-iohk (approved):

mkazlauskas (approved):

PR #282, opened 2022-06-14
fix(cardano-services): resolve package.json path when cli is run with ts-node

mkazlauskas:
# Context

`yarn cli` in `cardano-services` is broken

# Proposed Solution

Correctly resolve `package.json` path when run in `ts-node`


ssong-iohk (approved):
It works. thanks!

rhyslbw (approved):

PR #283, opened 2022-06-14
feat: transaction chaining

mkazlauskas:
# Context

It’d be useful to be able to submit a transaction that has an input that is an output of a transaction that didn’t get onto chain yet.

# Proposed Solution

- Include pending transaction change outputs in available utxo set [95c2671](https://github.com/input-output-hk/cardano-js-sdk/pull/283/commits/95c2671055627a7c6cb334adc1a05b6b43bcb738)
- Improve `ObservableWallet.balance` interface: with transaction chaining, deposit and rewards should be reflected to balance immediately and it no longer makes sense to have `balance.{total$,available$}.{deposit,rewards}`. Therefore there is now just a single field for deposit and rewards. [b8371f9](https://github.com/input-output-hk/cardano-js-sdk/pull/283/commits/b8371f97e151c2e9cb18e0ac431e9703fe490d26)

### DIFF
--- a/packages/wallet/src/cip30.ts
+++ b/packages/wallet/src/cip30.ts
@@ -71,7 +71,7 @@ export const createWalletApi = (
     logger.debug('getting balance');
     try {
       const wallet = await walletReady;
-      const value = await firstValueFrom(wallet.balance.available$);
+      const value = await firstValueFrom(wallet.balance.utxo.available$);
       return Buffer.from(coreToCsl.value(value).to_bytes()).toString('hex');
     } catch (error) {
       logger.error(error);
--- a/packages/wallet/test/integration/txChainingBalance.test.ts
+++ b/packages/wallet/test/integration/txChainingBalance.test.ts
@@ -1,0 +1,40 @@
+import { Cardano } from '@cardano-sdk/core';
+import { SingleAddressWallet } from '../../src';
+import { createWallet } from './util';
+import { firstValueFrom } from 'rxjs';
+
+describe('integration/txChainingBalance', () => {
+  let wallet: SingleAddressWallet;
+
+  beforeAll(async () => {
+    // Using mock TxSubmitProvider which doesn't do anything and instantly resolves
+    wallet = await createWallet();
+  });
+
+  it('available balance includes change outputs from pending transaction', async () => {
+    const utxo = await firstValueFrom(wallet.utxo.available$);
+    const balanceBefore = await firstValueFrom(wallet.balance.utxo.available$);
+    const outputCoins = utxo[0][1].value.coins + 1n; // will always select at least 2 utxo with mock utxo set
+    const output: Cardano.TxOut = {
+      address: Cardano.Address(
+        'addr_test1qzs0umu0s2ammmpw0hea0w2crtcymdjvvlqngpgqy76gpfnuzcjqw982pcftgx53fu5527z2cj2tkx2h8ux2vxsg475qp3y3vz'
+      ),
+      value: {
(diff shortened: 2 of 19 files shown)


packages/wallet/test/integration/txChainingBalance.test.ts
```diff
@@ -0,0 +1,32 @@
+import { Cardano } from '@cardano-sdk/core';
+import { SingleAddressWallet } from '../../src';
+import { createWallet } from './util';
+import { firstValueFrom } from 'rxjs';
+
+describe('integration/txChainingBalance', () => {
+  let wallet: SingleAddressWallet;
+
+  beforeAll(async () => {
+    wallet = await createWallet();
```
rhyslbw, line 10: Maybe add a comment here to ensure it's clear we're using a mocked tx submit provider

packages/wallet/src/cip30.ts
```diff
@@ -71,7 +71,13 @@ export const createWalletApi = (
     logger.debug('getting balance');
     try {
       const wallet = await walletReady;
-      const value = await firstValueFrom(wallet.balance.available$);
+      // Review: not including balance.rewardAccounts.{deposit,rewards}, should we?
```
rhyslbw, line 74: Unwithdrawn rewards are conceptually part of the balance, however, the[ CIP indicates](https://cips.cardano.org/cips/cip30/#apigetbalancepromisecborvalue) that it should just be the UTxOs

packages/wallet/src/cip30.ts
```diff
@@ -71,7 +71,13 @@ export const createWalletApi = (
...
+      // Not sure what the intended use case of getBalance is.
+      // We should include it if:
+      // - If it's to display the balance to the user
+      // - If it's to get quantity of coins delegated (although not sure if deposit is included)
```
rhyslbw, line 78: Deposits are held in a separate ada pot, and are not counted as part of the user's delegated funds.

rhyslbw (changes requested):
Looks good, and that's a nice refactor

packages/wallet/src/cip30.ts
mkazlauskas: Thanks, should've just taken a look at the cip :man_facepalming:

packages/wallet/test/integration/txChainingBalance.test.ts
mkazlauskas: How would another provider make it produce a false positive? :thinking: If you're thinking `wallet.balance` being updated from the provider then I don't think it could ever happen, even if the transaction got on-chain instantly upon submission and the wallet polling request matched the same tick, because it's only polling the `tip`, balance (utxo) request triggered after that

rhyslbw: Right, it wouldn't. The comment is just to communicate the behavior is isolated from any network transmission. I also immediately wondered which provider was being used, so it would clarify just this point alone.

mkazlauskas: Added a few comments in 4c94a3ff807c435c5aa27a04b46037fee3495d22

mkazlauskas: Further improved the test by asserting precondition in ccb6fa7aa7ac6b8c0dc4f6a23d12ef370a78a5d2

rhyslbw (approved):

rhyslbw (approved):
image

mirceahasegan (approved):

PR #284, opened 2022-06-14
refactor(cardano-services): order epoch and block queries by number instead of id

lgobbi-atix:
# Context

Currently, we are retrieving the last epoch using `ORDER BY id DESC LIMIT 1`, with `id` being the table id.

We should use `no` of the epoch record instead of `id` . Given that the epoch table index is created by `no` (`idx_epoc_no`)

# Proposed Solution

Modify block and epoch queries so they are ordered by number instead of `id`


rhyslbw (approved):
LGTM

mkazlauskas (approved):

IvayloAndonov (approved):

mkazlauskas (approved):

rhyslbw (approved):

PR #285, opened 2022-06-15
fix: messaging prop cache

mkazlauskas:
# Context

Extension messaging remote api object shouldn't recreate objects on every property access

```typescript
const obj = consumeRemoteApi(...)
obj.prop === obj.prop // false, should be true
```

# Proposed Solution

Cache remote api properties 44764aa6ef578d43b5726ba56a7d5c2f80958359

- chore(web-extension): bump chromedriver dependency version 9d042391c237836108f22c98044a447f1982a399
- refactor(web-extension)!: rename RemoteApiProperty.Observable->HotObservable 4bc99224d3cdcadc90729eecd8cb9ea2d6227438

### DIFF
--- a/packages/web-extension/e2e/extension/util.ts
+++ b/packages/web-extension/e2e/extension/util.ts
@@ -22,7 +22,7 @@ export interface BackgroundServices {
 }
 
 export const adaPriceProperties: RemoteApiProperties<BackgroundServices> = {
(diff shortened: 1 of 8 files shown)


rhyslbw (approved):
Nice work @mkazlauskas

mchappell (approved):

PR #286, opened 2022-06-15
feat: Enhance InMemoryCache 'get' method

IvayloAndonov:
# Context

In case when the cache is cleared and multiple simultaneous requests come in, we would run the DB query multiple times.
Raised by @mkazlauskas 

# Proposed Solution

Caching the promise instead of resolving the value, so the next invocation will instantly return the same promise.
In case of DB query fails -> `result.catch()` resets the cache so don't have to wait for the TTL

### DIFF
--- a/packages/cardano-services/src/InMemoryCache/InMemoryCache.ts
+++ b/packages/cardano-services/src/InMemoryCache/InMemoryCache.ts
@@ -27,9 +27,13 @@ export class InMemoryCache {
       return cachedValue;
     }
 
-    const result = await dbSyncQuery();
-    this.#cache.set(key, result, ttl);
-    return result;
+    const resultPromise = dbSyncQuery();
+    this.#cache.set(
+      key,
+      resultPromise.catch(() => this.#cache.del(key)),
+      ttl
+    );
+    return resultPromise;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


rhyslbw (changes requested):
It would be best to have a test demonstrate this

packages/cardano-services/src/InMemoryCache/InMemoryCache.ts
```diff
@@ -27,8 +27,12 @@ export class InMemoryCache {
       return cachedValue;
     }
 
-    const result = await dbSyncQuery();
-    this.#cache.set(key, result, ttl);
+    const result = dbSyncQuery();
```
rhyslbw, line 30:
```suggestion
    const resultPromise = dbSyncQuery();
```

rhyslbw (changes requested):
What about the test scenario of multiple requests being received before the cache is primed?

packages/cardano-services/src/InMemoryCache/InMemoryCache.ts
IvayloAndonov: [52a24a2](https://github.com/input-output-hk/cardano-js-sdk/pull/286/commits/52a24a29a6bc040355e2da5a676a9096e50fe987)

mkazlauskas (approved):
Looks good! The only improvement that comes to mind is a test that `.catch` resets cache, but I think it's good enough as-is.

Also, `perf` seems like a good commit type for this

IvayloAndonov:
> Looks good! The only improvement that comes to mind is a test that `.catch` resets cache, but I think it's good enough as-is.
> 
> Also, `perf` seems like a better commit type for this

Yeah, I'll squash the commits as `perf` once I've got another approval and its ready for merge

rhyslbw (dismissed):

mkazlauskas (approved):

PR #288, opened 2022-06-16
test(cardano-services): update test case for rewardsHttpProvider

jvieiro:
# Context

Test case “RewardsHttpService healthy state with rewardsHttpProvider rewardsHistory returns address rewards history with epochs 1” has an empty expected result.

# Proposed Solution

Change epochs filters and address of the request in order to expect a non empty response.
 
N/A

### DIFF
--- a/packages/cardano-services/test/Reward/RewardsHttpService.test.ts
+++ b/packages/cardano-services/test/Reward/RewardsHttpService.test.ts
@@ -174,12 +174,15 @@ describe('RewardsHttpService', () => {
           expect(response).toMatchSnapshot();
         });
         it('returns address rewards history with epochs ', async () => {
+          const accountWithRewardsAtEpoch76 = Cardano.RewardAccount(
+            'stake_test1up32f2hrv5ytqk8ad6e4apss5zrrjjlrkjhrksypn5g08fqrqf9gr'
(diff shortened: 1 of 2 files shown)


packages/cardano-services/test/Reward/RewardsHttpService.test.ts
```diff
@@ -174,12 +174,15 @@ describe('RewardsHttpService', () => {
...
+            'stake_test1up32f2hrv5ytqk8ad6e4apss5zrrjjlrkjhrksypn5g08fqrqf9gr'
+          );
           const response = await provider.rewardsHistory({
             epochs: {
-              lowerBound: 10,
-              upperBound: 11
+              lowerBound: 75,
+              upperBound: 76
             },
-            rewardAccounts: [rewardAcc]
```
mkazlauskas, line 182: Looks good, but would be great to rename at least one of `rewardAcc` or `rewardAccount` to communicate the intent/difference between the variables

rhyslbw: What's the issue?

mkazlauskas (approved):

rhyslbw (approved):

PR #289, opened 2022-06-16
test: typecheck e2e and hw tests

mkazlauskas:
# Context

e2e and hw tests are not being typechecked in CI

# Proposed Solution

add `test:build:verify` script to run `tsc` on `test/`


packages/blockfrost/test/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "../../../tsconfig.json",
+  "extends": "../../../test/tsconfig.json",
```
rhyslbw, line 2: :eyes:

packages/blockfrost/package.json
```diff
@@ -28,6 +28,7 @@
     "lint": "eslint --ignore-path ../../.eslintignore \"**/*.ts\"",
     "lint:fix": "eslint --fix --ignore-path ../../.eslintignore \"**/*.ts\"",
     "test": "jest -c ./jest.config.js",
+    "test:build:verify": "tsc --build ./test",
```
rhyslbw, line 31: This will produce compiled test output right? A cleanup will be needed if we can't use `--noEmit`, as we can then run this script locally prior to pushing to the remote.

rhyslbw (commented):
Yikes, that's a lot of misconfigured test `.tsconfig` inheritances :sweat_smile:

packages/blockfrost/package.json
mkazlauskas: No, I added ["noEmit": true](https://github.com/input-output-hk/cardano-js-sdk/pull/289/files#diff-c0d527bf7b08c376a02d793f515e4ae3ae5cdda654c6e66a75f8a99927fd23a8R4) in test tsconfig. Project references [don't build dependencies](https://www.typescriptlang.org/docs/handbook/project-references.html#caveats-for-project-references) without --build

packages/blockfrost/test/tsconfig.json
mkazlauskas: `"../../../test/tsconfig.json"` is a new one, so it's really a new configuration

packages/blockfrost/package.json
mkazlauskas: `-p` should actually work too as we're running `yarn build`, I was getting a compilation error due to having a reference to `../../cip2/test` which wasn't built before I replaced such references with `include`. Would you like me to change `--build` with `-p` and get rid of the new `/test/tsconfig.json`?

mkazlauskas: An argument for `"noEmit"` in tsconfig.json is that we are less likely to ever run into trashing /test/ with build artifacts

rhyslbw: > as we're running yarn build

In the [workflow](https://github.com/input-output-hk/cardano-js-sdk/blob/a4c048d6ffea6d79524eb5b708399b3523e66465/.github/workflows/continuous-integration.yaml#L36) ? It would be best to have this verification fully contained

rhyslbw (dismissed):
Thanks @mkazlauskas :rocket:

rhyslbw (approved):

PR #290, opened 2022-06-16
fix/rabbitmq-wrong-interface

iccicci:
# Context
The package `@cardano-sdk/rabbitmq` doesn't respect all the `cardano-js-sdk` standards and has a severely wrong error handling, in details:

- the `Promise` returned by `TxSubmitWorker.start` should resolve at the end of the init stage rather than when the `TxSubmitWorker` exits: ADP-1886
- the `TxSubmitWorker` should retry only those errors which can be retried rather than **all the errors**: ADP-1874
- the `TxSubmitWorker` should let know the `RabbitMqTxSubmitProvider` when a transaction is in mempool (or in case of error, propagate it) in order to let the `Promise` returned by `RabbitMqTxSubmitProvider.submitTx` to resolve when the transaction is in mempool or reject with the original error: ADP-1823

# Proposed Solution
The package `@cardano-sdk/rabbitmq` have been completely refactored.

- The required changes to implement the new `@cardano-sdk/rabbitmq` interface was applied to `@cardano-sdk/cardano-services`.
- The test mock in `@cardano-sdk/ogmios` was enriched to respond with an error which can be retried.
- Now all the time measured by the `@cardano-sdk/cardano-services` load test are meaningful.


packages/rabbitmq/src/TxSubmitWorker.ts
```diff
@@ -254,29 +258,80 @@ export class TxSubmitWorker {
...
+      this.#dependencies.logger!.info(`${moduleName}: submitting tx #${counter} id: ${txId}`);
+      this.#dependencies.logger!.debug(`${moduleName}: tx #${counter} dump:`, content.toString('hex'));
+      await this.#dependencies.txSubmitProvider.submitTx(txBody);
 
-      this.#dependencies.logger!.debug(`${moduleName}: ACKing RabbitMQ message ${counter}`);
+      this.#dependencies.logger!.debug(`${moduleName}: ACKing RabbitMQ message #${counter}`);
       this.#channel?.ack(message);
     } catch (error) {
+      errorDescription = await this.submitTxErrorHandler(error, counter, message);
+    } finally {
```
mkazlauskas, line 283: Since it's checking for `!errorDescription` it may be cleaner to just put everything from `finally` block into `catch`

packages/rabbitmq/src/errorsSerialization.ts
```diff
@@ -0,0 +1,49 @@
...
+  unknown: (details) => new Error(details.message),
+  ...Object.fromEntries(
+    TxSubmissionErrorsEntries.map(([key]) => [
+      key,
+      (details: ErrorDetails) => new Cardano.TxSubmissionErrors[key](JSON.parse(details.message!))
+    ])
+  )
+};
+
+export interface ErrorDescription extends ErrorDetails {
```
mkazlauskas, line 23: Could use [serialization utils](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/util/src/serializableObject.ts#L26) from `util` package.

Don't think you need to serialize `type` separately as the serialized error object contains it (`name`)

You could then deserialize into correct `instanceof` as it has [an option](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/util/src/serializableObject.ts#L19) for custom error prototype. Correct prototype is required for `isTxSubmissionError` to work and is lost with `JSON.stringify(err)`

packages/rabbitmq/src/rabbitmqTxSubmitProvider.ts
```diff
@@ -1,27 +1,62 @@
...
+const moduleName = 'RabbitMqTxSubmitProvider';
+
+/**
+ * Configuration options parameters for the RabbitMqTxSubmitProvider
+ */
+export interface RabbitMqTxSubmitProviderConfig {
+  /**
+   * Use the algorithm to get a dummy Tx Id from a Tx body. Used for tests
+   */
+  dummyTxId?: boolean;
```
mkazlauskas, line 17: Is this really necessary? Could just build an actual tx :thinking:

packages/rabbitmq/src/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const txBodyToId = (txBody: Buffer | Uint8Array, dummy?: boolean) => {
+  if (dummy) {
+    const buffer = txBody instanceof Buffer ? txBody : Buffer.from(txBody);
+
+    return buffer.toString('hex');
+  }
+
+  const buffer = txBody instanceof Buffer ? txBody : Buffer.from(txBody);
+  const txDecoded = CSL.Transaction.from_bytes(buffer);
```
mkazlauskas, line 13: I'm wondering if we should wrap this under a new util that would convert `Cardano.util.HexBlob` into `Cardano.NewTxAlonzo`. I think we should aim to not use CSL directly. What do you think @rhyslbw

mkazlauskas (changes requested):

packages/rabbitmq/src/TxSubmitWorker.ts
iccicci: The code piece inside the `finally` block needs to run also in case of success: can't be simply moved inside the catch `block`; we could find some other approach...
The response must be sent both in case of success and of not retriable error, the content of the result message should change depending on the success/error case.

packages/rabbitmq/src/errorsSerialization.ts
iccicci: I learned only now the existence of this util; thank you.
I'll check it.

packages/rabbitmq/src/TxSubmitWorker.ts
mkazlauskas: Oh I see, ok thanks

packages/rabbitmq/src/errorsSerialization.ts
mkazlauskas: You could use [this snippet](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services-client/src/TxSubmitProvider/txSubmitHttpProvider.ts#L12-L17) for `getErrorPrototype?: GetErrorPrototype;` in the serialization util, just instead of setting error prototype return it

packages/rabbitmq/src/rabbitmqTxSubmitProvider.ts
iccicci: No, it is not strictly required; I evaluated it as useful.
There are lots of tests already using fake transactions, without we should rework them all; for new tests without it we are forced to always use a real tx.
From my point of view it's just a choice, but my experience in the team is not enough to take this decision.
Please explicitly ask me to remove it, and I'll remove it.

packages/rabbitmq/src/errorsSerialization.ts
iccicci: Could you please check this again, @mkazlauskas ?
Thank you

packages/rabbitmq/src/utils.ts
```diff
@@ -0,0 +1,128 @@
...
+  if (innerError) addPrototypeToError(innerError);
+};
+
+/**
+ * Deserializes a serialized error
+ *
+ * @param serializedError the error to deserialize
+ */
+export const deserializeError = (serializedError: SerializedError): unknown => {
+  const ret = fromSerializableObject(serializedError) as DeserializedError;
```
mkazlauskas, line 52: Also, you can use the existing `getErrorPrototype` option instead of setting it manually:
```suggestion
  const getErrorPrototype = (err: unknown) => {
    if (typeof error === 'object' && typeof error?.name === 'string' && typeof error?.message === 'string') {
      const rawError = error as Cardano.TxSubmissionError;
      const txSubmissionErrorName = rawError.name as keyof typeof Cardano.TxSubmissionErrors;
      const ErrorClass = Cardano.TxSubmissionErrors[txSubmissionErrorName];
      if (ErrorClass) return ErrorClass.prototype;
    }
    return Error.prototype;
  }
  return fromSerializableObject(serializedError, {getErrorPrototype})
```

packages/rabbitmq/src/utils.ts
```diff
@@ -0,0 +1,128 @@
+import { CSL, Cardano, cslToCore } from '@cardano-sdk/core';
+import { OutsideOfValidityInterval } from '@cardano-ogmios/schema';
+import { fromSerializableObject, toSerializableObject } from '@cardano-sdk/util';
+
+export interface DeserializedError {
```
mkazlauskas, line 5: Wouldn't Cardano.TxSubmissionError work as a type instead of this?

packages/rabbitmq/src/utils.ts
```diff
@@ -0,0 +1,128 @@
...
+  innerError?: {
+    name: string;
+    message: string;
+  };
+  name: string;
+  message: string;
+}
+
+export interface SerializedError {
```
mkazlauskas, line 14: Is this type useful anywhere? If not, using `unknown` in place of it would work.

packages/rabbitmq/src/utils.ts
```diff
@@ -0,0 +1,128 @@
...
+      serializedError = { value: { message: JSON.stringify(err), name: 'Error' } };
+    } catch {
+      serializedError = { value: { message: 'Unknown', name: 'Error' } };
+    }
+  };
+
+  try {
+    serializedError = toSerializableObject(err) as SerializedError;
+  } catch {
+    lastChance();
```
mkazlauskas, line 79: That might be an overkill as it should never throw.

Either way you should probably return a UnknownTxSubmissionError in `lastChance` (which is one of the options in `Cardano.TxSubmissionError`) here, so we can assure that any error returned by this service would be a `TxSubmissionError`

mkazlauskas (changes requested):

packages/rabbitmq/src/errorsSerialization.ts
mkazlauskas: Posted another review, let's continue this conversation in new comment threads

packages/rabbitmq/src/utils.ts
iccicci: I tried but I wasn't able to use it, now I think I understood how the option works; I can try again to use it.
Thank you for the example.

packages/rabbitmq/src/utils.ts
iccicci: Actually, once applied [this change](https://github.com/input-output-hk/cardano-js-sdk/pull/290#discussion_r899802463), yes, it can be discarded.

packages/rabbitmq/src/utils.ts
iccicci: Its usefulness is to avoid using lots of `as` operators while accessing the attributes of a serializable error.
Looking at it once again, since the error is not yet serialized, but in the special form which can be serialized it's better to rename the type in `SerializableError`...
Once applied all the other changes requested in this review, I'll check how many time I'm going to use to decide if to keep it or remove it

mkazlauskas: Do you need to access attributes of serialized error anywhere? Why not just deserialize into a proper error type before doing that?

iccicci: Probably once applied all the changes, only inside the `getErrorPrototype` function

iccicci: Ah yes, I need to access attributes in order to check if the error is retriable or not

iccicci: Or maybe not... I could perform the check against the original error...

packages/rabbitmq/src/utils.ts
iccicci: removed

packages/rabbitmq/src/utils.ts
iccicci: A part from some minor changes since `error?.name` do not compile since `error: unknown`, it works, thank you

packages/rabbitmq/src/utils.ts
mkazlauskas: Right, it's probably best to serialize it just before sending

packages/rabbitmq/src/rabbitmqTxSubmitProvider.ts
rhyslbw: Agree with @mkazlauskas. It would also be a lot better not to have "dummy" things outside of the test code itself

packages/rabbitmq/src/utils.ts
rhyslbw: Agree. A util here makes sense

mkazlauskas: `Cardano.util.deserializeTx`?

packages/rabbitmq/src/rabbitmqTxSubmitProvider.ts
iccicci: removed

rhyslbw (changes requested):
This is looking very nice @iccicci :rocket: 
Just a few minor issues

packages/rabbitmq/src/TxSubmitWorker.ts
```diff
@@ -120,117 +129,107 @@ export class TxSubmitWorker {
...
-          await this.stop();
-          reject(err);
-        }
-      };
-
-      try {
-        this.#dependencies.logger!.info(`${moduleName} init: checking tx submission provider health status`);
+  async start() {
+    try {
+      this.#dependencies.logger!.info(`${moduleName} init: checking tx submission provider health status`);
```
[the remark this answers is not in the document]
mkazlauskas: Using `dummyLogger` as a default already got us in trouble once (pouchdb errors were silently swallowed). I suggest to make it a required dependency in any new code and only use dummyLogger in tests.

We could also create a ticket to refactor existing code to follow this pattern.

packages/cardano-services/test/entrypoints.txWorker.test.ts
```diff
@@ -46,54 +46,40 @@ describe('tx-worker entrypoints', () => {
     await listenPromise(ogmiosServer, { port });
     await ogmiosServerReady(ogmiosConnection);
     commonArgs = ['start-worker', '--logger-min-severity', 'error', '--ogmios-url', ogmiosConnection.address.webSocket];
-    commonEnv = { LOGGER_MIN_SEVERITY: 'error', OGMIOS_URL: ogmiosConnection.address.webSocket };
+    commonEnv = { LOGGER_MIN_SEVERITY: 'debug', OGMIOS_URL: ogmiosConnection.address.webSocket };
```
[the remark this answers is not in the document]
mkazlauskas: Agree, definitely not `debug` as default. Perhaps `warn` or `info` would be better?

packages/rabbitmq/src/rabbitmqTxSubmitProvider.ts
mkazlauskas: A commit hash in a comment would help reviewing requested changes @iccicci

packages/rabbitmq/src/TxSubmitWorker.ts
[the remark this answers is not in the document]
rhyslbw: ADP-1933

packages/cardano-services/test/load/load.test.ts
```diff
@@ -244,7 +245,16 @@ describe('load', () => {
           const coins = 1_000_000n + 1000n * BigInt(i);
           const tx = await wallet.initializeTx({ outputs: new Set([{ address, value: { coins } }]) });
 
-          submitPromises.push(wallet.submitTx(await wallet.finalizeTx(tx)));
+          submitPromises.push(
+            // eslint-disable-next-line no-loop-func
+            (async () => {
+              try {
```
mkazlauskas, line 251: Is this to be able to see the actual error? If so, it's a [jest bug](https://github.com/facebook/jest/issues/11617) that will hopefully be fixed one day, and we shouldn't be adapting our tests to it. There's a workaround: `node scripts/patch-jest-worker.js`

mkazlauskas (changes requested):

packages/cardano-services/test/entrypoints.txWorker.test.ts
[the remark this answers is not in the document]
rhyslbw: Yeah, `info` is probably the best option here. If it's too noisy for tests, it's too noisy for that level.

packages/cardano-services/test/load/load.test.ts
iccicci: I'm sorry but I have not clear about which piece of code you are referring. Another thing I've not clear is that the workaround you are proposing seems to be about `child_process` related stuff while the `Promise`s in `submitPromises` are about asynchronous tasks not working with `child_process`.

Anyway I added that `try/catch block` when I faced the problem related to value is conserved error we are facing on another thread.
It has two benefits (when a transaction submission unexpectedly fails):

1. the tests do not hangs
2. we can have the evidence of reason of the error

mkazlauskas: try/catch just seems odd there. How about: `await expect(promise).resolves.not.toThrow()`?

We can rely on test runner reports for failed expectations

packages/cardano-services/test/entrypoints.txWorker.test.ts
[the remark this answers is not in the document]
iccicci: I probably did this change to investigate some problem I had during the development.
I simply rollback this change as it is not required.

packages/rabbitmq/test/rabbitmqTxSubmitProvider.test.ts
```diff
@@ -24,29 +38,37 @@ describe('RabbitMqTxSubmitProvider', () => {
...
+      try {
+        const txs = await txsPromise;
+        provider = new RabbitMqTxSubmitProvider({ rabbitmqUrl }, { logger });
+        const resA = await provider.submitTx(txs[0].txBodyUint8Array);
+        // Called again to cover the idemopotent RabbitMqTxSubmitProvider.#ensureQueue() operation
+        const resB = await provider.submitTx(txs[1].txBodyUint8Array);
+        expect(resA).toBeUndefined();
+        expect(resB).toBeUndefined();
+      } catch (error) {
+        expect((error as { innerError: unknown }).innerError).toBeInstanceOf(ProviderError);
```
mkazlauskas, line 53:
```suggestion
        expect((error as ProviderError).innerError).toBeInstanceOf(ProviderError);
```

mkazlauskas (approved):
Nice work @iccicci!

@rhyslbw do we want a QA review on this?

rhyslbw (changes requested):
Fantastic work @iccicci :tada:
Please _autosquash_ and request final review from @ssong-iohk

PR #291, opened 2022-06-16
chore: fix lint-staged tsconfig

mkazlauskas:
# Context

`lint-staged` fails with `Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser`

# Proposed Solution

- consolidate .eslintrc.js into one 184c84982714e17b5a928633f9ec83ce3a22460b

- speed up lint 
  - update monorepo root lint script to run eslint once instead of for each package 184c84982714e17b5a928633f9ec83ce3a22460b
  - disable 'no-floating-promises' in a main `.eslintrc.js` so that we no longer need a fully-fledged tsc for linting 2ee69b29d343e74c1387b964f059c24ec4113a81
  - add `complete.eslintrc.js` that is used for `yarn lint` and has `no-floating-promises` enabled [ca6b32f](https://github.com/input-output-hk/cardano-js-sdk/pull/291/commits/ca6b32fca5976af290e0b45c6df11122603f1e78)
  - bump 'lint-staged' and update it's config to **only** lint changed files 2ee69b29d343e74c1387b964f059c24ec4113a81


rhyslbw (commented):
If we can run extra rules on-demand, outside of the pre-commit hook, then it would be the best of both worlds. The performance gain is impressive but losing `no-floating-promises` gives me a reason to pause since this rule does help to avoid bugs.

mkazlauskas:
Verified no-floating-promises fails CI: https://github.com/input-output-hk/cardano-js-sdk/runs/6931064808?check_suite_focus=true

rhyslbw (approved):
Great work @mkazlauskas :muscle:

mirceahasegan (approved):

PR #297, opened 2022-06-22
feat: dbsync nft metadata service

mkazlauskas:
# Context

Need to get on-chain NFT metadata as part of db-sync AssetProvider.

# Proposed Solution

- Refactor ChainHistory provider to be able to reuse metadata query in other providers c001b98ac6c851e0ab6c93b6e8963f153f08b570 
- Implement NFT metadata service according to [CIP25](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0025) rules and map it to core type 8e8bfbe20a12f35303b6f1e7b10dc8228e306b79

- chore(cardano-services): rm duplicate dependency 91365240f593e62a665d14417bcea2653e816f28
- feat(core): accept common ipfs hash without protocol for nft metadata uri a1fa80c528033893d8155da7e0cb13431b11297e
  - our test NFT was minted with IPFS hash in `image` field (without `scheme://`). It doesn't seem to be within CIP25 spec, but it is probably a common mistake, so we can just interpret values that look like common IPFS hashes


AngelCastilloB (dismissed):

packages/cardano-services/src/Asset/getNftMetadata.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export interface GetNftMetadataDependencies {
+  metadataService: MetadataService;
+  db: Pool;
+  logger: Logger;
+}
+
+export type AssetPolicyIdAndName = Pick<Asset.AssetInfo, 'name' | 'policyId'>;
+
+export const getNftMetadata = async (
```
mkazlauskas, line 15: Perhaps it would be better if there was a `NftMetadataService` interface + `DbSyncNftMetadata` implementation instead of this?

packages/cardano-services/src/Metadata/types.ts
```diff
@@ -0,0 +1,11 @@
+import { Cardano } from '@cardano-sdk/core';
+
+export interface MetadataService {
```
AngelCastilloB, line 3: I think it would be great to have a bit more documentation for the public interface of the modules. for example, what purpose/functionality a metadata service provides?

AngelCastilloB (commented):
The PR looks great, my only comment would be to add more documentation to the code, I believe is better to add it/improve it as we write the code, otherwise doing it retroactively becomes a titanic task.

packages/cardano-services/src/Metadata/types.ts
mkazlauskas: Added some doc comments in [9271def](https://github.com/input-output-hk/cardano-js-sdk/pull/297/commits/9271def01fc6cef2f05c20636733d1bafb82b37c)

rhyslbw (dismissed):
Nice work @mkazlauskas, please squash ready for SDET review

iadmytro (changes requested):

packages/cardano-services/test/Metadata/mappers.test.ts
```diff
@@ -0,0 +1,27 @@
+import { Cardano } from '@cardano-sdk/core';
+import { TxMetadataModel, mapTxMetadata } from '../../src/Metadata';
+
+describe('mapTxMetadata', () => {
+  test('map TxMetadataModel to Cardano.TxMetadata', () => {
```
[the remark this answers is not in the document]
mkazlauskas: Added some tests in [90f4076](https://github.com/input-output-hk/cardano-js-sdk/pull/297/commits/90f40765a462d9a2eec9335ecb57f4ba9c70d3d8). Note that this is not a new function, I just moved it over to a new module.

iadmytro (dismissed):

rhyslbw (approved):

rhyslbw (approved):

PR #298, opened 2022-06-22
feature/ADP-1901 - Ogmios time settings

lgobbi-atix:
# Context

Replace the [hard-coded network time settings](https://github.com/input-output-hk/cardano-js-sdk/blob/80c7be0764e7616091a94d8264485ffd89223087/packages/core/src/util/slotCalc.ts#L37-L49) with data fetched and cached, by extending the [InMemoryCache](https://github.com/input-output-hk/cardano-js-sdk/tree/master/packages/cardano-services/src/InMemoryCache) instance, from the node, via Ogmios' [StateQuery.eraSummaries()](https://github.com/CardanoSolutions/ogmios/blob/8b08ce856c5dda12f7918cd21718a8da5fc4904f/clients/TypeScript/packages/client/test/StateQuery/StateQuery.test.ts#L290-L300).

# Proposed Solution
- Refactor `InMemoryCache` class to a more generic value resolver
- Add ogmios provider to get time settings from `eraSummaries` query
- Remove hard coded time settings from `DbSyncNetworkInfo` and get it from ogmios provider 

- `InMemoryCache` parameters and related env variables renamed
- Re-structured `ogmios` package folders similar to `cardano-services` and `cardano-services-client` 
- Added `ogmiosTimeSettingsProvider.ts`
- `NetworkInfoHttpService` is now ogmios-dependent


rhyslbw:
I've rebased on #304 , so this must be merged first.

packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
```diff
@@ -14,52 +14,55 @@ import { NetworkInfoCacheKey } from '.';
...
 export interface NetworkInfoProviderProps {
   cardanoNodeConfigPath: string;
-  dbPollInterval: number;
+  pollInterval: number;
 }
 export interface NetworkInfoProviderDependencies {
   db: Pool;
   cache: InMemoryCache;
   logger?: Logger;
+  timeSettingsProvider: TimeSettingsProvider;
```
rhyslbw, line 27:
```suggestion
  stateQuery: StateQuery.StateQueryClient;
```

rhyslbw (changes requested):
I think it's better to depend on Ogmios directly with translation util to handle the format variance, rather than define a new provider, as this information is intended to be consumed via the existing `NetworkInfoProvider` only. It would be the first provider without an HTTP service, which is a hint IMO. Alternatively, we could add mini-protocol interfaces for the Ogmios client to then implement, but I'm not seeing the value given we consider Ogmios our native interface for state queries and chain-sync, and we already have a hard dependency on Ogmios within the core types.

We will also consider renaming `TimeSettings` to `EraSummaries` and remodel to align with the domain terminology and response shape, but that will be best as a separate refactor. 

``` ts
interface EraSummary {
  parameters:  {
    epochLength: number;
    slotLength: number;
  };
  start: {
    slot: number;
    time: Date;
  };
}
```

By knowing this it's easier to see why the ceremony with introducing a new provider is not required, and the translation util is there to just hydrate the date.

packages/core/src/Cardano/types/TxSubmissionErrors.ts
```diff
@@ -8,69 +8,70 @@ export class UnknownTxSubmissionError extends CustomError {
...
-  UnknownOrIncompleteWithdrawalsError: TxSubmission.errors.UnknownOrIncompleteWithdrawals.Error,
-  UnspendableDatumsError: TxSubmission.errors.UnspendableDatums.Error,
-  UnspendableScriptInputsError: TxSubmission.errors.UnspendableScriptInputs.Error,
-  UpdateWrongEpochError: TxSubmission.errors.UpdateWrongEpoch.Error,
-  ValidationTagMismatchError: TxSubmission.errors.ValidationTagMismatch.Error,
-  ValueNotConservedError: TxSubmission.errors.ValueNotConserved.Error,
-  WrongCertificateTypeError: TxSubmission.errors.WrongCertificateType.Error,
-  WrongPoolCertificateError: TxSubmission.errors.WrongPoolCertificate.Error,
-  WrongRetirementEpochError: TxSubmission.errors.WrongRetirementEpoch.Error
+  AddressAttributesTooLargeError: TxSubmission.submissionErrors.errors.AddressAttributesTooLarge.Error,
```
mkazlauskas, line 11: I think that would be a little cleaner:
```ts
const ogmiosErrors = TxSubmission.submissionErrors.errors;
```
```suggestion
  AddressAttributesTooLargeError: ogmiosErrors.AddressAttributesTooLarge.Error,
```

mkazlauskas (dismissed):
Looks good, just 1 optional minor suggestion

packages/core/src/Cardano/types/TxSubmissionErrors.ts
mkazlauskas: Posted this in wrong PR :face_with_head_bandage:

packages/cardano-services/src/InMemoryCache/InMemoryCache.ts
```diff
@@ -2,7 +2,7 @@ import { CACHE_TTL_DEFAULT } from './defaults';
 import NodeCache from 'node-cache';
 
 export type Key = string | number;
-export type DbSyncQuery<T> = () => Promise<T>;
+export type Action<T> = () => Promise<T>;
```
mkazlauskas, line 5: Could name it a little bit more specific
```suggestion
export type AsyncAction<T> = () => Promise<T>;
```

packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/utils.ts
```diff
@@ -1,19 +1,23 @@
-import { DbSyncQuery, InMemoryCache, UNLIMITED_CACHE_TTL } from '../../InMemoryCache';
+import { Action, InMemoryCache, UNLIMITED_CACHE_TTL } from '../../InMemoryCache';
 import { NetworkInfoCacheKey } from '.';
 import { Shutdown } from '@cardano-sdk/util';
 
-export const DB_POLL_INTERVAL_DEFAULT = 10_000;
+export const POLL_INTERVAL_DEFAULT = 10_000;
 
-export const pollDbSync = (cache: InMemoryCache, dbSyncQuery: DbSyncQuery<number>, interval: number): Shutdown => {
+export const pollService = (cache: InMemoryCache, action: Action<number>, interval: number): Shutdown => {
```
mkazlauskas, line 7:
```suggestion
export const epochPollService = (cache: InMemoryCache, action: Action<Cardano.Epoch>, interval: number): Shutdown => {
```

mkazlauskas (commented):
Just a few suggestions for names

packages/core/src/Provider/TimeSettingsProvider/types.ts
```diff
@@ -0,0 +1,14 @@
+import { TimeSettings } from '../../util';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import { Provider, ProviderError } from '../..';
+
+export interface TimeSettingsProvider extends Provider {
```
rhyslbw, line 6:
```suggestion
export interface CardanoNode {
```
+ move to `core/src/Cardano/types/CardanoNode.ts` 
+ Host and rename method to `StateQuery: { eraSummaries: () => Promise<EraSummary[]> }`

rhyslbw (changes requested):

packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
```diff
@@ -14,52 +14,55 @@ import { NetworkInfoCacheKey } from '.';
...
-  #dbPollInterval: number;
+  #pollInterval: number;
   #pollService: Shutdown | null;
+  #timeSettingsProvider: TimeSettingsProvider;
 
   constructor(
-    { cardanoNodeConfigPath, dbPollInterval }: NetworkInfoProviderProps,
-    { db, cache, logger = dummyLogger }: NetworkInfoProviderDependencies
+    { cardanoNodeConfigPath, pollInterval }: NetworkInfoProviderProps,
+    { db, cache, logger = dummyLogger, timeSettingsProvider }: NetworkInfoProviderDependencies
```
IvayloAndonov, line 40:
```suggestion
    { db, cache, logger = dummyLogger, stateQuery }: NetworkInfoProviderDependencies
```

packages/cardano-services/src/Program/loadHttpServer.ts
```diff
@@ -13,14 +13,14 @@ import { ProgramOptionDescriptions } from './ProgramOptionDescriptions';
...
+import { ogmiosTimeSettingsProvider, ogmiosTxSubmitProvider, urlToConnectionConfig } from '@cardano-sdk/ogmios';
 import Logger, { createLogger } from 'bunyan';
 import pg from 'pg';
 
 export interface HttpServerOptions extends CommonProgramOptions {
   dbConnectionString?: string;
-  dbQueriesCacheTtl: number;
-  dbPollInterval: number;
+  cacheTtl: number;
+  pollInterval: number;
```
IvayloAndonov, line 23:
```suggestion
  epochPollInterval: number;
```
?

packages/ogmios/test/mocks/mockOgmiosServer.ts
```diff
@@ -42,17 +42,72 @@ export interface MockOgmiosServerConfig {
...
-  submitTx: {
+  submitTx?: {
     response: {
       success: boolean;
       failWith?: {
         type: 'eraMismatch';
       };
     };
   };
+  timeSettings?: {
```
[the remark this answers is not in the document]
lgobbi-atix: I am a bit confused about what you mentioned in the other comment 
> We will also consider renaming `TimeSettings` to `EraSummaries` and remodel to align with the domain terminology and response shape

Are we getting rid of the old `TimeSettings` and now the network info would return `EraSummaries` instead? Or that was just for the `CardanoNode` implementation and network info should still map `EraSummaries` to `TimeSettings`?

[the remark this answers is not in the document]
rhyslbw: Yeah sorry @lgobbi-atix, this is a bit confusing.

> Are we getting rid of the old TimeSettings and now the network info would return EraSummaries instead?

We will, but it's not necessary to do that refactor here given changing the `NetworkInfoProvider` interface is out of scope. It's the same concept with a slight improvement (start `Date` vs relative timestamp), we just named it independently, so this is an alignment with the domain term for the new code being implemented.

> Or that was just for the CardanoNode implementation and network info should still map EraSummaries to TimeSettings?

Correct, for now.  

> If we are still mapping EraSummaries to TimeSettings I would also need to add StateQuery.systemStart to CardanoNode to calculate the dates

Yes, which you can call internally within `StateQuery.eraSummaries` to return what I've defined [here](https://github.com/input-output-hk/cardano-js-sdk/pull/298#pullrequestreview-1021022252) instead of a type that is practically a pointer to the Ogmios type.

packages/cardano-services/src/InMemoryCache/InMemoryCache.ts
lgobbi-atix: Done in 563e7b30ca19d1de3e518e4fc451137b91d658ca

packages/core/src/Provider/TimeSettingsProvider/types.ts
lgobbi-atix: Done in d2c0156242b9f6ed7591cee0c0acf75ea940e35e

packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
lgobbi-atix: Done in 563e7b30ca19d1de3e518e4fc451137b91d658ca

packages/cardano-services/src/Program/loadHttpServer.ts
lgobbi-atix: Done in 563e7b30ca19d1de3e518e4fc451137b91d658ca

packages/core/src/errors.ts
```diff
@@ -48,3 +48,14 @@ export class InvalidStringError extends CustomError {
     super(`Invalid string: "${expectation}"`);
   }
 }
+
+export enum CardanoNodeFailure {
```
rhyslbw, line 52: See https://github.com/input-output-hk/cardano-js-sdk/blob/0.3.0/packages/core/src/Cardano/types/TxSubmissionErrors.ts for reference of the existing pattern used for inheriting the Ogmios errors.

These are the list of errors we'll need to include
https://github.com/CardanoSolutions/ogmios/tree/master/clients/TypeScript/packages/client/src/errors

packages/ogmios/src/CardanoNode/cardanoNode.ts
```diff
@@ -0,0 +1,49 @@
...
+        logger.error.bind(logger)({ error: error.name, module: `cardanoNode.StateQuery.${methodName}` }, error.message);
+      },
+      logger.info.bind(logger),
+      { connection: connectionConfig, interactionType: 'LongRunning' }
+    );
+
+  return {
+    StateQuery: {
+      eraSummaries: async () => {
+        const interactionContext = await getInteractionContext('eraSummaries');
```
rhyslbw, line 33: Better to share a single interaction context across all queries, so best to convert this to a class and follow https://github.com/input-output-hk/cardano-graphql/blob/master/packages/api-cardano-db-hasura/src/CardanoNodeClient.ts for inspiration. There's no need to introduce the retry logic here, as we're expecting the process manager to only startup the services when Ogmios is ready.

packages/ogmios/src/CardanoNode/cardanoNode.ts
```diff
@@ -0,0 +1,49 @@
...
+        const stateQueryClient = await createStateQueryClient(interactionContext);
+        try {
+          const systemStart = await stateQueryClient.systemStart();
```
rhyslbw, line 36:
```suggestion
          const systemStart = await stateQueryClient.systemStart();
```
Since this is a `StateQuery`, it would be good to make it a public method too.

rhyslbw (changes requested):
Looking pretty good @lgobbi-atix, see inline feedback

packages/ogmios/src/CardanoNode/cardanoNode.ts
lgobbi-atix: Done in d43ad7b75626b7eacdfbf0e21296706e27acea08. I had to change `CardanoNode` interface as it is a class now though

packages/core/src/Cardano/types/CardanoNode.ts
```diff
@@ -0,0 +1,44 @@
...
+    epochLength: number;
+    slotLength: number;
+  };
+  start: {
+    slot: number;
+    time: Date;
+  };
+}
+
+export interface CardanoNode {
```
mkazlauskas, line 17: I think it would be better to put all these new CardanoNode-related types under `core/src/CardanoNode`, because:
- `core/src/Cardano` will most likely end up having just the types that map to new transaction CDDL
- imho using it as `CardanoNode` is nicer than `Cardano.CardanoNode`

packages/core/src/Cardano/types/CardanoNode.ts
```diff
@@ -0,0 +1,44 @@
...
+  /**
+   * Initialize CardanoNode instance
+   *
+   * @param {ConnectionConfig} connectionConfig Ogmios connection configuration
+   */
+  initialize: (connectionConfig: ConnectionConfig) => Promise<void>;
```
mkazlauskas, line 23: It would be better to not couple this interface with `@cardano-ogmios/client` - we have `packages/ogmios` for ogmios-specific implementations. Concrete implementation can have `connectionConfig: ConnectionConfig` injected via construcutor, and/or have additional method(s)

If we need to have life-cycle methods we could extend `Provider` because this interface is Provider-like and is used in a similar fashion within `cardano-services`.

packages/core/src/Cardano/types/CardanoNodeErrors.ts
```diff
@@ -0,0 +1,40 @@
...
+  QueryUnavailableInCurrentEraError,
+  ServerNotReady,
+  TipIsOriginError,
+  UnknownResultError,
+  WebSocketClosed
+};
+
+type CardanoNodeErrorName = keyof typeof CardanoNodeErrors;
+type CardanoNodeErrorClass = typeof CardanoNodeErrors[CardanoNodeErrorName];
+export type CardanoNodeError = InstanceType<CardanoNodeErrorClass> | UnknownCardanoNodeError;
```
mkazlauskas, line 40: Any reason not to do this to have a single type that we can use in doc comments and users can cast to in `catch`?

```suggestion
export type CardanoNodeError = InstanceType<CardanoNodeErrorClass> | UnknownCardanoNodeError | CardanoNodeNotInitializedError;
```

mkazlauskas (changes requested):

packages/core/src/Cardano/types/CardanoNode.ts
mkazlauskas: Looking good, let's just move `packages/core/src/Cardano/types/EraSummary.ts` as well.

packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
```diff
@@ -15,36 +15,47 @@ import { NetworkInfoBuilder } from './NetworkInfoBuilder';
...
-  dbPollInterval: number;
+  epochPollInterval: number;
 }
 export interface NetworkInfoProviderDependencies {
   db: Pool;
   cache: InMemoryCache;
   logger?: Logger;
+  cardanoNode: CardanoNode;
 }
 export class DbSyncNetworkInfoProvider extends DbSyncProvider implements NetworkInfoProvider {
```
mkazlauskas, line 39: The name of `DbSyncNetworkInfoProvider` is a little misleading since it's now also using CardanoNode. Don't have a good suggestion though.

packages/ogmios/src/CardanoNode/CardanoNode.ts
```diff
@@ -0,0 +1,76 @@
...
+import { mapEraSummary } from './mappers';
+
+type CardanoNodeState = 'initialized' | 'initializing' | null;
+
+/**
+ * Connects to the cardano node through ogmios
+ *
+ * @class CardanoNode
+ */
+export class CardanoNode implements ICardanoNode {
```
mkazlauskas, line 13:
```suggestion
export class OgmiosCardanoNode implements CardanoNode {
```

mkazlauskas (approved):
Just a few optional suggestions, nice work!

ssong-iohk (dismissed):
Currently failing tests in `entrypoints.test.ts` need to be fixed. Investigation is here: 

https://input-output-rnd.slack.com/archives/C02BW8GUNF7/p1657198361917709?thread_ts=1657183537.449479&cid=C02BW8GUNF7

AngelCastilloB (approved):
Looks good Rhys!

AngelCastilloB (dismissed):

AngelCastilloB (dismissed):

AngelCastilloB (approved):

IvayloAndonov (approved):
LGTM!

rhyslbw (approved):

PR #303, opened 2022-06-27
refactor: Remove HTTPService.create static methods and perform provider health checks during service initialization

IvayloAndonov:
# Context
We need to keep all services pure during initialization time (static `create` method). 
In order to do that, we need to remove real health checks inside `create` service method, more of that, the static `create` method becomes redundant and we can use `public constructor` directly for the service creation phase.

Provider's health checks should be performed during the service initialization phase and all services need to be aligned with that.

# Proposed Solution
See the changes introduced.

- make `HttpService` a concrete class, extending `RunnableModule`
- make all HTTP services constructable, remove the static `create` method from them
- perform provider health checks during service `initialize`
- hoist all repeatable methods from individual services to `HttpService` parent class
- rework `loadHttpServer` and `serviceMapFactory` as synchronous methods
- update relevant tests and mocks

### DIFF
--- a/packages/cardano-services/src/Http/HttpService.ts
+++ b/packages/cardano-services/src/Http/HttpService.ts
@@ -2,13 +2,15 @@ import { HealthCheckResponse, ProviderError, ProviderFailure } from '@cardano-sd
 import { HttpServer } from './HttpServer';
 import { Logger, dummyLogger } from 'ts-log';
 import { ProviderHandler } from '../util';
+import { RunnableModule } from '../RunnableModule';
 import express from 'express';
 
-export abstract class HttpService {
+export abstract class HttpService extends RunnableModule {
   public router: express.Router;
   public slug: string;
 
   protected constructor(slug: string, router: express.Router, logger = dummyLogger) {
+    super(slug, logger);
     this.router = router;
     this.slug = slug;
     const healthHandler = async (req: express.Request, res: express.Response) => {
@@ -27,14 +29,15 @@ export abstract class HttpService {
     this.router.post('/health', healthHandler);
   }
 
-  async start(): Promise<void> {
+  async startImpl(): Promise<void> {
     return Promise.resolve();
   }
 
-  async close(): Promise<void> {
+  async shutdownImpl(): Promise<void> {
     return Promise.resolve();
   }
 
+  protected abstract initializeImpl(): Promise<void>;
   protected abstract healthCheck(): Promise<HealthCheckResponse>;
 
   static routeHandler(logger: Logger): ProviderHandler {
--- a/packages/cardano-services/src/Http/HttpService.ts
+++ b/packages/cardano-services/src/Http/HttpService.ts
@@ -5,7 +5,7 @@ import { ProviderHandler } from '../util';
 import { RunnableModule } from '../RunnableModule';
 import express from 'express';
 
-export class HttpService extends RunnableModule {
+export abstract class HttpService extends RunnableModule {
   public router: express.Router;
   public slug: string;
   public provider: Provider;
--- a/packages/cardano-services/src/Http/HttpService.ts
+++ b/packages/cardano-services/src/Http/HttpService.ts
@@ -1,18 +1,21 @@
-import { HealthCheckResponse, ProviderError, ProviderFailure } from '@cardano-sdk/core';
+import { HealthCheckResponse, Provider, ProviderError, ProviderFailure } from '@cardano-sdk/core';
 import { HttpServer } from './HttpServer';
 import { Logger, dummyLogger } from 'ts-log';
 import { ProviderHandler } from '../util';
 import { RunnableModule } from '../RunnableModule';
 import express from 'express';
 
-export abstract class HttpService extends RunnableModule {
+export class HttpService extends RunnableModule {
   public router: express.Router;
   public slug: string;
+  public provider: Provider;
 
-  protected constructor(slug: string, router: express.Router, logger = dummyLogger) {
+  constructor(slug: string, provider: Provider, router: express.Router, logger = dummyLogger) {
     super(slug, logger);
     this.router = router;
     this.slug = slug;
+    this.provider = provider;
+
     const healthHandler = async (req: express.Request, res: express.Response) => {
       logger.debug('/health', { ip: req.ip });
       let body: HealthCheckResponse | Error['message'];
@@ -29,16 +32,23 @@ export abstract class HttpService extends RunnableModule {
     this.router.post('/health', healthHandler);
   }
 
+  async initializeImpl(): Promise<void> {
+    if (!(await this.healthCheck()).ok) {
+      throw new ProviderError(ProviderFailure.Unhealthy);
+    }
+  }
+
   async startImpl(): Promise<void> {
-    return Promise.resolve();
+    await this.provider.start?.();
   }
 
   async shutdownImpl(): Promise<void> {
-    return Promise.resolve();
+    await this.provider.close?.();
   }
 
-  protected abstract initializeImpl(): Promise<void>;
-  protected abstract healthCheck(): Promise<HealthCheckResponse>;
+  async healthCheck(): Promise<HealthCheckResponse> {
+    return await this.provider.healthCheck();
+  }
 
   static routeHandler(logger: Logger): ProviderHandler {
     return async (args, _r, res, _n, handler) => {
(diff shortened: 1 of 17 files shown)


packages/cardano-services/src/StakePool/StakePoolHttpService.ts
```diff
@@ -13,12 +13,17 @@ export interface StakePoolServiceDependencies {
...
+    { logger = dummyLogger, stakePoolProvider }: StakePoolServiceDependencies,
+    router: express.Router
+  ) {
     super(ServiceNames.StakePool, router, logger);
+    this.#stakePoolProvider = stakePoolProvider;
   }
 
   async healthCheck() {
-    return Promise.resolve({ ok: true });
+    return this.#stakePoolProvider.healthCheck();
```
rhyslbw, line 26: This is best submitted as a separate `fix` commit.

packages/cardano-services/test/Program/loadHttpServer.test.ts
```diff
@@ -71,7 +72,13 @@ describe('loadHttpServer', () => {
         async () =>
           await loadHttpServer({
             apiUrl,
-            serviceNames: [ServiceNames.StakePool]
+            serviceNames: [
```
rhyslbw, line 75: Also best if this part of the change is a separate `test` commit. The idea is to make atomic changes, and the test here is being improved by starting all of the available services. i.e it's an independent improvement

packages/cardano-services/test/Program/loadHttpServer.test.ts
```diff
@@ -98,20 +105,19 @@ describe('loadHttpServer', () => {
       await serverClosePromise(ogmiosServer);
     });
 
-    it('throws if any internal providers are unhealthy', async () => {
+    it('should not throws if any internal providers are unhealthy during http server initialization', async () => {
```
rhyslbw, line 108:
```suggestion
    it('should not throw if any internal providers are unhealthy during HTTP server initialization', async () => {
```

packages/cardano-services/test/TxSubmit/TxSubmitHttpService.test.ts
```diff
@@ -39,8 +39,8 @@ describe('TxSubmitHttpService', () => {
       };
     });
 
-    it('throws during initialization if the TxSubmitProvider is unhealthy', async () => {
-      await expect(() => TxSubmitHttpService.create({ txSubmitProvider })).rejects.toThrow(
+    it('should not throws during initialization if the TxSubmitProvider is unhealthy', async () => {
```
rhyslbw, line 42:
```suggestion
    it('should not throw during initialization if the TxSubmitProvider is unhealthy', async () => {
```

packages/cardano-services/test/entrypoints.test.ts
```diff
@@ -452,71 +452,6 @@ describe('entrypoints', () => {
       });
     });
 
-    describe('with unhealthy internal providers', () => {
```
rhyslbw, line 455: This is actually still a valid test. We don't want the service to start up if an internal provider is unhealthy

rhyslbw (changes requested):

packages/cardano-services/test/entrypoints.test.ts
IvayloAndonov: How this could be determined since we remove provider health checks during service init? Should we consider then moving them in [startImpl](https://github.com/input-output-hk/cardano-js-sdk/blob/a2bfcc62c25e71d78d07b961267d7ce9679b6cf4/packages/cardano-services/src/Http/HttpServer.ts#L83-L86) for instance?

rhyslbw: Yes, unhealthy at the point of _starting_ is the logical place for this to fail. It will likely need a refactor of the `HttpService`, maybe to expose a new method for checking the health of internal providers.

packages/cardano-services/test/Program/loadHttpServer.test.ts
IvayloAndonov: [d850ab4](https://github.com/input-output-hk/cardano-js-sdk/pull/303/commits/d850ab413ed691c2eafedf31fcd8644df4f4ac83)

rhyslbw (changes requested):
e71e7998452fbf9b73294e2e69ac6b8ab6f49c46 includes breaking changes, so the commit message needs to be updated to reflect this.

Have you considered making the `HttpService` a concrete class and hoisting the health check logic up to DRY up the code? Looks like we can also remove the static `create` method.

IvayloAndonov:
> [e71e799](https://github.com/input-output-hk/cardano-js-sdk/commit/e71e7998452fbf9b73294e2e69ac6b8ab6f49c46) includes breaking changes, so the commit message needs to be updated to reflect this.
> 
> Have you considered making the `HttpService` a concrete class and hoisting the health check logic up to DRY up the code? Looks like we can also remove the static `create` method.

I'll _reword_ the commit message to reflect the breaking change.

The removing `static create` method is considered to be introduced in a separate new commit soon, clicked re-review button by mistake yesterday.

About making `HttpService` a concrete class, yes it's a good option that helps with the hoisting of repeatable method implementations (ex. `initializeImpl`), then we need to extend and pass also `provider` to the `HttpService` already public constructor, how it sounds? Ex:
```export class HttpService extends RunnableModule {
  public router: express.Router;
  public slug: string;
  public provider: DbSyncProvider;

  constructor(slug: string, router: express.Router, provider: DbSyncProvider, logger = dummyLogger) {
    super(slug, logger);
    this.router = router;
    this.slug = slug;
    this.provider = provider;
```

IvayloAndonov:
The breaking change commits are amended as follows: 
- primary one: [060b5b8](https://github.com/input-output-hk/cardano-js-sdk/pull/303/commits/060b5b84db3ebe9d2ca7fd3ce842e77e4203f70a)
- fixup: [983810c](https://github.com/input-output-hk/cardano-js-sdk/pull/303/commits/983810cea1ffd3f91ccaa3a9e138ffba3fce49ba)

The `HttpService` refactoring is applied here [1bd7c23](https://github.com/input-output-hk/cardano-js-sdk/pull/303/commits/1bd7c231c885365d9d6158a04968a9fc7ba22dd9), seems much cleaner now, only it turned out the impact here is a bit higher than expected according to the initial task requirements.

rhyslbw:
>Might be worth updating the ticket and PR title accordingly to what's done?

Yes, PR description should reflect the final state of the PR we decided to merge. You can use ~strikeout~ if it's important to retain visibility, however, there's an edit record that is fine to rely on in most cases.

rhyslbw (approved):
Great work @IvayloAndonov 

Please squash and @mkazlauskas can perform final review prior to merge, as the test scope isn't changing.

IvayloAndonov:
Rebased with the latest master (conflict resolved) and the commits are squashed.

rhyslbw (dismissed):

mkazlauskas (approved):
Nice work @IvayloAndonov :tada:

packages/cardano-services/src/Http/HttpService.ts
```diff
@@ -1,18 +1,21 @@
-import { HealthCheckResponse, ProviderError, ProviderFailure } from '@cardano-sdk/core';
+import { HealthCheckResponse, Provider, ProviderError, ProviderFailure } from '@cardano-sdk/core';
 import { HttpServer } from './HttpServer';
 import { Logger, dummyLogger } from 'ts-log';
 import { ProviderHandler } from '../util';
 import { RunnableModule } from '../RunnableModule';
 import express from 'express';
 
-export abstract class HttpService extends RunnableModule {
```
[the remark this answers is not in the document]
rhyslbw: It was prompted [here](https://github.com/input-output-hk/cardano-js-sdk/pull/303#pullrequestreview-1021087382). I was thinking we could pass in a routes map and require a `Provider` to be passed in, so we don't have to extend for each provider type. I decided in my review it was unnecessary to take it that far here but missed the abstract definition still being removed.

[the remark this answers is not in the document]
rhyslbw: @IvayloAndonov, let's maybe just re-introduce the abstract definition to avoid degrading the constraints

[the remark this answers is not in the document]
IvayloAndonov: Re-introducing the abstract definition of HttpService is applied here as fixup commit 7503be8ae947c8c6428bfdea394997bc7463aa6e

mkazlauskas (approved):

rhyslbw (approved):

PR #306, opened 2022-06-28
refactor: generic NetworkInfoProvider.ledgerTip

mkazlauskas:
# Context

Implementing new providers for SingleAddressWallet requires resolving **all** fields defined in `core` package. This is not always possible.

# Proposed Solution

This PR demonstrates a generic parameter-based approach for more flexible types. It only implements the approach for `NetworkInfoProvider.ledgerTip` as a means to evaluate the approach. A full refactor would introduce a lot more changes, adding many more generic parameters to `ObservableWalllet` and `SingleAddressWallet`. See [this test](https://github.com/input-output-hk/cardano-js-sdk/blob/2e254cf7b1e4e8742e7849740f6d913215d41b64/packages/wallet/test/integration/partialWallet.test.ts) for usage example.

Relevant commits: 0632420abca4e9cd680bed26df37124dbd2a72b9 2e254cf7b1e4e8742e7849740f6d913215d41b64

# Other Changes Introduced

- chore(wallet): convert types.ts into directory module with index 485728f8e6d60c2b1a8b863ddeeb771f0d1ac83e

### DIFF
--- a/packages/wallet/src/services/util/equals.ts
+++ b/packages/wallet/src/services/util/equals.ts
@@ -1,5 +1,6 @@
 import { Cardano, EpochInfo, TimeSettings } from '@cardano-sdk/core';
 import { GroupedAddress } from '../../KeyManagement';
+import { WC } from '../../types';
 import isEqual from 'lodash/isEqual';
 
 export const strictEquals = <T>(a: T, b: T) => a === b;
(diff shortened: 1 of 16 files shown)


packages/wallet/src/services/util/equals.ts
```diff
@@ -11,7 +12,7 @@ export const arrayEquals = <T>(arrayA: T[], arrayB: T[], itemEquals: (a: T, b: T
 
 export const shallowArrayEquals = <T>(a: T[], b: T[]) => arrayEquals(a, b, strictEquals);
 
-export const tipEquals = (a: Cardano.Tip, b: Cardano.Tip) => a.hash === b.hash;
+export const tipEquals = (a: WC.Tip, b: WC.Tip) => a.slot === b.slot;
```
mkazlauskas, line 15: This change might be a bug if there's a rollback to the same slot but different block

mkazlauskas:
Closing in favor of #319

PR #308, opened 2022-06-30
refactor(wallet): inheritance-based partial providers

mkazlauskas:
# Context

An alternative approach to solve problem described in #306

Note: this PR is to be merged into #306 and not master.

# Proposed Solution

This PR reverts the generics-based approach commits and adds a commit to demonstrate a new approach: e61e600faec199ed99330319aa95fa5630475533

You may also look at the diff between the 2 branches to see the differences.


packages/wallet/src/types/ObservableWallet.ts
```diff
@@ -103,3 +106,7 @@ export interface ObservableWallet<Tip = WC.Tip> {
   submitTx(tx: Cardano.NewTxAlonzo): Promise<void>;
   shutdown(): void;
 }
+
+export interface ObservableWallet extends ObservableWalletCore {
+  readonly tip$: Observable<Cardano.Tip>;
```
mirceahasegan, line 111: I personally liked the generics approach better. Then, one can create types that have the generics already filled in. But it could be just personal preference 😄 
`export interface ObservableWalletCore<Tip = WC.Tip> {...}`
`export type ObservableWallet = ObservableWalletCore<Cardano.Tip>`

mkazlauskas:
Closing in favor of #306

PR #309, opened 2022-06-30
feat: cardano-services dependency port resolution using DNS SRV records

IvayloAndonov:
# Context
Add new `cardano-services` CLI and HttpServer arguments to support resolving the ports using DNS SRV records for:

**CLI:**
`--ogmios-srv-service-name` (alternative to `--ogmios-url`)

`--rabitmq-srv-service-name` (alternative to `--rabbitmq-url`)

`--postgres-srv-service-name` + `--postgres-db` + `--postgres-user` + `--postgres-password` (alternative to `--db-connection-string`)

**HTTP Server:**
`RABBITMQ_SRV_SERVICE_NAME` (alternative to `--RABBITMQ_URL`)

`OGMIOS_SRV_SERVICE_NAME` (alternative to`OGMIOS_URL`)

`POSTGRES_SRV_SERVICE_NAME` + `POSTGRES_DB` + `POSTGRES_USER` + `POSTGRES_PASSWORD` (alternative to `DB_CONNECTION_STRING`)

Use those new variables to resolve port of those services with [dns.resolveSrv](https://nodejs.org/docs/latest-v14.x/api/dns.html#dns_dns_resolvesrv_hostname_callback)

# Acceptance Criteria 
- If a connection error is thrown either during startup or during a request, the service should attempt to resolve another address using exponential back-off with no initial delay.

- The backoff factor should be configurable, with a sensible default.

- If a connection cannot be re-established after a configurable period of time in seconds, defaulting to **60**, the process should terminate. Exit process upon connection failure

- If the connection error is thrown at request time, the request must be held open until the timeout is reached.

- Try using a default back-off factor of **1.1**

# Proposed Solution
- Wrap client libraries of each service (_ogmios_, _rabbitmq_, _postgres_) to make dealing with failovers (re-resolving the port) opaque

- implement DNS resolve SRV records with exponential backoff retry logic
- create an abstraction over client libraries of each service (_postgres, rabbitmq, ogmios_) based on the provided program arguments
- For `Ogmios` and `RabbitMQ` client libraries, **service discovery** takes preference over the **url**  arg which will leave us with the ability to fall back on fixed defaults. When it comes to `Postgres` -> we throw an `InvalidArgsCombination` error if both are provided and `MissingProgramOption` error if neither one of them is provided.
- implement `createDnsResolver` factory method, in which `DnsResolver` instance is passed forward in order to re-resolve the port if connection failover occurs during runtime
- implement `resolveSrvRecord` which select the first random record from the DNS server resolved list
- use `Proxy` native module to intercept the provider's operations if the SRV service name is provided, it handles failovers produced by the `cardano-services` service dependencies
- extend `cli.ts` with:
-- `--service-discovery-backoff-factor`
-- `--service-discovery-backoff-timeout`
-- `--rabbitmq-srv-service-name`
-- `--ogmios-srv-service-name`
-- `--postgres-srv-service-name`
-- `--postgres-db`
-- `--postgres-user`
-- `--postgres-password`
- extend `run.ts` with:
-- `SERVICE_DISCOVERY_BACKOFF_FACTOR`
-- `SERVICE_DISCOVERY_BACKOFF_TIMEOUT`
-- `RABBITMQ_SRV_SERVICE_NAME`
-- `OGMIOS_SRV_SERVICE_NAME`
-- `POSTGRES_SRV_SERVICE_NAME`
-- `POSTGRES_DB`
-- `POSTGRES_USER`
-- `POSTGRES_PASSWORD`

### DIFF
--- a/packages/cardano-services/package.json
+++ b/packages/cardano-services/package.json
@@ -100,6 +100,7 @@
     "fraction.js": "^4.2.0",
     "json-bigint": "^1.0.0",
     "lodash": "^4.17.21",
+    "p-retry": "^4.2.0",
     "pg": "^8.7.3",
     "prom-client": "^14.0.1",
     "reflect-metadata": "~0.1.13",
--- a/packages/cardano-services/src/Program/ProgramOptionDescriptions.ts
+++ b/packages/cardano-services/src/Program/ProgramOptionDescriptions.ts
@@ -2,15 +2,18 @@ import { CommonOptionDescriptions } from '../ProgramsCommon';
 
 enum HttpServerOptionDescriptions {
   ApiUrl = 'API URL',
-  DbConnection = 'DB Connection',
+  DbConnection = 'DB Connection string',
   MetricsEnabled = 'Enable Prometheus Metrics',
   OgmiosUrl = 'Ogmios URL',
   RabbitMQUrl = 'RabbitMQ URL',
   UseQueue = 'Enables RabbitMQ',
   CardanoNodeConfigPath = 'Cardano node config path',
-  CacheTtl = 'Cache TTL in minutes between 1 and 2880',
-  EpochPollInterval = 'Epoch poll interval'
+  EpochPollInterval = 'Epoch poll interval',
+  PostgresSrvArgs = 'Postgres SRV service name, db, user and password'
 }
 
 export type ProgramOptionDescriptions = CommonOptionDescriptions | HttpServerOptionDescriptions;
-export const ProgramOptionDescriptions = { ...CommonOptionDescriptions, ...HttpServerOptionDescriptions };
+export const ProgramOptionDescriptions = {
+  ...CommonOptionDescriptions,
+  ...HttpServerOptionDescriptions
+};
--- a/packages/cardano-services/src/Program/loadHttpServer.ts
+++ b/packages/cardano-services/src/Program/loadHttpServer.ts
@@ -6,12 +6,18 @@ import { DbSyncNetworkInfoProvider, NetworkInfoHttpService } from '../NetworkInf
 import { DbSyncRewardsProvider, RewardsHttpService } from '../Rewards';
 import { DbSyncStakePoolProvider, StakePoolHttpService } from '../StakePool';
 import { DbSyncUtxoProvider, UtxoHttpService } from '../Utxo';
+import {
+  DnsResolver,
+  createDnsResolver,
+  getOgmiosTxSubmitProvider,
+  getPool,
+  getRabbitMqTxSubmitProvider
+} from './utils';
 import { HttpServer, HttpServerConfig, HttpService } from '../Http';
 import { InMemoryCache } from '../InMemoryCache';
 import { MissingProgramOption, UnknownServiceName } from './errors';
-import { OgmiosCardanoNode, ogmiosTxSubmitProvider, urlToConnectionConfig } from '@cardano-sdk/ogmios';
+import { OgmiosCardanoNode, urlToConnectionConfig } from '@cardano-sdk/ogmios';
 import { ProgramOptionDescriptions } from './ProgramOptionDescriptions';
-import { RabbitMqTxSubmitProvider } from '@cardano-sdk/rabbitmq';
 import { ServiceNames } from './ServiceNames';
 import { TxSubmitHttpService } from '../TxSubmit';
 import { createDbSyncMetadataService } from '../Metadata';
@@ -20,11 +26,14 @@ import pg from 'pg';
 
 export interface HttpServerOptions extends CommonProgramOptions {
   dbConnectionString?: string;
-  cacheTtl: number;
   epochPollInterval: number;
   cardanoNodeConfigPath?: string;
   metricsEnabled?: boolean;
   useQueue?: boolean;
+  postgresSrvServiceName?: string;
+  postgresDb?: string;
+  postgresUser?: string;
+  postgresPassword?: string;
 }
 
 export interface ProgramArgs {
@@ -37,22 +46,44 @@ export interface ProgramArgs {
     | ServiceNames.NetworkInfo
     | ServiceNames.Rewards
   )[];
+  /*
+    TODO: optimize passed options -> 'options' is always passed by default and shouldn't be optional field,
+    no need to check it with '.?' everywhere.Will be fixed within ADP-1990
+  */
   options?: HttpServerOptions;
 }
 
-const serviceMapFactory = (args: ProgramArgs, logger: Logger, cache: InMemoryCache, db?: pg.Pool) => ({
+const serviceMapFactory = (
+  args: ProgramArgs,
+  logger: Logger,
+  cache: InMemoryCache,
+  dnsResolver: DnsResolver,
+  db?: pg.Pool
+) => ({
   [ServiceNames.StakePool]: () => {
-    if (!db) throw new MissingProgramOption(ServiceNames.StakePool, ProgramOptionDescriptions.DbConnection);
+    if (!db)
+      throw new MissingProgramOption(ServiceNames.StakePool, [
+        ProgramOptionDescriptions.DbConnection,
+        ProgramOptionDescriptions.PostgresSrvArgs
+      ]);
 
     return new StakePoolHttpService({ logger, stakePoolProvider: new DbSyncStakePoolProvider(db, logger) });
   },
   [ServiceNames.Utxo]: () => {
-    if (!db) throw new MissingProgramOption(ServiceNames.Utxo, ProgramOptionDescriptions.DbConnection);
+    if (!db)
+      throw new MissingProgramOption(ServiceNames.Utxo, [
+        ProgramOptionDescriptions.DbConnection,
+        ProgramOptionDescriptions.PostgresSrvArgs
+      ]);
 
     return new UtxoHttpService({ logger, utxoProvider: new DbSyncUtxoProvider(db, logger) });
   },
   [ServiceNames.ChainHistory]: () => {
-    if (!db) throw new MissingProgramOption(ServiceNames.ChainHistory, ProgramOptionDescriptions.DbConnection);
+    if (!db)
+      throw new MissingProgramOption(ServiceNames.ChainHistory, [
+        ProgramOptionDescriptions.DbConnection,
+        ProgramOptionDescriptions.PostgresSrvArgs
+      ]);
 
     const metadataService = createDbSyncMetadataService(db, logger);
     return new ChainHistoryHttpService({
@@ -61,12 +92,20 @@ const serviceMapFactory = (args: ProgramArgs, logger: Logger, cache: InMemoryCac
     });
   },
   [ServiceNames.Rewards]: () => {
-    if (!db) throw new MissingProgramOption(ServiceNames.Rewards, ProgramOptionDescriptions.DbConnection);
+    if (!db)
+      throw new MissingProgramOption(ServiceNames.Rewards, [
+        ProgramOptionDescriptions.DbConnection,
+        ProgramOptionDescriptions.PostgresSrvArgs
+      ]);
 
     return new RewardsHttpService({ logger, rewardsProvider: new DbSyncRewardsProvider(db, logger) });
   },
   [ServiceNames.NetworkInfo]: () => {
-    if (!db) throw new MissingProgramOption(ServiceNames.NetworkInfo, ProgramOptionDescriptions.DbConnection);
+    if (!db)
+      throw new MissingProgramOption(ServiceNames.NetworkInfo, [
+        ProgramOptionDescriptions.DbConnection,
+        ProgramOptionDescriptions.PostgresSrvArgs
+      ]);
     if (args.options?.cardanoNodeConfigPath === undefined)
       throw new MissingProgramOption(ServiceNames.NetworkInfo, ProgramOptionDescriptions.CardanoNodeConfigPath);
     if (args.options?.ogmiosUrl === undefined)
@@ -82,34 +121,36 @@ const serviceMapFactory = (args: ProgramArgs, logger: Logger, cache: InMemoryCac
 
     return new NetworkInfoHttpService({ logger, networkInfoProvider });
   },
-  [ServiceNames.TxSubmit]: () => {
-    const txSubmitProvider =
-      args.options?.useQueue && args.options?.rabbitmqUrl
-        ? new RabbitMqTxSubmitProvider({ rabbitmqUrl: args.options.rabbitmqUrl })
-        : ogmiosTxSubmitProvider(urlToConnectionConfig(args.options?.ogmiosUrl));
+  [ServiceNames.TxSubmit]: async () => {
+    const txSubmitProvider = args.options?.useQueue
+      ? await getRabbitMqTxSubmitProvider(dnsResolver, logger, args.options)
+      : await getOgmiosTxSubmitProvider(dnsResolver, logger, args.options);
 
     return new TxSubmitHttpService({ logger, txSubmitProvider });
   }
 });
 
-export const loadHttpServer = (args: ProgramArgs): HttpServer => {
+export const loadHttpServer = async (args: ProgramArgs): Promise<HttpServer> => {
   const services: HttpService[] = [];
   const logger = createLogger({
     level: args.options?.loggerMinSeverity,
     name: 'http-server'
   });
 
-  const db = args.options?.dbConnectionString
-    ? new pg.Pool({ connectionString: args.options.dbConnectionString })
-    : undefined;
-
   const cache = new InMemoryCache(args.options?.cacheTtl);
-
-  const serviceMap = serviceMapFactory(args, logger, cache, db);
+  const dnsResolver = createDnsResolver(
+    {
+      factor: args.options?.serviceDiscoveryBackoffFactor,
+      maxRetryTime: args.options?.serviceDiscoveryTimeout
+    },
+    logger
+  );
+  const db = await getPool(dnsResolver, logger, args.options);
+  const serviceMap = serviceMapFactory(args, logger, cache, dnsResolver, db);
 
   for (const serviceName of args.serviceNames) {
     if (serviceMap[serviceName]) {
-      services.push(serviceMap[serviceName]());
+      services.push(await serviceMap[serviceName]());
     } else {
       throw new UnknownServiceName(serviceName);
     }
--- a/packages/cardano-services/src/Program/utils.ts
+++ b/packages/cardano-services/src/Program/utils.ts
@@ -1,0 +1,269 @@
+/* eslint-disable max-len */
+/* eslint-disable promise/no-nesting */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { ClientConfig, Pool, QueryConfig } from 'pg';
+import { CommonOptionDescriptions, CommonProgramOptions } from '../ProgramsCommon';
+import { HttpServerOptions } from './loadHttpServer';
+import { InvalidArgsCombination, MissingProgramOption } from './errors';
+import { Ogmios, ogmiosTxSubmitProvider, urlToConnectionConfig } from '@cardano-sdk/ogmios';
+import { ProgramOptionDescriptions } from './ProgramOptionDescriptions';
+import { ProviderFailure, TxSubmitProvider } from '@cardano-sdk/core';
+import { RabbitMqTxSubmitProvider } from '@cardano-sdk/rabbitmq';
+import { ServiceNames } from './ServiceNames';
+import Logger from 'bunyan';
+import dns, { SrvRecord } from 'dns';
+import pRetry, { FailedAttemptError } from 'p-retry';
+
+export const SERVICE_DISCOVERY_BACKOFF_FACTOR_DEFAULT = 1.1;
+export const SERVICE_DISCOVERY_TIMEOUT_DEFAULT = 60 * 1000;
+export const DNS_SRV_CACHE_KEY = 'DNS_SRV_Record';
+
+export type RetryBackoffConfig = {
+  factor?: number;
+  maxRetryTime?: number;
+};
+
+export const onFailedAttemptFor =
+  (serviceName: string, logger: Logger) =>
+  async ({ attemptNumber, message, retriesLeft }: FailedAttemptError) => {
+    const nextAction = retriesLeft > 0 ? 'retrying...' : 'exiting';
+    logger.trace(message);
+    logger.debug(
+      `Establishing connection to ${serviceName}: Attempt ${attemptNumber} of ${
+        attemptNumber + retriesLeft
+      }, ${nextAction}`
+    );
+    if (retriesLeft === 0) {
+      logger.error(message);
+      // Invokes onDeath() callback within cardano-services entrypoints, following by server.shutdown() and process.exit(1)
+      process.kill(process.pid, 'SIGTERM');
+    }
+  };
+
+// Select the first random record from the DNS server resolved list
+export const resolveSrvRecord = async (serviceName: string): Promise<SrvRecord> => {
+  const [srvRecord] = await dns.promises.resolveSrv(serviceName);
+  return srvRecord;
+};
+
+export const createDnsResolver = (config: RetryBackoffConfig, logger: Logger) => async (serviceName: string) =>
+  await pRetry(async () => await resolveSrvRecord(serviceName), {
+    factor: config.factor,
+    maxRetryTime: config.maxRetryTime,
+    onFailedAttempt: onFailedAttemptFor(serviceName, logger)
+  });
+
+export type DnsResolver = ReturnType<typeof createDnsResolver>;
+
+/**
+ * Creates a extended Pool client :
+ * - use passed srv service name in order to resolve the port
+ * - make dealing with failovers (re-resolving the port) opaque
+ * - use exponential backoff retry internally with default timeout and factor
+ * - intercept 'query' operation and handle connection errors runtime
+ * - all other operations are bind to pool object withoud modifications
+ *
+ * @returns pg.Pool instance
+ */
+export const getPoolWithServiceDiscovery = async (
+  dnsResolver: DnsResolver,
+  logger: Logger,
+  { host, database, password, user }: ClientConfig
+): Promise<Pool> => {
+  const { name, port } = await dnsResolver(host!);
+  let pool = new Pool({ database, host: name, password, port, user });
+
+  return new Proxy<Pool>({} as Pool, {
+    get(_, prop) {
+      if (prop === 'then') return;
+      if (prop === 'query') {
+        return (args: string | QueryConfig, values?: any) =>
+          pool.query(args, values).catch(async (error) => {
+            if (error.code && ['ENOTFOUND', 'ECONNREFUSED', 'ECONNRESET'].includes(error.code)) {
+              const record = await dnsResolver(host!);
+              logger.info(`DNS resolution for Postgres service, resolved with record: ${JSON.stringify(record)}`);
+              pool = new Pool({ database, host: record.name, password, port: record.port, user });
+              return await pool.query(args, values);
+            }
+            throw error;
+          });
+      }
+      // Bind if it is a function, no intercept operations
+      if (typeof pool[prop as keyof Pool] === 'function') {
+        const method = pool[prop as keyof Pool] as any;
+        return method.bind(pool);
+      }
+    }
+  });
+};
+
+export const getPool = async (
+  dnsResolver: DnsResolver,
+  logger: Logger,
+  options?: HttpServerOptions
+): Promise<Pool | undefined> => {
+  if (options?.dbConnectionString && options.postgresSrvServiceName)
+    throw new InvalidArgsCombination(ProgramOptionDescriptions.DbConnection, ProgramOptionDescriptions.PostgresSrvArgs);
+  if (options?.dbConnectionString) return new Pool({ connectionString: options.dbConnectionString });
+  if (options?.postgresSrvServiceName && options?.postgresUser && options.postgresDb && options.postgresPassword) {
+    return getPoolWithServiceDiscovery(dnsResolver, logger, {
+      database: options.postgresDb,
+      host: options.postgresSrvServiceName,
+      password: options.postgresPassword,
+      user: options.postgresUser
+    });
+  }
+  // If db connection string is not passed nor postgres srv service name
+  return undefined;
+};
+
+export const isOgmiosConnectionError = (error: any) => {
+  if (
+    (error?.innerError?.code && ['ENOTFOUND', 'ECONNREFUSED', 'ECONNRESET'].includes(error.innerError.code)) ||
+    error instanceof Ogmios.WebSocketClosed
+  ) {
+    return true;
+  }
+  return false;
+};
+
+/**
+ * Creates a extended TxSubmitProvider instance :
+ * - use passed srv service name in order to resolve the port
+ * - make dealing with failovers (re-resolving the port) opaque
+ * - use exponential backoff retry internally with default timeout and factor
+ * - intercept 'submitTx' operation and handle connection errors runtime
+ * - all other operations are bind to pool object withoud modifications
+ *
+ * @returns TxSubmitProvider instance
+ */
+export const ogmiosTxSubmitProviderWithDiscovery = async (
+  dnsResolver: DnsResolver,
+  logger: Logger,
+  serviceName: string
+): Promise<TxSubmitProvider> => {
+  const { name, port } = await dnsResolver(serviceName!);
+  let ogmiosProvider = ogmiosTxSubmitProvider({ host: name, port });
+
+  return new Proxy<TxSubmitProvider>({} as TxSubmitProvider, {
+    get(_, prop) {
+      if (prop === 'then') return;
+      if (prop === 'submitTx') {
+        return (args: Uint8Array) =>
+          ogmiosProvider.submitTx(args).catch(async (error) => {
+            if (isOgmiosConnectionError(error)) {
+              const record = await dnsResolver(serviceName!);
+              logger.info(`DNS resolution for Ogmios service, resolved with record: ${JSON.stringify(record)}`);
+              await ogmiosProvider
+                .close?.()
+                .catch((error_) => logger.warn(`Ogmios provider failed to close after DNS resolution: ${error_}`));
+              ogmiosProvider = ogmiosTxSubmitProvider({ host: record.name, port: record.port });
+              return await ogmiosProvider.submitTx(args);
+            }
+            throw error;
+          });
+      }
+      // Bind if it is a function, no intercept operations
+      if (typeof ogmiosProvider[prop as keyof TxSubmitProvider] === 'function') {
+        const method = ogmiosProvider[prop as keyof TxSubmitProvider] as any;
+        return method.bind(ogmiosProvider);
+      }
+    }
+  });
+};
+
+export const getOgmiosTxSubmitProvider = async (
+  dnsResolver: DnsResolver,
+  logger: Logger,
+  options?: CommonProgramOptions
+): Promise<TxSubmitProvider> => {
+  if (options?.ogmiosSrvServiceName)
+    return ogmiosTxSubmitProviderWithDiscovery(dnsResolver, logger, options.ogmiosSrvServiceName);
+  if (options?.ogmiosUrl) return ogmiosTxSubmitProvider(urlToConnectionConfig(options?.ogmiosUrl));
+  throw new MissingProgramOption(ServiceNames.TxSubmit, [
+    CommonOptionDescriptions.OgmiosUrl,
+    CommonOptionDescriptions.OgmiosSrvServiceName
+  ]);
+};
+
+export const srvRecordToRabbitmqURL = ({ name, port }: SrvRecord) => new URL(`amqp://${name}:${port}`);
+
+/**
+ * Creates a extended RabbitMqTxSubmitProvider instance :
+ * - use passed srv service name in order to resolve the port
+ * - make dealing with failovers (re-resolving the port) opaque
+ * - use exponential backoff retry internally with default timeout and factor
+ * - intercept 'submitTx' operation and handle connection errors runtime
+ * - all other operations are bind to pool object withoud modifications
+ *
+ * @returns RabbitMqTxSubmitProvider instance
+ */
+export const rabbitMqTxSubmitProviderWithDiscovery = async (
+  dnsResolver: DnsResolver,
+  logger: Logger,
+  serviceName: string
+): Promise<RabbitMqTxSubmitProvider> => {
+  const record = await dnsResolver(serviceName!);
+  let rabbitmqProvider = new RabbitMqTxSubmitProvider({
+    rabbitmqUrl: srvRecordToRabbitmqURL(record)
+  });
+
+  return new Proxy<RabbitMqTxSubmitProvider>({} as RabbitMqTxSubmitProvider, {
+    get(_, prop) {
+      if (prop === 'then') return;
+      if (prop === 'submitTx') {
+        return (args: Uint8Array) =>
+          rabbitmqProvider.submitTx(args).catch(async (error) => {
+            if (
+              error.innerError?.reason === ProviderFailure.ConnectionFailure ||
+              isOgmiosConnectionError(error.innerError)
+            ) {
+              const resolvedRecord = await dnsResolver(serviceName!);
+              logger.info(`DNS resolution for RabbitMQ service, resolved with record: ${JSON.stringify(record)}`);
+              await rabbitmqProvider
+                .close?.()
+                .catch((error_) => logger.warn(`RabbitMQ provider failed to close after DNS resolution: ${error_}`));
+              rabbitmqProvider = new RabbitMqTxSubmitProvider({
+                rabbitmqUrl: srvRecordToRabbitmqURL(resolvedRecord)
+              });
+              return await rabbitmqProvider.submitTx(args);
+            }
+            throw error;
+          });
+      }
+      // Bind if it is a function, no intercept operations
+      if (typeof rabbitmqProvider[prop as keyof RabbitMqTxSubmitProvider] === 'function') {
+        const method = rabbitmqProvider[prop as keyof RabbitMqTxSubmitProvider] as any;
+        return method.bind(rabbitmqProvider);
+      }
+    }
+  });
+};
+
+export const getRabbitMqTxSubmitProvider = async (
+  dnsResolver: DnsResolver,
+  logger: Logger,
+  options?: CommonProgramOptions
+): Promise<RabbitMqTxSubmitProvider> => {
+  if (options?.rabbitmqSrvServiceName)
+    return rabbitMqTxSubmitProviderWithDiscovery(dnsResolver, logger, options.rabbitmqSrvServiceName);
+  if (options?.rabbitmqUrl) return new RabbitMqTxSubmitProvider({ rabbitmqUrl: options.rabbitmqUrl });
+  throw new MissingProgramOption(ServiceNames.TxSubmit, [
+    CommonOptionDescriptions.RabbitMQUrl,
+    CommonOptionDescriptions.RabbitMQSrvServiceName
+  ]);
+};
+
+export const getRabbitMqUrl = async (dnsResolver: DnsResolver, args: CommonProgramOptions) => {
+  if (args.rabbitmqSrvServiceName) {
+    const record = await dnsResolver(args.rabbitmqSrvServiceName);
+    return srvRecordToRabbitmqURL(record);
+  }
+  if (args.rabbitmqUrl) {
+    return args.rabbitmqUrl;
+  }
+  throw new MissingProgramOption(ServiceNames.TxSubmit, [
+    CommonOptionDescriptions.RabbitMQUrl,
+    CommonOptionDescriptions.RabbitMQSrvServiceName
+  ]);
+};
--- a/packages/cardano-services/src/ProgramsCommon/CommonOptionDescriptions.ts
+++ b/packages/cardano-services/src/ProgramsCommon/CommonOptionDescriptions.ts
@@ -1,5 +1,10 @@
 export enum CommonOptionDescriptions {
   LoggerMinSeverity = 'Log level',
   OgmiosUrl = 'Ogmios URL',
-  RabbitMQUrl = 'RabbitMQ URL'
+  RabbitMQUrl = 'RabbitMQ URL',
+  OgmiosSrvServiceName = 'Ogmios SRV service name',
+  RabbitMQSrvServiceName = 'RabbitMQ SRV service name',
+  ServiceDiscoveryBackoffFactor = 'Exponential backoff factor for service discovery',
+  ServiceDiscoveryTimeout = 'Timeout for service discovery attempts',
+  CacheTtl = 'Cache TTL in minutes between 1 and 2880'
 }
--- a/packages/cardano-services/src/ProgramsCommon/options.ts
+++ b/packages/cardano-services/src/ProgramsCommon/options.ts
@@ -9,4 +9,9 @@ export interface CommonProgramOptions {
   loggerMinSeverity?: LogLevel;
   ogmiosUrl?: URL;
   rabbitmqUrl?: URL;
+  ogmiosSrvServiceName?: string;
+  rabbitmqSrvServiceName?: string;
+  serviceDiscoveryBackoffFactor?: number;
+  serviceDiscoveryTimeout?: number;
+  cacheTtl: number;
 }
--- a/packages/cardano-services/src/TxWorker/loadTxWorker.ts
+++ b/packages/cardano-services/src/TxWorker/loadTxWorker.ts
@@ -1,8 +1,7 @@
-import { CommonOptionDescriptions, CommonProgramOptions } from '../ProgramsCommon';
-import { MissingProgramOption, ServiceNames } from '../Program';
+import { CommonProgramOptions } from '../ProgramsCommon';
 import { TxSubmitWorker, TxSubmitWorkerConfig } from '@cardano-sdk/rabbitmq';
+import { createDnsResolver, getOgmiosTxSubmitProvider, getRabbitMqUrl } from '../Program';
 import { createLogger } from 'bunyan';
-import { ogmiosTxSubmitProvider, urlToConnectionConfig } from '@cardano-sdk/ogmios';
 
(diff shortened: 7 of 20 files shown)


packages/cardano-services/src/Program/ProgramOptionDescriptions.ts
```diff
@@ -9,7 +9,9 @@ enum HttpServerOptionDescriptions {
   UseQueue = 'Enables RabbitMQ',
   CardanoNodeConfigPath = 'Cardano node config path',
   DbQueriesCacheTtl = 'Db queries cache TTL in minutes between 1 and 2880',
-  DbPollInterval = 'Db poll interval'
+  DbPollInterval = 'Db poll interval',
+  RetryBackoffFactory = 'Exponential Back-Off retry factor',
```
rhyslbw, line 13:
```suggestion
  ServiceDiscoveryBackoffFactor = 'Exponential backoff factor for service discovery',
```

packages/cardano-services/src/Program/loadHttpServer.ts
```diff
@@ -14,14 +14,17 @@ import { RabbitMqTxSubmitProvider } from '@cardano-sdk/rabbitmq';
...
 import { ogmiosTxSubmitProvider, urlToConnectionConfig } from '@cardano-sdk/ogmios';
 import Logger, { createLogger } from 'bunyan';
 import pg from 'pg';
 
-export interface HttpServerOptions extends CommonProgramOptions {
+export interface HttpServerOptions extends CommonProgramOptions, SrvProgramOptions {
   dbConnectionString?: string;
   dbQueriesCacheTtl: number;
   dbPollInterval: number;
+  retryBackoffFactor: number;
```
rhyslbw, line 26:
```suggestion
  serviceDiscoveryBackoffFactor: number;
```
It's a little verbose, but adds the needed context

packages/cardano-services/src/Program/loadHttpServer.ts
```diff
@@ -14,14 +14,17 @@ import { RabbitMqTxSubmitProvider } from '@cardano-sdk/rabbitmq';
...
+  retryBackoffMaxTimeout: number;
```
rhyslbw, line 27:
```suggestion
  serviceDiscoveryTimeout: number;
```

packages/cardano-services/src/Program/utils.ts
```diff
@@ -0,0 +1,98 @@
...
+  };
+
+export const resolveDnsSrvWithExponentialBackoff = async (
+  hostname: string,
+  { factor, maxRetryTime }: RetryBackoffConfig,
+  logger: Logger
+) =>
+  await pRetry(
+    async () => {
+      // Shall we grab the first one always?
```
rhyslbw, line 38: Is this OK @blaggacao?
``` ts
const [record] = await dns.promises.resolveSrv(hostname);
```

packages/cardano-services/src/cli.ts
```diff
@@ -101,7 +102,23 @@ commonOptions(
     (interval) => Number.parseInt(interval, 10),
     DB_POLL_INTERVAL_DEFAULT
   )
+  .option(
+    '--retry-backoff-factor <retryBackoffFactor>',
```
rhyslbw, line 106:
```suggestion
    '--service-discovery-backoff-factor <serviceDiscoveryBackoffFactor>',
```

packages/cardano-services/src/cli.ts
```diff
@@ -101,7 +102,23 @@ commonOptions(
...
+    ProgramOptionDescriptions.RetryBackoffFactory,
+    (factor) => Number.parseFloat(factor),
+    RETRY_BACKOFF_FACTOR_DEFAULT
+  )
+  .option(
+    '--retry-backoff-max-timeout <retryBackoffMaxTimeout>',
```
rhyslbw, line 112:
```suggestion
    '--service-discovery-timeout <serviceDiscoveryTimeout>',
```

rhyslbw (changes requested):
Looking good @IvayloAndonov. Just some naming suggestions at this stage

packages/cardano-services/src/Program/utils.ts
```diff
@@ -0,0 +1,98 @@
...
+    if (retriesLeft === 0) {
+      logger.error(message);
+      // await server.shutdown(); ?
+      // eslint-disable-next-line unicorn/no-process-exit
+      process.exit(1);
+    }
+  };
+
+export const resolveDnsSrvWithExponentialBackoff = async (
+  hostname: string,
```
blaggacao, line 32:
```suggestion
  servicename: string,
```

packages/cardano-services/src/Program/utils.ts
blaggacao: A typical DNS SRV query returns a list of all services, eg:
```console
[root@ip-10-52-34-34:~]# dig _infra-node._tcp.service.consul SRV

; <<>> DiG 9.16.27 <<>> _infra-node._tcp.service.consul SRV
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 53825
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 10, AUTHORITY: 0, ADDITIONAL: 41

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;_infra-node._tcp.service.consul. IN	SRV

;; ANSWER SECTION:
_infra-node._tcp.service.consul. 0 IN	SRV	1 1 30911 0a185e5d.addr.eu-central-1.consul.
_infra-node._tcp.service.consul. 0 IN	SRV	1 1 22846 0a185e5d.addr.eu-central-1.consul.
_infra-node._tcp.service.consul. 0 IN	SRV	1 1 27727 0a3479fc.addr.eu-central-1.consul.
_infra-node._tcp.service.consul. 0 IN	SRV	1 1 22324 0a3479fc.addr.eu-central-1.consul.
_infra-node._tcp.service.consul. 0 IN	SRV	1 1 24536 0a18bd22.addr.eu-central-1.consul.
_infra-node._tcp.service.consul. 0 IN	SRV	1 1 27665 0a181be0.addr.eu-central-1.consul.
_infra-node._tcp.service.consul. 0 IN	SRV	1 1 28368 0a185e5d.addr.eu-central-1.consul.
_infra-node._tcp.service.consul. 0 IN	SRV	1 1 26681 0a18bd22.addr.eu-central-1.consul.
_infra-node._tcp.service.consul. 0 IN	SRV	1 1 29224 0a181be0.addr.eu-central-1.consul.
_infra-node._tcp.service.consul. 0 IN	SRV	1 1 31902 0a18bd22.addr.eu-central-1.consul.

;; ADDITIONAL SECTION:
0a185e5d.addr.eu-central-1.consul. 0 IN	A	10.24.94.93
ip-10-24-94-93.eu-central-1.compute.internal.node.eu-central-1.consul. 0 IN TXT	"region=eu-central-1"
ip-10-24-94-93.eu-central-1.compute.internal.node.eu-central-1.consul. 0 IN TXT	"consul-network-segment="
ip-10-24-94-93.eu-central-1.compute.internal.node.eu-central-1.consul. 0 IN TXT	"node_name=client-eu-central-1-c5-2xlarge-infra"
0a185e5d.addr.eu-central-1.consul. 0 IN	A	10.24.94.93
ip-10-24-94-93.eu-central-1.compute.internal.node.eu-central-1.consul. 0 IN TXT	"region=eu-central-1"
ip-10-24-94-93.eu-central-1.compute.internal.node.eu-central-1.consul. 0 IN TXT	"consul-network-segment="
ip-10-24-94-93.eu-central-1.compute.internal.node.eu-central-1.consul. 0 IN TXT	"node_name=client-eu-central-1-c5-2xlarge-infra"
0a3479fc.addr.eu-central-1.consul. 0 IN	A	10.52.121.252
ip-10-52-121-252.us-east-2.compute.internal.node.eu-central-1.consul. 0	IN TXT "region=eu-central-1"
ip-10-52-121-252.us-east-2.compute.internal.node.eu-central-1.consul. 0	IN TXT "consul-network-segment="
ip-10-52-121-252.us-east-2.compute.internal.node.eu-central-1.consul. 0	IN TXT "node_name=client-us-east-2-c5-2xlarge-infra"
0a3479fc.addr.eu-central-1.consul. 0 IN	A	10.52.121.252
ip-10-52-121-252.us-east-2.compute.internal.node.eu-central-1.consul. 0	IN TXT "node_name=client-us-east-2-c5-2xlarge-infra"
ip-10-52-121-252.us-east-2.compute.internal.node.eu-central-1.consul. 0	IN TXT "region=eu-central-1"
ip-10-52-121-252.us-east-2.compute.internal.node.eu-central-1.consul. 0	IN TXT "consul-network-segment="
0a18bd22.addr.eu-central-1.consul. 0 IN	A	10.24.189.34
ip-10-24-189-34.eu-central-1.compute.internal.node.eu-central-1.consul.	0 IN TXT "node_name=client-eu-central-1-c5-2xlarge-infra"
ip-10-24-189-34.eu-central-1.compute.internal.node.eu-central-1.consul.	0 IN TXT "region=eu-central-1"
ip-10-24-189-34.eu-central-1.compute.internal.node.eu-central-1.consul.	0 IN TXT "consul-network-segment="
0a181be0.addr.eu-central-1.consul. 0 IN	A	10.24.27.224
ip-10-24-27-224.eu-central-1.compute.internal.node.eu-central-1.consul.	0 IN TXT "consul-network-segment="
ip-10-24-27-224.eu-central-1.compute.internal.node.eu-central-1.consul.	0 IN TXT "node_name=client-eu-central-1-c5-2xlarge-infra"
ip-10-24-27-224.eu-central-1.compute.internal.node.eu-central-1.consul.	0 IN TXT "region=eu-central-1"
0a185e5d.addr.eu-central-1.consul. 0 IN	A	10.24.94.93
ip-10-24-94-93.eu-central-1.compute.internal.node.eu-central-1.consul. 0 IN TXT	"node_name=client-eu-central-1-c5-2xlarge-infra"
ip-10-24-94-93.eu-central-1.compute.internal.node.eu-central-1.consul. 0 IN TXT	"region=eu-central-1"
ip-10-24-94-93.eu-central-1.compute.internal.node.eu-central-1.consul. 0 IN TXT	"consul-network-segment="
0a18bd22.addr.eu-central-1.consul. 0 IN	A	10.24.189.34
ip-10-24-189-34.eu-central-1.compute.internal.node.eu-central-1.consul.	0 IN TXT "node_name=client-eu-central-1-c5-2xlarge-infra"
ip-10-24-189-34.eu-central-1.compute.internal.node.eu-central-1.consul.	0 IN TXT "region=eu-central-1"
ip-10-24-189-34.eu-central-1.compute.internal.node.eu-central-1.consul.	0 IN TXT "consul-network-segment="
0a181be0.addr.eu-central-1.consul. 0 IN	A	10.24.27.224
ip-10-24-27-224.eu-central-1.compute.internal.node.eu-central-1.consul.	0 IN TXT "node_name=client-eu-central-1-c5-2xlarge-infra"
ip-10-24-27-224.eu-central-1.compute.internal.node.eu-central-1.consul.	0 IN TXT "region=eu-central-1"
ip-10-24-27-224.eu-central-1.compute.internal.node.eu-central-1.consul.	0 IN TXT "consul-network-segment="
0a18bd22.addr.eu-central-1.consul. 0 IN	A	10.24.189.34
ip-10-24-189-34.eu-central-1.compute.internal.node.eu-central-1.consul.	0 IN TXT "node_name=client-eu-central-1-c5-2xlarge-infra"
ip-10-24-189-34.eu-central-1.compute.internal.node.eu-central-1.consul.	0 IN TXT "region=eu-central-1"
ip-10-24-189-34.eu-central-1.compute.internal.node.eu-central-1.consul.	0 IN TXT "consul-network-segment="

;; Query time: 209 msec
;; SERVER: 127.0.0.1#53(127.0.0.1)
;; WHEN: Fri Jul 01 05:23:25 UTC 2022
;; MSG SIZE  rcvd: 2123
```

blaggacao: I would pick one random and stick with it.  That gives us _some_ very naive load balancing.

blaggacao: The proper solution would be a ring hash:
- https://medium.com/swlh/load-balancing-and-consistent-hashing-5fe0156035e1
- https://blog.getambassador.io/load-balancing-strategies-in-kubernetes-l4-round-robin-l7-round-robin-ring-hash-and-more-6a5b81595d6c

blaggacao: Ring Hash is a good solution when some "stickiness" is desired, because doing so would increase performance.

blaggacao: Side Note: Consul allows us to route the requests in the same data-center only, in case you wondered :smile:

blaggacao:
This is coming along well, I'm excited.

In general `hostname` is the wrong ontology though, since `serviceName` are precisely one thing not: hosts :smile:

packages/cardano-services/src/Program/utils.ts
IvayloAndonov: It says: `The ring hash approach is used for both “sticky sessions” (where a cookie is set to ensure that all requests from a client arrive at the same pod) and for “session affinity” (which relies on client IP or some other piece of client state).` so should we use some sort of caching the resolved srv record `port` and try to stick with that one in case of on re-resolving or just get the first resolved record would do the job at first stage?

rhyslbw: Let's just do random selection and make it sticky for **reconnects** by storing a reference in memory. I think it's acceptable since a reconnection timeout indicates there's a complete outage, which is unlikely, and stickiness doesn't really matter in that context.

packages/cardano-services/src/Program/utils.ts
```diff
@@ -0,0 +1,98 @@
...
+    if (retriesLeft === 0) {
+      logger.error(message);
+      // await server.shutdown(); ?
+      // eslint-disable-next-line unicorn/no-process-exit
+      process.exit(1);
+    }
+  };
+
+export const resolveDnsSrvWithExponentialBackoff = async (
+  servicename: string,
```
rhyslbw, line 32:
```suggestion
  serviceName: string,
```

rhyslbw (changes requested):

packages/cardano-services/src/cli.ts
IvayloAndonov: https://github.com/input-output-hk/cardano-js-sdk/pull/309/commits/0a73034d6d30a95a36abf1fb7a2d069b2ce3cf1e

packages/cardano-services/src/cli.ts
IvayloAndonov: https://github.com/input-output-hk/cardano-js-sdk/pull/309/commits/0a73034d6d30a95a36abf1fb7a2d069b2ce3cf1e

packages/cardano-services/src/Program/loadHttpServer.ts
IvayloAndonov: https://github.com/input-output-hk/cardano-js-sdk/pull/309/commits/0a73034d6d30a95a36abf1fb7a2d069b2ce3cf1e

packages/cardano-services/src/Program/loadHttpServer.ts
IvayloAndonov: https://github.com/input-output-hk/cardano-js-sdk/pull/309/commits/0a73034d6d30a95a36abf1fb7a2d069b2ce3cf1e

packages/cardano-services/src/Program/ProgramOptionDescriptions.ts
IvayloAndonov: https://github.com/input-output-hk/cardano-js-sdk/pull/309/commits/0a73034d6d30a95a36abf1fb7a2d069b2ce3cf1e

packages/cardano-services/src/Program/utils.ts
IvayloAndonov: https://github.com/input-output-hk/cardano-js-sdk/pull/309/commits/879931b375aa24e7ea3c4304e8a6c7c67b07e8ab

packages/cardano-services/src/Program/utils.ts
IvayloAndonov: Storing a random selection of resolved DNS address in memory is applied here https://github.com/input-output-hk/cardano-js-sdk/pull/309/commits/b0f80d11b65d8de72b1188aa4c1a33489fb90643

packages/cardano-services/src/cli.ts
```diff
@@ -101,16 +105,40 @@ commonOptions(
...
+  .option('--postgres-uid <postgresUser>', ProgramOptionDescriptions.DbConnection)
+  .option('--postgres-password <postgresPassword>', ProgramOptionDescriptions.DbConnection)
   .action(async (serviceNames: ServiceNames[], options: { apiUrl: URL } & HttpServerOptions) => {
     const { apiUrl, ...rest } = options;
-    const server = await loadHttpServer({ apiUrl: apiUrl || API_URL_DEFAULT, options: rest, serviceNames });
-    await server.initialize();
-    await server.start();
-    onDeath(async () => {
-      await server.shutdown();
+    try {
```
IvayloAndonov, line 129: Align it with `run.ts` - `loadHttpServer` is wrapped within a try-catch block

packages/cardano-services/src/Program/utils.ts
IvayloAndonov: I need pair of eyes about the correctness of the applied logic  [getStikyAddressWithDnsSrv](https://github.com/input-output-hk/cardano-js-sdk/pull/309/commits/b9663759a1c458ba742e717761b68bace447ffc6#diff-5935fb07258956001ade74f0edc794b473e06615d67973cc4d416ec157357be3R42-R59)

rhyslbw:
>  Use case: Ogmios and RabbitMQ URLs args default to values if they are not passed.
option b) we could leave the defaulting and just take srv service name presence with priority over the URL 

Let's give service discovery config preference, which will leave us with the ability to fall back on fixed defaults.

> Conflict with the introduced [DB credentials args](https://github.com/input-output-hk/cardano-js-sdk/pull/294/files#diff-cd616b155b9729672e02921d8863a908cbaf8d92d935ca96f9708a083c87959b) by https://github.com/input-output-hk/cardano-js-sdk/pull/294
I guess need to remove POSTGRES_NAME, POSTGRES_USER, POSTGRES_PASSWORD and use those ones from files 

No, loading from files should always be an option for the ENV-based run scripts, as it serves as a more secure way rather than setting environment variables with the value in plain text. I'm OK with `POSTGRES_PASSWORD` etc being introduced alongside `POSTGRES_PASSWORD_FILE` so that we're consistent with the CLI.

packages/cardano-services/src/cli.ts
rhyslbw: Good catch :muscle:

IvayloAndonov:
The long list of merge conflicts with _master_ is being resolved under [71cb21b](https://github.com/input-output-hk/cardano-js-sdk/pull/309/commits/71cb21b8ce481c4d03ef5192be7fc93faac0ef0b)

packages/cardano-services/test/Program/utils.test.ts
```diff
@@ -0,0 +1,373 @@
...
+        await httpServer.initialize();
+        await httpServer.start();
+      });
+
+      afterAll(async () => {
+        await httpServer.shutdown();
+        await serverClosePromise(ogmiosServer);
+      });
+
+      it('txSubmitProvider should be a instance of Proxy ', () => {
```
rhyslbw, line 270:
```suggestion
      it('txSubmitProvider should not be a instance of Proxy ', () => {
```

packages/cardano-services/test/entrypoints.test.ts
```diff
@@ -298,6 +313,80 @@ describe('entrypoints', () => {
         });
       });
 
+      describe('specifying a PostgreSQL-dependent services with providing service discovery args', () => {
```
rhyslbw, line 316:
```suggestion
      describe('specifying PostgreSQL-dependent services with service discovery args', () => {
```

packages/cardano-services/test/entrypoints.test.ts
```diff
@@ -298,6 +313,80 @@ describe('entrypoints', () => {
...
+          });
+
+          proc.on('exit', (code) => {
+            expect(code).toBe(1);
+            expect(spy).toHaveBeenCalled();
+            done();
+          });
+        });
+
+        it('run throws dns srv error and exits with code 1', (done) => {
```
rhyslbw, line 363:
```suggestion
        it('run throws DNS SRV error and exits with code 1', (done) => {
```
Also check for other cases

packages/cardano-services/test/entrypoints.test.ts
```diff
@@ -384,6 +473,69 @@ describe('entrypoints', () => {
       });
     });
 
+    describe('specifying a Ogmios-dependent service with providing service discovery name', () => {
```
rhyslbw, line 476:
```suggestion
    describe('specifying an Ogmios-dependent service with service discovery args', () => {
```

packages/cardano-services/test/entrypoints.test.ts
```diff
@@ -447,6 +599,71 @@ describe('entrypoints', () => {
       });
     });
 
+    describe('specifying a RabbitMQ-dependent service with providing service discovery name', () => {
```
rhyslbw, line 602:
```suggestion
    describe('specifying a RabbitMQ-dependent service with service discovery args', () => {
```

packages/cardano-services/src/Program/utils.ts
```diff
@@ -0,0 +1,236 @@
...
+    );
+    if (retriesLeft === 0) {
+      logger.error(message);
+      // Invokes onDeath() callback within cardano-services entrypoints, following by server.shutdown() and process.exit(1)
+      process.kill(process.pid, 'SIGTERM');
+    }
+  };
+
+// Get a random selection of dns srv resolved service address initially and make it sticky for reconnects by storing a reference in memory
+export const getStikyAddressWithDnsSrv = async (serviceName: string, cache: InMemoryCache): Promise<SrvRecord> => {
```
rhyslbw, line 43:
```suggestion
export const resolveMaybeCachedSrvRecord = async (serviceName: string, cache: InMemoryCache): Promise<SrvRecord> => {
```
I think we can probably depend on the concept of a cached record, rather than use the term _sticky_.

packages/cardano-services/src/Program/utils.ts
```diff
@@ -0,0 +1,236 @@
...
+  const stickyAddress = cache.getVal<SrvRecord>(`${DNS_SRV_CACHE_KEY}/${serviceName}`);
```
rhyslbw, line 44:
```suggestion
  const cachedSrvRecord = cache.getVal<SrvRecord>(`${DNS_SRV_CACHE_KEY}/${serviceName}`);
```

packages/cardano-services/src/Program/utils.ts
```diff
@@ -0,0 +1,236 @@
...
+  if (!stickyAddress) {
+    const [address] = await dns.promises.resolveSrv(serviceName);
```
rhyslbw, line 46:
```suggestion
    const [srvRecord] = await dns.promises.resolveSrv(serviceName);
```

packages/cardano-services/src/Program/utils.ts
```diff
@@ -0,0 +1,236 @@
...
+      }, ${nextAction}`
+    );
+    if (retriesLeft === 0) {
+      logger.error(message);
+      // Invokes onDeath() callback within cardano-services entrypoints, following by server.shutdown() and process.exit(1)
+      process.kill(process.pid, 'SIGTERM');
+    }
+  };
+
+// Get a random selection of dns srv resolved service address initially and make it sticky for reconnects by storing a reference in memory
```
rhyslbw, line 42:
```suggestion
// Return the cached record if present for stickiness with the service dependency, otherwise, select a random record from the DNS server
```

rhyslbw (changes requested):
Looking good @IvayloAndonov 

I think it would be easier to grok if each of the service dependencies were isolated in their own files under `Program/services/`, but that's a minor point we can leave for now. Some inline feedback

packages/cardano-services/src/Program/utils.ts
```diff
@@ -1,58 +1,89 @@
...
-      maxRetryTime,
-      onFailedAttempt: onFailedAttemptFor(`Establishing connection to ${serviceName}`, logger)
-    }
-  );
+// Get a random selection of dns srv resolved service address initially and make it sticky for reconnects by storing a reference in memory
+export const getStikyAddressWithDnsSrv = async (serviceName: string, cache: InMemoryCache): Promise<SrvRecord> => {
+  const stickyAddress = cache.getVal<SrvRecord>(`${DNS_SRV_CACHE_KEY}/${serviceName}`);
+  if (!stickyAddress) {
+    const [address] = await dns.promises.resolveSrv(serviceName);
+    cache.set(`${DNS_SRV_CACHE_KEY}/${serviceName}`, address, UNLIMITED_CACHE_TTL);
```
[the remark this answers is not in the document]
IvayloAndonov: As I see from the commit it will default to `#ttlDefault` which is [120 mins](https://github.com/input-output-hk/cardano-js-sdk/blob/a2bfcc62c25e71d78d07b961267d7ce9679b6cf4/packages/cardano-services/src/InMemoryCache/defaults.ts#L1) but in this case, here we would like to cache it during the whole `cardano-services` instance lifetime, no? @rhyslbw

packages/cardano-services/src/Program/utils.ts
```diff
@@ -76,23 +107,143 @@ export const getSrvPool = async (
...
+      // Bind if it is a function, no intercept operations
+      if (typeof ogmiosProvider[prop as keyof TxSubmitProvider] === 'function') {
+        const method = ogmiosProvider[prop as keyof TxSubmitProvider] as any;
+        return method.bind(ogmiosProvider);
+      }
+    }
+  });
+};
+
+export const getCardanoNodeProvider = async (
```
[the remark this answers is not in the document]
IvayloAndonov: This is already resolved from prev fixup commit, probably you started the review a moments before the commit it is pushed https://github.com/input-output-hk/cardano-js-sdk/pull/309/commits/b8a743bdfb21aa6c0ae265eb42ca9d1ffbf73d78

packages/cardano-services/test/Program/utils.test.ts
```diff
@@ -0,0 +1,373 @@
...
+    let config: HttpServerConfig;
+    let service: NetworkInfoHttpService;
+    let networkInfoProvider: DbSyncNetworkInfoProvider;
+
+    beforeAll(async () => {
+      db = await getPool(dnsSrvResolve, {
+        cacheTtl: 10_000,
+        dbConnectionString: process.env.DB_CONNECTION_STRING,
+        epochPollInterval: 1000,
+        serviceDiscoveryBackoffFactor: 1.1,
```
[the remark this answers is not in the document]
IvayloAndonov: By definition some of the [HTTP server options](https://github.com/input-output-hk/cardano-js-sdk/blob/b8a743bdfb21aa6c0ae265eb42ca9d1ffbf73d78/packages/cardano-services/src/Program/loadHttpServer.ts#L27-L36) are required as they always have default values if they are not provided as args to the program, so this is some sort of alignment with the `HttpServerOptions` interface in the tests

packages/cardano-services/src/Program/utils.ts
IvayloAndonov: I was wondering here what would be better as terminology since both seems valid: resolved _addresses_ (based on the DNS [docs](https://nodejs.org/docs/latest-v14.x/api/dns.html#dns_dns_resolvesrv_hostname_callback) "The addresses argument passed to the callback function will be an array of objects..." OR _records_ (based on the `dns.SrvRecord` type name itself) ? To me personally, address sounds more intuitive but totally fine to use _record_ as well? @rhyslbw

packages/cardano-services/test/Program/utils.test.ts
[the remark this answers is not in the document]
rhyslbw: Seems like `loadHttpServer` `args` would be better to not require `serviceDiscoveryBackoffFactor` or `serviceDiscoveryBackoffTimeout`, as they're only used with service discovery :thinking:

packages/cardano-services/src/Program/utils.ts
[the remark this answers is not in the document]
rhyslbw: Ah yes, sorry, I thought it was going to use the underlying library default

packages/cardano-services/test/Program/utils.test.ts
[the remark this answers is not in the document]
IvayloAndonov: Agree on that, the thing I intended to express from `HttpServerOptions` interface is that they are not required but they will be always provided regardless of which service/provider will use them. 

Following your comment, `cacheTtl` and `epochPollInterval` should be optional as well as they are used if the NetworkInfo service is attached to the server?

packages/cardano-services/src/Program/utils.ts
rhyslbw: It's inconsistent in the docs. Here's the promise-based description:
image

Let's go with the more descriptive option of `srvRecord`.

IvayloAndonov:
> Looking good @IvayloAndonov
> 
> ... Some inline feedback

The feedback is addressed under https://github.com/input-output-hk/cardano-js-sdk/pull/309/commits/ed5ffb2ed8b96c20cec45b53a4786662fd2d2333 .Just one comment left from my side [link](https://github.com/input-output-hk/cardano-js-sdk/pull/309#discussion_r917659417).
Please check the initial naming [suggestions](https://github.com/input-output-hk/cardano-js-sdk/pull/309#pullrequestreview-1025704549) which are being resolved as well.

IvayloAndonov:
Extending [loadTxWorker](https://github.com/input-output-hk/cardano-js-sdk/blob/cb61f300158ce69dab382778fc1e39861abd69a4/packages/cardano-services/src/TxWorker/loadTxWorker.ts#L14) and moving all service discovery related args to common options (used by both the HTTP server and worker) changes are introduced under https://github.com/input-output-hk/cardano-js-sdk/pull/309/commits/90cdd8de1f1266f46e67ed93b08c2c91466634a1 commit.

@iccicci would be great if you take a look as well, thank you!

packages/cardano-services/src/Program/utils.ts
```diff
@@ -0,0 +1,249 @@
...
+    rabbitmqUrl: srvRecordToRabbitmqURL(record)
+  });
+
+  return new Proxy<RabbitMqTxSubmitProvider>({} as RabbitMqTxSubmitProvider, {
+    get(_, prop) {
+      if (prop === 'then') return;
+      if (prop === 'submitTx') {
+        return (args: Uint8Array) =>
+          rabbitmqProvider.submitTx(args).catch(async (error) => {
+            if (error instanceof ProviderError && error.innerError === ProviderFailure.ConnectionFailure) {
```
iccicci, line 206: Really nice implementation!!

The only thing to note: in case the condition in this `if` is not matched, the error should be propagated to the caller.
I'm not expert with this construct: it could be it is propagated implicitly... I'm simply rising the flag as I can't understand how (since I don't see this explicitly).

packages/cardano-services/src/TxWorker/loadTxWorker.ts
```diff
@@ -12,12 +12,19 @@ export interface TxWorkerArgs {
 }
 
 export const loadTxWorker = async (args: TxWorkerArgs) => {
-  const { loggerMinSeverity, ogmiosUrl, rabbitmqUrl } = args.options;
-  const txSubmitProvider = ogmiosTxSubmitProvider(urlToConnectionConfig(ogmiosUrl));
+  const { loggerMinSeverity, serviceDiscoveryBackoffFactor, serviceDiscoveryBackoffTimeout, cacheTtl } = args.options;
   const logger = createLogger({ level: loggerMinSeverity, name: 'tx-worker' });
+  const cache = new InMemoryCache(cacheTtl);
+  const dnsResolver = createDnsResolver(
```
iccicci, line 18: The entire file LGTM. I simply have not clear how this can work when the related options are not provided from command line... but looking at the tests, it is clear that or it works or tests will fail!

packages/cardano-services/src/Program/utils.ts
IvayloAndonov: Good catch @iccicci , the missing line of code is introduced here https://github.com/input-output-hk/cardano-js-sdk/pull/309/commits/cd4da53c23624baa6b7bf944ec079ae9ed82fac1 : in case the error doesn't match the condition for connection error within the `Proxy` -> re-throw the error unchanged which propagates the error to the caller.

Unfortunately, I can not see currently how this case could be unit tested...
With service discovery, we use `Proxy` abstraction over the dependency services where in case of connection error runtime, we always try to [re-connect to a new instance](https://github.com/input-output-hk/cardano-js-sdk/blob/cd4da53c23624baa6b7bf944ec079ae9ed82fac1/packages/cardano-services/src/Program/utils.ts#L95-L99) internally of given dependency (_postgres,rabbitmq or ogmios_), or in this case: any other error type to be propagated unchanged. The main issue is with mocking such behavior with the underlying dependencies. (`pg.Pool`, `rabbitmqTxProvider` , `ogmiosTcProvider`)

As I am aware, simulating such behavior (killing a given healthy and running instance and expecting to recover/restart in some way) is associated with chaos testing with appropriate tools. Any suggestions are welcomed!

packages/cardano-services/test/Program/utils.test.ts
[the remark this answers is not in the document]
rhyslbw: Please capture this issue as follow-up work since it's not critical to address here

packages/cardano-services/src/cli.ts
```diff
@@ -71,6 +72,21 @@ const commonOptions = (command: Command) =>
...
+    .option('--rabbitmq-srv-service-name <rabbitmqSrvServiceName>', ProgramOptionDescriptions.RabbitMQSrvServiceName)
+    .option('--cache-ttl <cacheTtl>', ProgramOptionDescriptions.CacheTtl, cacheTtlValidator, CACHE_TTL_DEFAULT)
+    .option(
+      '--service-discovery-backoff-factor <serviceDiscoveryBackoffFactor>',
+      ProgramOptionDescriptions.ServiceDiscoveryBackoffFactor,
+      (factor) => Number.parseFloat(factor),
+      SERVICE_DISCOVERY_BACKOFF_FACTOR_DEFAULT
+    )
+    .option(
+      '--service-discovery-backoff-timeout <serviceDiscoveryBackoffTimeout>',
```
rhyslbw, line 86:
```suggestion
      '--service-discovery-timeout <serviceDiscoveryTimeout>',
```

packages/cardano-services/src/ProgramsCommon/options.ts
```diff
@@ -9,4 +9,9 @@ export interface CommonProgramOptions {
   loggerMinSeverity?: LogLevel;
   ogmiosUrl?: URL;
   rabbitmqUrl?: URL;
+  ogmiosSrvServiceName?: string;
+  rabbitmqSrvServiceName?: string;
+  serviceDiscoveryBackoffFactor?: number;
+  serviceDiscoveryBackoffTimeout?: number;
```
rhyslbw, line 15:
```suggestion
  serviceDiscoveryTimeout?: number;
```

rhyslbw (changes requested):

packages/cardano-services/src/ProgramsCommon/options.ts
IvayloAndonov: bd44f5e3e91165c3ca6bda66d1a81e9e18de3a83

packages/cardano-services/src/cli.ts
IvayloAndonov: bd44f5e3e91165c3ca6bda66d1a81e9e18de3a83

packages/cardano-services/src/ProgramsCommon/CommonOptionDescriptions.ts
```diff
@@ -1,5 +1,10 @@
 export enum CommonOptionDescriptions {
   LoggerMinSeverity = 'Log level',
   OgmiosUrl = 'Ogmios URL',
-  RabbitMQUrl = 'RabbitMQ URL'
+  RabbitMQUrl = 'RabbitMQ URL',
+  OgmiosSrvServiceName = 'Ogmios SRV service name',
+  RabbitMQSrvServiceName = 'RabbitMQ SRV service name',
+  ServiceDiscoveryBackoffFactor = 'Exponential backoff factor for service discovery',
+  ServiceDiscoveryTimeout = 'Exponential backoff max timeout for service discovery',
```
rhyslbw, line 8:
```suggestion
  ServiceDiscoveryTimeout = 'Timeout for service discovery attempts',
```

packages/cardano-services/src/run.ts
```diff
@@ -27,13 +36,21 @@ const envSpecs = {
...
+  POSTGRES_PASSWORD: envalid.str({ default: undefined }),
   POSTGRES_PASSWORD_FILE: existingFileValidator({ default: undefined }),
   POSTGRES_PORT: envalid.num({ default: undefined }),
+  POSTGRES_SRV_SERVICE_NAME: envalid.str({ default: undefined }),
+  POSTGRES_USER: envalid.str({ default: undefined }),
   POSTGRES_USER_FILE: existingFileValidator({ default: undefined }),
+  RABBITMQ_SRV_SERVICE_NAME: envalid.str({ default: undefined }),
   RABBITMQ_URL: envalid.url({ default: RABBITMQ_URL_DEFAULT }),
+  SERVICE_DISCOVERY_BACKOFF_FACTOR: envalid.num({ default: SERVICE_DISCOVERY_BACKOFF_FACTOR_DEFAULT }),
+  SERVICE_DISCOVERY_BACKOFF_TIMEOUT: envalid.num({ default: SERVICE_DISCOVERY_TIMEOUT_DEFAULT }),
```
rhyslbw, line 53:
```suggestion
  SERVICE_DISCOVERY_TIMEOUT: envalid.num({ default: SERVICE_DISCOVERY_TIMEOUT_DEFAULT }),
```

packages/cardano-services/src/run.ts
```diff
@@ -45,10 +62,18 @@ void (async () => {
...
   const apiUrl = new URL(env.API_URL);
   const ogmiosUrl = new URL(env.OGMIOS_URL);
+  const ogmiosSrvServiceName = env.OGMIOS_SRV_SERVICE_NAME;
   const rabbitmqUrl = new URL(env.RABBITMQ_URL);
+  const rabbitmqSrvServiceName = env.RABBITMQ_SRV_SERVICE_NAME;
   const cardanoNodeConfigPath = env.CARDANO_NODE_CONFIG_PATH;
   const cacheTtl = env.CACHE_TTL;
   const epochPollInterval = env.EPOCH_POLL_INTERVAL;
+  const serviceDiscoveryBackoffFactor = env.SERVICE_DISCOVERY_BACKOFF_FACTOR;
+  const serviceDiscoveryTimeout = env.SERVICE_DISCOVERY_BACKOFF_TIMEOUT;
```
rhyslbw, line 72:
```suggestion
  const serviceDiscoveryTimeout = env.SERVICE_DISCOVERY_TIMEOUT;
```

rhyslbw (changes requested):
Just a few more to resolve :smile:

packages/cardano-services/src/TxWorker/loadTxWorker.ts
iccicci: Ok, now it's clear, thank you

packages/cardano-services/src/Program/utils.ts
iccicci: I could suggest to start an ogmios mock like [this one](https://github.com/input-output-hk/cardano-js-sdk/blob/2cd240513a4632da7c6fef9274e65d025e803048/packages/rabbitmq/test/TxSubmitWorker.test.ts#L152).
This should be enough to test `ogmiosTxSubmitProviderWithDiscovery` while to test `rabbitMqTxSubmitProviderWithDiscovery` a txWorker would be required as well (while running `cardano-services` tests are running a RabbitMQ server is running as well inside docker).

IvayloAndonov: The failovers cases are covered under https://github.com/input-output-hk/cardano-js-sdk/pull/309/commits/87b50012fd1e88e722358e8f89bf17c797db9abc

packages/cardano-services/test/Program/loadHttpServer.test.ts
```diff
@@ -67,48 +102,223 @@ describe('loadHttpServer', () => {
...
+            serviceDiscoveryTimeout,
+            useQueue: true
+          },
+          serviceNames: [ServiceNames.TxSubmit]
+        });
+
+        expect(httpServer).toBeInstanceOf(HttpServer);
+      });
+
+      it('throws if a service is nominated without providing rabbitmq url nor service discovery name', async () => {
```
rhyslbw, line 283: Can we not interpret this to use the default RabbitMQ URL?

packages/cardano-services/test/Program/loadHttpServer.test.ts
```diff
@@ -67,48 +102,223 @@ describe('loadHttpServer', () => {
...
+            })
+        ).rejects.toThrow(
+          new MissingProgramOption(ServiceNames.StakePool, [
+            ProgramOptionDescriptions.DbConnection,
+            ProgramOptionDescriptions.PostgresSrvArgs
+          ])
+        );
+      });
+
+      it('throws if a service is nominated without providing db connection string nor service discovery args', async () => {
```
rhyslbw, line 151: Can we not interpret this to use the default PostgreSQL connection string?

iccicci (approved):
Good refactoring and optimization of the tests we worked on together; now they are much more clear!

rhyslbw (dismissed):
Well done @IvayloAndonov :rocket:

iccicci (dismissed):
Good job!

blaggacao (approved):
Based on my shadowing of this PR: LGTM (no detailed code review, I'm no JS expert)

packages/cardano-services/package.json
```diff
@@ -112,4 +113,4 @@
     "LICENSE",
     "NOTICE"
   ]
-}
+}
```
rhyslbw, line 116: Please revert the line break removal

packages/e2e/package.json
```diff
@@ -110,4 +110,4 @@
     "webpack": "^5.72.0",
     "webpack-cli": "^4.9.2"
   }
-}
+}
```
rhyslbw, line 113: Please remove

rhyslbw (changes requested):
I think squashing 69ff160aa519418108b45abf791d560e6bea0a8d into 69ff160aa519418108b45abf791d560e6bea0a8d with a commit message headed with **feat(cardano-services): service discovery via DNS resolution** will be more appropriate than splitting them. 

I've highlighted a couple of out-of-scope changes that look to have been introduced by your editor. We should have this covered by our lint config, but until then, please avoid committing these, as we may end up flipping back and forth depending on who changed the file last.

Looking good otherwise!

packages/ogmios/src/index.ts
```diff
@@ -1,4 +1,4 @@
 export * from './Provider';
 export * from './CardanoNode';
 export * from './util';
-export { createConnectionObject, Connection, ConnectionConfig } from '@cardano-ogmios/client';
+export { createConnectionObject, Connection, ConnectionConfig, WebSocketClosed } from '@cardano-ogmios/client';
```
rhyslbw, line 4: As of #330, this is now a single export as `Ogmios`

packages/cardano-services/src/Program/utils.ts
```diff
@@ -0,0 +1,250 @@
...
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { ClientConfig, Pool, QueryConfig } from 'pg';
+import { CommonOptionDescriptions, CommonProgramOptions } from '../ProgramsCommon';
+import { HttpServerOptions } from './loadHttpServer';
+import { InvalidArgsCombination, MissingProgramOption } from './errors';
+import { ProgramOptionDescriptions } from './ProgramOptionDescriptions';
+import { ProviderFailure, TxSubmitProvider } from '@cardano-sdk/core';
+import { RabbitMqTxSubmitProvider } from '@cardano-sdk/rabbitmq';
+import { ServiceNames } from './ServiceNames';
+import { WebSocketClosed, ogmiosTxSubmitProvider, urlToConnectionConfig } from '@cardano-sdk/ogmios';
```
rhyslbw, line 11:
```suggestion
import { Ogmios } from '@cardano-sdk/ogmios';
```

packages/cardano-services/src/Program/utils.ts
```diff
@@ -0,0 +1,250 @@
...
+    });
+  }
+  // If db connection string is not passed nor postgres srv service name
+  return undefined;
+};
+
+export const isOgmiosConnectionError = (error: any) => {
+  if (
+    (error?.innerError?.code && ['ENOTFOUND', 'ECONNREFUSED'].includes(error.innerError.code)) ||
+    error instanceof WebSocketClosed
```
rhyslbw, line 116:
```suggestion
    error instanceof Ogmios.WebSocketClosed
```

rhyslbw (changes requested):

packages/cardano-services/src/Program/utils.ts
```diff
@@ -0,0 +1,250 @@
...
+  const { name, port } = await dnsResolver(host!);
+  let pool = new Pool({ database, host: name, password, port, user });
+
+  return new Proxy<Pool>({} as Pool, {
+    get(_, prop) {
+      if (prop === 'then') return;
+      if (prop === 'query') {
+        return (args: string | QueryConfig, values?: any) =>
+          pool.query(args, values).catch(async (error) => {
+            if (error.code && ['ENOTFOUND', 'ECONNREFUSED'].includes(error.code)) {
```
mkazlauskas, line 80: I'm wondering if we also need to watch for `ECONNRESET` :thinking:

packages/cardano-services/src/Program/utils.ts
```diff
@@ -0,0 +1,250 @@
...
+
+  return new Proxy<TxSubmitProvider>({} as TxSubmitProvider, {
+    get(_, prop) {
+      if (prop === 'then') return;
+      if (prop === 'submitTx') {
+        return (args: Uint8Array) =>
+          ogmiosProvider.submitTx(args).catch(async (error) => {
+            if (isOgmiosConnectionError(error)) {
+              const record = await dnsResolver(serviceName!);
+              ogmiosProvider = ogmiosTxSubmitProvider({ host: record.name, port: record.port });
```
mkazlauskas, line 148: We should attempt to close the old provider, same with rabbitmq provider
```suggestion
              await ogmiosProvider.close().catch(() => void 0); // maybe log a message if failed to close
              ogmiosProvider = ogmiosTxSubmitProvider({ host: record.name, port: record.port });
```

packages/cardano-services/src/TxWorker/loadTxWorker.ts
```diff
@@ -12,12 +11,17 @@ export interface TxWorkerArgs {
...
+    logger
+  );
+  const txSubmitProvider = await getOgmiosTxSubmitProvider(dnsResolver, args.options);
+  const url = await getRabbitMqUrl(dnsResolver, args.options);
 
-  // Ensure rabbitmqUrl is not undefined
-  if (!rabbitmqUrl) throw new MissingProgramOption(ServiceNames.TxSubmit, CommonOptionDescriptions.RabbitMQUrl);
-
-  return new TxSubmitWorker({ ...args.options, rabbitmqUrl }, { logger, txSubmitProvider });
+  return new TxSubmitWorker({ ...args.options, rabbitmqUrl: url }, { logger, txSubmitProvider });
```
mkazlauskas, line 26: We might want to also wrap this in a proxy - re-resolve if rabbitmq connection of the worker fails

mkazlauskas (changes requested):
Looking good, just a few things I noticed

packages/cardano-services/src/Program/utils.ts
IvayloAndonov: I've seen only the `ENOTFOUND`/ `ECONNREFUSED` during local testing, but sounds reasonable to watch for `ECONNRESET` as well just in case (`A connection was forcibly closed by a peer`). Will add it

packages/cardano-services/src/Program/utils.ts
mkazlauskas: `Provider` also has a `start?` but that doesn't seem to be used on any provider. Should we remove it from the interface? @rhyslbw @iccicci

packages/cardano-services/src/TxWorker/loadTxWorker.ts
IvayloAndonov: Yep makes sense, just need to verify which method needs to be intercepted by the proxy, I guess `start()` where connects to the rabbitmq. 
Do you think we should introduce it within this PR or as a smaller separate task? @mkazlauskas

mkazlauskas: It can fail after starting though. Maybe `this.#channel?.get` :thinking: Let's do this as separate task, maybe @iccicci can help to figure out the best way to do this for rabbitmq

packages/cardano-services/src/Program/utils.ts
rhyslbw: Let's leave it as-is, then consider it as part of the decorator refactor

packages/cardano-services/src/Program/utils.ts
IvayloAndonov: 8fd7c5267f8250848f954350f58bfef3735aea12

packages/cardano-services/src/Program/utils.ts
IvayloAndonov: 8fd7c5267f8250848f954350f58bfef3735aea12

packages/ogmios/src/index.ts
IvayloAndonov: All Ogmios imports-related diffs are directly resolved after rebasing with master, af689c041ef2ff4ef70ca8d5452c2416a7906baa

packages/e2e/package.json
IvayloAndonov: `e2e package.json` change is entirely omitted after rebasing with master, af689c041ef2ff4ef70ca8d5452c2416a7906baa

rhyslbw (approved):

PR #302, opened 2022-06-27
imp: build reproducibly

blaggacao:

PR #312, opened 2022-07-04
chore(core): add missing dependencies

blaggacao:


packages/cardano-services/package.json
```diff
@@ -61,7 +65,6 @@
     "dockerode": "^3.3.1",
     "dockerode-utils": "^0.0.7",
     "get-port-please": "^2.5.0",
-    "npm-run-all": "^4.1.5",
```
rhyslbw, line 64: :thinking: Usually [locally referenced](https://github.com/input-output-hk/cardano-js-sdk/pull/312/commits/5764345374eb2300a35f7ebb9252fc665cb6ec69#diff-718fd329469aa32e693b48626b09b792498085607dae1c0f8404fe5a8970d848L28) devDeps need to be defined in the local `package.json`, as the executables are unpacked locally.

blaggacao: This now works due to emulation: https://github.com/input-output-hk/cardano-js-sdk/pull/302/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R28

Included into the build dependencies: https://github.com/input-output-hk/cardano-js-sdk/pull/302/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R38

packages/core/src/Cardano/types/TxSubmissionErrors.ts
```diff
@@ -1,5 +1,6 @@
 import { CustomError } from 'ts-custom-error';
 import { TxSubmission } from '@cardano-ogmios/client';
+import _ from '@cardano-ogmios/schema';
```
mkazlauskas, line 3: What was the issue here? :thinking: Seems like there's some difference in `tsc` (using different version in nix? not using the same `tsconfig.json`?)
```suggestion
import '@cardano-ogmios/schema';
```

packages/util-dev/package.json
```diff
@@ -34,9 +34,15 @@
     "test:debug": "DEBUG=true yarn test",
     "test:e2e": "shx echo 'test:e2e' command not implemented yet"
   },
+  "devDependencies": {
+    "shx": "^0.3.3"
+  },
   "dependencies": {
     "@cardano-sdk/core": "0.3.0",
     "rxjs": "^7.4.0",
+    "ts-jest": "^27.1.3",
```
mkazlauskas, line 43: These 3 should go under `devDependencies`

packages/util-dev/src/stubLogger.ts
```diff
@@ -1,4 +1,5 @@
 import { Logger } from 'ts-log';
+import _ from 'jest';
```
mkazlauskas, line 2: `util-dev` should only be used in `package/test` project and you shouldn't need to build this
```suggestion
import 'jest';
```

mkazlauskas (changes requested):

packages/util-dev/src/stubLogger.ts
blaggacao: There is no distinction between build time and test time dependencies and various devDependencies are effectively build time dependencies, such as `node-run-all` or `shx` or `typescript`, for some examples. Hence, we need all dev dependencies in lack of a more granular distinction.

blaggacao:
@mkazlauskas I adopted your suggestions. thank you! :smile:

packages/core/src/Cardano/types/TxSubmissionErrors.ts
blaggacao: Some type is not imported from the schema, but used.

blaggacao: `tsc` is exactly the same version as specified ...

packages/util-dev/src/stubLogger.ts
mkazlauskas: Right, but `util-dev` should only be referenced by `[package]/test/tsconfig.json` and you should only need to build using `[package]/src/tsconfig.json` - so this `stubLogger.ts` file shouldn't be used by `tsc`. I'm wondering if you could reduce the number of packages to build by only building `cardano-services` and it's dependencies as [referenced](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services/src/tsconfig.json#L7) by it's tsconfig.json.

packages/core/src/Cardano/types/TxSubmissionErrors.ts
mkazlauskas: Would be great to understand what's the difference between the 2 environments that makes those extra imports needed :thinking:

rhyslbw: This seems like a workaround for bringing the Ogmios TypeDefs into scope, as this import is not required to satisfy the compiler given it can access it via `TxSubmission`.

packages/cardano-services/package.json
rhyslbw: Sure, but we still need it defined here (and any other `package.json` that uses it)

rhyslbw:
Closed in favor of #312

PR #314, opened 2022-07-04
[WIP]Feat/adp 1899 prototype local test network

AngelCastilloB:
# Context

Prototype an environment to demonstrate the use of a local network for our end-to-end.

# Proposed Solution
Create a local network and a set of tests to demonstrate how we can use a local test network to run our end-to-end tests faster and more reliably.

- Local test network was introduced.
- Set of test cases demonstrating how to use the local test network.


packages/e2e/test/faucet/faucet.test.ts
```diff
@@ -30,18 +79,44 @@ describe('CardanoWalletFaucetProvider', () => {
...
-      ],
-      [33_000_000, 22_000_000]
-    );
+    wallet.transactions.history$
+      .pipe(filter((txs) => txs.filter((tx) => tx.id === signedTx.id).length === 1))
+      .pipe(map((txs) => txs[0]))
+      .subscribe((transaction) => {
+        logger.debug(transaction);
+        done();
+      });
```
mkazlauskas, line 120:
```suggestion
      });
    wallet.shutdown();
```

PR #317, opened 2022-07-05
Chore/tip tracker and cold obs wallet args

mirceahasegan:
# Context

These changes are needed in preparation for adding connection status events.

# Proposed Solution
1. Use named parameters for coldObservableProvider to avoid having too many positional args, especially since connection status will also be a parameter.
1. Lengthy named SyncableIntervalPersistentDocumentTrackerSubject is renamed to TipTracker because it is only used for this purpose. There are also some refactoring here:
  - the type is no longer generic, it is a Cardano.Tip
  - equals is internal to the TipTracker (no longer configurable). 

### DIFF
--- a/packages/wallet/src/SingleAddressWallet.ts
+++ b/packages/wallet/src/SingleAddressWallet.ts
@@ -22,7 +22,7 @@ import {
   FailedTx,
   PersistentDocumentTrackerSubject,
   PollingConfig,
-  SyncableIntervalPersistentDocumentTrackerSubject,
+  TipTracker,
   TrackedAssetProvider,
   TrackedChainHistoryProvider,
   TrackedNetworkInfoProvider,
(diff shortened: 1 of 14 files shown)


mirceahasegan:
I was not able to run the e2e tests

mkazlauskas (changes requested):
Looks good, I verified that e2e tests are still passing.

Commit ~~scope~~type should probably be `refactor` instead of `chore`, I don't think breaking chores should be a thing :sweat_smile:

mirceahasegan:
> 

I have got the commit message rename ready and will force push at the end of the day to allow other review comments to come in.

mirceahasegan:
commit messages are updated to `refactor`

rhyslbw (approved):

mkazlauskas (approved):

PR #318, opened 2022-07-05
feat(cardano-services): adds token metadata service

iccicci:
# Context

# Proposed Solution


packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  async getTokenMetadata(subject: string): Promise<Asset.TokenMetadata | null> {
+    if (this.#cache[subject]) return this.#cache[subject];
+
+    this.#logger.debug(`Requesting for metatada for subject "${subject}"`);
+    const metadata: Asset.TokenMetadata = { version: '1.0' };
+
+    try {
+      const { data } = await axios.get(`https://tokens.cardano.org/metadata/${subject}`);
```
rhyslbw, line 41: 1. The URL should also be passed in via the constructor.
2. Make a batch POST request like [here](https://github.com/input-output-hk/cardano-graphql/blob/ca5d556bcf0c2bd0b022ee3c293851b7ddd8f6f7/packages/api-cardano-db-hasura/src/MetadataClient.ts#L40-L50), also specifying the properties we know about.

https://input-output-hk.github.io/offchain-metadata-tools/api/latest/ -> `Batch Metadata Query`

packages/cardano-services/src/Asset/types.ts
```diff
@@ -16,3 +16,15 @@ export interface NftMetadataService {
...
+/**
+ * Service to get CIP-35 token metadata for a given subject
+ */
+export interface TokenMetadataService {
+  /**
+   * Get CIP-35 token metadata for a given subject
+   *
+   * @returns CIP-35 token metadata, `null` if not found
+   */
+  getTokenMetadata(subject: string): Promise<Asset.TokenMetadata | null>;
```
rhyslbw, line 29:
```suggestion
  getTokenMetadata(assetIds: AssetId[]): Promise<Asset.TokenMetadata[] | null>;
```
The Cardano Token Registry, on which we're basing our interface, supports batch fetching in a single request.

packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
```diff
@@ -0,0 +1,40 @@
+import { Asset } from '@cardano-sdk/core';
+import { CardanoTokenRegistry } from '../../src/Asset';
+import { dummyLogger } from 'ts-log';
+
+describe('CardanoTokenRegistry', () => {
+  const tokenRegistry = new CardanoTokenRegistry({ logger: dummyLogger });
+  const validSubject = 'f43a62fdc3965df486de8a0d32fe800963589c41b38946602a0dc53541474958';
```
rhyslbw, line 7:
```suggestion
  const validAssetId = 'f43a62fdc3965df486de8a0d32fe800963589c41b38946602a0dc53541474958';
```

packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  it('returns null for unexisting subject', async () => {
```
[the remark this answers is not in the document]
rhyslbw: This will return an empty array when using the batch request

rhyslbw (changes requested):

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
```diff
@@ -0,0 +1,56 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  #checkStringValue(data: any, attribute: string) {
+    if (attribute in data && typeof data[attribute].value === 'string') return data[attribute].value as string;
+  }
+
+  async getTokenMetadata(subject: string): Promise<Asset.TokenMetadata | null> {
+    if (this.#cache[subject]) return this.#cache[subject];
+
+    this.#logger.debug(`Requesting for metatada for subject "${subject}"`);
+    const metadata: Asset.TokenMetadata = { version: '1.0' };
```
rhyslbw, line 38: Can probably just omit this hard-coded value given it's the default: https://github.com/input-output-hk/cardano-js-sdk/blob/1a873ec182c901813feae8244c86c0346bb70022/packages/core/src/Asset/types/TokenMetadata.ts#L58

packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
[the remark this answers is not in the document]
iccicci: Rather than an empty array I'm returning an array of `null`s to be able to distinguish the assets with metadata between the ones without in case of requests with mixed good and bad asset ids.
I need confirmation this is a good option.
Thank you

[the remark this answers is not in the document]
rhyslbw: Yes, that's a suitable return value.

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
```diff
@@ -0,0 +1,104 @@
+import { Asset, Cardano } from '@cardano-sdk/core';
+import { Logger } from 'ts-log';
+import { TokenMetadataService } from './types';
```
rhyslbw, line 3: The interface would be better sourced in the `core` package, although I'm not sure where. Maybe `core/src/Asset/types/TokenMetadataService.ts`

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
```diff
@@ -0,0 +1,104 @@
...
+  #checkStringValue(data: any, attribute: string) {
+    if (attribute in data && typeof data[attribute].value === 'string') return data[attribute].value as string;
+  }
+
+  async getTokenMetadata(assetIds: Cardano.AssetId[]): Promise<(Asset.TokenMetadata | null)[]> {
+    const assetIdsToRequest: Cardano.AssetId[] = [];
+    // eslint-disable-next-line unicorn/no-new-array
+    const ret: (Asset.TokenMetadata | null)[] = new Array(assetIds.length).fill(null);
+
+    this.#logger.info(`Requested asset metatada for "${assetIds}"`);
```
rhyslbw, line 59:
```suggestion
    this.#logger.debug(`Requested asset metatada for "${assetIds}"`);
```

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
```diff
@@ -0,0 +1,104 @@
...
+export interface CardanoTokenRegistryDependencies {
+  logger: Logger;
+}
+
+/**
+ * TokenMetadataService implementation using Cardano Token Registry public API
+ */
+export class CardanoTokenRegistry implements TokenMetadataService {
+  #axiosClient: AxiosInstance;
+  #cache: Record<string, Asset.TokenMetadata | null> = {};
```
rhyslbw, line 31: You can use the [InMemoryCache](https://github.com/input-output-hk/cardano-js-sdk/blob/feature/ADP-1901-Ogmios-time-settings/packages/cardano-services/src/InMemoryCache/InMemoryCache.ts) (this branch includes a refactor to make it generic compared to what's in _master_)

packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
```diff
@@ -0,0 +1,46 @@
...
+      ticker: 'AGIX',
+      url: 'https://singularitynet.io/'
+    };
+
+    expect(rest).toEqual(result);
+  });
+
+  it('returns cached metadata for already requested subject', async () => {
+    // We need to get metadata at the beginning of the test, to make work
+    // yarn test -t "returns cached metadata for already requested subject"
```
rhyslbw, line 37: It would be better to spy on methods that access the cache, which will be easier after implementing `InMemoryCache`

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
```diff
@@ -0,0 +1,104 @@
...
+    for (const [i, assetId] of assetIds.entries()) {
+      const stringAssetId = assetId.toString();
+
+      if (this.#cache[stringAssetId] !== undefined) {
+        this.#logger.debug(`Using chached asset metatada value for "${stringAssetId}"`);
+        ret[i] = this.#cache[stringAssetId];
+      } else assetIdsToRequest.push(assetId);
+    }
+
+    // All metadatas was taken from cache
```
rhyslbw, line 70:
```suggestion
    // All metadata was taken from cache
```

packages/cardano-services/src/Asset/types.ts
```diff
@@ -16,3 +16,15 @@ export interface NftMetadataService {
    */
   getNftMetadata(asset: AssetPolicyIdAndName): Promise<Asset.NftMetadata | undefined>;
 }
+
+/**
+ * Service to get CIP-35 token metadata for a given subject
```
rhyslbw, line 21:
```suggestion
 * Service to get CIP-38? token metadata for a given subject
```
I've just noticed 35 was claimed by another CIP first, so this one has tentatively been named 38

packages/cardano-services/src/Asset/types.ts
```diff
@@ -16,3 +16,15 @@ export interface NftMetadataService {
...
+ */
+export interface TokenMetadataService {
+  /**
+   * Get CIP-35 token metadata for a given subject
```
rhyslbw, line 25:
```suggestion
   * Get CIP-38? token metadata for a given subject
```

packages/cardano-services/src/Asset/types.ts
```diff
@@ -16,3 +16,15 @@ export interface NftMetadataService {
...
+   *
+   * @returns CIP-35 token metadata, `null` if not found
```
rhyslbw, line 27:
```suggestion
   * @returns CIP-38? token metadata, `null` if not found
```

rhyslbw (changes requested):

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
mkazlauskas: @rhyslbw do you have some other use case of this interface in mind? This is not the first case when `util-services`-like package would be useful (`RunnableModule` is another example).

If we move this then we should also move `NftMetadataService` as they are quite similar.

If there's no immediate need, I'd say just leave them in `cardano-services` until some other package needs to use them.

iccicci: More or less I had the same thoughts as @mkazlauskas

rhyslbw: No other use-cases at the moment, so let's leave it here

packages/cardano-services/src/Asset/types.ts
```diff
@@ -16,3 +16,15 @@ export interface NftMetadataService {
    */
   getNftMetadata(asset: AssetPolicyIdAndName): Promise<Asset.NftMetadata | undefined>;
 }
+
+/**
+ * Service to get CIP-38? token metadata for a given subject
+ */
+export interface TokenMetadataService {
```
iccicci, line 23: While working on ADP-1606 I have a doubt, could it be better to call this `TokenMetadataProvider` ?

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
```diff
@@ -0,0 +1,142 @@
...
+      metadataServerUri: DEFAULT_METADATA_SERVER_URI,
+      ...config
+    };
+
+    this.#cache = cache || new InMemoryCache(defaultConfig.cacheTTL);
+    this.#axiosClient = axios.create({ baseURL: defaultConfig.metadataServerUri });
+    this.#logger = logger;
+  }
+
+  shutdown() {
```
iccicci, line 76: While working on ADP-1606 I noticed that providers should expose close `method` rather than a `shutdown` one ([ref here](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services/src/Http/HttpService.ts#L46)). How do you think it is better to proceed?

packages/cardano-services/src/Asset/types.ts
mkazlauskas: Probably best to leave it as-is for now. All "Provider" interfaces are currently specified in `core` package. When we write docs we need to come up with a good definition of what exactly a cardano-js-sdk Provider means.

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
mkazlauskas: We should unify the terminology, but let's not worry about it in this ticket.

packages/cardano-services/src/Asset/types.ts
rhyslbw: The naming is intentionally aligned with https://github.com/input-output-hk/cardano-js-sdk/blob/6431f57ee5d0624784e9ae6c6b6a73b35a514ee0/packages/cardano-services/src/Metadata/types.ts#L3-L5

packages/cardano-services/src/Asset/types.ts
```diff
@@ -16,3 +16,15 @@ export interface NftMetadataService {
...
+  /**
+   * Get CIP-38? token metadata for a given subject
+   *
+   * @returns CIP-38? token metadata, `null` if not found
+   */
+  getTokenMetadata(assetIds: Cardano.AssetId[]): Promise<(Asset.TokenMetadata | null)[]>;
```
mkazlauskas, line 29: Since AssetProvider only has getAsset (single asset), this could be simplified.
```suggestion
  getTokenMetadata(assetId: Cardano.AssetId): Promise<Asset.TokenMetadata | null>;
```

mkazlauskas (approved):
Looks good, just 1 optional suggestion below

packages/cardano-services/src/Asset/types.ts
rhyslbw: I think it's better to make the interface match the service capabilities, as we'll be making batch requests if moving from a lazy cache to one based on discovery at the point of minting.

rhyslbw (dismissed):
Great work @iccicci :handshake:

packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    it('returns null for nonexisting AssetId', async () => {
+      expect(await tokenRegistry.getTokenMetadata([nonvalidAssetId])).toEqual([null]);
+    });
+
+    it('returns metadata when subject exists', async () => {
+      const [metadata] = await tokenRegistry.getTokenMetadata([validAssetId]);
+
+      expect(metadata).not.toBeNull();
+      if (!metadata) return;
```
iadmytro, line 23: assertion will interrupt the test when something is wrong

packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
```diff
@@ -0,0 +1,74 @@
...
+  const nonvalidAssetId = Cardano.AssetId('0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef');
+  const validAssetId = Cardano.AssetId('f43a62fdc3965df486de8a0d32fe800963589c41b38946602a0dc53541474958');
+
+  describe('return value', () => {
+    const tokenRegistry = new CardanoTokenRegistry({ logger: dummyLogger });
+
+    afterAll(() => tokenRegistry.shutdown());
+
+    it('returns null for nonexisting AssetId', async () => {
+      expect(await tokenRegistry.getTokenMetadata([nonvalidAssetId])).toEqual([null]);
```
iadmytro, line 16: maybe it's better to have just empty list?

rhyslbw: The value is there to cover partial match for the complete query, e.g `[null, { subject: 'abc', ... }]`

iccicci: Not only. I chosen this result type also to be immediately compliant with [this comment](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/core/src/Asset/types/AssetInfo.ts#L26).

packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
iccicci: Just to give you the reason of line 23.

The only meaning of line 23 is because it is partes by TypeScript while line 22 is not: till line 23 (included) `metadata` is of type `Asset.TokenMetadata | null` while from line 24 on `metadata` is of type `Asset.TokenMetadata`.

Once said that I can remove line 23 and use the exclamation mark operator after `metadata`... BTW it is used only once later...

iadmytro: That was ...   unexpected 😀 
I'd vote for exclamation mark usage.

iccicci: Since both @rhyslbw and @iadmytro are still asking for clarifications about this, it is clear I have to apologize for my bad/wrong communication.

Please let me recap: I agree removing line 23 and using the question mark operator, I already applied the change but I still need tu push it.

iccicci:
I had to perform a force push to change the first commit message as requested by @rhyslbw [here](https://github.com/input-output-hk/cardano-js-sdk/pull/321#pullrequestreview-1034037416).

Whit this push I also resolved three conversation from [PR-321](https://github.com/input-output-hk/cardano-js-sdk/pull/321), as they refer more to this PR rather than to the other one:

- [remove custom data type coercion](https://github.com/input-output-hk/cardano-js-sdk/pull/321#discussion_r917715777)
- [eslint/no-unused-vars](https://github.com/input-output-hk/cardano-js-sdk/pull/321#discussion_r917720538)
- [test icon as well](https://github.com/input-output-hk/cardano-js-sdk/pull/321#discussion_r917722496)

rhyslbw (approved):

packages/cardano-services/src/Asset/types.ts
```diff
@@ -16,3 +17,15 @@ export interface NftMetadataService {
...
+/**
+ * Service to get CIP-38? token metadata for a given subject
+ */
+export interface TokenMetadataService extends Shutdown {
+  /**
+   * Get CIP-38? token metadata for a given subject
+   *
+   * @returns CIP-38? token metadata, `null` if not found
+   */
+  getTokenMetadata(assetIds: Cardano.AssetId[]): Promise<(Asset.TokenMetadata | null | undefined)[]>;
```
mkazlauskas, line 30: Please document what `undefined` means. Seems to be indicating an error. Wouldn't throwing an error be more suitable instead of returning undefineds then?

mkazlauskas (changes requested):

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
```diff
@@ -0,0 +1,171 @@
...
+    this.#cache.shutdown();
```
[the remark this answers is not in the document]
rhyslbw: Yes, the management of the dependency should be limited to when the current context created a state change, ie. a class instance started it so it's now responsible to stop it.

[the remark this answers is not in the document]
mkazlauskas: Right. In this specific case it would be good to have an explicit `InMemoryCache.initialize` or `.start`, so that it's more obvious on where to shut it down.

packages/cardano-services/src/Asset/types.ts
iccicci: Yes @mkazlauskas ,

I taken [this line](https://github.com/input-output-hk/cardano-js-sdk/blob/75acb80b2310a183dca71b94d091b422d5ca6672/packages/core/src/Asset/types/AssetInfo.ts#L26) as specification: the only reason I can see why a token metadata could be not loaded is in case of error; so in case of error I'm returning `undefined` and logging the error.

But sorry, in your comment you are asking to better document the reason for `undefined` and proposing to change the interface to throw an error (which would also mean to remove `undefined` from possible returned values): the two actions are in contrast; which one is the preferred?

Thank you

mkazlauskas: The line is only relevant on client-side. Token metadata was loaded async, so `undefined` could be used to distinguish assets that are:
- Not tokens (no token metadata) = `null`
- Don't know if it's a token or not (still loading) = `undefined`

iccicci: Sorry @mkazlauskas ,

applying this change I need to change [this line](https://github.com/input-output-hk/cardano-js-sdk/blob/feat/ADP-1605-token-metadata-provider/packages/cardano-services/src/Asset/CardanoTokenRegistry.ts#L123) with a `throw` statement: could you please suggest me the right error to throw?

Thank you

mkazlauskas: Probably ProviderError{Unknown} that has real error in 'innerError' if it's just some internal error

iccicci: More or less the same question for the errors in the two `catch` blocks, @mkazlauskas ,

if I remember correctly we usually do not simply propagate the errors, but we encapsulate them in a SDK error type (at least once I used `ProviderError`). Should I use the same pattern?

iccicci: Resolved in cc5145c3eaf007707b35a34896ff093b42a30a85

could you please review this again @mkazlauskas  ? Thank you

mkazlauskas (changes requested):
Looks good @iccicci, just a minor issue with message I noticed. You can update it together with squash of the fixup

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
```diff
@@ -0,0 +1,180 @@
...
+  ticker?: StringValue;
+  url?: StringValue;
+}
+
+const toProviderError = (error: unknown, details: string) => {
+  if (error instanceof ProviderError) return error;
+
+  const message = error instanceof Error ? `${error.message} ` : '';
+
+  return new ProviderError(ProviderFailure.Unknown, error, `${message} ${details}`);
```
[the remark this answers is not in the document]
iccicci: Thank you @mkazlauskas !!!
I already added the space in the message only when required; thank you.

`git commit --fixup` is my friend :wink: resolved in de02ff45f4b4f430e208299d01d769ee6948af9e

mkazlauskas (approved):
Nice work!

rhyslbw (dismissed):

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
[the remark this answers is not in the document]
iccicci: As I told in another thread @rhyslbw and @mkazlauskas , I don't like to force the caller to perform _not required calls_ as `InMemoryCache.initialize()` and `InMemoryCache.start()` would be going in that direction.
Anyway if that's the requirement for our standard, this reason is much more stronger than anyone else bound to what I like or not.

Before proceeding, however, we have an open discussion about the definition of `Provider` and another one about the right placement of `RunningModule`... I think it's better first of all to consolidate the standard, then we could refactor both `InMemoryCache` and `CardanoTokenRegistry` classes.

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  shutdown() {
+    this.#cache.shutdown();
+  }
+
+  async getTokenMetadata(assetIds: Cardano.AssetId[]) {
+    this.#logger.debug(`Requested asset metatada for "${assetIds}"`);
+
+    const { assetIdsToRequest, ret } = this.getTokenMetadataFromCache(assetIds);
```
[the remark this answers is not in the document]
iccicci: Resolved in 32d7bd4b91d4a76f0ca6890008f8825245004fd5

could you please review this again @rhyslbw and @IvayloAndonov ? Thank you

IvayloAndonov (approved):
Nice work!

packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
```diff
@@ -0,0 +1,158 @@
...
+
+    const tokenRegistry = new CardanoTokenRegistry({ cache: new TestInMemoryCache(1), logger: dummyLogger });
+    const [firstResult] = await tokenRegistry.getTokenMetadata([validAssetId]);
+    const [secondResult] = await tokenRegistry.getTokenMetadata([validAssetId]);
+
+    expect(gotValues[0]).toBeUndefined();
+    expect(gotValues[1]).toEqual(firstResult);
+    expect(firstResult).toEqual(secondResult);
+
+    tokenRegistry.shutdown();
```
iadmytro, line 103: `shutdown`  will not be executed in case of failed assertion

iccicci: Resolved in d7831d36cf8126eab8f875065a6aa581259e2141

could you please review this again @iadmytro ? Thank you

iadmytro (approved):

mkazlauskas (dismissed):

rhyslbw (approved):

mkazlauskas (approved):

PR #319, opened 2022-07-05
refactor: modular wallet utils

mkazlauskas:
# Context

- Implementing new providers for `SingleAddressWallet` requires resolving **all** fields defined in core package. This is not always possible.
- Using `ObservableWallet` interface as a base in a custom wallet application in not always possible for similar reasons.
- Creating a new wallet interface and basing implementation on SDK utils is also complicated, because most utils depend on either full `Cardano.*` types or full `ObservableWallet`, which (again) could not necessarily be resolved and might be difficult or impossible to do without losing type safety that is provided by TypeScript.

# Proposed Solution

- Refactor `WalletUtil` as an example of what we would eventually do with the rest of the SDK utils 0339afdc86494cea7d760383c869a0258deafc07:
  - Replace `ObservableWallet` arg with minimal `{protocolParameters$, utxo}`
  - Export `OutputValidator` and `InputResolver` utils separately
- (:exclamation:) Add a new documentation-style test as an example of how custom wallets could be implemented utilizing pieces of the SDK and not necessarily a fully-fledged SingleAddressWallet 8c231cd051e97e88fca5d081bd21a98c48945c29 
- Propose 1 more change in scope of this PR to give a decent minimal subset of ObservableWallet type required by cip30 d2a24cee3d10d0f1532b4179e872e0361e2c99fb

### DIFF
--- a/packages/wallet/src/services/WalletUtil.ts
+++ b/packages/wallet/src/services/WalletUtil.ts
@@ -1,50 +1,106 @@
 import { BigIntMath } from '@cardano-sdk/util';
 import { Cardano, ProtocolParametersRequiredByWallet } from '@cardano-sdk/core';
-import { ObservableWallet, OutputValidation } from '../types';
+import { Observable, firstValueFrom } from 'rxjs';
+import { OutputValidation } from '../types';
 import { ResolveInputAddress } from '../KeyManagement';
 import { computeMinimumCoinQuantity, tokenBundleSizeExceedsLimit } from '@cardano-sdk/cip2';
-import { firstValueFrom } from 'rxjs';
 import { txInEquals } from './util';
 
-/**
- * @returns common wallet utility functions that are aware of wallet state and computes useful things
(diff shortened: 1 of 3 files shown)


mirceahasegan (approved):
I think this is the perfect route. Allowing the SDK users to selectively choose the functionality they need and use composition to create the wallets, is the most flexible approach from my point of view.
Nicely done @mkazlauskas ! 🥇

AngelCastilloB:
I like this approach @mkazlauskas :), its pretty flexible and addresses problem perfectly.

rhyslbw (dismissed):
This looks like a very flexible approach 🦾
Narrowing the util param types is a good refactor on it's own.

rhyslbw (dismissed):
Nice work @mkazlauskas :tada:

mirceahasegan (dismissed):

iadmytro (approved):

PR #320, opened 2022-07-05
Feature/adp 1806 optimize polling strategy to utilize navigator on line

mirceahasegan:
# Context

ADP-1806 Utilize Navigator.onLine and window.online/offline events to optimize polling strategy

# Proposed Solution
- `ConnectionStatusTracker` observable type is used to notify connection state changes.
- `createSimpleConnectionStatusTracker` is used by default in SingleAddressWallet:
  - When running in browser, window online/offline events are used to determine if the connection is up or down.
  - When running in Node, the connection will always be reported as up, because there is no simple mechanism to determine if the active interface is connected or not.
- Users can provide their own `ConnectionStatusTracker` observable to `SingleAddressWallet`
- Providers exponential backoff is canceled while connection is down
- `TipTracker` will emit only while connection is up

### DIFF
--- a/packages/wallet/src/SingleAddressWallet.ts
+++ b/packages/wallet/src/SingleAddressWallet.ts
@@ -18,6 +18,8 @@ import {
 import { Assets, InitializeTxProps, InitializeTxResult, ObservableWallet, SignDataProps, SyncStatus } from './types';
 import {
   BalanceTracker,
+  ConnectionStatus,
+  ConnectionStatusTracker,
   DelegationTracker,
   FailedTx,
   PersistentDocumentTrackerSubject,
@@ -38,6 +40,7 @@ import {
   createBalanceTracker,
   createDelegationTracker,
   createProviderStatusTracker,
+  createSimpleConnectionStatusTracker,
   createTransactionsTracker,
   createUtxoTracker,
   createWalletUtil,
@@ -89,6 +92,7 @@ export interface SingleAddressWalletDependencies {
   readonly inputSelector?: InputSelector;
   readonly stores?: WalletStores;
   readonly logger?: Logger;
+  readonly connectionStatusTracker$?: ConnectionStatusTracker;
 }
 
 export class SingleAddressWallet implements ObservableWallet {
@@ -125,6 +129,7 @@ export class SingleAddressWallet implements ObservableWallet {
   readonly util: WalletUtil;
   readonly rewardsProvider: TrackedRewardsProvider;
 
+  // eslint-disable-next-line max-statements
   constructor(
     {
       name,
@@ -149,7 +154,8 @@ export class SingleAddressWallet implements ObservableWallet {
       rewardsProvider,
       logger = dummyLogger,
       inputSelector = roundRobinRandomImprove(),
-      stores = createInMemoryWalletStores()
+      stores = createInMemoryWalletStores(),
+      connectionStatusTracker$ = createSimpleConnectionStatusTracker()
     }: SingleAddressWalletDependencies
   ) {
     this.#logger = logger;
@@ -193,10 +199,13 @@ export class SingleAddressWallet implements ObservableWallet {
       )
     );
     this.name = name;
+    const cancel$ = connectionStatusTracker$.pipe(filter((status) => status === ConnectionStatus.down));
     this.#tip$ = this.tip$ = new TipTracker({
+      connectionStatus$: connectionStatusTracker$,
       maxPollInterval: maxInterval,
       minPollInterval: pollInterval,
       provider$: coldObservableProvider({
+        cancel$,
         provider: this.networkInfoProvider.ledgerTip,
         retryBackoffConfig
       }),
@@ -210,6 +219,7 @@ export class SingleAddressWallet implements ObservableWallet {
       // This is a little complicated since there is a circular dependency.
       // Some logic is needed to initiate a fetch if epoch is not available in store already.
       coldObservableProvider({
+        cancel$,
         equals: deepEquals,
         provider: this.networkInfoProvider.timeSettings,
         retryBackoffConfig,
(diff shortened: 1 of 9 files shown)


packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -125,6 +125,7 @@ export class SingleAddressWallet implements ObservableWallet {
   readonly util: WalletUtil;
   readonly rewardsProvider: TrackedRewardsProvider;
 
+  // eslint-disable-next-line max-statements
```
mirceahasegan, line 132: Please help with an alternate solution to too many statements in constructor

packages/wallet/src/services/util/connectionStatusTracker.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export enum ConnectionStatus {
+  down = 0,
+  up
+}
+
+export type ConnectionStatusTracker = Observable<ConnectionStatus>;
+
+export interface ConnectionStatusTrackerInternals {
+  isNodeEnv?: () => boolean;
```
mirceahasegan, line 11: I think these could simply be properties. I'll try to refactor...
I'm changing this PR to draft, but please take a look 😄

packages/wallet/test/services/ProviderTracker/TipTracker.test.ts
```diff
@@ -95,6 +100,43 @@ describe('TipTracker', () => {
     });
   });
 
+  it('syncStatus updates emitted only once connection is back up', () => {
```
mirceahasegan, line 103: Please take a closer look here. I need another pair of eyes to validate the correctness of the test. Thanks!

packages/wallet/src/SingleAddressWallet.ts
rhyslbw: I think it's reasonable to merge with this, as reworking the interface is out of scope, and there will be some attention here when refactoring to support the narrowed types.

rhyslbw: This is reasonable to leave since refactoring the constructor is better left for the planned rework relating to narrowed internal types.

packages/wallet/src/services/util/connectionStatusTracker.ts
mirceahasegan: I have simplified the internals here: [6607665](https://github.com/input-output-hk/cardano-js-sdk/pull/320/commits/6607665dbd8d0208919a3c998227ab9f5f067951)

AngelCastilloB (commented):
The PR looks good!

packages/wallet/src/services/util/connectionStatusTracker.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Returns an observable that emits the online status of the browser.
+ * When running in Node, it always emits 'up'
+ *
+ * @returns {ConnectionStatusTracker} ConnectionStatusTracker
+ */
+export const createSimpleConnectionStatusTracker = ({
+  isNodeEnv = typeof window === 'undefined',
+  online$ = isNodeEnv ? of(true) : fromEvent(window, 'online'),
```
mkazlauskas, line 35: This will emit twice in node.js due to `startWith(initialStatus)`
```suggestion
  online$ = isNodeEnv ? EMPTY : fromEvent(window, 'online'),
```

packages/wallet/test/services/ProviderTracker/TipTracker.test.ts
```diff
@@ -95,6 +100,43 @@ describe('TipTracker', () => {
...
+      });
+      expectObservable(tracker$.asObservable()).toBe('x---a', mockTips);
+    });
+  });
+
+  it('syncStatus timeout while connection is down does not emit tip updates', () => {
+    createTestScheduler().run(({ cold, hot, expectObservable }) => {
+      const syncStatus: Partial<SyncStatus> = { isSettled$: hot('10ms |') };
+      const connectionStatusMock$: Observable<ConnectionStatus> = hot('d--|', { d: ConnectionStatus.down });
+      const provider$ = stubObservableProvider<Cardano.Tip>(cold('x|', mockTips));
```
mkazlauskas, line 127: Maybe we could also assert that provider$ wasn't subscribed (there's a `expectSubscriptions`)

packages/wallet/test/services/ProviderTracker/TipTracker.test.ts
mkazlauskas: This test looks good to me

mkazlauskas (changes requested):

packages/wallet/src/services/util/connectionStatusTracker.ts
mirceahasegan: > ce in node.js due

The current impl returns a simple BehaviorSubject for the node scenario, ignoring online$/offline$ streams.
I think this is a problem. Since we give access to the internals, we should use them in both node/browser envs.
I will make the changes

mkazlauskas (approved):

AngelCastilloB (approved):
The PR looks good!

mkazlauskas (dismissed):

rhyslbw (changes requested):
Looks great! Please squash to minimal commits.

packages/wallet/test/mocks/mockUtxoProvider.ts
```diff
@@ -68,7 +68,7 @@ export const utxo2 = utxo.slice(1);
  *
  * returns UtxoProvider-compatible object
  */
-export const mockUtxoProvider = (): UtxoProvider => ({
+export const mockUtxoProvider = () => ({
```
mirceahasegan, line 71: I think this is a mistake. I will remove it as part of the squash

mirceahasegan: I have added the type back and updated the test code:
```
utxoProvider.utxoByAddresses = jest
      .fn()
      .mockImplementationOnce(() => mocks.utxo)
      .mockImplementation(
        () =>
          new Promise(() => {
            // Make sure wallet never settles
          })
      );
```

mirceahasegan:
> Looks great! Please squash to minimal commits.

Done

mkazlauskas (approved):

rhyslbw (approved):
Great work @mirceahasegan! 💪

PR #321, opened 2022-07-08
feat(cardano-services): adds asset http service

iccicci:
# Context

We need to add an http service to serve asset related functions.

# Proposed Solution

Added `AssetHttpService` and `TokenMetadataProvider` to expose `CardanoTokenRegistry.getTokenMetadata`.

### DIFF
--- a/packages/cardano-services/src/Asset/AssetHttpService.ts
+++ b/packages/cardano-services/src/Asset/AssetHttpService.ts
@@ -1,0 +1,46 @@
+import * as OpenApiValidator from 'express-openapi-validator';
+import { AssetProvider } from '@cardano-sdk/core';
+import { HttpService } from '../Http';
+import { Logger } from 'ts-log';
+import { ServiceNames } from '../Program';
+import { providerHandler } from '../util';
+import express from 'express';
+import path from 'path';
+
+/**
+ * Dependencies that are need to create AssetHttpService
+ */
+export interface AssetHttpServiceDependencies {
+  /**
+   * The asset provider to fetch data
+   */
+  assetProvider: AssetProvider;
+
+  /**
+   * The logger object
+   */
+  logger: Logger;
+}
+
+/**
+ * The Asset Http Service
+ */
+export class AssetHttpService extends HttpService {
+  constructor({ assetProvider, logger }: AssetHttpServiceDependencies, router: express.Router = express.Router()) {
+    super(ServiceNames.Asset, assetProvider, router, logger);
+
+    const apiSpec = path.join(__dirname, 'openApi.json');
+    router.use(
+      OpenApiValidator.middleware({
+        apiSpec,
+        ignoreUndocumented: true,
+        validateRequests: true,
+        validateResponses: true
+      })
+    );
+    router.post(
+      '/get-asset',
+      providerHandler(assetProvider.getAsset.bind(assetProvider))(HttpService.routeHandler(logger), logger)
+    );
+  }
+}
--- a/packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
+++ b/packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
@@ -25,6 +25,14 @@ interface TokenMetadataServiceRecord {
   url?: StringValue;
 }
 
+const propertiesToChange: Record<string, string> = { description: 'desc', logo: 'icon' };
+export const toCoreTokenMetadata = (record: TokenMetadataServiceRecord) =>
+  Object.fromEntries(
+    Object.entries(record)
+      .filter(([key]) => ['decimals', 'description', 'logo', 'name', 'ticker', 'url'].includes(key))
+      .map(([key, value]) => [propertiesToChange[key] || key, value.value])
+  ) as Asset.TokenMetadata;
+
 const toProviderError = (error: unknown, details: string) => {
   if (error instanceof ProviderError) return error;
 
@@ -125,7 +133,7 @@ export class CardanoTokenRegistry implements TokenMetadataService {
 
           if (subject) {
             const assetId = Cardano.AssetId(subject);
-            const metadata = this.parseTokenMetadataServiceRecord(record);
+            const metadata = toCoreTokenMetadata(record);
 
             tokenMetadata[assetIds.indexOf(assetId)] = metadata;
             this.#cache.set(assetId.toString(), metadata);
@@ -163,19 +171,5 @@ export class CardanoTokenRegistry implements TokenMetadataService {
     }
 
     return [assetIdsToRequest, cachedTokenMetadata] as const;
-  }
-
-  private parseTokenMetadataServiceRecord(record: TokenMetadataServiceRecord) {
-    const { decimals, description, logo, name, ticker, url } = record;
-    const metadata: Asset.TokenMetadata = {};
-
-    metadata.decimals = decimals?.value;
-    metadata.desc = description?.value;
-    metadata.icon = logo?.value;
-    metadata.name = name?.value;
-    metadata.ticker = ticker?.value;
-    metadata.url = url?.value;
-
-    return metadata;
   }
 }
--- a/packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
+++ b/packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
@@ -4,7 +4,8 @@ import { Logger } from 'ts-log';
 import { TokenMetadataService } from './types';
 import axios, { AxiosInstance } from 'axios';
 
-const DEFAULT_METADATA_SERVER_URI = 'https://tokens.cardano.org';
+export const DEFAULT_TOKEN_METADATA_CACHE_TTL = 600;
+export const DEFAULT_TOKEN_METADATA_SERVER_URL = 'https://tokens.cardano.org';
 
 interface NumberValue {
   value?: number;
@@ -39,17 +40,17 @@ export interface CardanoTokenRegistryConfiguration {
   /**
    * The cache TTL in seconds. Default: 10 minutes.
    */
-  cacheTTL?: number;
+  tokenMetadataCacheTTL?: number;
 
   /**
    * The Cardano Token Registry public API base URL. Default: https://tokens.cardano.org
    */
-  metadataServerUri?: string;
+  tokenMetadataServerUrl?: string;
 }
 
 interface CardanoTokenRegistryConfigurationWithRequired extends CardanoTokenRegistryConfiguration {
-  cacheTTL: number;
-  metadataServerUri: string;
+  tokenMetadataCacheTTL: number;
+  tokenMetadataServerUrl: string;
 }
 
 /**
@@ -88,13 +89,13 @@ export class CardanoTokenRegistry implements TokenMetadataService {
 
   constructor({ cache, logger }: CardanoTokenRegistryDependencies, config: CardanoTokenRegistryConfiguration = {}) {
     const defaultConfig: CardanoTokenRegistryConfigurationWithRequired = {
-      cacheTTL: 600,
-      metadataServerUri: DEFAULT_METADATA_SERVER_URI,
+      tokenMetadataCacheTTL: DEFAULT_TOKEN_METADATA_CACHE_TTL,
+      tokenMetadataServerUrl: DEFAULT_TOKEN_METADATA_SERVER_URL,
       ...config
     };
 
-    this.#cache = cache || new InMemoryCache(defaultConfig.cacheTTL);
-    this.#axiosClient = axios.create({ baseURL: defaultConfig.metadataServerUri });
+    this.#cache = cache || new InMemoryCache(defaultConfig.tokenMetadataCacheTTL);
+    this.#axiosClient = axios.create({ baseURL: defaultConfig.tokenMetadataServerUrl });
     this.#logger = logger;
   }
 
--- a/packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
+++ b/packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
@@ -139,7 +139,7 @@ export class CardanoTokenRegistry implements TokenMetadataService {
         }
       }
     } catch (error) {
-      throw toProviderError(error, 'while fetching metadata from token registriy');
+      throw toProviderError(error, 'while fetching metadata from token registry');
     }
 
     return tokenMetadata;
@@ -156,7 +156,7 @@ export class CardanoTokenRegistry implements TokenMetadataService {
       const cachedMetadata = this.#cache.getVal<Asset.TokenMetadata>(stringAssetId);
 
       if (cachedMetadata) {
-        this.#logger.debug(`Using chached asset metatada value for "${stringAssetId}"`);
+        this.#logger.debug(`Using cached asset metatada value for "${stringAssetId}"`);
         cachedTokenMetadata[i] = cachedMetadata;
       } else assetIdsToRequest.push(assetId);
     }
--- a/packages/cardano-services/src/Asset/openApi.json
+++ b/packages/cardano-services/src/Asset/openApi.json
@@ -1,0 +1,119 @@
+{
+  "openapi": "3.0.0",
+  "info": {
+    "title": "Asset",
+    "license": {
+      "name": "Apache 2.0",
+      "url": "http://www.apache.org/licenses/LICENSE-2.0.html"
+    },
+    "version": "1.0.0"
+  },
+  "paths": {
+    "/asset/get-asset": {
+      "post": {
+        "summary": "Get Asset Info",
+        "operationId": "getAssetInfo",
+        "requestBody": {
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "#/components/schemas/GetAssetRequest"
+              }
+            }
+          },
+          "required": true
+        },
+        "responses": {
+          "200": {
+            "description": "success operation",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/AssetInfo"
+                }
+              }
+            }
+          },
+          "400": {
+            "description": "invalid request",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object"
+                }
+              }
+            }
+          },
+          "404": {
+            "description": "not found",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object"
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  },
+  "components": {
+    "schemas": {
+      "AssetInfo": {
+        "type": "object",
+        "properties": {
+          "assetId": {
+            "type": "string"
+          },
+          "fingerprint": {
+            "type": "string"
+          },
+          "mintOrBurnCount": {
+            "type": "number"
+          },
+          "name": {
+            "type": "string"
+          },
+          "policyId": {
+            "type": "string"
+          }
+        }
+      },
+      "ExtraData": {
+        "type": "object",
+        "properties": {
+          "history": {
+            "type": "boolean"
+          },
+          "nftMetadata": {
+            "type": "boolean"
+          },
+          "tokenMetadata": {
+            "type": "boolean"
+          }
+        }
+      },
+      "GetAssetRequest": {
+        "required": ["args"],
+        "type": "object",
+        "properties": {
+          "args": {
+            "type": "array",
+            "minItems": 1,
+            "items": {
+              "oneOf": [
+                {
+                  "type": "string"
+                },
+                {
+                  "$ref": "#/components/schemas/ExtraData"
+                }
+              ]
+            }
+          }
+        }
+      }
+    }
+  }
+}
--- a/packages/cardano-services/src/Http/HttpService.ts
+++ b/packages/cardano-services/src/Http/HttpService.ts
@@ -56,6 +56,13 @@ export abstract class HttpService extends RunnableModule {
         return HttpServer.sendJSON(res, await handler(...args));
       } catch (error) {
         logger.error(error);
+
+        if (error instanceof ProviderError) {
+          const code = error.reason === ProviderFailure.NotFound ? 404 : 500;
+
+          return HttpServer.sendJSON(res, error, code);
+        }
+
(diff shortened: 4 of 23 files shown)


packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
```diff
@@ -0,0 +1,141 @@
...
+      } else assetIdsToRequest.push(assetId);
+    }
+
+    // All metadata was taken from cache
+    if (assetIdsToRequest.length === 0) return ret;
+
+    this.#logger.debug(`Fetching asset metatada for "${assetIdsToRequest}"`);
+
+    try {
+      const response = await this.#axiosClient.post('metadata/query', {
```
rhyslbw, line 113:
```suggestion
      const response = await this.#axiosClient.post<{ subjects: (Asset.TokenMetadata & { subject: string })[] }>('metadata/query', {
```
See if you can remove the custom data type coercion by providing a type upfront. I've used `Asset.TokenMetadata` as a demo, but you'll need to define the API response type given the property names vary a little.

packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    });
+
+    it('returns metadata when subject exists', async () => {
+      const [metadata] = await tokenRegistry.getTokenMetadata([validAssetId]);
+
+      expect(metadata).not.toBeNull();
+      if (!metadata) return;
+
+      // Skip icon from test as it is a long long string
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
rhyslbw, line 26: Which variable is unused?

packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
```diff
@@ -0,0 +1,74 @@
...
+      expect(await tokenRegistry.getTokenMetadata([nonvalidAssetId])).toEqual([null]);
+    });
+
+    it('returns metadata when subject exists', async () => {
+      const [metadata] = await tokenRegistry.getTokenMetadata([validAssetId]);
+
+      expect(metadata).not.toBeNull();
+      if (!metadata) return;
+
+      // Skip icon from test as it is a long long string
```
rhyslbw, line 25: You could instead just assert a value is returned, otherwise, we're not covering the complete API

packages/cardano-services/src/Asset/AssetHttpService.ts
```diff
@@ -0,0 +1,55 @@
...
+export interface AssetHttpServiceDependencies {
+  /**
+   * The logger object
+   */
+  logger: Logger;
+
+  /**
+   * The token metadata provider. Default: CardanoTokenRegistry instance
+   */
+  tokenMetadataProvider?: TokenMetadataProvider;
```
rhyslbw, line 23:
```suggestion
  assetProvider?: AssetProvider;
```
Token metadata is just one part of the asset provider

packages/cardano-services/src/Asset/types.ts
```diff
@@ -16,3 +17,27 @@ export interface NftMetadataService {
...
+   *
+   * @returns CIP-38? token metadata, `null` if not found
+   */
+  getTokenMetadata(assetIds: Cardano.AssetId[]): Promise<(Asset.TokenMetadata | null)[]>;
+}
+
+/**
+ * Provider to get CIP-38? token metadata for a given subject
+ */
+export interface TokenMetadataProvider extends Provider {
```
rhyslbw, line 36: This interface is not required

rhyslbw (changes requested):
We're not looking to add another provider here @iccicci, so the HTTP server needs to depend on `AssetProvider`. What's missing is a `DbSyncAssetProvider`, which will require some additional DB queries, the `DbSyncNftMetadataService` (dependency can be on `NftMetadataService`), and your `TokenMetadataService` to resolve the `AssetInfo` type.

ac9901a419222336250bdd88b728dae8780155bf is a `refactor` commit.

packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
iccicci: `icon` itself: its declaration is required only to extract the `rest` object.

packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
iccicci: Ok, that also resolve the previous discussion!

packages/cardano-services/src/Asset/AssetHttpService.ts
mkazlauskas: Yep, AssetProvider is the only external interface here. It has NftMetadataService and TokenMetadataService as internal dependencies.

rhyslbw:
> I could guess that `assetId` is the input

:heavy_check_mark: 

> ... and `name` can be taken from TokenMetadata.name;

:x:  `name` is the sub-policy identifier, is on-chain, distinct from the metadata property. See below

The following can be queried from the [multi_asset](The remaining on-chain data can be fetched from this table: https://github.com/input-output-hk/cardano-db-sync/blob/master/doc/schema.md#multi_asset) table:

- `policyId` -> `policy`
- `name`
- `fingerprint`

The following can be accessed from the [ma_tx_mint](https://github.com/input-output-hk/cardano-db-sync/blob/master/doc/schema.md#ma_tx_mint) table:

- `mintOrBurnCount` = the number of times the asset has been minted or burned, so a count of the records in this table.
 - `history` - The type explains this
- `quantity` - Number of assets minted - burned


The [ChainHistory](https://github.com/input-output-hk/cardano-js-sdk/tree/master/packages/cardano-services/src/ChainHistory/DbSyncChainHistory) `cardano-db-sync` implementation is probably the best reference for making these queries.

packages/cardano-services/src/Asset/openApi.json
```diff
@@ -0,0 +1,102 @@
...
+            }
+          }
+        }
+      }
+    }
+  },
+  "components": {
+    "schemas": {
+      "TokenMetadata": {
+        "required": ["decimals", "desc", "icon", "name", "ticker", "url"],
```
iccicci, line 54: I'm not sure about this line.
Sorry @rhyslbw  and @mkazlauskas , could you please confirm which are the attributes we can trust will be for sure in the source token API? I know, this line validates the output of our http server, but the presence of these attributes strictly depends on the  presence in the token API.

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
iccicci: I found no defined `TokenMetadataServiceRecord`.
If you are suggesting it as the new type name (ok), the `Partial<>` should not be required as I can define all the properties as optional.

packages/cardano-services/src/Asset/AssetHttpService.ts
iccicci: cd7458e should resolve this

packages/cardano-services/src/Asset/types.ts
iccicci: cd7458e should resolve this

packages/cardano-services/test/Asset/AssetHttpService.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        try {
+          await axios.post(`${apiUrlBase}/get-asset`, { args: [['test']] });
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        } catch (error: any) {
+          expect(error.response.status).toBe(400);
+          expect(error.message).toBe(BAD_REQUEST_STRING);
+        }
+      });
+
+      it('internal server error for not existing existing asset id', async () => {
```
iccicci, line 72: I don't think this is the expected behavior...

packages/cardano-services/src/Asset/openApi.json
```diff
@@ -0,0 +1,109 @@
...
+      },
+      "GetAssetRequest": {
+        "required": ["args"],
+        "type": "object",
+        "properties": {
+          "args": {
+            "type": "array",
+            "minItems": 1,
+            "items": {
+              "oneOf": [
```
iccicci, line 95: Here I used `oneOf` in order to accept `[string, ExtraData]`, but this will also allow `[ExtraData, string]` (... or `[string, string]`).
Who could I change this in order accept only `[string, ExtraData]`?
Thank you

ssong-iohk (dismissed):

packages/cardano-services/src/Asset/AssetHttpService.ts
```diff
@@ -0,0 +1,67 @@
...
+export interface AssetHttpServiceDependencies {
+  /**
+   * The db-sync database Pgpool
+   */
+  db: Pool;
+
+  /**
+   * The logger object
+   */
+  logger: Logger;
```
rhyslbw, line 26:
```suggestion
  logger: Logger;
  assetProvider: AssetProvider;
```

The concrete asset provider class should be instantiated and passed in here as a dependency. This is an example of where it will happen, with a provider that has an additional service needed:

https://github.com/input-output-hk/cardano-js-sdk/blob/72d83fb203d3dc417f8acc3894d3db2a398513fb/packages/cardano-services/src/Program/loadHttpServer.ts#L81-L93

packages/cardano-services/src/Asset/AssetHttpService.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/**
+ * The Asset Http Service
+ */
+export class AssetHttpService extends HttpService {
+  #tokenMetadataService: CardanoTokenRegistry;
```
rhyslbw, line 33:
```suggestion
```

packages/cardano-services/src/Asset/openApi.json
rhyslbw: You can leave this for now, as we'll be refactoring the providers to use a single object argument rather than depending on the position in the argument array.

packages/cardano-services/test/Asset/AssetHttpService.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        try {
+          const res = await axios.post(`${apiUrlBase}/get-asset`, {
+            args: ['0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef']
+          });
+          expect(res.data[0]).toEqual({});
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        } catch (error: any) {
+          expect(error.response.status).toBe(500);
```
rhyslbw, line 80:
```suggestion
          expect(error.response.status).toBe(404);
```

rhyslbw (changes requested):

packages/cardano-services/test/Asset/AssetHttpService.test.ts
mkazlauskas: Nope, it should be ProviderError{ProviderFailure.NotFound}

iccicci: Let's go on in [this discussion](https://github.com/input-output-hk/cardano-js-sdk/pull/321#discussion_r924385192) as it is the same topic.

packages/cardano-services/test/Asset/AssetHttpService.test.ts
iccicci: Sorry @rhyslbw and @mkazlauskas ,

I was trying to propagate `ProviderError{ProviderFailure.NotFound}` through the HTTP response as we said [here](https://github.com/input-output-hk/cardano-js-sdk/pull/321#discussion_r924423559); is there already some helper in the sdk to do that?

I mean the [router](https://github.com/input-output-hk/cardano-js-sdk/blob/feat/ADP-1606-asset-http-service/packages/cardano-services/src/Asset/AssetHttpService.ts#L57-L60) is using sdk standard; the [getAssist](https://github.com/input-output-hk/cardano-js-sdk/blob/feat/ADP-1606-asset-http-service/packages/cardano-services/test/Asset/DbSyncAssetProvider.test.ts#L40-L42) method already throw the right error...

If I have to write some custom code, could you please point me a code sample to see how to propagate a `ProviderError` through the HTTP response?

Thank you

mkazlauskas: [This](https://github.com/input-output-hk/cardano-js-sdk/blob/cd7458e7c2550f5c7a6d98b3dfacb8d0ea5ce2ea/packages/cardano-services/src/Http/HttpService.ts#L59) catch should work

packages/cardano-services/src/Asset/AssetHttpService.ts
iccicci: f872fec9f should resolve this.

packages/cardano-services/src/Asset/AssetHttpService.ts
iccicci: f872fec9f should resolve this.

packages/cardano-services/test/Asset/AssetHttpService.test.ts
iccicci: Hi @rhyslbw and @mkazlauskas ,

it seems the given `catch` didn't exactly fit as it was always responding with http status code `500`; anyway f872fec9f introduces the small change to let it work as required.

packages/cardano-services/src/Http/HttpService.ts
```diff
@@ -56,7 +56,16 @@ export abstract class HttpService extends RunnableModule {
...
+
+        let code = 500;
+        let providerError: ProviderError;
+
+        if (error instanceof ProviderError) {
+          providerError = error;
+          if (error.reason === ProviderFailure.NotFound) code = 404;
+        } else providerError = new ProviderError(ProviderFailure.Unhealthy, error);
+
+        return HttpServer.sendJSON(res, providerError, code);
```
mkazlauskas, line 68: I think it's best to avoid using mutable variables - immutable style tends to be less error-prone.

Mutable variables have their use cases (sometimes just fits the problem well, or as performance optimization), but I don't think this is one of those cases.

```suggestion
        if (error instanceof ProviderError) {
          const code = error.reason === ProviderFailure.NotFound ? 404 : 500;
          return HttpServer.sendJSON(res, error, code);
        }
        const providerError = new ProviderError(ProviderFailure.Unhealthy, error);
        return HttpServer.sendJSON(res, providerError, 500);
```

mkazlauskas (changes requested):

packages/cardano-services/src/Http/HttpService.ts
iccicci: Hi @mkazlauskas ;

67efd3a71 should resolve this

mkazlauskas (approved):
Great work!

packages/cardano-services/src/Asset/AssetHttpService.ts
```diff
@@ -0,0 +1,46 @@
...
+import path from 'path';
+
+/**
+ * Dependencies that are need to create AssetHttpService
+ */
+export interface AssetHttpServiceDependencies {
+  /**
+   * The asset provider to fetch data
+   */
+  assetProvider: Provider & AssetProvider;
```
rhyslbw, line 17:
```suggestion
  assetProvider: AssetProvider;
```
Let's merge https://github.com/input-output-hk/cardano-js-sdk/pull/346 first, then you can make this change

rhyslbw (changes requested):
As this PR is touching the program configuration in `loadHttpServer.ts`, it would be best to now also include the configuration mapping within the entrypoints (`cli.ts` and `run.ts`).

packages/cardano-services/src/Asset/openApi.json
```diff
@@ -0,0 +1,119 @@
+{
+  "openapi": "3.0.0",
+  "info": {
+    "title": "UTxO",
```
rhyslbw, line 4:
```suggestion
    "title": "Asset",
```

rhyslbw (changes requested):

packages/cardano-services/src/Asset/openApi.json
iccicci: Hi @rhyslbw ;

905559c21 solves this (and rebase as well)

iccicci:
> As this PR is touching the program configuration in `loadHttpServer.ts`, it would be best to now also include the configuration mapping within the entrypoints (`cli.ts` and `run.ts`).

Hi @rhyslbw , 8b0ddf112 should solve this.

rhyslbw (changes requested):
Getting close now!

Now that we're seeing the token metadata configuration next to the existing entrypoint config, I think we need more specificity by appending "token". e.g. `TOKEN_METADATA_SERVER_URI` in place of `METADATA_SERVER_URI` or  `TokenMetadataCacheTtl` in place of `MetadataCacheTtl`. This is best done as a separate `refactor(cardano-services)!:` commit.

iccicci:
Hi @rhyslbw and @mkazlauskas ;

I had to rebase the branch in order to introduce some new commits...
3a488e2 should solve all the remaining open topics.

packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
```diff
@@ -35,10 +35,42 @@ const mockTokenRegistry = async (handler: () => { body: unknown; code?: number }
...
+
 describe('CardanoTokenRegistry', () => {
   const notValidAssetId = Cardano.AssetId('0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef');
   const validAssetId = Cardano.AssetId('f43a62fdc3965df486de8a0d32fe800963589c41b38946602a0dc53541474958');
 
+  describe('parseTokenMetadataServiceRecord', () => {
+    it('complete attributes', () =>
+      expect(
+        parseTokenMetadataServiceRecord({
+          decimals: { value: 23 },
```
rhyslbw, line 86: Missing `subject`. Best to add it to the second test case too.

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
```diff
@@ -24,6 +25,14 @@ interface TokenMetadataServiceRecord {
   url?: StringValue;
 }
 
+const propertiesToChange: Record<string, string> = { description: 'desc', logo: 'icon' };
+export const parseTokenMetadataServiceRecord = (record: TokenMetadataServiceRecord) =>
```
rhyslbw, line 29:
```suggestion
export const toCoreTokenMetadata = (record: TokenMetadataServiceRecord) =>
```
This tells us what we get from the function :-)

packages/cardano-services/test/Program/utils.test.ts
```diff
@@ -495,17 +496,20 @@ describe('Service dependency abstractions', () => {
...
-          rabbitmqSrvServiceName: process.env.RABBITMQ_SRV_SERVICE_NAME,
-          serviceDiscoveryBackoffFactor: 1.1,
-          serviceDiscoveryTimeout: 1000
-        }
-      });
+      txSubmitWorker = await loadTxWorker(
+        {
+          options: {
+            cacheTtl: 10_000,
+            loggerMinSeverity: 'error',
```
rhyslbw, line 503:
```suggestion
```
This is now not needed since you're passing the `dummyLogger`

packages/cardano-services/src/run.ts
```diff
@@ -52,6 +53,10 @@ const envSpecs = {
   SERVICE_DISCOVERY_BACKOFF_FACTOR: envalid.num({ default: SERVICE_DISCOVERY_BACKOFF_FACTOR_DEFAULT }),
   SERVICE_DISCOVERY_TIMEOUT: envalid.num({ default: SERVICE_DISCOVERY_TIMEOUT_DEFAULT }),
   SERVICE_NAMES: envalid.str({ example: Object.values(ServiceNames).toString() }),
+  TOKEN_METADATA_CACHE_TTL: envalid.makeValidator(cacheTtlValidator)(
+    envalid.num({ default: DEFAULT_TOKEN_METADATA_CACHE_TTL })
+  ),
+  TOKEN_METADATA_SERVER_URI: envalid.url({ default: DEFAULT_TOKEN_METADATA_SERVER_URI }),
```
rhyslbw, line 59:
```suggestion
  TOKEN_METADATA_SERVER_URL: envalid.url({ default: DEFAULT_TOKEN_METADATA_SERVER_URL }),
```

I've just noticed the inconsistency, which we should correct. Since it's already in _master_, you can do a separate `refactor` commit to align it with our other variables that use "URL". Sorry for not picking this up in the prior PRs

rhyslbw (changes requested):
Nice work @iccicci, I can see some good improvements here, particularly the test mock.

Just a few minor things before we can wrap this up.

packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
iccicci: 434b443db should solve this

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
iccicci: d9ebbe08c should solve this

packages/cardano-services/test/Program/utils.test.ts
iccicci: 6f3d639c8 should solve this

packages/cardano-services/src/run.ts
iccicci: 4a2bf47fe refactors exactly the same variable (plus some more). 0bbc2e63c (fixup for the same commit) should solve this

rhyslbw (dismissed):
Please squash in preparation for merging :rocket:

rhyslbw (approved):
Looks good @iccicci!

I noticed a couple of grammar issues, which I've fixed in f85e6f72543e4998bf578e7fd3b5b24a45509e33, as it was easier than making the change request at this point in the review process. I also squashed the two spelling fix commits into 6c9cd7e68f5f7878c76c3c47cf152e6b8c374bf7 and removed the scope. Once you have reviewed my change, please squash and push.

rhyslbw (approved):

PR #322, opened 2022-07-11
chore(cardano-services): update queries to match new dbSync v13 schema

AngelCastilloB:
# Context

We must bump cardano-services to be compatible with the new Vasil Hardfork.

# Proposed Solution

Update queries to match the new DBsync v13 schema.

dbSync queries were updated


packages/core/src/Provider/NetworkInfoProvider/types.ts
```diff
@@ -3,7 +3,7 @@ import { Cardano, TimeSettings } from '../..';
 export type ProtocolParametersRequiredByWallet = Required<
   Pick<
     Cardano.ProtocolParametersAlonzo,
-    | 'coinsPerUtxoWord'
```
mkazlauskas, line 6: It's [used](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cip2/src/selectionConstraints.ts#L44) when computing minimum coin quantity. Is this just a rename or is it the `perByte` value? We might need to also adjust the computation.

packages/ogmios/package.json
```diff
@@ -40,7 +40,7 @@
     "ws": "^8.5.0"
   },
   "dependencies": {
-    "@cardano-ogmios/client": "5.4.0",
+    "@cardano-ogmios/client": "5.5.1",
```
rhyslbw, line 43:
```suggestion
    "@cardano-ogmios/client": "5.5.2",
```

https://github.com/CardanoSolutions/ogmios/releases/tag/v5.5.2

packages/cardano-services/test/ChainHistory/DbSyncChainHistoryProvider/__snapshots__/ChainHistoryBuilder.test.ts.snap
```diff
@@ -99,7 +99,7 @@ Map {
 
 exports[`ChainHistoryBuilder queryProtocolParams query protocol parameters from last epoch 1`] = `
 Object {
-  "coinsPerUtxoWord": NaN,
+  "coinsPerUtxoSize": NaN,
```
rhyslbw, line 102: Jest Snapshot need updating?

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/queries.ts
```diff
@@ -66,7 +66,7 @@ pools_delegates AS (
             LIMIT 1
       )
    JOIN stake_address sa ON 
-     sa.hash_raw  = pu.reward_addr 
+     sa.hash_raw  = pu.reward_addr_id
```
rhyslbw, line 69: These are not valid joins. `pu.reward_addr_id` is referencing the `sa.id` (`addr_id`) column

rhyslbw (changes requested):
The `cardano-node-ogmios` service in the `cardano-services` Docker Compose file needs to be bumped to the latest version.

packages/cardano-services/test/ChainHistory/DbSyncChainHistoryProvider/__snapshots__/ChainHistoryBuilder.test.ts.snap
```diff
@@ -99,7 +99,7 @@ Map {
 
 exports[`ChainHistoryBuilder queryProtocolParams query protocol parameters from last epoch 1`] = `
 Object {
-  "coinsPerUtxoWord": NaN,
+  "coinsPerUtxoByte": NaN,
```
rhyslbw, line 102: The value isn't changing, so it's not part of this PR's scope, but let's make sure to log this for follow-up work

packages/cardano-services/test/tsconfig.json
```diff
@@ -28,6 +28,18 @@
     },
     {
       "path": "../../wallet/src"
+    },
```
rhyslbw, line 31: Nice catch

rhyslbw (approved):
LGTM :rocket:

packages/cardano-services/test/ChainHistory/DbSyncChainHistoryProvider/__snapshots__/ChainHistoryBuilder.test.ts.snap
mkazlauskas, line 102: Is `NaN` ok here? Maybe we should include a Babbage block to test that this param exists.

I'm wondering why it was `NaN` before too, maybe this is an actual bug :thinking:

packages/cip2/src/selectionConstraints.ts
```diff
@@ -10,6 +10,8 @@ import {
 } from './types';
 import { ProtocolParametersRequiredByInputSelection } from '.';
 
+const coinsPerUTxOByteToCoinsPerUTxOWord = (coinsPerByte: number): number => coinsPerByte * 8;
```
mkazlauskas, line 13: @AngelCastilloB what is the source for `*8` :thinking:? I'm not confident that's correct, because:
Current value of this parameter in testnet is `4310*8=34480`
Old value (perWord) was `34482`, which is slightly higher than that.

However, when running one of our e2e tests (using old param) that had 1M coins in an output I got a tx submission error, which doesn't make sense as minimum coin should have become lower instead of higher :thinking:

mkazlauskas: Ok, let's do a `*9` then so that we at least get enough coin?

AngelCastilloB: @mkazlauskas Yes the multiplication by 8 is an approximation (the conversion is done the same way by the ledger team see [here](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/babbage/impl/src/Cardano/Ledger/Babbage/Translation.hs#L191)).

It doesn't yields the same value because the old value was not a multiple of 8, for now this is the only way we have to approximate the old value, however, as @rhyslbw mentioned, this is temporary until we update de CLS that supports the new calculations which require this new coinsPerByte value.

What is the test that is failing? I can take a look :).

rhyslbw (approved):

PR #324, opened 2022-07-11
feat: add missing transaction fields

lgobbi-atix:
# Context

Some transaction body fields have not yet been defined in our[ core type](https://github.com/input-output-hk/cardano-js-sdk/blob/6431f57ee5d0624784e9ae6c6b6a73b35a514ee0/packages/core/src/Cardano/types/Transaction.ts#L37-L48), so the task here is to complete the mapping to all available properties of the current era (Babbage is out of scope, as it requires the unreleased v11 of the CSL)

Add support for those fields we're missing in [createTransactionInternals](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/wallet/src/Transaction/createTransactionInternals.ts), [cslToCore](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/core/src/CSL/cslToCore/cslToCore.ts), [coreToCsl](https://github.com/input-output-hk/cardano-js-sdk/tree/master/packages/core/src/CSL/coreToCsl), [finalizeTx](https://github.com/input-output-hk/cardano-js-sdk/blob/6431f57ee5d0624784e9ae6c6b6a73b35a514ee0/packages/wallet/src/SingleAddressWallet.ts#L329-L361):

- collaterals
- mint
- requiredExtraSignatures
- scriptIntegrityHash
- witness

# Proposed Solution
- Add missing fields to `CreateTxInternalsProps` and `InitializeTxProps`
- Add missing fields to `createTransactionInternals`, `initializeTx` and `finalizeTx`
- Modify tests to include the new fields

### DIFF
--- a/packages/wallet/src/SingleAddressWallet.ts
+++ b/packages/wallet/src/SingleAddressWallet.ts
@@ -332,7 +332,11 @@ export class SingleAddressWallet implements ObservableWallet {
       auxiliaryData: props.auxiliaryData,
       certificates: props.certificates,
       changeAddress,
+      collaterals: props.collaterals,
       inputSelection,
+      mint: props.mint,
+      requiredExtraSignatures: props.requiredExtraSignatures,
+      scriptIntegrityHash: props.scriptIntegrityHash,
       validityInterval,
       withdrawals: props.withdrawals
     });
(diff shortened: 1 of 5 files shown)


rhyslbw (changes requested):
Thanks @lgobbi-atix

The commit can be scope to `wallet`, which I'll do if we get approval from other reviewers.

rhyslbw (approved):

ssong-iohk (approved):

ssong-iohk (approved):

mkazlauskas (approved):

PR #325, opened 2022-07-11
fix(wallet): remove assets no longer available in total balance

mirceahasegan:
# Context

`SingleAddressWallet.assets$` not removing assets when they're no longer available in `balance$`

# Proposed Solution
Remove no longer available assets.

### DIFF
--- a/packages/wallet/src/services/AssetsTracker.ts
+++ b/packages/wallet/src/services/AssetsTracker.ts
@@ -1,9 +1,9 @@
 import { Asset, Cardano } from '@cardano-sdk/core';
 import { BalanceTracker } from './types';
+import { Observable, forkJoin, map, mergeMap, of, tap } from 'rxjs';
 import { RetryBackoffConfig } from 'backoff-rxjs';
 import { TrackedAssetProvider } from './ProviderTracker';
 import { coldObservableProvider } from './util';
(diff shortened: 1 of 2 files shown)


mirceahasegan:
I trying a different approach as this one seems a bit messy

mkazlauskas (dismissed):
Sourcing ` balanceAssetIds$` twice feels a little odd but looks good enough for me. If you can find a cleaner solution - great, otherwise I'm happy with this as-is.

mirceahasegan:
I have added an alternate solution that comes with the following benefits:
- Does not subscribe twice to balanceTracker
- Does not verify the emitted map for equality thanks to it not repeating balanceTracker emitted values
- balanceTracker updates are emitted after metadata was fetched for all assets instead of fetching one by one and emitting incremental Map objects
- Does not 'remove', instead it rebuilds the map using only the assetIds from balance, with the asset info from the cached assets result

mkazlauskas (approved):
Well done! :tada:

mirceahasegan:
I have rebased on master and squashed the changes.

mkazlauskas (approved):

rhyslbw (approved):
Great work @mirceahasegan 💪

PR #326, opened 2022-07-12
feat: implement cip30 getCollateral

jvieiro:
# Context

Implements CIP 30 getCollateral method

# Proposed Solution
 
Implements getCollateral method according to what is described [here](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0030#apigetcollateralparams--amount-cborcoin--promisetransactionunspentoutput--null)

### DIFF
--- a/packages/cip30/src/WalletApi/types.ts
+++ b/packages/cip30/src/WalletApi/types.ts
@@ -63,7 +63,7 @@ export type GetUtxos = (amount?: Cbor, paginate?: Paginate) => Promise<Cbor[] | 
  * for transactions with plutus script inputs (pure ADA-only UTxOs).
  * @throws ApiError
  */
-export type GetCollateral = (params: { amount: Cbor }) => Promise<Cbor[] | null>;
+export type GetCollateral = (params?: { amount?: Cbor }) => Promise<Cbor[] | null>;
 
 /**
  * Returns the total balance available of the wallet.
--- a/packages/wallet/src/cip30.ts
+++ b/packages/wallet/src/cip30.ts
@@ -51,9 +51,25 @@ export type CallbackConfirmation = (
   args: SignDataCallbackParams | SignTxCallbackParams | SubmitTxCallbackParams
 ) => Promise<boolean>;
 
+interface CslInterface {
+  to_bytes(): Uint8Array;
+}
+
 const mapCallbackFailure = (err: unknown, logger?: Logger) => {
   logger?.error(err);
   return false;
+};
+
+const MAX_COLLATERAL_AMOUNT = CSL.BigNum.from_str('5000000');
+
+const cslToCbor = (csl: CslInterface) => Buffer.from(csl.to_bytes()).toString('hex');
+
+const compareUtxos = (utxo: Cardano.Utxo, comparedTo: Cardano.Utxo) => {
+  const currentCoin = utxo[1].value.coins;
+  const comparedToCoin = comparedTo[1].value.coins;
+  if (currentCoin < comparedToCoin) return -1;
+  if (currentCoin > comparedToCoin) return 1;
+  return 0;
 };
 
 export const createWalletApi = (
@@ -92,9 +108,38 @@ export const createWalletApi = (
       throw new ApiError(500, 'Nope');
     }
   },
-  getCollateral: async () => {
-    logger.warn('getCollateral is not implemented');
-    return null;
+  getCollateral: async ({ amount = cslToCbor(MAX_COLLATERAL_AMOUNT) }: { amount?: Cbor } = {}): Promise<
+    Cbor[] | null
+    // eslint-disable-next-line sonarjs/cognitive-complexity
+  > => {
+    logger.debug('getting collateral');
+    const wallet = await walletReady;
+    let unspendables = (await firstValueFrom(wallet.utxo.unspendable$)).sort(compareUtxos);
+    if (unspendables.some((utxo) => utxo[1].value.assets && utxo[1].value.assets.size > 0))
+      throw new ApiError(400, 'not enough coins in unspendable UTxOs');
+    if (amount) {
+      try {
+        const filterAmount = CSL.BigNum.from_bytes(Buffer.from(amount, 'hex'));
+        if (filterAmount.compare(MAX_COLLATERAL_AMOUNT) > 0) throw new ApiError(400, 'given amount is too big');
+        const utxos = [];
+        let totalCoins = CSL.BigNum.from_str('0');
+        for (const utxo of unspendables) {
+          const coin = CSL.BigNum.from_str(utxo[1].value.coins.toString());
+          totalCoins = totalCoins.checked_add(coin);
+          utxos.push(utxo);
+          if (totalCoins.compare(filterAmount) !== -1) break;
+        }
+        if (totalCoins.compare(filterAmount) === -1) throw new ApiError(400, 'not enough coins in unspendable UTxOs');
+        unspendables = utxos;
+      } catch (error) {
+        logger.error(error);
+        if (error instanceof ApiError) {
+          throw error;
+        }
+        throw new ApiError(500, 'Nope');
+      }
+    }
+    return coreToCsl.utxo(unspendables).map(cslToCbor);
   },
   getNetworkId: async (): Promise<Cardano.NetworkId> => {
     logger.debug('getting networkId');
@@ -140,7 +185,6 @@ export const createWalletApi = (
   getUtxos: async (amount?: Cbor, paginate?: Paginate): Promise<Cbor[] | undefined> => {
     const wallet = await walletReady;
     let utxos = await firstValueFrom(wallet.utxo.available$);
-
     if (amount) {
       try {
         const filterAmount = CSL.Value.from_bytes(Buffer.from(amount, 'hex'));
@@ -166,7 +210,7 @@ export const createWalletApi = (
       utxos = utxos.slice(paginate.page * paginate.limit, paginate.page * paginate.limit + paginate.limit);
     }
 
-    return Promise.resolve(coreToCsl.utxo(utxos).map((utxo) => Buffer.from(utxo.to_bytes()).toString('hex')));
+    return coreToCsl.utxo(utxos).map(cslToCbor);
   },
   signData: async (addr: Cardano.Address, payload: Bytes): Promise<Cip30DataSignature> => {
     logger.debug('signData');
--- a/packages/wallet/src/cip30.ts
+++ b/packages/wallet/src/cip30.ts
@@ -1,4 +1,5 @@
 import {
+  APIErrorCode,
   ApiError,
   Bytes,
   Cbor,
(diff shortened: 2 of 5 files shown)


packages/wallet/src/cip30.ts
```diff
@@ -92,9 +102,23 @@ export const createWalletApi = (
       throw new ApiError(500, 'Nope');
     }
   },
-  getCollateral: async () => {
-    logger.warn('getCollateral is not implemented');
-    return null;
+  getCollateral: async (params: { amount: Cbor }): Promise<Cbor[] | null> => {
```
rhyslbw, line 105: The spec says `null` is a valid return type, however, `getUtxos` was implemented with `undefined` instead :thinking:

packages/wallet/src/cip30.ts
```diff
@@ -92,9 +102,23 @@ export const createWalletApi = (
...
+    const amount = bigIntAmount > MAX_COLLATERAL_AMOUNT ? MAX_COLLATERAL_AMOUNT : bigIntAmount;
+    const utxos = [];
+    let totalSum = BigInt('0');
+    for (const utxo of unspendables) {
+      const utxoAmount = utxo[1].value.coins;
+      totalSum += utxoAmount;
+      utxos.push(utxo);
+      if (totalSum >= amount) break;
+    }
+    if (totalSum < amount) throw new ApiError(400, 'unspendable utxos are not enough');
```
rhyslbw, line 120:
```suggestion
    if (totalSum < amount) throw new ApiError(400, 'not enough coins in unspendable UTxOs');
```

packages/wallet/src/cip30.ts
mkazlauskas: Yep seems like a bug with getUtxos

packages/wallet/src/cip30.ts
```diff
@@ -92,9 +102,23 @@ export const createWalletApi = (
...
-  getCollateral: async () => {
-    logger.warn('getCollateral is not implemented');
-    return null;
+  getCollateral: async (params: { amount: Cbor }): Promise<Cbor[] | null> => {
+    logger.debug('getting collateral');
+    const wallet = await walletReady;
+    // unspendable utxo should not be used by anything else other than collateral
+    const unspendables = (await firstValueFrom(wallet.utxo.unspendable$)).sort(compareUtxos);
+    const bigIntAmount = BigInt(params.amount);
+    const amount = bigIntAmount > MAX_COLLATERAL_AMOUNT ? MAX_COLLATERAL_AMOUNT : bigIntAmount;
```
mkazlauskas, line 111: I think this breaks the contract. Maybe throw `APIError` instead

packages/wallet/src/cip30.ts
mkazlauskas, line 105: This might be a thing:
> NOTE: some wallets may be ignoring the amount parameter, in which case it might be possible to call the function without it, but this behavior is not recommended!)

```suggestion
  getCollateral: async ({ amount = cbor(MAX_COLLATERAL_AMOUNT)}: { amount?: Cbor } = {}): Promise<Cbor[] | null> => {
```

mkazlauskas (changes requested):

packages/wallet/src/cip30.ts
mkazlauskas: Hm no,
- CIP specifies the amount as CBOR so we need to use `CSL.BigNum` to parse it (I guess, it's not 100% clear what type is it, could be CSL.Value as well but let's just go with BigNum). So actually this is also incorrect: `const bigIntAmount = BigInt(amount);`
- The NOTE says that it could be called without arguments, so we need a default `{}`

```
getCollateral: async ({ amount = CSL.BigNum.fromStr(MAX_COLLATERAL_AMOUNT.toString())}: { amount?: Cbor } = {}): Promise<Cbor[] | null>
```

jvieiro: Since the contract expects an amount of type `string` it should be `MAX_COLLATERAL_AMOUNT.to_str() `(assuming `MAX_COLLATERAL_AMOUNT = CSL.BigNum.from_str('5000000');` ) and internally we have to map it to `BigNum` to operate.
Also, what type of string are we expecting? I know `Cbor = string` but this represents a hex string, right? So we have to build the object `CSL.BigNum.from_bytes(Buffer.from(amount,'hex'))`

mkazlauskas: Seems right, internally you can probably do `BigInt(BigNum.toString())`

mkazlauskas (approved):
Looks good! Just a couple of conflicts with master to resolve

rhyslbw (approved):
Nice work @jvieiro :rocket:

jvieiro:
> Looks good! Just a couple of conflicts with master to resolve

Done!

packages/wallet/src/cip30.ts
```diff
@@ -92,9 +104,33 @@ export const createWalletApi = (
...
+        unspendables = utxos;
+      } catch (error) {
+        logger.error(error);
+        if (error instanceof ApiError) {
+          throw error;
+        }
+        throw new ApiError(500, 'Nope');
+      }
+    }
+    return Promise.resolve(coreToCsl.utxo(unspendables).map((utxo) => Buffer.from(utxo.to_bytes()).toString('hex')));
```
mkazlauskas, line 133: Looks good, just noticed it has a redundant `Promise.resolve` (in `getUtxos` as well). Maybe wrap it in `utxosToCbor` fn to convert the return value of both functions.

```suggestion
    return coreToCsl.utxo(unspendables).map((utxo) => Buffer.from(utxo.to_bytes()).toString('hex'));
```

jvieiro: Good catch, I've forgot to change the type of the parameter

mkazlauskas (dismissed):

rhyslbw (dismissed):

packages/wallet/test/integration/cip30mapping.test.ts
```diff
@@ -49,6 +55,14 @@ describe('cip30', () => {
       ).not.toThrow();
     });
 
+    test('api.getCollateral', async () => {
```
iadmytro, line 58: Missing cases:
- no utxo in the wallet
- combined utxo value in the wallet is less the amount
- combined utxo value more than amount and less then max_amount (more that 1 utxo needed to satisfy amount)
- amount more then max_amount
- amount is 0

iadmytro (changes requested):

packages/wallet/src/cip30.ts
```diff
@@ -92,9 +108,35 @@ export const createWalletApi = (
...
+          if (totalCoins.compare(filterAmount) !== -1) break;
+        }
+        if (totalCoins.compare(filterAmount) === -1) throw new ApiError(400, 'not enough coins in unspendable UTxOs');
+        unspendables = utxos;
+      } catch (error) {
+        logger.error(error);
+        if (error instanceof ApiError) {
+          throw error;
+        }
+        throw new ApiError(500, 'Nope');
```
ssong-iohk, line 136:
```suggestion
        throw new ApiError(500, 'Nope'); // We can define more meaningful error message such as 'Error occurred during validating given amount of collateral' 
```

ssong-iohk (commented):
Leave a suggestion but not strongly request changes. :)

packages/wallet/src/cip30.ts
```diff
@@ -51,11 +51,27 @@ export type CallbackConfirmation = (
...
+interface CslInterface {
+  to_bytes(): Uint8Array;
+}
+
 const mapCallbackFailure = (err: unknown, logger?: Logger) => {
   logger?.error(err);
   return false;
 };
 
+const MAX_COLLATERAL_AMOUNT = CSL.BigNum.from_str('5000000');
```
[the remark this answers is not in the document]
jvieiro: Yes it is

packages/wallet/test/integration/cip30mapping.test.ts
jvieiro: Added

packages/cip30/src/WalletApi/types.ts
```diff
@@ -63,7 +63,7 @@ export type GetUtxos = (amount?: Cbor, paginate?: Paginate) => Promise<Cbor[] |
  * for transactions with plutus script inputs (pure ADA-only UTxOs).
  * @throws ApiError
  */
-export type GetCollateral = (params: { amount: Cbor }) => Promise<Cbor[] | null>;
+export type GetCollateral = (params: { amount?: Cbor }) => Promise<Cbor[] | null>;
```
jvieiro, line 66: @mkazlauskas this should look like this, right?

packages/wallet/src/cip30.ts
jvieiro: I wouldn't say it is being use to validate the amount. I think it is used to handle unexpected errors. So, message would be something like 'Unexpected error occured'. It should be changed in other methods to be constant.

packages/cip30/src/WalletApi/types.ts
mkazlauskas: Spec says dApps might call this withoutu arguments: `params?: { amount?: Cbor }`

packages/wallet/test/integration/cip30mapping.test.ts
jvieiro: Isn't this being checked with this expect?: 
```
expect(() =>
  cslToCore.utxo(utxos!.map((utxo) => CSL.TransactionUnspentOutput.from_bytes(Buffer.from(utxo, 'hex'))))
).not.toThrow();
```

mkazlauskas: I think by 'pure' @iadmytro means that they must have no assets (coin only)

jvieiro: So, unspendables returned by the walled should be checkd? Shouldn't this be done when setting them?

jvieiro: Alright. I've added this new validation

packages/wallet/src/cip30.ts
```diff
@@ -92,9 +108,38 @@ export const createWalletApi = (
...
-    return null;
+  getCollateral: async ({ amount = cslToCbor(MAX_COLLATERAL_AMOUNT) }: { amount?: Cbor } = {}): Promise<
+    Cbor[] | null
+    // eslint-disable-next-line sonarjs/cognitive-complexity
+  > => {
+    logger.debug('getting collateral');
+    const wallet = await walletReady;
+    let unspendables = (await firstValueFrom(wallet.utxo.unspendable$)).sort(compareUtxos);
+    if (unspendables.some((utxo) => utxo[1].value.assets && utxo[1].value.assets.size > 0))
+      throw new ApiError(400, 'unspendable UTxOs must not contain assets');
```
mkazlauskas, line 119:
```suggestion
      throw new ApiError(APIErrorCode.InternalError, 'unspendable UTxOs must not contain assets when used as collateral');
```

packages/wallet/src/cip30.ts
```diff
@@ -92,9 +108,38 @@ export const createWalletApi = (
...
+    if (amount) {
+      try {
+        const filterAmount = CSL.BigNum.from_bytes(Buffer.from(amount, 'hex'));
+        if (filterAmount.compare(MAX_COLLATERAL_AMOUNT) > 0) throw new ApiError(400, 'given amount is too big');
```
mkazlauskas, line 123:
```suggestion
        if (filterAmount.compare(MAX_COLLATERAL_AMOUNT) > 0) throw new ApiError(APIErrorCode.InvalidRequest, 'requested amount is too big');
```

packages/wallet/src/cip30.ts
```diff
@@ -92,9 +108,38 @@ export const createWalletApi = (
...
+        const utxos = [];
+        let totalCoins = CSL.BigNum.from_str('0');
+        for (const utxo of unspendables) {
+          const coin = CSL.BigNum.from_str(utxo[1].value.coins.toString());
+          totalCoins = totalCoins.checked_add(coin);
+          utxos.push(utxo);
+          if (totalCoins.compare(filterAmount) !== -1) break;
+        }
+        if (totalCoins.compare(filterAmount) === -1) throw new ApiError(400, 'not enough coins in unspendable UTxOs');
```
mkazlauskas, line 132:
```suggestion
        if (totalCoins.compare(filterAmount) === -1) throw new ApiError(APIErrorCode.Refused, 'not enough coins in configured collateral UTxOs');
```

mkazlauskas (changes requested):

mkazlauskas:
We're using some other invalid ApiError codes, @jvieiro could you please update those as well? The only valid values for code is in `APIErrorCode` enum.

jvieiro:
> We're using some invalid ApiError codes in other functions, it's a bug. @jvieiro could you please update those as well in a separate commit? The only valid values for error code is in `APIErrorCode` enum.

Just added!

mkazlauskas (approved):
Looks good!

iadmytro (approved):

PR #330, opened 2022-07-13
Chore/modernise golden test generator

rhyslbw:
# Context
This package had a direct dependency on an older version of the Ogmios client, and the linting was being skipped.

# Proposed Solution
- Import the Ogmios deps from `@cardano-sdk/ogmios`, updating the implementation for minimum compatibility.
- Fix lint issues, reinstate linting scripts


mkazlauskas (approved):
Looks good!

PR #333, opened 2022-07-14
Chore/Upgrade to Yarn 3

rhyslbw:
# Context
[Prior](https://github.com/input-output-hk/cardano-js-sdk/pull/245) [attempts](https://github.com/input-output-hk/cardano-js-sdk/pull/302) at implementing a Nix reproducible build have failed due to the shortcomings of Yarn 1 as a resolver and enforcer of project dependencies. Being able to build the SDK with Nix is a requirement for backend deployments, but also unlocks capabilities within the development and CI environments.

# Proposed Solution
- Migrate from Yarn 1 to Yarn 3.
- Make inferred dependencies explicit.
- Remove erroneous use of `got` in `cardano-service`, which was missed in the refactor to `axios`.

- This PR also included bumping a number of dependencies to address security alerts.
- `nanoid` also requires patching when running within Jest 28.


blaggacao:
@rhyslbw Can you rebase and run just `yarn`?

rhyslbw:
Yes, sorry @blaggacao, there was a bad rebase merge (package ordering), and I failed to merge and push the the lock file changes. Both issues have been corrected in 716f67830c5eefe7da3965f6eadbf326f4fc67f4 , so you can drop be29a70873fb0e10c0e5352488b1f33382548b43 and rebase.

rhyslbw:
I'd like to squash f91b6929a5ca2ffa0da03b75feeacf6ef5480eb8 into 84da7bdd37f41eedfb6c180b7bb4dc4f358b4005, but left it separate until it's been reviewed.

packages/cardano-services/test/StakePool/StakePoolHttpService.test.ts
```diff
@@ -15,7 +15,7 @@ import { DbSyncStakePoolProvider, HttpServer, HttpServerConfig, StakePoolHttpSer
 import { Pool } from 'pg';
 import { doServerRequest } from '../util';
 import { getPort } from 'get-port-please';
-import { stakePoolHttpProvider } from '@cardano-sdk/cardano-services-client';
+import { stakePoolHttpProvider } from '../../../cardano-services-client';
```
AngelCastilloB, line 18: I think we should use @package-name when using modules from different packages and relative paths when using modules from the same package; looks cleaner IMHO; also, we are doing so in other packages, so it's also more consistent, is there any reason why this needed to change?

rhyslbw: It's also my preference, but I suspect there's an issue with our tsconfig that is preventing referenced imports in tests, as we've already had to refactor some prior to this change.

> we are doing so in other packages

In source code it's fine, but all tests should now have relative path imports.

AngelCastilloB (approved):
LGTM!

packages/cardano-services/test/StakePool/StakePoolHttpService.test.ts
mkazlauskas: If we think `from '@package'` syntax has advantages we could go back to using path mappings in tsconfig. Should be a separate task though

mkazlauskas (dismissed):
Looks good! :rocket:

packages/cardano-services/test/StakePool/StakePoolHttpService.test.ts
rhyslbw: `ADP-2032`

AngelCastilloB (approved):

mkazlauskas (dismissed):

AngelCastilloB (approved):

mkazlauskas (approved):

PR #336, opened 2022-07-15
nix/take 3

blaggacao:
This allows to use this package in other Nix based systems.

Note that this is focused on cardano-services, and `nix run` will call its CLI.

The package is based on https://github.com/stephank/yarn-plugin-nixify that is included in Yarn config and automatically generates Nix expression that can be used for packaging in Nix.

This could be improved using [isolated builds](https://github.com/stephank/yarn-plugin-nixify/blob/main/ISOLATED_BUILDS.md) but it is incompatible with `nodeLinker: node-modules` that is set during transition to Plug-n-Play.


rhyslbw (changes requested):
Nice work @blaggacao and @YorikSar 🤝 

We're using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/), so when it comes time to squash to a single commit, please use a prefix of `build:`.

mkazlauskas (commented):
Looks good! yarn plugin made it much cleaner :sunglasses:

Just need to squash into `build:` commit as @rhyslbw mentioned

~~Also a question: do we need to commit .yarn/plugins/yarn-plugin-nixify.cjs since it's location is specified in .yarnrc?~~

YorikSar:
@mkazlauskas I don't really know what is the established way for Yarn plugins, but I would prefer to have it in repo to not depend on external systems during build and development.

rhyslbw (approved):

mkazlauskas (approved):

PR #338, opened 2022-07-16
fix(wallet)!: fetch time settings only when epoch changes (ADP-1682)

mirceahasegan:
Time settings is fetched once when epoch or time settings is
subscribed to, then refetched every time the epoch changes.

BREAKING CHANGE:
  - currentEpochTracker creates an observable that emits a tuple of
    TimeSettings and EpochInfo, instead of just EpochInfo.

# Context

TimeSettings is currently fetched every block, because it's using tipBlockHeight$ as trigger/source.
It can only change once per epoch, so we're doing some redundant fetching here.
Challenge here is that computing epoch # depends on knowing current TimeSettings, so it's a circular dependency.

# Proposed Solution

`TimeSettingsManualTracker` wraps the provider and fetches once on subscription, while having a manual `refetch()` method to allow manual refetching.
`EpochTracker` subscribes to `TimeSettingsManualTracker`, thus triggering the initial subscription.
On epoch change, `EpochTracker` call `refetch()` on `TimeSettingsManualTracker`.
`EpochTracker` always emits a tuple of `[TimeSettings, EpochInfo]`.
`SingleAddressWallet` `currentEpoch$` and `timeSettings$` are both using `EpochTracker` as observable source, extracting the `TimeSettings` and `EpochInfo` respectively.

currentEpochTracker creates an observable that emits a tuple of TimeSettings and EpochInfo, instead of just EpochInfo.

### DIFF
--- a/packages/wallet/src/SingleAddressWallet.ts
+++ b/packages/wallet/src/SingleAddressWallet.ts
@@ -51,12 +51,8 @@ import {
   groupedAddressesEquals
 } from './services';
 import { BehaviorObservable, TrackerSubject } from '@cardano-sdk/util-rxjs';
-import { Cip30DataSignature } from '@cardano-sdk/cip30';
-import { InputSelector, defaultSelectionConstraints, roundRobinRandomImprove } from '@cardano-sdk/cip2';
-import { Logger, dummyLogger } from 'ts-log';
-import { RetryBackoffConfig } from 'backoff-rxjs';
-import { Shutdown } from '@cardano-sdk/util';
 import {
+  BehaviorSubject,
   Subject,
   combineLatest,
   concat,
@@ -68,6 +64,11 @@ import {
   take,
   tap
 } from 'rxjs';
+import { Cip30DataSignature } from '@cardano-sdk/cip30';
+import { InputSelector, defaultSelectionConstraints, roundRobinRandomImprove } from '@cardano-sdk/cip2';
+import { Logger, dummyLogger } from 'ts-log';
+import { RetryBackoffConfig } from 'backoff-rxjs';
+import { Shutdown } from '@cardano-sdk/util';
 import { TrackedUtxoProvider } from './services/ProviderTracker/TrackedUtxoProvider';
 import { TxInternals, createTransactionInternals, ensureValidityInterval } from './Transaction';
 import { WalletStores, createInMemoryWalletStores } from './persistence';
@@ -104,6 +105,7 @@ export class SingleAddressWallet implements ObservableWallet {
     pending$: new Subject<Cardano.NewTxAlonzo>(),
     submitting$: new Subject<Cardano.NewTxAlonzo>()
   };
+
   readonly keyAgent: AsyncKeyAgent;
   readonly currentEpoch$: TrackerSubject<EpochInfo>;
   readonly txSubmitProvider: TrackedTxSubmitProvider;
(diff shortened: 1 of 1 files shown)


packages/wallet/src/services/TimeSettingsManualTracker.ts
```diff
@@ -0,0 +1,45 @@
+import { BehaviorSubject } from 'rxjs';
```
mirceahasegan, line 1: I omitted tests for this file intentionally because:
- It is only a wrapper over coldObservable.
- Testing it would mean doing similar mocks as in the coldObservable tests, which would dup the work there.
- An alternative would be to mock the coldObservable itself, leaving nothing to test but a BehaviourSubject.

mkazlauskas (changes requested):

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -213,23 +213,25 @@ export class SingleAddressWallet implements ObservableWallet {
...
+      cancel$,
+      provider: this.networkInfoProvider.timeSettings,
+      retryBackoffConfig,
+      store: stores.timeSettings
+    });
+
+    // epochAndTimeSettingsTracker uses initial time settings fetch to create the epoch info
+    // Afterwards, each epoch change also triggers refetch of time settings
+    const epochAndTimeSettingsTracker = currentEpochTracker(this.tip$, timeSettingsManualTracker);
+    this.currentEpoch$ = new TrackerSubject(epochAndTimeSettingsTracker.pipe(map(([, currentEpoch]) => currentEpoch)));
```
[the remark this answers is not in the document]
mirceahasegan: > Could maybe keep the refetch logic within SingleAddressWallet?

I have done the requested changes locally and it is indeed less invasive this way, by keeping the changes local to SingleAddressWallet. Thank you @mkazlauskas !
One problem I am facing now is that some tests are failing due to timeouts. I will investigate and push after I figure out why.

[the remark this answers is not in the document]
mkazlauskas: You likely need to complete() the subject in SingleAddressWallet.shutdown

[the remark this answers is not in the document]
mirceahasegan: I have done the changes.
I had to force push since the changes were not a fixup, but a different approach, so there was no value in keeping incremental changes (also i rebased on master)

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -213,23 +217,31 @@ export class SingleAddressWallet implements ObservableWallet {
...
     });
     const tipBlockHeight$ = distinctBlock(this.tip$);
-    this.timeSettings$ = new PersistentDocumentTrackerSubject(
-      // TODO: it only really needs to be fetched once per epoch,
-      // so we should replace the trigger from tipBlockHeight$ to epoch$.
-      // This is a little complicated since there is a circular dependency.
-      // Some logic is needed to initiate a fetch if epoch is not available in store already.
+
+    this.#timeSettingsTrigger = new BehaviorSubject<void>(undefined);
+    this.#timeSettingsManualTracker$ = new PersistentDocumentTrackerSubject(
```
mkazlauskas, line 222: Would this work?

```suggestion
    this.timeSettings$ = new PersistentDocumentTrackerSubject(
```

mirceahasegan: I don't think it will work in this case: timeSettings$ is subscribed to, but epoch$ is not subscribed. In this case, timeSettings$ will not be refetched

mirceahasegan: Updates are in [33d730e](https://github.com/input-output-hk/cardano-js-sdk/pull/338/commits/33d730e6c21ad834e9feec984870b90b526b2cb8)

rhyslbw (approved):
Great result @mirceahasegan!

mkazlauskas (approved):
Looks good!

mkazlauskas (approved):

rhyslbw (approved):

PR #341, opened 2022-07-19
test: cardano-services tests racing for binding the port

jvieiro:
# Context

yarn test sometimes fail with an error from httpServer.start() when it cannot bind to port.
yarn test --runInBand seems to be working well, so the tests are probably racing to bind the same port.

# Proposed Solution

Refactor tests and replace `getPort()` per `getRandomPort()` and remove `--runInBand` flag from `test` script


mkazlauskas (commented):
Replacing`getPort` with `getRandomPort` lowers the probability of binding the same port, but does not really solve the problem, unless it's guarantees to not return the same port twice. Does it?

rhyslbw:
> Does it?

:shrug: There's no tests expressing the behaviour of `getRandomPort` https://github.com/unjs/get-port-please/blob/main/test/index.test.ts, but I think we still have a potential race condition here.

PR #344, opened 2022-07-19
feat: service discovery error handling for TxSubmitWorker

IvayloAndonov:
# Context
Since dependency port resolution using DNS SRV records for the HTTP server is in place, we need also to wrap [TxSubmitWorker](https://github.com/input-output-hk/cardano-js-sdk/blob/8fd7c5267f8250848f954350f58bfef3735aea12/packages/cardano-services/src/TxWorker/loadTxWorker.ts#L26) within abstraction which to perform DNS resolution with retry logic if a connection error occurs between the `worker` and `rabbitmq` broker.

# Proposed Solution
- `TxSubmitWorker` class extends `EventEmitter` to be able to raise events, in our case connection-related.
- Implement `getRunningTxSubmitWorker` abstraction which determines how to instantiate TxSubmitWorker based on the provided args ( with or without service discovery )
- Implement `startTxSubmitWorkerWithDiscovery` which manages worker instances and handles connection-related errors through a listener, returns a `RunningTxSubmitWorker`, with an argument `workerFactory:WorkerFactory`
- Implement `createWorkerFactoryWithDiscovery` - creates a factory that produces `TxSubmitWorker` with service discovery

This approach gives us flexibility and control over the testing and worker factory mock.

### DIFF
--- a/packages/cardano-services/src/TxWorker/utils.ts
+++ b/packages/cardano-services/src/TxWorker/utils.ts
@@ -1,0 +1,98 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/no-shadow */
+/* eslint-disable no-use-before-define */
+import { CONNECTION_ERROR_EVENT, TxSubmitWorker } from '@cardano-sdk/rabbitmq';
+import { CommonOptionDescriptions } from '../ProgramsCommon';
+import { DnsResolver, MissingProgramOption, ServiceNames, srvRecordToRabbitmqURL } from '../Program';
+import { Logger } from 'ts-log';
+import { TxSubmitProvider } from '@cardano-sdk/core';
+import { TxWorkerOptions } from './loadTxWorker';
+
+type WorkerFactory = () => Promise<TxSubmitWorker>;
+
+/**
+ * Create a worker factory with service discovery
+ *
+ * @param {DnsResolver} dnsResolver used for dns resoluton
+ * @param {TxSubmitProvider} txSubmitProvider tx submit provider 'ogmiosTxSubmitProvider'
+ * @param {Logger} logger common logger
+ * @param {TxWorkerOptions} options needed for tx worker initialization
+ * @returns {WorkerFactory} WorkerFactory with service discovery, returning a 'TxSubmitWorker' instance
+ */
+export const createWorkerFactoryWithDiscovery =
+  (
+    dnsResolver: DnsResolver,
+    txSubmitProvider: TxSubmitProvider,
+    logger: Logger,
+    options: TxWorkerOptions
+  ): WorkerFactory =>
+  async () => {
+    const record = await dnsResolver(options.rabbitmqSrvServiceName!);
+    return new TxSubmitWorker(
+      { ...options, rabbitmqUrl: srvRecordToRabbitmqURL(record) },
+      { logger, txSubmitProvider }
+    );
+  };
+
+/**
+ * Create and start a new worker instance with registered listener for connection error events
+ *
+ * @param {WorkerFactory} workerFactory creates a worker instance with service discovery
+ */
+export const createAndStartNewWorker = async (workerFactory: WorkerFactory) => {
+  const newWorker = await workerFactory();
+  attachOnConnectionErrorHandler(newWorker, workerFactory);
+  await newWorker.start();
+};
+
+/**
+ * With service discovery, register a listener for 'connection-error' event type
+ * emitted in the case of a connection-related runtime error
+ *
+ * @param {TxSubmitWorker} worker TxSubmitWorker instance
+ * @param {WorkerFactory} factory creates a worker instance with service discovery
+ */
+export const attachOnConnectionErrorHandler = (worker: TxSubmitWorker, factory: WorkerFactory) => {
+  worker.on(CONNECTION_ERROR_EVENT, () => createAndStartNewWorker(factory));
+};
+
+export type RunningTxSubmitWorker = Pick<TxSubmitWorker, 'stop' | 'getStatus'>;
+
+/**
+ * An abstraction which starts and manages restarts of the worker instance with service discovery
+ *
+ * @param {WorkerFactory} workerFactory creates a worker instance with service discovery
+ * @returns {RunningTxSubmitWorker} RunningTxSubmitWorker instance
+ */
+export const startTxSubmitWorkerWithDiscovery = async (
+  workerFactory: WorkerFactory
+): Promise<RunningTxSubmitWorker> => {
+  let worker: TxSubmitWorker;
+  await createAndStartNewWorker(async () => (worker = await workerFactory()));
+
+  return {
+    getStatus: () => worker.getStatus(),
+    stop: () => worker.stop()
+  };
+};
+
+export const getRunningTxSubmitWorker = async (
+  dnsResolver: DnsResolver,
+  txSubmitProvider: TxSubmitProvider,
+  logger: Logger,
+  options?: TxWorkerOptions
+): Promise<RunningTxSubmitWorker> => {
+  if (options?.rabbitmqSrvServiceName)
+    return startTxSubmitWorkerWithDiscovery(
+      createWorkerFactoryWithDiscovery(dnsResolver, txSubmitProvider, logger, options)
+    );
+  if (options?.rabbitmqUrl) {
+    const worker = new TxSubmitWorker({ ...options, rabbitmqUrl: options.rabbitmqUrl }, { logger, txSubmitProvider });
+    await worker.start();
+    return worker;
+  }
+  throw new MissingProgramOption(ServiceNames.TxSubmit, [
+    CommonOptionDescriptions.RabbitMQUrl,
+    CommonOptionDescriptions.RabbitMQSrvServiceName
+  ]);
+};
--- a/packages/cardano-services/src/TxWorker/utils.ts
+++ b/packages/cardano-services/src/TxWorker/utils.ts
@@ -5,7 +5,7 @@ import { CONNECTION_ERROR_EVENT, TxSubmitWorker } from '@cardano-sdk/rabbitmq';
 import { CommonOptionDescriptions } from '../ProgramsCommon';
 import { DnsResolver, MissingProgramOption, ServiceNames, srvRecordToRabbitmqURL } from '../Program';
 import { Logger } from 'ts-log';
-import { TxSubmitProvider } from '@cardano-sdk/core';
+import { ProviderError, ProviderFailure, TxSubmitProvider } from '@cardano-sdk/core';
 import { TxWorkerOptions } from './loadTxWorker';
 
 type WorkerFactory = () => Promise<TxSubmitWorker>;
@@ -13,7 +13,7 @@ type WorkerFactory = () => Promise<TxSubmitWorker>;
 /**
  * Create a worker factory with service discovery
  *
- * @param {DnsResolver} dnsResolver used for dns resoluton
+ * @param {DnsResolver} dnsResolver used for DNS resolution
  * @param {TxSubmitProvider} txSubmitProvider tx submit provider 'ogmiosTxSubmitProvider'
  * @param {Logger} logger common logger
  * @param {TxWorkerOptions} options needed for tx worker initialization
@@ -76,6 +76,16 @@ export const startTxSubmitWorkerWithDiscovery = async (
   };
 };
 
+/**
+ * Create and return a running worker instance with static service config or service discovery
+ *
+ * @param {DnsResolver} dnsResolver used for DNS resolution
+ * @param {TxSubmitProvider} txSubmitProvider tx submit provider 'ogmiosTxSubmitProvider'
+ * @param {Logger} logger common logger
+ * @param {TxWorkerOptions} options needed for tx worker initialization
+ * @returns {RunningTxSubmitWorker} RunningTxSubmitWorker instance
+ * @throws {MissingProgramOption} error if neither URL nor service name is provided
+ */
 export const getRunningTxSubmitWorker = async (
   dnsResolver: DnsResolver,
   txSubmitProvider: TxSubmitProvider,
@@ -88,6 +98,10 @@ export const getRunningTxSubmitWorker = async (
     );
   if (options?.rabbitmqUrl) {
     const worker = new TxSubmitWorker({ ...options, rabbitmqUrl: options.rabbitmqUrl }, { logger, txSubmitProvider });
+    worker.on(CONNECTION_ERROR_EVENT, (error) => {
+      logger.error(`Worker received a connection error event, terminating the process caused by: ${error}`);
+      throw new ProviderError(ProviderFailure.ConnectionFailure, error);
+    });
     await worker.start();
     return worker;
   }
--- a/packages/cardano-services/test/TxWorker/utils.test.ts
+++ b/packages/cardano-services/test/TxWorker/utils.test.ts
@@ -1,0 +1,177 @@
+/* eslint-disable max-len */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/no-empty-function */
+import { Connection } from '@cardano-ogmios/client';
+import { Ogmios } from '@cardano-sdk/ogmios';
+import { RABBITMQ_URL_DEFAULT, createDnsResolver, getOgmiosTxSubmitProvider } from '../../src';
+import {
+  RunningTxSubmitWorker,
+  getRunningTxSubmitWorker,
+  startTxSubmitWorkerWithDiscovery
+} from '../../src/TxWorker/utils';
+import { SrvRecord } from 'dns';
+import { TxSubmitProvider } from '@cardano-sdk/core';
+import { createLogger } from 'bunyan';
+import { createMockOgmiosServer } from '../../../ogmios/test/mocks/mockOgmiosServer';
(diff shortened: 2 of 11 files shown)


packages/cardano-services/src/TxWorker/utils.ts
```diff
@@ -0,0 +1,79 @@
...
+        return () =>
+          worker.start().catch(async (error) => {
+            if (error.innerError?.reason === ProviderFailure.ConnectionFailure) {
+              const resolvedRecord = await dnsResolver(options.rabbitmqSrvServiceName!);
+              logger.info(
+                `TxSubmit worker: DNS resolution for RabbitMQ service, resolved with record: ${JSON.stringify(
+                  resolvedRecord
+                )}`
+              );
+              worker = new TxSubmitWorker(
```
IvayloAndonov, line 47: `worker.close()` is already executed right before this error is propagated

packages/cardano-services/src/TxWorker/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+import { CONNECTION_ERROR_EVENT, TxSubmitWorker } from '@cardano-sdk/rabbitmq';
+import { CommonOptionDescriptions } from '../ProgramsCommon';
+import { DnsResolver, MissingProgramOption, ServiceNames, srvRecordToRabbitmqURL } from '../Program';
+import { Logger } from 'ts-log';
+import { TxSubmitProvider } from '@cardano-sdk/core';
+import { TxWorkerOptions } from './loadTxWorker';
+
+type WorkerFactory = () => Promise<TxSubmitWorker>;
+
+export const createWorkerFactoryWithDiscovery =
```
rhyslbw, line 13: Please add docs to the exported functions in this file

packages/rabbitmq/test/TxSubmitWorker.test.ts
```diff
@@ -80,6 +81,24 @@ describe('TxSubmitWorker', () => {
     await expect(worker.start()).rejects.toBeInstanceOf(ProviderError);
   });
 
+  it('emits event if unable to connect to the RabbitMQ broker', async () => {
+    mock = createMockOgmiosServer({
```
rhyslbw, line 85:
```suggestion
    expect.assertions(1);
    mock = createMockOgmiosServer({
```

packages/cardano-services/test/TxWorker/utils.test.ts
```diff
@@ -0,0 +1,131 @@
...
+
+      const runnableWorker = await startTxSubmitWorkerWithDiscovery(workerFactory);
+      await runnableWorker.stop();
+
+      expect(start).toBeCalledTimes(1);
+      expect(stop).toBeCalledTimes(1);
+      expect(workerFactory).toBeCalledTimes(1);
+    });
+
+    it('start the worker and then fail the listener for connection error events, should create and start a new instance of a running worker', async () => {
```
rhyslbw:
```suggestion
    it('should create and start a new instance of TxWorker when a broker connection error event is received', async () => {
```
I'm a little unsure about how the test works though given there's no use of `CONNECTION_ERROR_EVENT` within the assertion :thinking:

packages/cardano-services/test/TxWorker/utils.test.ts
```diff
@@ -0,0 +1,131 @@
...
+        serviceDiscoveryTimeout: 1000
+      });
+
+      await expect(dnsResolverMock).toBeCalledTimes(1);
+      expect(txSubmitWorker.getStatus()).toEqual('connected');
+    });
+  });
+
+  describe('startTxSubmitWorkerWithDiscovery', () => {
+    it('starts and stops the worker', async () => {
```
rhyslbw, line 102: This is a bit misleading, as the test reads "startTxSubmitWorkerWithDiscovery starts and stops the worker". 
```suggestion
    it('returns a started worker, which can then be stopped', async () => {
```

rhyslbw (changes requested):

packages/cardano-services/test/TxWorker/utils.test.ts
IvayloAndonov: We assign `jest.fn()` func to the event listener [here](https://github.com/input-output-hk/cardano-js-sdk/blob/848dc62bb5432762b5939454f75fe62db33aa7e5/packages/cardano-services/test/TxWorker/utils.test.ts#L127) and after that manually invoke it [here](https://github.com/input-output-hk/cardano-js-sdk/blob/848dc62bb5432762b5939454f75fe62db33aa7e5/packages/cardano-services/test/TxWorker/utils.test.ts#L141) which simulates that the [event is emitted](https://github.com/input-output-hk/cardano-js-sdk/blob/39d22556809a465c07e6615c628707ff7ee19c29/packages/cardano-services/src/TxWorker/utils.ts#L38)

rhyslbw: How about renaming `failListener` as `failWithEvent(CONNECTION_ERROR_EVENT)`, even if it's just to better communicate this simulation?

packages/rabbitmq/src/TxSubmitWorker.ts
```diff
@@ -109,13 +118,18 @@ export class TxSubmitWorker {
...
    */
-  private async errorHandler(isAsync: boolean, err: unknown) {
+  private async errorHandler(isAsync: boolean, err?: any) {
     if (err) {
       this.logError(err, isAsync);
       this.#status = 'error';
       await this.stop();
+
+      // Emit event due to connection error in order to notify the worker abstraction listener
+      if (isConnectionError(err)) {
```
iccicci, line 130: I'm not so sure about this check.
We are not sure that `isConnectionError` includes all network related `Error`s that `amqplib` can throw and leave us prone to changes in the `amqplib` itself.
More than this `TxSubmitWorker.errorHandler` closes the worker regardless from the the error being a connection related one or not.
For the same reason I would have used the standard `'error'` event name.

iccicci (dismissed):
LGTM

packages/rabbitmq/src/TxSubmitWorker.ts
IvayloAndonov: The thing here is that since we reuse [errorHander](https://github.com/input-output-hk/cardano-js-sdk/blob/39d22556809a465c07e6615c628707ff7ee19c29/packages/rabbitmq/src/TxSubmitWorker.ts#L116-L133) for both on startup and runtime error handling, I noticed that there could be passed as an `error` a `ProviderError` one as well, raised by [provider](https://github.com/input-output-hk/cardano-js-sdk/blob/39d22556809a465c07e6615c628707ff7ee19c29/packages/rabbitmq/src/TxSubmitWorker.ts#L165) not a healthy response. 
So my question here is requirements oriented, do we want to distinguish the errors (connection/non-connection) or to emit all types of errors from the worker error handler? 🤔

iccicci: Since the worker is closed anyway I would restart it anyway as well, otherwise we could have situations where the worker is stopped and not restarted: we will have in our hands a worker which is not running and which nobody will never restart.

IvayloAndonov: RIght, in that case where we always stop the worker caused by any type of error and put it into `idle` state, we should then handle the restarts via the worker abstraction with DNS resolution (regardless `ProviderError` error coming from Ogmios provider or errors coming from the `rabbitmq` broker ). Without service discovery - we entrust the restarts to nix/infrastructure we have.

packages/cardano-services/test/TxWorker/utils.test.ts
rhyslbw: Let's go with this instead, since the argument isn't used, so just creates misdirection

``` ts
let simulateConnectionError: any = jest.fn();

on(_: string, listener: any) {
          simulateConnectionError = listener;
 },

simulateConnectionError();
```

packages/cardano-services/src/TxWorker/utils.ts
IvayloAndonov: https://github.com/input-output-hk/cardano-js-sdk/pull/344/commits/89a1a0d3cf617df8746f6493e70edfa219226835

packages/rabbitmq/test/TxSubmitWorker.test.ts
IvayloAndonov: https://github.com/input-output-hk/cardano-js-sdk/pull/344/commits/89a1a0d3cf617df8746f6493e70edfa219226835

packages/cardano-services/test/TxWorker/utils.test.ts
IvayloAndonov: https://github.com/input-output-hk/cardano-js-sdk/pull/344/commits/89a1a0d3cf617df8746f6493e70edfa219226835

packages/cardano-services/test/TxWorker/utils.test.ts
IvayloAndonov: https://github.com/input-output-hk/cardano-js-sdk/pull/344/commits/89a1a0d3cf617df8746f6493e70edfa219226835

packages/cardano-services/src/TxWorker/utils.ts
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Create and start a new worker instance with registered listener for connection error events
+ *
+ * @param {WorkerFactory} workerFactory creates a worker instance with service discovery
+ * @returns {TxSubmitWorker} TxSubmitWorker instance
+ */
+export const createAndStartNewWorker = async (workerFactory: WorkerFactory) => {
+  const newWorker = await workerFactory();
+  attachOnConnectionErrorHandler(newWorker, workerFactory);
+  await newWorker.start();
```
mkazlauskas, line 46: This can throw. We might want to apply exponential backoff retry strategy to re-resolve and recreate the worker when it happens

mkazlauskas (changes requested):

packages/cardano-services/src/TxWorker/utils.ts
mkazlauskas: Depends on error. If it's a connection error then I think we want to wait for a second and re-resolve.

IvayloAndonov: in addition: if `start()` throw caused by connection error - the worker will emit a new event and the abstraction will try to re-resolve and [create new instance](https://github.com/input-output-hk/cardano-js-sdk/blob/89a1a0d3cf617df8746f6493e70edfa219226835/packages/cardano-services/src/TxWorker/utils.ts#L73)
Could you check our discussion [here](https://github.com/input-output-hk/cardano-js-sdk/pull/344#discussion_r926754387) and say your opinion about the emitting events for all kind of errors, I think it makes sense

mkazlauskas: It will emit the event, but it will also throw an error which I believe is an unhandled promise rejection right now.

> and say your opinion about the emitting events for all kind of errors instead of only connection-related ones

I think it should just throw (to terminate the process) on any other errors and rely on external infrastructure to restart. If there's some unexpected error state (e.g. faulty/destroyed txSubmitProvider) re-resolving might not help. Running forever in a faulty state would be pretty bad.

IvayloAndonov: Based on our brainstorming at today's daily, the error handling refactoring of `TxSubmitWorker` is introduced within https://github.com/input-output-hk/cardano-js-sdk/pull/344/commits/5d02656de2f4649f9528084a9dfcbe2b49990a7d , please review the changes when you have time @mkazlauskas  @iccicci  :)

packages/cardano-services/src/TxWorker/utils.ts
```diff
@@ -0,0 +1,100 @@
...
+import { Logger } from 'ts-log';
+import { TxSubmitProvider } from '@cardano-sdk/core';
+import { TxWorkerOptions } from './loadTxWorker';
+
+type WorkerFactory = () => Promise<TxSubmitWorker>;
+
+/**
+ * Create a worker factory with service discovery
+ *
+ * @param {DnsResolver} dnsResolver used for dns resoluton
```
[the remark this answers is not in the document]
IvayloAndonov: https://github.com/input-output-hk/cardano-js-sdk/pull/344/commits/b52a85413514d6a614a1c28eb7a6bb204581c7d7

packages/cardano-services/src/TxWorker/utils.ts
```diff
@@ -0,0 +1,100 @@
...
+ * @param {WorkerFactory} factory creates a worker instance with service discovery
+ */
+export const attachOnConnectionErrorHandler = (worker: TxSubmitWorker, factory: WorkerFactory) => {
+  worker.on(CONNECTION_ERROR_EVENT, () => createAndStartNewWorker(factory));
+};
+
+export type RunningTxSubmitWorker = Pick<TxSubmitWorker, 'stop' | 'getStatus'>;
+
+/**
+ * An abstraction which srarts and manages restarts of the worker instance with service discovery
```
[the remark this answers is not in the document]
IvayloAndonov: https://github.com/input-output-hk/cardano-js-sdk/pull/344/commits/b52a85413514d6a614a1c28eb7a6bb204581c7d7

packages/cardano-services/src/TxWorker/utils.ts
```diff
@@ -0,0 +1,100 @@
...
+  let worker: TxSubmitWorker;
+  await createAndStartNewWorker(async () => (worker = await workerFactory()));
+
+  return {
+    getStatus: () => worker.getStatus(),
+    stop: () => worker.stop()
+  };
+};
+
+export const getRunningTxSubmitWorker = async (
```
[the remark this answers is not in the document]
IvayloAndonov: https://github.com/input-output-hk/cardano-js-sdk/pull/344/commits/b52a85413514d6a614a1c28eb7a6bb204581c7d7

packages/rabbitmq/test/TxSubmitWorker.test.ts
```diff
@@ -76,8 +79,13 @@ describe('TxSubmitWorker', () => {
     await listenPromise(mock, port);
 
     worker = new TxSubmitWorker({ rabbitmqUrl: BAD_CONNECTION_URL }, { logger, txSubmitProvider });
+    const emitEventSpy = jest.spyOn(worker, 'emitEvent');
 
-    await expect(worker.start()).rejects.toBeInstanceOf(ProviderError);
+    try {
+      await expect(worker.start()).resolves.toBeUndefined();
```
mkazlauskas, line 85: `expect` in both try and catch is confusing
```suggestion
      await worker.start();
```

mkazlauskas (changes requested):
Looks good, just 1 small change request

packages/rabbitmq/test/TxSubmitWorker.test.ts
IvayloAndonov: https://github.com/input-output-hk/cardano-js-sdk/pull/344/commits/a84f4e1b695d3c14e0231a2170e67e0f65db966b I think now better explains the behaviour, with both resolving and emitting event

mkazlauskas (approved):
Nice work @IvayloAndonov :tada: Will need to rebase on master once @rhyslbw resolves requested changes

packages/cardano-services/src/TxWorker/utils.ts
```diff
@@ -0,0 +1,112 @@
...
+import { Logger } from 'ts-log';
+import { ProviderError, ProviderFailure, TxSubmitProvider } from '@cardano-sdk/core';
+import { TxWorkerOptions } from './loadTxWorker';
+
+type WorkerFactory = () => Promise<TxSubmitWorker>;
+
+/**
+ * Create a worker factory with service discovery
+ *
+ * @param {DnsResolver} dnsResolver used for DNS resoluton
```
rhyslbw, line 16:
```suggestion
 * @param {DnsResolver} dnsResolver used for DNS resolution
```

packages/cardano-services/src/TxWorker/utils.ts
```diff
@@ -0,0 +1,112 @@
...
+  return {
+    getStatus: () => worker.getStatus(),
+    stop: () => worker.stop()
+  };
+};
+
+/**
+ * Create and return a running worker instance with static service config or service discovery
+ *
+ * @param {DnsResolver} dnsResolver used for DNS resoluton
```
rhyslbw, line 82:
```suggestion
 * @param {DnsResolver} dnsResolver used for DNS resolution
```

packages/cardano-services/src/TxWorker/utils.ts
```diff
@@ -0,0 +1,112 @@
...
+ * @param {TxSubmitProvider} txSubmitProvider tx submit provider 'ogmiosTxSubmitProvider'
+ * @param {Logger} logger common logger
+ * @param {TxWorkerOptions} options needed for tx worker initialization
+ * @returns {RunningTxSubmitWorker} RunningTxSubmitWorker instance
+ * @throws {MissingProgramOption} error if neither url nor service name is provided
```
rhyslbw, line 87:
```suggestion
 * @throws {MissingProgramOption} error if neither URL nor service name is provided
```

rhyslbw (changes requested):

rhyslbw (approved):
Nice work @IvayloAndonov

iccicci (dismissed):
LGTM

IvayloAndonov:
Squashed and rebased on top of master

rhyslbw (approved):
:tada:

PR #345, opened 2022-07-19
fix(blockfrost): handle arrays in image and description fields

jvieiro:
# Context

When using Blockfrost provider, NFT gets rejected with array in image field when parsing a transaction and Array in description field shown as undefined.

# Proposed Solution

Create AssetInfo.TokenMetadata only from response.metadata, since onchain_metadata refers to cip 25.

### DIFF
--- a/packages/blockfrost/src/blockfrostAssetProvider.ts
+++ b/packages/blockfrost/src/blockfrostAssetProvider.ts
@@ -6,22 +6,14 @@ import { BlockFrostAPI, Responses } from '@blockfrost/blockfrost-js';
 import { blockfrostMetadataToTxMetadata, fetchSequentially, toProviderError } from './util';
 import omit from 'lodash/omit';
 
-const mapMetadata = (
-  onChain: Responses['asset']['onchain_metadata'],
-  offChain: Responses['asset']['metadata']
-): Asset.TokenMetadata | null => {
-  const metadata = { ...onChain, ...offChain };
+const mapMetadata = (offChain: Responses['asset']['metadata']): Asset.TokenMetadata | null => {
+  const metadata = { ...offChain };
   if (Object.values(metadata).every((value) => value === undefined || value === null)) return null;
   return {
-    ...replaceNullsWithUndefineds(omit(metadata, ['logo', 'image'])),
+    ...replaceNullsWithUndefineds(omit(metadata, ['logo'])),
     desc: metadata.description,
     // The other type option is any[] - not sure what it means, omitting if no string.
-    icon:
-      typeof metadata.logo === 'string'
-        ? metadata.logo
-        : typeof metadata.image === 'string'
-        ? metadata.image
-        : undefined
+    icon: typeof metadata.logo === 'string' ? metadata.logo : undefined
   };
 };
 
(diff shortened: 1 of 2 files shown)


packages/core/src/Asset/types/TokenMetadata.ts
```diff
@@ -30,7 +30,7 @@ export interface TokenMetadata {
    * When implementing the parser, it is recommended to also check 'image' and 'logo'
    * properties for backwards compatibility.
    */
-  icon?: string;
+  icon?: string | string[];
```
mkazlauskas, line 33: Is this valid? Can't see anything about arrays in the [CIP](https://github.com/cardano-foundation/CIPs/pull/137/files#diff-f478f61a4b67bc7fee1d41bbb427028a7615beaacf726efc9289104e64f26ffaR64)

jvieiro: Thanks, I've changed it

packages/blockfrost/test/blockfrostAssetProvider.test.ts
```diff
@@ -205,5 +205,56 @@ describe('blockfrostAssetProvider', () => {
...
+
+    test('description and image as array', async () => {
+      BlockFrostAPI.prototype.assetsById = jest.fn().mockResolvedValue({
+        asset: '8153f8be9f05b2f32b481bbf7af877f592160b39e87f5f55c8ab035f4e46543031',
+        asset_name: '4e46543031',
+        fingerprint: 'asset1h2z87pq2tr4ksxl5nkzd2ltrrtd6vvmf5nnn46',
+        initial_mint_tx_hash: 'cb2cf241d03f0f6c523a2064b51de783db590f2871da24ce89698db2bef5fb39',
+        metadata: null,
+        mint_or_burn_count: 1,
+        onchain_metadata: {
```
mkazlauskas, line 217: I'm wondering if this is actually the cip25 NFT metadata. If it is then we shouldn't use it as cip38 token metadata :thinking: maybe let's add some e2e test(s) (in `packages/e2e/test/blockfrost`) to fetch metadata of some token and some nft

mkazlauskas: So I think the underlying bug is that we're trying to map cip38 (AssetInfo.tokenMetadata) into cip25 (AssetInfo.nftMetadata). We shouldn't be doing that

jvieiro: So, `response.onchain_metadata `shouldn't be used to create `AssetInfo.tokenMetadata,` just `response.metadata`?

packages/blockfrost/src/blockfrostAssetProvider.ts
```diff
@@ -6,6 +6,13 @@ import { BlockFrostAPI, Responses } from '@blockfrost/blockfrost-js';
 import { blockfrostMetadataToTxMetadata, fetchSequentially, toProviderError } from './util';
 import omit from 'lodash/omit';
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const getIcon = (logo?: string | null, image?: string | any[]) => {
+  if (typeof logo === 'string') return logo;
```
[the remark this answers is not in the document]
jvieiro: It comes from https://github.com/blockfrost/blockfrost-js/blob/master/packages/blockfrost-js/src/types/OpenApi.ts#L4959

packages/blockfrost/test/blockfrostAssetProvider.test.ts
mkazlauskas: Yes, `response.onchain_metadata` is `AssetInfo.nftMetadata`. We already have a different implementation to get `nftMetadata`, so lets just keep that for now to not break anything.

mkazlauskas (dismissed):
LGTM. Will need to squash into 1 commit after all approvals

iadmytro (approved):

jvieiro:
> LGTM. Will need to squash into 1 commit after all approvals

Squashed!

PR #348, opened 2022-07-22
docs(cardano-services): add basics and a couple of examples

rhyslbw:
# Context
We're missing basic docs to assist with deploying the services. Prior to a docs site, we just need the basics in the package README.

# Proposed Solution
- Introduces the package, listing some notable features.
- Examples include:
  - using direct or queue-based tx submission
  - static or discovered config

### DIFF
--- a/packages/cardano-services/README.md
+++ b/packages/cardano-services/README.md
@@ -1,7 +1,103 @@
 # Cardano JS SDK | Cardano GraphQL Services
+Libraries and program entrypoints for services to facilitate remote data and submit access using 
+[Provider] interfaces over HTTP, with _optional_ queue-based transaction submission; The 
+[TxSubmitHttpService] can be configured to submit directly via [Ogmios], or via a [RabbitMQ] broker,
+with one or more workers handling submission and response job creation. Data is sourced from 
+[Cardano DB Sync], the local [Cardano Node] via [Ogmios] Local State Queries, genesis files, and 
+remote sources.
+
+## Features
+- [CLI] or run scripts for the [HTTP server](./src/run.ts) and [worker](./src/startWorker.ts) with 
+configuration via environment variables and _optional_ loading of secrets from disk.
+- Service port discovery via DNS resolution, or static configuration.
+- Fault-tolerant transaction submission via persistent queue, or direct submission.
+- _Optional_ [Prometheus] metrics available at `/metrics`
+- Data sourced from Cardano DB Sync PostgreSQL, Local State Queries, genesis files, and remote 
+  sources.
+
+## Services
+
+The services require instances of [Cardano Node] and [Ogmios] as a minimum, with 
+[Cardano DB Sync] and [RabbitMQ] dependent on the run command. Please refer to
+[docker-compose.json](./docker-compose.yml) for the current supported version of each service 
+dependency.
+
+### HTTP Server
+The HTTP server can be started with one or more provider modules by name, segmented by URL path. 
+Run the [CLI] with `start-server --help` to see the full list of options, or inspect the 
+environment variables within [./src/run.ts](./src/run.ts). 
+
+### Worker
+A worker must be started when opting for queue-based transaction submission.
+Run the [CLI] with `start-worker --help` to see the full list of options, or inspect the
+environment variables within [./src/startWorker.ts](./src/startWorker.ts).
+
+
+## Examples
+
+_The following examples require the [install and build] steps to be completed._
+
+#### All Providers | Static Service Config | Direct Tx Submission
+
+- The server will expose all [Provider] HTTP services
+- Transactions will be submitted directly via [Ogmios], running at `ws://localhost:1338`.
+- Connects to [PostgreSQL] service running at `localhost:5432`
+- HTTP API exposed using a custom API URL
+
+``` console
+./dist/cjs/cli.js \
+  start-server \
+    --api-url http://localhost:6000 \
+    --cardano-node-config-path ./config/cardano-node/config.json \
+    --ogmios-url  ws://localhost:1338 \
+    --db-connection-string postgresql://somePgUser:somePassword@localhost:5432/someDbName \
+    asset,chain-history,stake-pool,tx-submit,network-info,utxo
+```
+
+#### All Providers | Service Discovery | Queued Tx Submission | Metrics
+
+- The server will expose all [Provider] HTTP services
+- Transactions will be queued by the HTTP service, with the worker completing the
+  submission. The HTTP service receives the submission result via a dedicated channel to
+  complete the HTTP request.
+- Ports for [Ogmios], [PostgreSQL], and [RabbitMQ], discovered using DNS resolution.
+- HTTP API exposed using a custom API URL
+- Prometheus metrics exporter enabled at http://localhost:6000/metrics
+
+``` console
+./dist/cjs/cli.js \
+  start-server \
+    --api-url http://localhost:6000 \
+    --cardano-node-config-path ./config/cardano-node/config.json \
+    --enable-metrics true \
+    --ogmios-srv-service-name  some-domain-for-ogmios \
+    --postgres-db someDbName \
+    --postgres-password somePassword \
+    --postgres-srv-service-name \
+    --postgres-user somePgUser \
+    --rabbitmq-srv-service-name some-domain-for-rabbitmq \
+    --use-queue \
+    asset,chain-history,stake-pool,tx-submit,network-info,utxo
+```
+``` console
+./dist/cjs/cli.js \
+  start-worker \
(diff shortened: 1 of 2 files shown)


AngelCastilloB:
I think we should also mention that you need to have dbSync running; In general, maybe we should have a section specifying which external services are expected to be running and why/how they are being used by our services (DBSycn, Cardano Node, Ogmios, RabbitMQ).

Also, it would be great to have a section describing all the available services and what they do. If you don't know much about this project and just open the repository is not very clear what it does or what it can be used for (I mean the Cardano JS SDK in general).

IvayloAndonov:
Might be a usage example with the other entrypoint - run scripts for the HTTP server would be useful as well? (`run:http-server`)

packages/cardano-services/README.md
```diff
@@ -1,7 +1,92 @@
...
+- Transactions will be submitted directly via [Ogmios], running at `ws://localhost:1338`.
+- Connects to [PostgreSQL] service running at `localhost:5432`
+- HTTP API exposed using a custom API URL
+
+``` console
+./dist/cli.js \
+  start-server \
+    --api-url http://localhost:6000 \
+    --cardano-node-config-path ./config/cardano-node/config.json \
+    --ogmios-url  ws://localhost:1338 \
```
IvayloAndonov, line 43: Shouldn't we use in the sample port `1337` as the `docker-compose` says [here](https://github.com/input-output-hk/cardano-js-sdk/blob/b64fd99daffa96f14873992379aba0fa77b2e272/packages/cardano-services/docker-compose.yml#L39-L40) ?

packages/cardano-services/README.md
```diff
@@ -1,7 +1,92 @@
...
+    --db-connection-string postgresql://somePgUser:somePassword@localhost:5432/someDbName \
+    asset,chain-history,stake-pool,tx-submit,network-info,utxo
+```
```
YorikSar, line 46: Note that CLI args are not very automation friendly. I found that `cli.js` doesn't handle environment variables quite as good as `run.js`/`startWorker.js` do.

packages/cardano-services/README.md
```diff
@@ -1,7 +1,92 @@
...
+
+#### All Providers | Static Service Config | Direct Tx Submission
+
+- The server will expose all [Provider] HTTP services
+- Transactions will be submitted directly via [Ogmios], running at `ws://localhost:1338`.
+- Connects to [PostgreSQL] service running at `localhost:5432`
+- HTTP API exposed using a custom API URL
+
+``` console
+./dist/cli.js \
```
YorikSar, line 39: These end up in `./dist/cjs` folder. By the way, in my build these files don't have execute permissions set.

packages/cardano-services/README.md
```diff
@@ -1,7 +1,92 @@
...
+configuration via environment variables and _optional_ loading of secrets from disk.
+- Service port discovery via DNS resolution, or static configuration.
+- Fault-tolerant transaction submission via persistent queue, or direct submission.
+- _Optional_ [Prometheus] metrics available at `/metrics`
+
+## Services
+### HTTP Server
+The HTTP server can be started with one or more provider modules by name, segmented by URL path. 
+Run the [CLI] with `start-server --help` to see the full list of options, or inspect the 
+environment variables within [./src/run.ts](./src/run.ts).
```
YorikSar, line 18: That file doesn't have any comments about what does each environment variable actually mean. I think it would be very useful to have them listed here with descriptions and default values.

packages/cardano-services/README.md
```diff
@@ -1,7 +1,92 @@
...
+    --rabbitmq-srv-service-name some-domain-for-rabbitmq \
+    --use-queue \
+    asset,chain-history,stake-pool,tx-submit,network-info,utxo
+```
+``` console
+./dist/cli.js \
+  start-worker \
+    --ogmios-srv-service-name  some-domain-for-ogmios \
+    --rabbitmq-srv-service-name some-domain-for-rabbitmq
+```
```
YorikSar, line 79: Some visual representation of interaction between all components for both deployment profiles would go a long way. You can use https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/ to do this on GitHub.

YorikSar (changes requested):
Thanks a lot for this doc, it clarifies a lot of things. I think it would be enough for me if I was just starting. Couple of possible improvements inline.

packages/cardano-services/README.md
rhyslbw: That would be confusing as it would then not be required given `ws://localhost:1337` is the default

packages/cardano-services/README.md
rhyslbw: > Note that CLI args are not very automation friendly

Right, the idea of these docs was to demonstrate the program concept, which the CLI offers a slightly cleaner presentation IMO, and there's also `--help` to call. It's a fine balance between too little documentation and too much, the latter of which becomes a maintenance nightmare without some type of docs generation that can be automated :thinking:. That's why we've tended to just point to the Docker Compose file, integration tests, or instruct to run the CLI help, but it's time to meet in the middle with the API stability close.

packages/cardano-services/README.md
rhyslbw: Fixed the invalid path in 3f2085911a5932c3484183befdb0da7d9354e98f, and executable permissions in 1c1eeb2240c98bf53ce5007406e094f82f2a255d

packages/cardano-services/README.md
rhyslbw: Rendering in MD would be nice, but for now, I can more easily add some links to our hosted docs, which are automatically generated and deployed: https://input-output-hk.github.io/cardano-js-sdk/interfaces/_cardano_sdk_cardano_services.HttpServerOptions.html.

packages/cardano-services/README.md
rhyslbw: I've purposely omitted architecture diagrams, as it's being worked on in the lower application context, which we intend to hoist once complete.

packages/cardano-services/README.md
YorikSar: I don't think autogenerated docs can get to the level of usability as hand-crafted ones. Currently they still don't include any description about what each option mean, how it's translated to environment variables and what would be a sensible way to set them.

packages/cardano-services/README.md
YorikSar: I think CLI should support environment variables and we should document both. Then user (i.e. me) don't have to guess which .js file to run and which env variables to pass if I want to rely on them.

packages/cardano-services/README.md
YorikSar: Well, this README is targeted at GitHub readers, I assume. It can link to better more in-depth architecture description, but it could still contain some pictures to get a quick understanding of what's going on.

packages/cardano-services/README.md
rhyslbw: We'll be consolidating the two entrypoints in the next sprint. There was a technical limitation with the CLI library, but that's been removed with the introduction of a new feature :tada:

packages/cardano-services/README.md
rhyslbw: This will be covered by the task mentioned here: https://github.com/input-output-hk/cardano-js-sdk/pull/348#discussion_r936281539

packages/cardano-services/README.md
rhyslbw: We'll consider this in the next round of improvements once the other documentation has been considered for adoption.

rhyslbw:
> I think we should also mention that you need to have dbSync running; In general, maybe we should have a section specifying which external services are expected to be running and why/how they are being used by our services (DBSycn, Cardano Node, Ogmios, RabbitMQ).

@AngelCastilloB I've pushed another commit to discuss the service dependencies: https://github.com/input-output-hk/cardano-js-sdk/pull/348/commits/6d9cae7c2b69097fc91942a6d7012b9d6aa8fcda. Happy with this? I'd like to keep this brief until we can introduce the more sophisticated architecture docs.

AngelCastilloB:
> > I think we should also mention that you need to have dbSync running; In general, maybe we should have a section specifying which external services are expected to be running and why/how they are being used by our services (DBSycn, Cardano Node, Ogmios, RabbitMQ).
> 
> @AngelCastilloB I've pushed another commit to discuss the service dependencies: [6d9cae7](https://github.com/input-output-hk/cardano-js-sdk/commit/6d9cae7c2b69097fc91942a6d7012b9d6aa8fcda). Happy with this? I'd like to keep this brief until we can introduce the more sophisticated architecture docs.

Looks great @rhyslbw

AngelCastilloB (approved):

PR #349, opened 2022-07-22
Test/fix mac support on download script

AngelCastilloB:
# Context

The local network setup is failing to start on Mac OS. 

# Proposed Solution
Fix the setup scripts to accommodate Mac OS differences.

The setup scripts have been updated to download the macOS binaries for the cardano-cli and update the alonzo genesis file hash on the node config.json

### DIFF
--- a/packages/e2e/local-network/scripts/install.sh
+++ b/packages/e2e/local-network/scripts/install.sh
@@ -18,7 +18,7 @@ trap clean EXIT
 
 VERSION="1.35.0"
 LINUX_BUILD="16159630"
-MACOS_BUILD="16159630"
+MACOS_BUILD="17006848"
 
 rm -rf bin
 mkdir -p bin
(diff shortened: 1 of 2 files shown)


ssong-iohk:
Fix works so I can run local-network with/without dockerisation in MacOS successfully. 
Thank you so much!

packages/e2e/local-network/scripts/install.sh
```diff
@@ -18,7 +18,7 @@ trap clean EXIT
 
 VERSION="1.35.0"
 LINUX_BUILD="16159630"
-MACOS_BUILD="16159630"
```
rhyslbw, line 21: Was this a copy/paste error, or an issue with the build?

rhyslbw (changes requested):
Great to see this is helping unblock @ssong-iohk 

These should actually be two distinct commits though, as one is correcting the macOS build link, and the other adding Alonzo support.
image

packages/e2e/local-network/scripts/install.sh
AngelCastilloB: copy/past error

mkazlauskas (dismissed):
This looks good to me. Is there a reason we're not merging it?

mkazlauskas (approved):

rhyslbw (approved):

PR #351, opened 2022-07-25
Feature/adp 1208 resubmit transactions on rollback

mirceahasegan:
# Context

Retry rollbacked transactions.

# Proposed Solution
1. Emit rollback transactions along with the historical transactions
2. Store confirmed transactions that are not yet stable
3. Retry confirmed transactions that were rolledback

### DIFF
--- a/packages/wallet/src/SingleAddressWallet.ts
+++ b/packages/wallet/src/SingleAddressWallet.ts
@@ -53,14 +53,19 @@ import {
 import { BehaviorObservable, TrackerSubject } from '@cardano-sdk/util-rxjs';
 import {
   BehaviorSubject,
+  EMPTY,
   Subject,
+  Subscription,
+  catchError,
   combineLatest,
   concat,
   distinctUntilChanged,
   filter,
   firstValueFrom,
+  from,
   lastValueFrom,
   map,
+  mergeMap,
   take,
   tap
 } from 'rxjs';
@@ -73,6 +78,7 @@ import { TrackedUtxoProvider } from './services/ProviderTracker/TrackedUtxoProvi
 import { TxInternals, createTransactionInternals, ensureValidityInterval } from './Transaction';
 import { WalletStores, createInMemoryWalletStores } from './persistence';
 import { cip30signData } from './KeyManagement/cip8';
+import { createTransactionReemitter } from './services/TransactionReemitter';
 import isEqual from 'lodash/isEqual';
 
 export interface SingleAddressWalletProps {
@@ -105,6 +111,7 @@ export class SingleAddressWallet implements ObservableWallet {
     pending$: new Subject<Cardano.NewTxAlonzo>(),
     submitting$: new Subject<Cardano.NewTxAlonzo>()
   };
+  #resubmitSubscription: Subscription;
 
   readonly keyAgent: AsyncKeyAgent;
   readonly currentEpoch$: TrackerSubject<EpochInfo>;
@@ -288,6 +295,23 @@ export class SingleAddressWallet implements ObservableWallet {
       tip$: this.tip$,
       transactionsHistoryStore: stores.transactions
     });
+
+    this.#resubmitSubscription = createTransactionReemitter({
+      confirmed$: this.transactions.outgoing.confirmed$,
+      rollback$: this.transactions.rollback$,
+      store: stores.volatileTransactions,
+      submitting$: this.transactions.outgoing.submitting$,
+      tipSlot$: this.tip$.pipe(map((tip) => tip.slot))
+    })
+      .pipe(
+        mergeMap((transaction) => from(this.submitTx(transaction))),
+        catchError((err) => {
+          this.#logger.error('Failed to resubmit transaction', err);
+          return EMPTY;
+        })
+      )
+      .subscribe();
+
     this.utxo = createUtxoTracker({
       addresses$,
       retryBackoffConfig,
@@ -408,6 +432,7 @@ export class SingleAddressWallet implements ObservableWallet {
     this.#newTransactions.failedToSubmit$.complete();
     this.#newTransactions.pending$.complete();
     this.#newTransactions.submitting$.complete();
+    this.#resubmitSubscription.unsubscribe();
   }
 
   #prepareTx(props: InitializeTxProps) {
--- a/packages/wallet/src/persistence/pouchdbStores/pouchdbWalletStores.ts
+++ b/packages/wallet/src/persistence/pouchdbStores/pouchdbWalletStores.ts
@@ -10,6 +10,7 @@ import {
 import { EMPTY, combineLatest, map } from 'rxjs';
 import { GroupedAddress } from '../../KeyManagement';
 import { Logger } from 'ts-log';
+import { NewTxAlonzoWithSlot } from '../../services';
 import { PouchdbCollectionStore } from './PouchdbCollectionStore';
 import { PouchdbDocumentStore } from './PouchdbDocumentStore';
 import { PouchdbKeyValueStore } from './PouchdbKeyValueStore';
@@ -25,6 +26,7 @@ export class PouchdbTimeSettingsStore extends PouchdbDocumentStore<TimeSettings[
 export class PouchdbAssetsStore extends PouchdbDocumentStore<Assets> {}
 export class PouchdbAddressesStore extends PouchdbDocumentStore<GroupedAddress[]> {}
 export class PouchdbInFlightTransactionsStore extends PouchdbDocumentStore<Cardano.NewTxAlonzo[]> {}
+export class PouchdbVolatileTransactionsStore extends PouchdbDocumentStore<NewTxAlonzoWithSlot[]> {}
 
 export class PouchdbTransactionsStore extends PouchdbCollectionStore<Cardano.TxAlonzo> {}
 export class PouchdbUtxoStore extends PouchdbCollectionStore<Cardano.Utxo> {}
@@ -56,7 +58,6 @@ export const createPouchdbWalletStores = (
         return combineLatest([
           destroyDocumentsDb,
           this.transactions.destroy(),
-          this.inFlightTransactions.destroy(),
           this.utxo.destroy(),
           this.unspendableUtxo.destroy(),
           this.rewardsHistory.destroy(),
@@ -90,6 +91,7 @@ export const createPouchdbWalletStores = (
       logger
     ),
     unspendableUtxo: new PouchdbUtxoStore({ dbName: `${baseDbName}UnspendableUtxo` }, logger),
-    utxo: new PouchdbUtxoStore({ dbName: `${baseDbName}Utxo` }, logger)
+    utxo: new PouchdbUtxoStore({ dbName: `${baseDbName}Utxo` }, logger),
+    volatileTransactions: new PouchdbVolatileTransactionsStore(docsDbName, 'volatileTransactions', logger)
   };
 };
--- a/packages/wallet/src/persistence/types.ts
+++ b/packages/wallet/src/persistence/types.ts
@@ -8,6 +8,7 @@ import {
   TimeSettings
 } from '@cardano-sdk/core';
 import { GroupedAddress } from '../KeyManagement';
+import { NewTxAlonzoWithSlot } from '../services';
 import { Observable } from 'rxjs';
 
 export interface Destroyable {
@@ -81,6 +82,7 @@ export interface WalletStores extends Destroyable {
   unspendableUtxo: CollectionStore<Cardano.Utxo>;
   transactions: OrderedCollectionStore<Cardano.TxAlonzo>;
   inFlightTransactions: DocumentStore<Cardano.NewTxAlonzo[]>;
+  volatileTransactions: DocumentStore<NewTxAlonzoWithSlot[]>;
   rewardsHistory: KeyValueStore<Cardano.RewardAccount, EpochRewards[]>;
   rewardsBalances: KeyValueStore<Cardano.RewardAccount, Cardano.Lovelace>;
   stakePools: KeyValueStore<Cardano.PoolId, Cardano.StakePool>;
--- a/packages/wallet/src/services/TransactionReemitter.ts
+++ b/packages/wallet/src/services/TransactionReemitter.ts
@@ -1,0 +1,113 @@
+import { Cardano } from '@cardano-sdk/core';
+import { Logger, dummyLogger } from 'ts-log';
+import { Observable, filter, from, map, merge, mergeMap, scan, withLatestFrom } from 'rxjs';
+
+import { CustomError } from 'ts-custom-error';
+import { DocumentStore } from '../persistence';
+import { NewTxAlonzoWithSlot } from './types';
+
+export enum TransactionReemitErrorCode {
+  invalidHereafter = 'invalidHereafter',
+  notFound = 'notFound'
+}
+class TransactionReemitError extends CustomError {
+  code: TransactionReemitErrorCode;
+  public constructor(code: TransactionReemitErrorCode, message: string) {
+    super(message);
+    this.code = code;
+  }
+}
+
+interface TransactionReemitterProps {
+  rollback$: Observable<Cardano.TxAlonzo>;
+  confirmed$: Observable<NewTxAlonzoWithSlot>;
+  submitting$: Observable<Cardano.NewTxAlonzo>;
+  store: DocumentStore<NewTxAlonzoWithSlot[]>;
+  tipSlot$: Observable<Cardano.Slot>;
+  stabilityWindowSlotsCount?: number;
+  logger?: Logger;
+}
+
+enum txSource {
+  store,
+  confirmed,
+  submitting
+}
+
+// 3k/f (where k is the security parameter in genesis, and f is the active slot co-efficient parameter
+// in genesis that determines the probability for amount of blocks created in an epoch.)
+const kStabilityWindowSlotsCount = 129_600; // 3k/f on current mainnet
+
+export const createTransactionReemitter = ({
+  rollback$,
+  confirmed$,
+  submitting$,
+  store,
+  tipSlot$,
+  stabilityWindowSlotsCount = kStabilityWindowSlotsCount,
+  logger = dummyLogger
+}: TransactionReemitterProps): Observable<Cardano.NewTxAlonzo> => {
+  const volatileTransactions$ = merge(
+    store.get().pipe(
+      mergeMap((txs) => from(txs)),
+      map((tx) => ({ source: txSource.store, tx }))
+    ),
+    confirmed$.pipe(map((tx) => ({ source: txSource.confirmed, tx }))),
+    submitting$.pipe(map((tx) => ({ source: txSource.submitting, tx: { ...tx, slot: null! } })))
+  ).pipe(
+    scan((volatiles, { tx, source }) => {
+      switch (source) {
+        case txSource.store: {
+          // Do not calculate stability window for old transactions coming from the store
+          volatiles = [...volatiles, tx];
+          break;
+        }
+        case txSource.submitting: {
+          // Transactions in submitting are the ones reemitted. Remove them from volatiles
+          volatiles = volatiles.filter((v) => v.id !== tx.id);
+          store.set(volatiles);
+          break;
+        }
+        case txSource.confirmed: {
+          const oldestAcceptedSlot = tx.slot > stabilityWindowSlotsCount ? tx.slot - stabilityWindowSlotsCount : 0;
+          // Remove transactions considered stable
+          volatiles = [...volatiles.filter(({ slot }) => slot > oldestAcceptedSlot), tx];
+          store.set(volatiles);
+          break;
+        }
+      }
+      return volatiles;
+    }, [] as NewTxAlonzoWithSlot[])
+  );
+
+  return rollback$.pipe(
+    withLatestFrom(tipSlot$),
+    map(([tx, tipSlot]) => {
+      const invalidHereafter = tx.body?.validityInterval?.invalidHereafter;
+      if (invalidHereafter && tipSlot > invalidHereafter) {
+        const err = new TransactionReemitError(
+          TransactionReemitErrorCode.invalidHereafter,
+          `Rolled back transaction with id ${tx.id} is no longer valid`
+        );
+        logger.error(err.message, err.code);
+        return;
+      }
+      return tx;
+    }),
+    filter((tx) => !!tx),
+    withLatestFrom(volatileTransactions$),
+    map(([tx, volatiles]) => {
+      // Get the confirmed NewTxAlonzo transaction to be retried
+      const reemitTx = volatiles.find((txVolatile) => txVolatile.id === tx!.id);
+      if (!reemitTx) {
+        const err = new TransactionReemitError(
+          TransactionReemitErrorCode.notFound,
+          `Could not find confirmed transaction with id ${tx!.id} that was rolled back`
+        );
+        logger.error(err.message, err.code);
+      }
+      return reemitTx!;
+    }),
+    filter((tx) => !!tx)
+  );
+};
--- a/packages/wallet/src/services/TransactionsTracker.ts
+++ b/packages/wallet/src/services/TransactionsTracker.ts
@@ -23,7 +23,7 @@ import {
   takeUntil,
   tap
 } from 'rxjs';
-import { FailedTx, TransactionFailure, TransactionsTracker } from './types';
+import { FailedTx, NewTxAlonzoWithSlot, TransactionFailure, TransactionsTracker } from './types';
 import { RetryBackoffConfig } from 'backoff-rxjs';
 import { Shutdown } from '@cardano-sdk/util';
 import { TrackerSubject } from '@cardano-sdk/util-rxjs';
@@ -47,7 +47,8 @@ export interface TransactionsTrackerProps {
 }
 
 export interface TransactionsTrackerInternals {
-  transactionsSource$?: Observable<Cardano.TxAlonzo[]>;
+  transactionsSource$: Observable<Cardano.TxAlonzo[]>;
+  rollback$: Observable<Cardano.TxAlonzo>;
 }
 
 export const createAddressTransactionsProvider = (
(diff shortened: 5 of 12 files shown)


mirceahasegan:
Unit tests and refactoring will follow

packages/wallet/src/services/TransactionsTracker.ts
```diff
@@ -47,7 +47,7 @@ export interface TransactionsTrackerProps {
 }
 
 export interface TransactionsTrackerInternals {
-  transactionsSource$?: Observable<Cardano.TxAlonzo[]>;
+  transactionsSource$?: Observable<{ transactions: Cardano.TxAlonzo[]; rollback: Cardano.TxAlonzo[] }>;
```
mkazlauskas, line 50: This is ok as it's internal impl details, but maybe separate observables here would be cleaner. What do you think @mirceahasegan ?

```suggestion
  transactionsSource$?: Observable<Cardano.TxAlonzo[]>;
  rollback$?: Observable<Cardano.TxAlonzo>;
```

packages/wallet/src/services/types.ts
```diff
@@ -51,14 +51,17 @@ export interface FailedTx {
   error?: Cardano.TxSubmissionError;
 }
 
+export type ConfirmedNewTxAlonzoWithSlot = Cardano.NewTxAlonzo & { slot: Cardano.PartialBlockHeader['slot'] };
+
 export interface TransactionsTracker {
   readonly history$: Observable<Cardano.TxAlonzo[]>;
+  readonly rollback$: Observable<Cardano.TxAlonzo[]>;
```
mkazlauskas, line 58: I think an array here is confusing. We only need 1 notification per rollback since we're setting up the resubmitter on startup.

```suggestion
  readonly rollback$: Observable<Cardano.TxAlonzo>;
```

packages/wallet/src/persistence/types.ts
```diff
@@ -81,6 +82,8 @@ export interface WalletStores extends Destroyable {
   unspendableUtxo: CollectionStore<Cardano.Utxo>;
   transactions: OrderedCollectionStore<Cardano.TxAlonzo>;
   inFlightTransactions: DocumentStore<Cardano.NewTxAlonzo[]>;
+  // REVIEW: i'm not sure what type of store would be suitable here
+  confirmedTransactions: DocumentStore<ConfirmedNewTxAlonzoWithSlot[]>;
```
mkazlauskas, line 86: Would be great to come up with a better name for this store, to indicate that those are the transactions of not 100% confidence (within rollbackable range)

mkazlauskas (changes requested):

packages/wallet/src/services/RollbackResubmitter.ts
```diff
@@ -0,0 +1,50 @@
...
+  store
+    .get()
+    .pipe(
+      switchMap((transactions) => {
+        // Load stored transactions
+        confirmedTransactions = [...transactions];
+        return confirmed$;
+      })
+    )
+    .subscribe((t) => {
```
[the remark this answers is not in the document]
mkazlauskas: > Then you won't be able to do [this](https://github.com/input-output-hk/cardano-js-sdk/pull/351/files#diff-fa763eb5d53c8a5623e10eae183f3f7a96bfa37d3762cb856fb324b803906a3fR45), but I think it's fine as we will filter them out anyway after stability window. I'd prefer cleaner implementation over slight optimization.

Actually I realized that this would cause a problem with duplicates (resubmitted and then reconfirmed transactions), so you do need to remove them. Let's keep your implementation of maintaining the array, just maybe refactor it a little bit to do side effects with `tap` and keeping stored transactions in sync with the array:

```ts
confirmedTransactions = confirmedTransactions.filter(({ slot }) => slot > oldestAcceptedSlot);
store.set(confirmedTransactions);
```

packages/wallet/src/services/RollbackResubmitter.ts
```diff
@@ -0,0 +1,50 @@
+import { Cardano } from '@cardano-sdk/core';
+import { ConfirmedNewTxAlonzoWithSlot } from './types';
+import { DocumentStore } from '../persistence';
+import { Observable, map, switchMap, tap } from 'rxjs';
+
+// REVIEW: explore how to get the stability window from the network info
```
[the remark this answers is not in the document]
mirceahasegan: Done: [ADP-2077](https://input-output.atlassian.net/browse/ADP-2077)

packages/wallet/src/services/TransactionsTracker.ts
mirceahasegan: An less invasive to this would be to keep the transactionsSource unchanged (as it is now in master), and use pairwise operator to determine which transactions were removed since the last emission.
The only problem would be if the array of transaction arrays gets really big.
What do you think?

packages/wallet/src/services/types.ts
mirceahasegan: I think there could be multiple rolled back transactions detected in the same [tipHeight iteration](https://github.com/input-output-hk/cardano-js-sdk/pull/351/files#diff-b015172e50668407821a584b9a03f1c2c70ccf60bd27d3832335c85360d86d96).
But I can turn it into a stream before processing.

packages/wallet/src/services/TransactionsTracker.ts
mkazlauskas: Good idea, but I'm also fine with returning it from same function that returns `transactionsSource$`.

> The only problem would be if the array of transaction arrays gets really big.

You can still make it efficient because transactions are supposed to be ordered, so you can just start iterating from end. Not concerned about memory consumption :sweat_smile:

packages/wallet/src/services/types.ts
mkazlauskas: Yep there can be, just emit them all one by one then

packages/wallet/src/persistence/types.ts
mirceahasegan: A few options:
`stabilityWindowTransactions`
`revertableTransactions`
`pendingChainConfirmationTransactions`

mkazlauskas: I like `stabilityWindowTransactions`

packages/wallet/src/services/RollbackResubmitter.ts
```diff
@@ -0,0 +1,50 @@
...
+// 3k/f (where k is the security parameter in genesis, and f is the active slot co-efficient parameter
+// in genesis that determines the probability for amount of blocks created in an epoch.)
+const kStabilityWindowSlotsCount = 129_600; // 3k/f
+
+export const createResubmitRollbacks = (
```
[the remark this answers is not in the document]
rhyslbw: `createTransactionReemitter` :sweat_smile: ?

packages/wallet/src/services/TransactionsTracker.ts
mirceahasegan: > 

If we return it from the same function that returns transactionsSource$, it is easier to emit every time a rollback is detected. The comparing approach is useful only if used externally.

mirceahasegan:
Changes made fixups complicated and unreadable, so I force pushed the requested changes.
Still pending:
1. Unit tests for new code
2. Fix some new problems with the existing unit tests

packages/wallet/src/persistence/types.ts
mirceahasegan: Updated in [0286cb7](https://github.com/input-output-hk/cardano-js-sdk/pull/351/commits/0286cb7b5c3844765304acc03f3163345df65fa2)

packages/wallet/src/services/types.ts
mirceahasegan: Updated in [0286cb7](https://github.com/input-output-hk/cardano-js-sdk/pull/351/commits/0286cb7b5c3844765304acc03f3163345df65fa2)

packages/wallet/src/services/TransactionsTracker.ts
mirceahasegan: Updated in [0286cb7](https://github.com/input-output-hk/cardano-js-sdk/pull/351/commits/0286cb7b5c3844765304acc03f3163345df65fa2)

packages/wallet/src/services/RollbackResubmitter.ts
```diff
@@ -0,0 +1,50 @@
...
+      // REVIEW: we could do this check for every tip$ update, but I think it would be enough to do it for every
+      // confirmed transaction
+      const oldestAcceptedSlot = t.slot > stabilityWindowSlotsCount ? t.slot > stabilityWindowSlotsCount : 0;
+      // Remove transactions considered stable
+      confirmedTransactions.filter(({ slot }) => slot > oldestAcceptedSlot);
+      store.set(confirmedTransactions);
+    });
+
+  // REVIEW: throw some error or log something in case there is no entry in confirmed for the rollback?
+  // REVIEW: shutdown method should exist to close the confirmedTransactions$ tracker
```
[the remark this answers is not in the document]
mirceahasegan: Updated as described above. See [0286cb7](https://github.com/input-output-hk/cardano-js-sdk/pull/351/commits/0286cb7b5c3844765304acc03f3163345df65fa2)

packages/wallet/src/services/RollbackResubmitter.ts
```diff
@@ -0,0 +1,50 @@
...
+      confirmedTransactions.push(t);
+      // REVIEW: we could do this check for every tip$ update, but I think it would be enough to do it for every
+      // confirmed transaction
+      const oldestAcceptedSlot = t.slot > stabilityWindowSlotsCount ? t.slot > stabilityWindowSlotsCount : 0;
+      // Remove transactions considered stable
+      confirmedTransactions.filter(({ slot }) => slot > oldestAcceptedSlot);
+      store.set(confirmedTransactions);
+    });
+
+  // REVIEW: throw some error or log something in case there is no entry in confirmed for the rollback?
```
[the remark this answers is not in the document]
mirceahasegan: todo. Not sure how to add the logger (I assume as a parameter)

packages/wallet/src/services/RollbackResubmitter.ts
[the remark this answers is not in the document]
mirceahasegan: Added `submitted$` transactions in the logic. See [0286cb7](https://github.com/input-output-hk/cardano-js-sdk/pull/351/commits/0286cb7b5c3844765304acc03f3163345df65fa2)

packages/wallet/src/services/RollbackResubmitter.ts
```diff
@@ -0,0 +1,50 @@
...
+import { DocumentStore } from '../persistence';
+import { Observable, map, switchMap, tap } from 'rxjs';
+
+// REVIEW: explore how to get the stability window from the network info
+// 3k/f (where k is the security parameter in genesis, and f is the active slot co-efficient parameter
+// in genesis that determines the probability for amount of blocks created in an epoch.)
+const kStabilityWindowSlotsCount = 129_600; // 3k/f
+
+export const createResubmitRollbacks = (
+  rollback$: Observable<Cardano.TxAlonzo[]>,
```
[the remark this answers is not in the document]
mirceahasegan: Updated in [0286cb7](https://github.com/input-output-hk/cardano-js-sdk/pull/351/commits/0286cb7b5c3844765304acc03f3163345df65fa2)

packages/wallet/src/services/RollbackResubmitter.ts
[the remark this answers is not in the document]
mirceahasegan: Updated in [0286cb7](https://github.com/input-output-hk/cardano-js-sdk/pull/351/commits/0286cb7b5c3844765304acc03f3163345df65fa2)

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -283,6 +290,21 @@ export class SingleAddressWallet implements ObservableWallet {
...
+    this.#rollbackSubscription.add(
+      createTransactionReemitter({
+        confirmed$: this.transactions.outgoing.confirmed$,
+        rollback$: this.transactions.rollback$,
+        store: stores.volatileTransactions,
+        submitting$: this.transactions.outgoing.submitting$
+      })
+        .pipe(
+          mergeMap((rollbackTransaction) => from(this.submitTx(rollbackTransaction))),
+          catchError(() => EMPTY) // REVIEW: ignoring errors; already handled as part of the transactions errors
```
mirceahasegan, line 303: Is it ok to ignore the errors?

packages/wallet/src/services/TransactionReemitter.ts
```diff
@@ -0,0 +1,66 @@
...
+        }
+      }
+    }, [] as NewTxAlonzoWithSlot[])
+  );
+
+  return rollback$.pipe(
+    withLatestFrom(volatileTransactions$),
+    map(
+      ([{ id }, volatiles]) =>
+        // REVIEW: all examples of logger imply passing it as parameter. Is that the only way to do it?
```
mirceahasegan, line 60: Logger question :)

mirceahasegan:
Rebased on top of master and fixed existing unit tests

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -288,6 +295,21 @@ export class SingleAddressWallet implements ObservableWallet {
...
```
rhyslbw, line 308: Is it worth logging as `debug` level?

packages/wallet/src/services/TransactionReemitter.ts
rhyslbw: We've not really established a pattern for logging within the wallet package. Do you see an issue with passing it through this way?

packages/wallet/src/SingleAddressWallet.ts
rhyslbw: Maybe worth logging as `debug` level?

packages/wallet/src/services/TransactionReemitter.ts
mirceahasegan: No. It's just that I have add one more param and I wasn't sure if there is a better way to do it.
I will do that

mirceahasegan:
On squash, I will add a better commit message and include the breaking changes.

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -288,6 +295,24 @@ export class SingleAddressWallet implements ObservableWallet {
       tip$: this.tip$,
       transactionsHistoryStore: stores.transactions
     });
+
+    this.#rollbackSubscription.add(
```
mirceahasegan, line 299: @mkazlauskas , do you think a test is needed for this code too?

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -288,6 +295,24 @@ export class SingleAddressWallet implements ObservableWallet {
...
+      createTransactionReemitter({
+        confirmed$: this.transactions.outgoing.confirmed$,
+        rollback$: this.transactions.rollback$,
+        store: stores.volatileTransactions,
+        submitting$: this.transactions.outgoing.submitting$
+      })
+        .pipe(
+          mergeMap((rollbackTransaction) => from(this.submitTx(rollbackTransaction))),
+          catchError((err) => {
+            this.#logger.debug('Failed to resubmit rolled back transaction', err);
```
mkazlauskas, line 309:
```suggestion
            this.#logger.error('Failed to resubmit rolled back transaction', err);
```

packages/wallet/src/SingleAddressWallet.ts
mkazlauskas, line 299: I think this would be a little clearer
```suggestion
    this.#rollbackSubscription =  createTransactionReemitter(...).subscribe(async (rollbackTransaction) => {})
```

packages/wallet/src/persistence/pouchdbStores/pouchdbWalletStores.ts
```diff
@@ -90,6 +93,7 @@ export const createPouchdbWalletStores = (
       logger
     ),
     unspendableUtxo: new PouchdbUtxoStore({ dbName: `${baseDbName}UnspendableUtxo` }, logger),
-    utxo: new PouchdbUtxoStore({ dbName: `${baseDbName}Utxo` }, logger)
+    utxo: new PouchdbUtxoStore({ dbName: `${baseDbName}Utxo` }, logger),
+    volatileTransactions: new PouchdbConfirmedTransactionsStore(docsDbName, 'confirmedTransactions', logger)
```
mkazlauskas, line 97:
```suggestion
    volatileTransactions: new PouchdbConfirmedTransactionsStore(docsDbName, 'volatileTransactions', logger)
```

packages/wallet/src/persistence/pouchdbStores/pouchdbWalletStores.ts
```diff
@@ -25,6 +26,7 @@ export class PouchdbTimeSettingsStore extends PouchdbDocumentStore<TimeSettings[
 export class PouchdbAssetsStore extends PouchdbDocumentStore<Assets> {}
 export class PouchdbAddressesStore extends PouchdbDocumentStore<GroupedAddress[]> {}
 export class PouchdbInFlightTransactionsStore extends PouchdbDocumentStore<Cardano.NewTxAlonzo[]> {}
+export class PouchdbConfirmedTransactionsStore extends PouchdbDocumentStore<NewTxAlonzoWithSlot[]> {}
```
mkazlauskas, line 29:
```suggestion
export class PouchdbVolatileTransactionsStore extends PouchdbDocumentStore<NewTxAlonzoWithSlot[]> {}
```

packages/wallet/src/services/TransactionReemitter.ts
```diff
@@ -0,0 +1,81 @@
...
+        case txSource.submitting: {
+          // Transactions in submitting are the ones reemitted. Remove them from volatiles
+          volatiles = volatiles.filter((v) => v.id !== tx.id);
+          store.set(volatiles);
+          break;
+        }
+        case txSource.confirmed: {
+          const oldestAcceptedSlot = tx.slot > stabilityWindowSlotsCount ? tx.slot - stabilityWindowSlotsCount : 0;
+          // Remove transactions considered stable
+          volatiles = [...volatiles.filter(({ slot }) => slot > oldestAcceptedSlot), tx];
```
mkazlauskas, line 58: Might as well filter out transactions where slot is > of [validityInterval](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/core/src/Cardano/types/Transaction.ts#L42), cause those will fail resubmission anyway

packages/wallet/src/services/TransactionReemitter.ts
```diff
@@ -0,0 +1,81 @@
...
+      }
+      return volatiles;
+    }, [] as NewTxAlonzoWithSlot[])
+  );
+
+  return rollback$.pipe(
+    withLatestFrom(volatileTransactions$),
+    tap(
+      ([{ id }, volatiles]) =>
+        volatiles.some((tx) => tx.id === id) ||
```
mkazlauskas, line 71: If you `map` before `tap` then you won't need to iterate the array again

mkazlauskas (changes requested):
Looking good, just a few minor changes needed

packages/wallet/src/SingleAddressWallet.ts
mkazlauskas: Yes, that would be great - I suggest `test/SingleAddressWallet/rolback.test.ts`

You can inject stub ChainHistoryProvider with multiple mockResolvedValueOnce and mock a TxSubmitProvider. Then await for `firstValueFrom(transactions.history$.pipe(filter(txs => txs.length === n)))` before asserting calls on the TxSubmitProvider

packages/wallet/src/SingleAddressWallet.ts
mirceahasegan: This was per @rhyslbw suggestion to use the `.debug` level. I assume any errors would be caught/logged by the normal tx sumit flow. This was just an extra logging step to traceback the error to rollbacks.

packages/wallet/src/SingleAddressWallet.ts
mirceahasegan: > I think this would be a little clearer

The advantage of `add` (or at least the way I've used it in the past), is that you don't have to check if the subscription exists and subscribed to, in order to unsubscribe it.
Offtopic: it's also a great way to bundle multiple subscriptions that you want to cleanup at a later time.

packages/wallet/src/services/TransactionReemitter.ts
mirceahasegan: > If you `map` before `tap` then you won't need to iterate the array again

I did it like this because after map I lose the `id`, and I wanted to log it.
I could also emit the id from `map`, then strip it in another map. Something like this:
```
map // id -> [id, volatileTx]
tap // log if tx is undefined
map // [id, tx] -> tx

File: packages/wallet/src/SingleAddressWallet.ts (hunk 14 above)
Comment (mkazlauskas) [reply packages/wallet/src/SingleAddressWallet.ts:309]: Depends on which tx submit provider, I think this should be logging an error as this is where the `submitTx` is called :thinking: Thoughts @rhyslbw ?

File: packages/wallet/src/SingleAddressWallet.ts (hunk 13 above)
Comment (mkazlauskas) [reply packages/wallet/src/SingleAddressWallet.ts:299]: I see, but in this case subscription is being set in constructor so it will always exist.

> Offtopic: it's also a great way to bundle multiple subscriptions that you want to cleanup at a later time.

Yep that's a nice way to bundle subscriptions

File: packages/wallet/src/services/TransactionReemitter.ts (hunk 18 above)
Comment (mkazlauskas) [reply packages/wallet/src/services/TransactionReemitter.ts:71]: Good idea, 2 maps is better than iterating the array twice

File: packages/wallet/src/SingleAddressWallet.ts (hunk 14 above)
Comment (rhyslbw) [reply packages/wallet/src/SingleAddressWallet.ts:309]: Yeah, we need to consider the appropriate place to log errors. I'm not sure I have a good answer here, as I think it makes sense for the `Provider` to do it, but there's no way to ensure it :thinking: 

Currently, we're not in these, so best to use `error` over `debug` here:
- https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/blockfrost/src/blockfrostTxSubmitProvider.ts
- https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services-client/src/TxSubmitProvider/txSubmitHttpProvider.ts
- https://github.com/input-output-hk/cardano-js-sdk/blob/980c2227d39af5b2b3476597a4fb74615f4e62aa/packages/ogmios/src/Provider/TxSubmitProvider/ogmiosTxSubmitProvider.ts#L57-L59

Comment (mirceahasegan) [reply packages/wallet/src/SingleAddressWallet.ts]: Done in: [ddb6af8](https://github.com/input-output-hk/cardano-js-sdk/pull/351/commits/ddb6af800a591149f87a2668b4da04d0470a29c7)

File: packages/wallet/src/SingleAddressWallet.ts (hunk 13 above)
Comment (mirceahasegan) [reply packages/wallet/src/SingleAddressWallet.ts]: Done in: [ddb6af8](https://github.com/input-output-hk/cardano-js-sdk/pull/351/commits/ddb6af800a591149f87a2668b4da04d0470a29c7)

File: packages/wallet/src/persistence/pouchdbStores/pouchdbWalletStores.ts (hunk 15 above)
Comment (mirceahasegan) [reply packages/wallet/src/persistence/pouchdbStores/pouchdbWalletStores.ts]: Done in: [ddb6af8](https://github.com/input-output-hk/cardano-js-sdk/pull/351/commits/ddb6af800a591149f87a2668b4da04d0470a29c7)

File: packages/wallet/src/persistence/pouchdbStores/pouchdbWalletStores.ts (hunk 16 above)
Comment (mirceahasegan) [reply packages/wallet/src/persistence/pouchdbStores/pouchdbWalletStores.ts]: Done in: [ddb6af8](https://github.com/input-output-hk/cardano-js-sdk/pull/351/commits/ddb6af800a591149f87a2668b4da04d0470a29c7)

File: packages/wallet/src/services/TransactionReemitter.ts (hunk 18 above)
Comment (mirceahasegan) [reply packages/wallet/src/services/TransactionReemitter.ts]: Done in: [ddb6af8](https://github.com/input-output-hk/cardano-js-sdk/pull/351/commits/ddb6af800a591149f87a2668b4da04d0470a29c7)

File: packages/wallet/src/SingleAddressWallet.ts [hunk 19]
```diff
@@ -105,6 +111,7 @@ export class SingleAddressWallet implements ObservableWallet {
     pending$: new Subject<Cardano.NewTxAlonzo>(),
     submitting$: new Subject<Cardano.NewTxAlonzo>()
   };
+  #resubmitSubscription = new Subscription();
```
mkazlauskas, line 114:
```suggestion
  #resubmitSubscription: Subscription;
```

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -288,6 +295,22 @@ export class SingleAddressWallet implements ObservableWallet {
...
+      confirmed$: this.transactions.outgoing.confirmed$,
+      rollback$: this.transactions.rollback$,
+      store: stores.volatileTransactions,
+      submitting$: this.transactions.outgoing.submitting$
+    })
+      .pipe(
+        mergeMap((transaction) => from(this.submitTx(transaction))),
+        catchError((err) => {
+          this.#logger.error('Failed to resubmit transaction', err);
+          return EMPTY;
```
mkazlauskas, line 309: This is not ideal as SingleAddressWallet silently (end-users won't read the log) fails the transaction here.

We probably want to add some logic:
- resubmit on certain errors (probably with some timeout and maybe considering online status as well): `ProviderFailure.{Unknown,Unhealthy,ConnectionFailure}`
- emit something on any other errors, so that user could inform the user about failed transaction

We might want to do this in a separate ticket, we haven't considered this when estimating. What do you think @rhyslbw ?

mkazlauskas (changes requested):

packages/wallet/src/SingleAddressWallet.ts
mirceahasegan: Fixed here: [eede564](https://github.com/input-output-hk/cardano-js-sdk/pull/351/commits/eede5641ef0ee8cde08bcecdb6215f95158a9357)

packages/wallet/src/services/TransactionReemitter.ts
mirceahasegan: Fixed here [19b46b1](https://github.com/input-output-hk/cardano-js-sdk/pull/351/commits/19b46b15c9ebd6072e2e4e00645e24a10b648e45)

packages/wallet/src/SingleAddressWallet.ts
mirceahasegan: Done here: [afa2d65](https://github.com/input-output-hk/cardano-js-sdk/pull/351/commits/afa2d650c5e559cf6ad8847d6fc89ace1f4827e1)

packages/wallet/test/services/TransactionReemiter.test.ts
```diff
@@ -110,21 +123,26 @@ describe('TransactionReemiter', () => {
...
         confirmed$,
+        logger,
         rollback$,
         store,
-        submitting$
+        submitting$,
+        tipSlot$
       });
-      expectObservable(transactionReemiter).toBe('--a--c|', { a: volatileA, c: volatileC });
+      expectObservable(transactionReemiter).toBe('--a---', { a: volatileA });
```
mkazlauskas, line 132: Could you please add 1 more valid transaction to be emitted after `c`? Just want to make sure that observable is still running after the error.

mkazlauskas (changes requested):
Looking good, just 1 minor change request

packages/wallet/test/services/TransactionReemiter.test.ts
mirceahasegan: Great catch @mkazlauskas. Thanks!
There's a bug there, it does not continue running.
I'm working on the fix

mkazlauskas (approved):

Could you please create follow-up ticket(s) for [this](https://github.com/input-output-hk/cardano-js-sdk/pull/351#discussion_r932232480):
  - emit something when transaction couldn't be submitted (although it's already emitting on failed$ observable, maybe that's good enough?)
  - re-submit on recoverable errors (connection failure etc.)
  
 Also, maybe for resubmitting when previously submitted transaction didn't get into history$ for too long (e.g. due to timing of rollback - when tx was submitted successfully but never got to history$)

rhyslbw (dismissed):
Great work @mirceahasegan :rocket: 
Please squash ready for merge

packages/wallet/src/persistence/pouchdbStores/pouchdbWalletStores.ts
```diff
@@ -57,6 +59,7 @@ export const createPouchdbWalletStores = (
           destroyDocumentsDb,
           this.transactions.destroy(),
           this.inFlightTransactions.destroy(),
+          this.volatileTransactions.destroy(),
```
mkazlauskas, line 62: This line is not needed: all document stores are destroyed via `destroyDocumentsDb`

mkazlauskas (changes requested):

packages/wallet/src/persistence/pouchdbStores/pouchdbWalletStores.ts
mirceahasegan: 👍 
Fixed and squashed in one go: https://github.com/input-output-hk/cardano-js-sdk/commit/2a4ccb0abead34481e817f807850d29e77d7340a#diff-47afda8b8869c2531c861dbf59d047a481b4765d18b8e5f8ba0a7b19ada7c02dL59

rhyslbw (approved):

mkazlauskas (approved):

PR #352, opened 2022-07-25
refactor(cardano-services):  CLI improvements in cardano-services

IvayloAndonov:
# Context

A) There’s currently an inconsistency with our PostgreSQL configuration. Static configuration is named `--db-connection-string` , yet DNS resolved is --postgres-db --postgres-user etc. I suggest renaming to `--postgres-connection-string`. This change should also be reflected in the environment variables and all internal interfaces and variables.

B)
Separate these:

82f386cb-0124-4ab3-91e8-f326fa12867a

These options are both booleans, yet one appears with <>

8d5c3232-373e-4d14-893c-e459978f52e4

Duplicate option, plus it’s a little vague about what the cache is for:

17c6c275-72a6-4342-b539-bb848515f658

# Proposed Solution

Screenshot 2022-07-25 at 13 52 12

+ check **Important Changes Introduced**

- rename `DB_CONNECTION_STRING` to `POSTGRES_CONNECTION_STRING`; `--db-connection-string` to `--postgres-connection-string`
- rename `CACHE_TTL` to `DB_CACHE_TTL`; `--cache-ttl` to `--db-cache-ttl`
- fix _enableMetrics_ var name (previously _metricsEnabled_)
- remove duplicated `cache-ttl`
- rename `cache-ttl` to `db-cache-ttl` and remove it from common args (since the worker does not depend on the cache for DNS resolution)
- break down `ProgramOptionDescriptions.PostgresSrvArgs` into separate postgres related args descriptions

# Additional notes

We still keep the [combined](https://github.com/input-output-hk/cardano-js-sdk/pull/352/files#diff-f22112ab0dd5aa721e1cd32145b150caffbf1a01a2707166f208568286b4e2c0R14-R15) service discovery option description for postgres (db,user,password,srv name) for the [dbConnection check](https://github.com/input-output-hk/cardano-js-sdk/blob/0f532252b77e9fca84033ec8dd34d610de816d7c/packages/cardano-services/src/Program/loadHttpServer.ts#L71-L75) as at that point we do not know which env exactly is missing. This would be eliminated once we have consolidated entry points (run scripts and cli) refactoring done so that the check will be moved beforehand as part of env vars/args validation via Commander.js


packages/cardano-services/src/ProgramsCommon/CommonOptionDescriptions.ts
```diff
@@ -1,5 +1,5 @@
 export enum CommonOptionDescriptions {
-  CacheTtl = 'Cache TTL in minutes between 1 and 2880',
+  DbCacheTtl = 'Cache TTL in minutes between 1 and 2880, an option for database related operations',
```
mkazlauskas, line 2: Would be nice to list what exactly is being cached

mkazlauskas (dismissed):
Looks good!

packages/cardano-services/src/Program/ProgramOptionDescriptions.ts
```diff
@@ -3,11 +3,16 @@ import { CommonOptionDescriptions } from '../ProgramsCommon';
 enum HttpServerOptionDescriptions {
   ApiUrl = 'API URL',
   CardanoNodeConfigPath = 'Cardano node config path',
-  DbConnection = 'DB Connection string',
   EpochPollInterval = 'Epoch poll interval',
   MetricsEnabled = 'Enable Prometheus Metrics',
```
rhyslbw, line 7:
```suggestion
  EnableMetrics = 'Enable Prometheus Metrics',
```

rhyslbw (changes requested):

packages/cardano-services/src/ProgramsCommon/CommonOptionDescriptions.ts
IvayloAndonov: I think the list would become longer for this static description, as currently, we cache 4 network info related values, 1 epoch-related, and soon more stake-pool related cached values.

IvayloAndonov: Probably could be created a cache description generation util which gets a config list of cached keys and returns a composed description, not sure how it will look with `cli --help` though :D

packages/cardano-services/src/Program/ProgramOptionDescriptions.ts
IvayloAndonov: [05880f1](https://github.com/input-output-hk/cardano-js-sdk/pull/352/commits/05880f1b446edfa537478786bc33ba2fc1aa674b)

packages/cardano-services/src/ProgramsCommon/CommonOptionDescriptions.ts
mkazlauskas: Yeah, nevermind if it's not practical

packages/cardano-services/src/Program/ProgramOptionDescriptions.ts
```diff
@@ -3,11 +3,16 @@ import { CommonOptionDescriptions } from '../ProgramsCommon';
...
   ApiUrl = 'API URL',
   CardanoNodeConfigPath = 'Cardano node config path',
-  DbConnection = 'DB Connection string',
   EpochPollInterval = 'Epoch poll interval',
-  MetricsEnabled = 'Enable Prometheus Metrics',
+  EnableMetrics = 'Enable Prometheus Metrics',
   OgmiosUrl = 'Ogmios URL',
-  PostgresSrvArgs = 'Postgres SRV service name, db, user and password',
+  PostgresConnectionString = 'Postgres Connection string',
+  PostgresSrvServiceName = 'Postgres SRV service name with service discovery',
```
rhyslbw, line 10:
```suggestion
  PostgresSrvServiceName = 'PostgreSQL SRV service name when using service discovery',
```

packages/cardano-services/src/Program/ProgramOptionDescriptions.ts
```diff
@@ -3,11 +3,16 @@ import { CommonOptionDescriptions } from '../ProgramsCommon';
...
+  PostgresDb = 'Postgres db name with service discovery',
```
rhyslbw, line 11:
```suggestion
  PostgresDb = 'PostgreSQL database name when using service discovery',
```

packages/cardano-services/src/Program/ProgramOptionDescriptions.ts
```diff
@@ -3,11 +3,16 @@ import { CommonOptionDescriptions } from '../ProgramsCommon';
...
+  PostgresUser = 'Postgres user with service discovery',
```
rhyslbw, line 12:
```suggestion
  PostgresUser = 'PostgreSQL user when using service discovery',
```

packages/cardano-services/src/Program/ProgramOptionDescriptions.ts
```diff
@@ -3,11 +3,16 @@ import { CommonOptionDescriptions } from '../ProgramsCommon';
...
+  PostgresPassword = 'Postgres password with service discovery',
```
rhyslbw, line 13:
```suggestion
  PostgresPassword = 'PostgreSQL password when using service discovery',
```

rhyslbw (changes requested):

rhyslbw (approved):

rhyslbw (dismissed):

rhyslbw (approved):

PR #353, opened 2022-07-25
ci: nightly publish automation

mkazlauskas:
# Context

Publishing packages to npm is currently done quite rarely and there are users who would benefit from more frequent deployments. Existing process includes a manual chore and we want to reduce the overhead of having to worry about publishing the packages.

# Proposed Solution

- Install and setup [lerna](https://lerna.js.org/) with npm scripts to publish:
  - `nightly` releases from `nightly` branch (that is kept in sync with `master`)
  - `beta` releases from `x-y-stable` branch
  - stable releases from `x-y-stable` branch
- Add GH workflow to publish `nightly` periodically (every midnight UTC)

See 8865db62d988ed69f9efb907fc9705b2a11c3c9c and 4adf3d842aef06b9fcea42e0f8346883fbde382b commit notes


.github/workflows/publish-nightly.yaml
```diff
@@ -0,0 +1,48 @@
...
+        submodules: true
+      run: git rebase master
+
+    - name: 🧰 Setup Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: 14.17
+        registry-url: 'https://registry.npmjs.org'
+
+    # Alternatively (if we don't like the workflow failures)
```
rhyslbw, line 30: Let's go with the alternative, as a failure indicates something is wrong and needs attention.

rhyslbw (changes requested):
Overall this looks great, but the workflow failure isn't ideal.

.github/workflows/publish-nightly.yaml
mkazlauskas: Done in 8fbf2d3d81f571ece14746db9f0a77d2f7fd6751

rhyslbw (dismissed):

rhyslbw (approved):
:rocket:

PR #354, opened 2022-07-25
Feat/support secure db connection

jvieiro:
# Context

Support secure db-connection

# Proposed Solution

Add a configuration option to the cardano-services entrypoints, which maps into the construction of Pool

### DIFF
--- a/packages/cardano-services/src/Program/utils.ts
+++ b/packages/cardano-services/src/Program/utils.ts
@@ -12,6 +12,7 @@ import { ProviderFailure, TxSubmitProvider } from '@cardano-sdk/core';
 import { RabbitMqTxSubmitProvider } from '@cardano-sdk/rabbitmq';
 import { ServiceNames } from './ServiceNames';
 import dns, { SrvRecord } from 'dns';
+import fs from 'fs';
 import pRetry, { FailedAttemptError } from 'p-retry';
 
 export const SERVICE_DISCOVERY_BACKOFF_FACTOR_DEFAULT = 1.1;
@@ -68,10 +69,10 @@ export type DnsResolver = ReturnType<typeof createDnsResolver>;
 export const getPoolWithServiceDiscovery = async (
   dnsResolver: DnsResolver,
   logger: Logger,
-  { host, database, password, user }: ClientConfig
+  { host, database, password, ssl, user }: ClientConfig
 ): Promise<Pool> => {
   const { name, port } = await dnsResolver(host!);
-  let pool = new Pool({ database, host: name, password, port, user });
+  let pool = new Pool({ database, host: name, password, port, ssl, user });
 
   return new Proxy<Pool>({} as Pool, {
     get(_, prop) {
@@ -82,7 +83,7 @@ export const getPoolWithServiceDiscovery = async (
             if (error.code && ['ENOTFOUND', 'ECONNREFUSED', 'ECONNRESET'].includes(error.code)) {
               const record = await dnsResolver(host!);
               logger.info(`DNS resolution for Postgres service, resolved with record: ${JSON.stringify(record)}`);
-              pool = new Pool({ database, host: record.name, password, port: record.port, user });
+              pool = new Pool({ database, host: record.name, password, port: record.port, ssl, user });
               return await pool.query(args, values);
             }
             throw error;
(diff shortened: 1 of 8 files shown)


packages/cardano-services/src/Program/utils.ts
```diff
@@ -55,6 +56,13 @@ export const createDnsResolver = (config: RetryBackoffConfig, logger: Logger) =>
 
 export type DnsResolver = ReturnType<typeof createDnsResolver>;
 
+const httpServerOptionsToPostgresSsl = (options?: HttpServerOptions): ClientConfig['ssl'] =>
+  options?.postgresSsl
+    ? options.postgresSslCaFile
+      ? { ca: options.postgresSslCaFile }
```
YorikSar, line 62: This file needs to be read before passing to ssl library.

YorikSar (changes requested):

packages/cardano-services/src/Program/utils.ts
YorikSar: See https://github.com/input-output-hk/cardano-js-sdk/commit/51bee6ea0038d9de0e448e684fa002d4502521a6#diff-5935fb07258956001ade74f0edc794b473e06615d67973cc4d416ec157357be3R70-R71 for example

mkazlauskas (commented):
An entrypoints test would be nice. Maybe asserting that output (error) contains some specific text like [here](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services/test/entrypoints.test.ts#L563)

packages/cardano-services/src/Program/utils.ts
rhyslbw: In my quick refactor, I loaded it [here](https://github.com/input-output-hk/cardano-js-sdk/pull/354/files#diff-cd616b155b9729672e02921d8863a908cbaf8d92d935ca96f9708a083c87959bR92), but yes, doing it in the util will cover both entrypoints (we still have). @jvieiro please make this change, along with the test requested by @mkazlauskas below.

rhyslbw (approved):
Looks good @jvieiro. Please squash to a single commit, and @YorikSar can cherry-pick to ensure it's working with the nix infra before we merge.

packages/cardano-services/test/entrypoints.test.ts
```diff
@@ -863,5 +865,137 @@ describe('entrypoints', () => {
...
+                POSTGRES_CONNECTION_STRING: postgresConnectionString,
+                POSTGRES_SSL_CA_FILE: postgresSslCaFile,
+                SERVICE_NAMES: ServiceNames.NetworkInfo
+              },
+              stdio: 'pipe'
+            });
+            proc.stderr!.on('data', (data) => {
+              spy();
+              // eslint-disable-next-line no-console
+              console.log(data.toString());
```
rhyslbw, line 988:
```suggestion
```

packages/cardano-services/test/entrypoints.test.ts
```diff
@@ -863,5 +865,137 @@ describe('entrypoints', () => {
...
+              cardanoNodeConfigPath,
+              ServiceNames.NetworkInfo,
+              '--postgres-ssl-ca-file',
+              postgresSslCaFile
+            ],
+            { stdio: 'pipe' }
+          );
+          proc.stderr!.on('data', (data) => {
+            spy();
+            expect(data.toString().includes('self signed certificate')).toEqual(true);
```
rhyslbw, line 962: Where is this text coming from?

rhyslbw (changes requested):
Just noticed a couple of issues with the test. The happy path assertion here might be a bit tricky, but as long as we're seeing an error that indicates the config resulted in the file contents being loaded, it's acceptable.

packages/cardano-services/test/entrypoints.test.ts
jvieiro: The error thrown is due to an invalid certificate `Error: DEPTH_ZERO_SELF_SIGNED_CERT: self signed certificate`, a comment claryfing this can be added. 
If we want to avoid this error I've seen a workaround by setting `process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0"` but I'm not convinced with that.

rhyslbw: OK. I think this is fine given it demonstrates the content of the file is being loaded and passed to the underlying library.

jvieiro: Updated to `The server does not support SSL connections`

rhyslbw (approved):

mkazlauskas (approved):
Looks good!

PR #355, opened 2022-07-26
fix: malformed string and add missing service to Docker defaults

rhyslbw:
# Context
The Docker files are currently not covered by e2e tests in CI, so there were some bugs with configuration structure that slipped though, and the latest HTTP service, `asset`, was not appended to the default.

# Proposed Solution
- Fix the malformed string
- Add `asset` to the list


AngelCastilloB (approved):
LGTM!

PR #357, opened 2022-07-26
ci: add on workflow_dispatch in nightly publish workflow

mkazlauskas:
# Context

Need to be able to manually trigger nightly publish workflow


rhyslbw (commented):
Maybe this workflow is better named just nightly? `nightly.jobs.publish`

rhyslbw (approved):

PR #358, opened 2022-07-26
Fix/adp 2081 allow patching axios with custom adapter

AngelCastilloB:
# Context

The Cardano services HTTP clients cannot be used in the browser environment because they depend on Axios.

# Proposed Solution

Allow the HTTP clients to receive an Axios adapter so we can replace the default adapter with one that uses fetch instead.

It is now possible to optionally pass an Axios adapter to the HTTP clients as an argument.
It is now possible to pass a logger instance to the HTTP clients as an argument.

### DIFF
--- a/packages/cardano-services-client/src/AssetInfoProvider/assetInfoHttpProvider.ts
+++ b/packages/cardano-services-client/src/AssetInfoProvider/assetInfoHttpProvider.ts
@@ -1,4 +1,5 @@
 import { AssetProvider } from '@cardano-sdk/core';
+import { AxiosAdapter } from 'axios';
 import { HttpProviderConfigPaths, createHttpProvider } from '../HttpProvider';
 
 export const defaultAssetProviderPaths: HttpProviderConfigPaths<AssetProvider> = {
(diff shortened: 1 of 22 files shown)


packages/cardano-services-client/src/AssetInfoProvider/assetInfoHttpProvider.ts
```diff
@@ -10,9 +12,19 @@ export const defaultAssetProviderPaths: HttpProviderConfigPaths<AssetProvider> =
...
+ * @param paths A mapping between provider method names and url paths. Paths have to use /leadingSlash
+ * @param adapter This adapter that allows to you to modify the way Axios make requests.
+ * @param logger The logger instance.
  */
-export const assetInfoHttpProvider = (baseUrl: string, paths = defaultAssetProviderPaths): AssetProvider =>
+export const assetInfoHttpProvider = (
+  baseUrl: string,
+  paths = defaultAssetProviderPaths,
+  adapter?: AxiosAdapter,
+  logger = dummyLogger
```
iccicci, line 23: Is the scope of ADP-1933 only server side packages?
Otherwise `dummyLogger` (or dummy stuff in general) should be used only in tests.
The same comment apply to other files as well.

iccicci (dismissed):
LGTM.
I have only a minor question.

packages/cardano-services-client/src/AssetInfoProvider/assetInfoHttpProvider.ts
rhyslbw: > Is the scope of ADP-1933 only server side packages?

No. Good catch @iccicci

mkazlauskas: Right, so logger should be a required param

AngelCastilloB: addressed here https://github.com/input-output-hk/cardano-js-sdk/commit/78c257176933121ad83152ddb17d1e555b0f8b35 , logger is now a required parameter, also, in order to remove positional parameters in the services factory methods, I created a new config object.

mkazlauskas (dismissed):
Nice work!

rhyslbw (approved):

PR #360, opened 2022-07-27
feat: Service discovery support for OgmiosCardanoNode

IvayloAndonov:
# Context

We need to wrap [OgmiosCardanoNode](https://github.com/input-output-hk/cardano-js-sdk/blob/08d9ed2b6aa20cf4df2a063f046f4e5ca28c6bd5/packages/ogmios/src/CardanoNode/OgmiosCardanoNode.ts#L13) class within an abstraction that supports DNS resolution and error handling, similar to Pool, Ogmios, and RabbitMQ provider's abstractions.

`OgmiosCardanoNode` is a newly created class in `ogmios` package needed for querying on-chain data through Ogmios cardano-node dependency.

With the latest version, `network-info` service depends on OgmiosCardanoNode class in order to fetch time-settings data dynamically. Currently, it consumes only `ogmiosUrl` .

# Proposed Solution
N/A

- regroup service dependencies into files: `Program/services/`
- create `getOgmiosCardanoNode` abstraction which instantiate `OgmiosCardanoNode` instance based on the provided program argmuments
- create `ogmiosCardanoNodeWithDiscovery` with DNS resolution, retry logic and error handling
- extend `mockOgmiosServer` with `connectionError` type


IvayloAndonov:
Closed in favor of https://github.com/input-output-hk/cardano-js-sdk/pull/362

PR #362, opened 2022-07-28
feat: Service discovery support for OgmiosCardanoNode

IvayloAndonov:
# Context

We need to wrap [OgmiosCardanoNode](https://github.com/input-output-hk/cardano-js-sdk/blob/08d9ed2b6aa20cf4df2a063f046f4e5ca28c6bd5/packages/ogmios/src/CardanoNode/OgmiosCardanoNode.ts#L13) class within an abstraction that supports DNS resolution and error handling, similar to Postgres, Ogmios, and RabbitMQ provider's abstractions.

`OgmiosCardanoNode` is a newly created class in `ogmios` package needed for querying on-chain data through Ogmios cardano-node dependency.

With the latest version, `network-info` service depends on `OgmiosCardanoNode` class in order to fetch time-settings data dynamically. Currently, it consumes only `ogmiosUrl` .

# Proposed Solution
- create `getOgmiosCardanoNode` abstraction which instantiate `OgmiosCardanoNode` instance based on the provided program argmuments
- create `ogmiosCardanoNodeWithDiscovery` with DNS resolution, retry logic and error handling
- extend `mockOgmiosServer` with `connectionError` type
- reorganize service dependencies into files: `Program/services/`
- reorganize tests
- add missing dev dependency `ts-node` required for running `yarn run:http-server` and `yarn run:http-server:debug`
- expose `isConnectionError` from `util` package

N/A


rhyslbw (changes requested):
Nice work @IvayloAndonov. Please rebase first to address the merge conflict, then apply the suggested change, as @YorikSar can then test this branch along with another recently merged[ feature](https://github.com/input-output-hk/cardano-js-sdk/pull/354) of interest.

packages/cardano-services/src/Program/utils.ts
```diff
@@ -55,6 +55,19 @@ export const createDnsResolver = (config: RetryBackoffConfig, logger: Logger) =>
 
 export type DnsResolver = ReturnType<typeof createDnsResolver>;
 
+export const connectionErrorCodes = ['ENOTFOUND', 'ECONNREFUSED', 'ECONNRESET'];
```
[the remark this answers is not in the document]
IvayloAndonov: Since the repeated logic is used across two different packages, where we prefer to export the common function, into `util` or `util-dev` package maybe ?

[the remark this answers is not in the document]
rhyslbw: `util` :+1:

Any package named `*-dev` is a `devDependency` BTW, so that wouldn't be an option here.

packages/cardano-services/src/Program/services/ogmios.ts
```diff
@@ -0,0 +1,151 @@
...
+      if (prop === 'eraSummaries') {
+        return () =>
+          ogmiosCardanoNode.eraSummaries().catch(async (error) => {
+            if (isConnectionError(error)) {
+              const record = await dnsResolver(serviceName!);
+              logger.info(`DNS resolution for Ogmios service, resolved with record: ${JSON.stringify(record)}`);
+              await ogmiosCardanoNode
+                .shutdown?.()
+                .catch((error_) =>
+                  logger.warn(`Ogmios cardano node failed to shutdown after DNS resolution: ${error_}`)
```
mkazlauskas, line 121:
```suggestion
                  logger.warn(`Ogmios cardano node failed to shutdown after connection error: ${error_}`)
```

packages/cardano-services/src/Program/services/ogmios.ts
```diff
@@ -0,0 +1,151 @@
...
+                );
+              ogmiosCardanoNode = new OgmiosCardanoNode({ host: record.name, port: record.port }, logger);
+              return await ogmiosCardanoNode.initialize();
+            }
+            throw error;
+          });
+      }
+      if (prop === 'eraSummaries') {
```
mkazlauskas, line 112: I think we also need to override `systemStart`. Can probably add `'initialize'` as well to dedupe some code.

```suggestion
      const isCardanoNodeOperation = (prop: unknown): prop is 'eraSummaries' | 'systemStart' => ['eraSummaries', 'systemStart'].includes(prop);
      if (isCardanoNodeOperation(prop)) {
```

packages/cardano-services/src/Program/services/ogmios.ts
```diff
@@ -0,0 +1,151 @@
...
+        return () =>
+          ogmiosCardanoNode.eraSummaries().catch(async (error) => {
```
mkazlauskas, line 114:
```suggestion
          ogmiosCardanoNode[prop]().catch(async (error) => {
```

packages/cardano-services/src/Program/services/ogmios.ts
```diff
@@ -0,0 +1,151 @@
...
+              const record = await dnsResolver(serviceName!);
+              logger.info(`DNS resolution for Ogmios service, resolved with record: ${JSON.stringify(record)}`);
+              await ogmiosCardanoNode
+                .shutdown?.()
+                .catch((error_) =>
+                  logger.warn(`Ogmios cardano node failed to shutdown after DNS resolution: ${error_}`)
+                );
+              ogmiosCardanoNode = new OgmiosCardanoNode({ host: record.name, port: record.port }, logger);
+              await ogmiosCardanoNode.initialize();
+              return await ogmiosCardanoNode.eraSummaries();
```
mkazlauskas, line 125:
```suggestion
              return await ogmiosCardanoNode[prop]();
```

mkazlauskas (changes requested):

packages/cardano-services/src/Program/services/ogmios.ts
IvayloAndonov: I initially checked that [systemStart](https://github.com/input-output-hk/cardano-js-sdk/blob/08d9ed2b6aa20cf4df2a063f046f4e5ca28c6bd5/packages/ogmios/src/CardanoNode/OgmiosCardanoNode.ts#L65) has no real usage and refs except in the tests. The `stateQueryClient.systemStart()` is directly used within [eraSummaries](https://github.com/input-output-hk/cardano-js-sdk/blob/08d9ed2b6aa20cf4df2a063f046f4e5ca28c6bd5/packages/ogmios/src/CardanoNode/OgmiosCardanoNode.ts#L51-L63) method, that's why didn't override Cardano Node one. Shall we override it then just in case? @mkazlauskas

mkazlauskas: Yes please, if it gets used then it'll be a bug. Or make it private.

IvayloAndonov: _> Can probably add 'initialize' as well_
It differs from the other operations, as there it needs to invoke first (`initialize`) and then the given operation (`systemStart` or `eraSummaries`), where if `initialize` fails we return it after DNS resolution

mkazlauskas: I see, maybe you could hoist [these](https://github.com/input-output-hk/cardano-js-sdk/blob/e90f99d6ad7ac47669c50b388b2a1b137d5de3aa/packages/cardano-services/src/Program/services/ogmios.ts#L99-L106) in a function like `recreateCardanoNode` then

packages/cardano-services/src/Program/utils.ts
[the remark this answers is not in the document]
IvayloAndonov: 629e967ee41670107a9f2fb26e0328366fadfa51

packages/cardano-services/src/Program/services/ogmios.ts
IvayloAndonov: The changes are applied under: 629e967ee41670107a9f2fb26e0328366fadfa51

packages/cardano-services/src/Program/services/ogmios.ts
```diff
@@ -0,0 +1,155 @@
...
+  let ogmiosCardanoNode = new OgmiosCardanoNode({ host: name, port }, logger);
+
+  return new Proxy<OgmiosCardanoNode>({} as OgmiosCardanoNode, {
+    get(_, prop) {
+      if (prop === 'then') return;
+      if (prop === 'initialize') {
+        return () =>
+          ogmiosCardanoNode.initialize().catch(async (error) => {
+            if (isConnectionError(error)) {
+              ogmiosCardanoNode = await recreateOgmiosCardanoNode(serviceName, ogmiosCardanoNode, dnsResolver, logger);
```
mkazlauskas, line 117: If you put `recreateOgmiosCardanoNode` inside `ogmiosCardanoNodeWithDiscovery` then you wouldn't need to pass these args. Just a note, this is good as-is.

mkazlauskas (dismissed):
Looks good, just need to put `isConnectionError` in a separate commit with `util` scope. You can update the error message when squashing.

packages/cardano-services/src/Program/services/ogmios.ts
IvayloAndonov: yep, I've just put it outside to reduce the lines of `ogmiosTxSubmitProviderWithDiscovery` even more.

packages/cardano-services/src/Program/services/ogmios.ts
IvayloAndonov: Renamed under https://github.com/input-output-hk/cardano-js-sdk/commit/ac1bf85a52ee993e9acf3cc81cf392d42a045088

IvayloAndonov:
> Looks good, just need to put `isConnectionError` in a separate commit with util scope. You can update the error message when squashing.

Both are done:
- util package change: [56e05a1](https://github.com/input-output-hk/cardano-js-sdk/commit/56e05a1eff65a058b48f786fdd6bebe5949b1aa7) ,
- updated usage: [ac1bf85](https://github.com/input-output-hk/cardano-js-sdk/commit/ac1bf85a52ee993e9acf3cc81cf392d42a045088) + applied feedback

packages/cardano-services/src/Program/utils.ts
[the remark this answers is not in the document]
rhyslbw: How about adding a `network` directory to set some context?

`./src/network/isConnectionError.ts`

rhyslbw (dismissed):
Ready for squashing!

IvayloAndonov:
Squashed after CI passes

rhyslbw (approved):

mkazlauskas (approved):
Nice work @IvayloAndonov :tada:

PR #364, opened 2022-07-28
chore: Docker Compose config fixes

rhyslbw:
# Context
This is a change to support cardano-db-sync@13, but it was missed or lost in the process of
upgrading. The snapshot expects this new path, otherwise, it will fail during unpacking. This change will only cause existing containers to be rebuilt, which will happen next time the stack is started without intervention. There were also a couple of other issues that I've corrected, see the commit log for details.

### DIFF
--- a/packages/cardano-services/docker-compose.yml
+++ b/packages/cardano-services/docker-compose.yml
@@ -68,7 +68,7 @@ services:
       - postgres_db
     volumes:
       - ./config/network/${NETWORK:-mainnet}:/config
-      - db-sync-data:/var/lib/cdbsync
+      - db-sync-data:/var/lib/cexplorer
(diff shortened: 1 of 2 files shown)


iccicci (approved):
I trust you. I take this more as an info than as review request. Thank you.

IvayloAndonov (approved):

iccicci:
Sorry @rhyslbw ; I saw that only now.

Why an empty `docker-compose.yml` file in the root? Is it a refuse or is there a specific reason?

Thank you.

rhyslbw:
Sorry reviewers, I pulled the trigger on requests too early. Updated the description, but please see each commit for details

mkazlauskas (approved):
LGTM

PR #365, opened 2022-07-28
refactor(cardano-services): use client instead of http server on http…

jvieiro:
… service tests

# Context

Making raw HTTP requests is useful to make assertions about the HTTP server, but for domain tests, we can be using the cardano-services-client modules to abstract the transport concerns.

# Proposed Solution

Replace tests that were using HTTP request and not making assertions about the HTTP server per the corresponding cardano-service-client module

### DIFF
--- a/packages/cardano-services/test/ChainHistory/__snapshots__/ChainHistoryHttpService.test.ts.snap
+++ b/packages/cardano-services/test/ChainHistory/__snapshots__/ChainHistoryHttpService.test.ts.snap
@@ -4,65 +4,39 @@ exports[`ChainHistoryHttpService healthy state /blocks/by-hashes server and snap
 Array [
   Object {
     "confirmations": 1957883,
-    "date": Object {
-      "__type": "Date",
-      "value": 1595995096000,
-    },
+    "date": 2020-07-29T03:58:16.000Z,
     "epoch": 74,
     "epochSlot": 27480,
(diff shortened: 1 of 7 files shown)


packages/cardano-services/test/ChainHistory/__snapshots__/ChainHistoryHttpService.test.ts.snap
```diff
@@ -4,65 +4,39 @@ exports[`ChainHistoryHttpService healthy state /blocks/by-hashes server and snap
...
       "slot": 1625880,
     },
-    "nextBlock": Object {
-      "__type": "undefined",
-    },
-    "previousBlock": Object {
-      "__type": "undefined",
-    },
+    "nextBlock": undefined,
+    "previousBlock": undefined,
```
mkazlauskas, line 17: We might want to add some consecutive blocks to our database snapshot to test this

mkazlauskas (approved):
LGTM!

rhyslbw (approved):

packages/cardano-services/test/ChainHistory/__snapshots__/ChainHistoryHttpService.test.ts.snap
rhyslbw: ADP-2096

PR #367, opened 2022-07-29
fix(cardano-services):  empty array and default condition for  identi…

jvieiro:
…fier filters - search stakePool

# Context

There are two cases related to `identifier` filter that throws syntax error when hitting `stake-pool/search` endpoint
Body request samples are: 

- `{"args": [{"filters": {"identifier": {"values":[{"name": "TestPool", "ticker": "TEST"}]}}}]}`

- `{"args": [{ "filters": { "identifier": { "values": [] } } }]}`

# Proposed Solution

- [x] Fix the current logic to get the default filter's condition

- [x] Ignore `identifier` filter if `values` array is empty

### DIFF
--- a/packages/cardano-services/test/StakePool/StakePoolHttpService.test.ts
+++ b/packages/cardano-services/test/StakePool/StakePoolHttpService.test.ts
@@ -261,17 +261,46 @@ describe('StakePoolHttpService', () => {
           expect(responseWithOrCondition).toMatchSnapshot();
           expect(responseWithAndCondition).toEqual(responseWithAndCondition);
         });
+        it('no given condition equals to OR condition', async () => {
+          const req: StakePoolQueryOptions = {
+            filters: {
+              identifier: { values: [{ name: 'Unknown Name', ticker: 'TEST' }] }
+            }
+          };
+          const response = await doStakePoolRequest<[StakePoolQueryOptions], StakePoolSearchResults>(url, [req]);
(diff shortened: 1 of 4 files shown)


packages/cardano-services/test/StakePool/StakePoolHttpService.test.ts
```diff
@@ -261,6 +261,20 @@ describe('StakePoolHttpService', () => {
           expect(responseWithOrCondition).toMatchSnapshot();
           expect(responseWithAndCondition).toEqual(responseWithAndCondition);
         });
+        it('no given condition equals to OR condition', async () => {
+          const req: StakePoolQueryOptions = {
+            filters: {
+              identifier: { values: [{ name: 'TestPool', ticker: 'TEST' }] }
```
rhyslbw, line 267: This is the same expected result for an `AND` condition.

rhyslbw (changes requested):

AngelCastilloB:
LGTM

AngelCastilloB (approved):

packages/cardano-services/test/StakePool/StakePoolHttpService.test.ts
rhyslbw: 1957b10ef42034931ecf954fa66bf898f141cc8d

AngelCastilloB (approved):

AngelCastilloB (approved):

PR #368, opened 2022-08-02
fix(cardano-services): fixed a division by 0 on APY calculation

AngelCastilloB:
# Context

If the network is configured to have an epoch length duration of less than one day, the APY calculation query throws a division by 0 error.

# Proposed Solution

Make sure the division by zero can not occur in the query.

A small change in the query was introduced to avoid the division by 0.


IvayloAndonov (approved):
LGTM! Not sure if this could be covered with a test though.

rhyslbw:
> LGTM! Not sure if this could be covered with a test though.

It's covered by the local network e2e testing, which has the sub 1 day epochs

rhyslbw (approved):

PR #369, opened 2022-08-02
test(e2e): refactor delegation e2e test to use a dynamic pool selection

rhyslbw:
# Context
The current implementation using a configurable pair of pool IDs was done when we had the `StakePoolProvider` mocked. Since this is no longer a condition, we can now fetch active pools in the network to delegate to.

# Proposed Solution
- Refactor the `getWallet` function to return the provider instances, enabling tests to also use them directly.
- Query the StakePoolProvider to select an active one randomly, filtering the ensure it's not the same as the current delegation.

- `POOL_ID_1` and `POOL_ID_2` are no longer required, so you can now remove from git ignored `.env`s

### DIFF
--- a/packages/e2e/env/.env.wallet
+++ b/packages/e2e/env/.env.wallet
@@ -15,7 +15,3 @@ TX_SUBMIT_PROVIDER=blockfrost
 UTXO_PROVIDER=blockfrost
 WALLET_PROVIDER=blockfrost
 STAKE_POOL_PROVIDER=stub
-
-# Test Params
-POOL_ID_1=pool1euf2nh92ehqfw7rpd4s9qgq34z8dg4pvfqhjmhggmzk95gcd402
-POOL_ID_2=pool1fghrkl620rl3g54ezv56weeuwlyce2tdannm2hphs62syf3vyyh
--- a/packages/e2e/test/wallet/SingleAddressWallet/delegation.test.ts
+++ b/packages/e2e/test/wallet/SingleAddressWallet/delegation.test.ts
@@ -67,8 +67,8 @@ const waitForTx = async (wallet: ObservableWallet, { hash }: Transaction.TxInter
 };
 
 describe('SingleAddressWallet/delegation', () => {
-  let wallet1: ObservableWallet;
-  let wallet2: ObservableWallet;
+  let wallet1: Awaited<ReturnType<typeof getWallet>>;
+  let wallet2: Awaited<ReturnType<typeof getWallet>>;
 
   beforeAll(async () => {
     jest.setTimeout(180_000);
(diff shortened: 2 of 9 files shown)


AngelCastilloB:
``` javascript
  POOL_ID_1: envalid.str(),
  POOL_ID_2: envalid.str(),
```

Must be removed from [environment.ts](https://github.com/input-output-hk/cardano-js-sdk/blob/af50e199bd048565ad4e73299a361b7c5147520a/packages/e2e/test/wallet/environment.ts#L14)

packages/e2e/test/wallet/SingleAddressWallet/delegation.test.ts
```diff
@@ -67,31 +58,45 @@ const waitForTx = async (wallet: ObservableWallet, { hash }: Transaction.TxInter
...
+      ? [wallet1.wallet, wallet2.wallet]
+      : [wallet2.wallet, wallet1.wallet];
+  };
+
+  const chooseDifferentPoolIdRandomly = async (
+    delegateeBefore1stTx: Cardano.PoolId | undefined
+  ): Promise<Cardano.PoolId> => {
+    const activePools = await wallet1.providers.stakePoolProvider.queryStakePools({
+      filters: { status: [Cardano.StakePoolStatus.Active] },
+      pagination: { limit: 5, startAt: 0 }
```
mkazlauskas, line 90: That should be slightly faster
```suggestion
      pagination: { limit: 2, startAt: 0 }
```

mkazlauskas (changes requested):

packages/e2e/test/wallet/SingleAddressWallet/delegation.test.ts
rhyslbw: Done in https://github.com/input-output-hk/cardano-js-sdk/pull/369/commits/3b886a7a6529b71445e0928d92921d9b732aa2b1

packages/e2e/env/.env.wallet
```diff
@@ -15,7 +15,3 @@ TX_SUBMIT_PROVIDER=blockfrost
 UTXO_PROVIDER=blockfrost
 WALLET_PROVIDER=blockfrost
 STAKE_POOL_PROVIDER=stub
```
ssong-iohk, line 17: Is it still `stub`? Or is `STAKE_POOL_PROVIDER` no longer in use?

rhyslbw: There's still a stub, but it will be removed as part of @AngelCastilloB's work. He'll update this example within ADP-2055, which is his current WIP. He has confirmed this refactor is working on his local branch using the local network too BTW

ssong-iohk: If dependent work is WIP, I think below stub related parameters should be remained as it will make the tests are failed when running against `testnet`.

ssong-iohk (approved):
Thanks @AngelCastilloB

packages/e2e/env/.env.wallet
rhyslbw: Still `stub` when using Blockfrost, but this will be parked soon, so testing here is limited to the `cardano-services` providers. @AngelCastilloB will be updating the examples in his branch.

PR #370, opened 2022-08-02
feat: Resolve poor performance of endpoints in "stake-pool"

IvayloAndonov:
# Context

Similar to `network-info` related query, it shows massive slowness when getting the response of:
- http://localhost:3000/stake-pool/search 
- http://localhost:3000/stake-pool/stats 

when using `testnet` DB regardless of applying “sort” and “filters”.
# Proposed Solution
Introduced caching strategy:
- cache each sub query individually depends on its cache invalidation indicator (TTL or epoch rollover detected by polling)
- store all cached core stake pool objects in `fromCache` map as an optimized way to store pools data if already fetched from a previous query, which reduces drastically re-fetching of the same pools across queries with different arguments (filters/sort/pagination). We cache the mapped core stake pool object ( type `Cardano.StakePool` )
- introduce `idsToFetch` map which holds all pools that need to be fetched from DB source
- wrap all sub-queries related with extra stake pool info  within `queryExtraPoolsData` which processes the `idsToFetch` map and ensures all pools of extra data are being provided
- rename and extend `toCoreStakePool` mapping to `toStakePoolResults`, now it returns composed result type: `{results, poolsToCache}`,  as `poolsToCache` is being cached afterward, core type [StakePoolSearchResults](https://github.com/input-output-hk/cardano-js-sdk/blob/22e55db3bdd45f401bc30dae27a06c841870cec5/packages/core/src/Provider/StakePoolProvider/types/StakePoolProvider.ts#L55) is not touched
- clear the whole in-memory cache instance on epoch rollover
- cache keys are composed using string literals, query cache keys are generated by `namespace` + `query name` + `options` which ensure the uniqness of the key, for single stake pool cache key is used `namespace` + `hashId`
- `stakePoolStats` is cached with default TTL value (2 hours)

- rework `EpochPollService` as a singleton class, used across services
- extend `StakePoolProvider` with `cache` and `epochPollService`
- refactor sub queries and optimize caching of pools


packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
```diff
@@ -52,10 +50,10 @@ export class DbSyncNetworkInfoProvider extends DbSyncProvider implements Network
     super(db);
     this.#logger = logger;
     this.#cache = cache;
+    this.#cardanoNode = cardanoNode;
     this.#builder = new NetworkInfoBuilder(db, logger);
     this.#genesisDataReady = loadGenesisData(cardanoNodeConfigPath);
-    this.#epochPollInterval = epochPollInterval;
-    this.#cardanoNode = cardanoNode;
+    this.#epochPollService = EpochPollService.create(cache, () => this.#builder.queryLatestEpoch(), epochPollInterval);
```
mkazlauskas, line 56: It would be better to inject this as part of `NetworkInfoProviderDependencies`

packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
```diff
@@ -115,19 +113,13 @@ export class DbSyncNetworkInfoProvider extends DbSyncProvider implements Network
   }
 
   async start(): Promise<void> {
+    this.#epochPollService.start();
```
mkazlauskas, line 116: If it's injected as per previous comment, then it should be started/shutdown where it's created instead of here

```suggestion
```

packages/cardano-services/src/util/polling/EpochPollService.ts
```diff
@@ -0,0 +1,64 @@
...
+      this.#cache.set<number>(CURRENT_EPOCH_CACHE_KEY, lastEpoch, UNLIMITED_CACHE_TTL);
+    }
+  }
+
+  /**
+   * Creates a single instance of EpochPollService
+   *
+   * @returns {EpochPollService} EpochPollService instance
+   */
+  static create(cache: InMemoryCache, asyncAction: AsyncAction<number>, interval: number): EpochPollService {
```
mkazlauskas, line 42: Since it takes in an asyncAction and interval, singleton doesn't make much sense here. It should at least group instances by `asyncAction`. I suggest to not do singleton and just inject the same instance as dependency to every dependent.

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePool.ts
```diff
@@ -1,21 +1,48 @@
...
+  constructor(
+    { epochPollInterval }: StakePoolProviderProps,
+    { db, cache, logger = dummyLogger }: StakePoolProviderDependencies
+  ) {
     super(db);
-    this.#builder = new StakePoolBuilder(db, logger);
     this.#logger = logger;
+    this.#cache = cache;
+    this.#builder = new StakePoolBuilder(db, logger);
+    this.#epochPollService = EpochPollService.create(cache, () => this.#builder.getLastEpoch(), epochPollInterval);
```
mkazlauskas, line 45: This should also be injected as part of `StakePoolProviderDependencies`

mkazlauskas (changes requested):
Would be great to run all DbSyncStakePool tests twice and assert that we get the same result as on the 1st call when it comes from cache

packages/cardano-services/src/util/polling/EpochPollService.ts
IvayloAndonov: The `asyncAction` was one of the providers' builder queries, which at the point of poll service `create` the builder was not yet created. The idea was each of the providers to know about its asyncAction only, but since it's a common query then could be moved into epoch poll service scope/directory and passed during poll service init/start in `loadHttpServer.ts`

packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
IvayloAndonov: I think we should keep the service's load within [serviceMapFactory](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services/src/Program/loadHttpServer.ts#L56) pure, without sending DB operations (same reason that we removed `healthChecks` on service init), so probably keeping `this.#epochPollService.start()` invoked by [service startImpl()](https://github.com/input-output-hk/cardano-js-sdk/blob/cc82bc27539e3ff07f7c2d5816fa7e70c32d06ac/packages/cardano-services/src/Http/HttpService.ts#L41-L43) after initialization is fine?

packages/cardano-services/src/util/polling/EpochPollService.ts
IvayloAndonov: Another thing is that `db` could be `undefined` at point of `epochPollService` initialization if we create one instance in [loadHttpServer](https://github.com/input-output-hk/cardano-js-sdk/blob/419252a9082b710ad42012dd4c01cc397de8e658/packages/cardano-services/src/Program/loadHttpServer.ts#L134) and pass it down to every dependant. So we might want to benefit from `withDb()` decorator and instantiate epoch service within each dependant service map [StakePool](https://github.com/input-output-hk/cardano-js-sdk/blob/419252a9082b710ad42012dd4c01cc397de8e658/packages/cardano-services/src/Program/loadHttpServer.ts#L87-L93) and [NetworkInfo](https://github.com/input-output-hk/cardano-js-sdk/blob/419252a9082b710ad42012dd4c01cc397de8e658/packages/cardano-services/src/Program/loadHttpServer.ts#L107-L122) but then the singleton helps in that case? Any thoughts @mkazlauskas

packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
IvayloAndonov: https://github.com/input-output-hk/cardano-js-sdk/pull/370/commits/7c8cd1f53bc0bfecbb7fe3af6c33fa0b409288f6

packages/cardano-services/src/util/polling/EpochPollService.ts
IvayloAndonov: The changes are applied under https://github.com/input-output-hk/cardano-js-sdk/pull/370/commits/7c8cd1f53bc0bfecbb7fe3af6c33fa0b409288f6 , as we agreed - use `memoize` for single instance creation of `DbSyncEpochPollService` on each dependant service load method.

IvayloAndonov:
Rebased with _master_, fixup commits are squashed, ready for reviews and QA testing.

packages/cardano-services/src/util/polling/DbSyncEpochPollService.ts
```diff
@@ -0,0 +1,62 @@
...
+import { Pool, QueryResult } from 'pg';
+import { findLastEpoch } from './queries';
+
+export const EPOCH_POLL_INTERVAL_DEFAULT = 10_000;
+export const CURRENT_EPOCH_CACHE_KEY = 'current_epoch';
+
+/**
+ * Class to handle epoch rollover through db polling
+ */
+export class DbSyncEpochPollService {
```
mkazlauskas, line 12: Would be nice to also export an interface that dependents can depend on instead of this specific class. Maybe move cache clearing logic out of it too? Otherwise it should be callled EpochRolloverCacheCleaner or sth. How about
```suggestion
export interface EpochMonitor {
  onEpochRollover(callback: (epoch: Cardano.Epoch) => void));
}
export class DbSyncEpochPollService implements EpochMonitor {
```

Or just `Observable<Cardano.Epoch>`, but cardano-services does not depend on rxjs so maybe not the best idea to add this dependency just for this :sweat_smile: 

Thoughts @rhyslbw @IvayloAndonov ?

mkazlauskas (commented):
Looks pretty good @IvayloAndonov . Just 1 idea below to change the concept of EpochPollService a little bit (to be discussed).

packages/cardano-services/src/util/polling/DbSyncEpochPollService.ts
rhyslbw: I like the interface idea as a way to complete this initial feature. I think we should be looking at Observerable patterns in `cardano-services` too, and this is a good candidate, but it would be best to assess the whole design and plan it out as a refactor when we have capacity.

packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
rhyslbw: Agree. It's reasonable for the `NetworkInfoProvider` to maintain the lifecycle of this service.

mkazlauskas: If NetworkInfoProvider controls the lifecycle, then I think it should be the one creating the service, which is not ideal as it can logically be a separate service.

How about **not** having an explicit start & shutdown, but instead polling if there is at least 1 callback registered? Would then also need a way to unsubscribe, like:
```ts
export type Unsubscribe = () => void;
export interface EpochMonitor {
  onEpochRollover(callback: (epoch: Cardano.Epoch) => void)): Unsubscribe;
}
```

rhyslbw: OK, yeah, that's a nice solution. Each consumer can then have control of the monitor's function within it's lifecycle without impacting the others. Could also be:

```ts
export type Disposer = () => void;
```
... but the pattern is the important aspect. Nice one @mkazlauskas

mkazlauskas: If there's an intent of introducing rxjs then we could use a type that's compatible with [Unsubscribable](https://github.com/ReactiveX/rxjs/blob/5d0e41ee46337f56dba6d9a1ef938002b261ccac/src/internal/types.ts#L65)

packages/cardano-services/src/util/polling/DbSyncEpochPollService.ts
IvayloAndonov: `EpochMonitor` interface as dependency is a good suggestion, but I'm not so competent on the idea of Observable usage with `rxjs` though

mkazlauskas: Let's keep rxjs out of it for now as @rhyslbw suggested, let's just apply the pattern with callbacks

IvayloAndonov: Callbacks patter is applied under https://github.com/input-output-hk/cardano-js-sdk/pull/370/commits/9a6e57e6440feb8df6d8ccaaeaa681c64f5da672 @mkazlauskas

packages/cardano-services/src/util/polling/types.ts
```diff
@@ -0,0 +1,6 @@
+export type Disposer = () => void;
+
+export interface EpochMonitor {
+  onEpochRollover(callback: () => void): Disposer;
+  getEpoch(): Promise<number>;
```
mkazlauskas, line 5: Maybe let's remove this from the interface (keeping it on DbSyncEpochPollService), because it's quite confusing as it's implemented now. I would expect this to return current epoch (make a query) instead of return last known cached epoch.

```suggestion
```

packages/cardano-services/src/util/polling/DbSyncEpochPollService.ts
```diff
@@ -0,0 +1,99 @@
...
+      }
+    }
+  }
+
+  /**
+   * Gets the last epoch number stored in db
+   *
+   * @returns {number} epoch number
+   */
+  async #getLastEpoch() {
```
mkazlauskas, line 59:
```suggestion
  async #queryLastEpoch() {
```

packages/cardano-services/src/util/polling/DbSyncEpochPollService.ts
```diff
@@ -0,0 +1,99 @@
...
+    return () => {
+      this.#callbacks.splice(this.#callbacks.indexOf(cb), 1);
+      if (this.#callbacks.length === 0) this.#shutdown();
+    };
+  }
+
+  /**
+   * Get current epoch
+   */
+  async getEpoch() {
```
mkazlauskas, line 96: Since this is a public method let's make the name more descriptive. Also, `this.#currentEpoch` could be `undefined` if this is called before registering any callback - let's just resolve `null` in this case.

```suggestion
  async getLastKnownEpoch(): Promise<Cardano.Epoch | null> {
```

packages/cardano-services/test/util/polling/DbSyncEpochPollService.test.ts
```diff
@@ -0,0 +1,59 @@
...
+          ['id', 'out_sum', 'fees', 'tx_count', 'blk_count', 'no', 'start_time', 'end_time'],
+          [greaterEpoch, 58_389_393_484_858, 43_424_552, 55_666, 10_000, greaterEpoch, '2022-05-28', '2022-06-02']
+        );
+
+        await sleep(epochPollInterval * 2);
+
+        expect(firstRegisteredCallback).toHaveBeenCalled();
+        expect(secondRegisteredCallback).toHaveBeenCalled();
+        expect(await epochMonitor.getEpoch()).toEqual(greaterEpoch);
+        expect(cache.keys().length).toEqual(0);
```
mkazlauskas, line 51: This test doesn't seem to be using `cache` for anything

```suggestion
```

mkazlauskas (changes requested):
Can we also update stake pool service tests to assert that 2nd request (hitting the cache) returns the same response? For example given [this](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services/test/StakePool/StakePoolHttpService.test.ts#L166):

Add this to the test:
```ts
const cachedResponseWithPagination = await provider.queryStakePools(reqWithPagination);
expect(cachedResponseWithPagination).toEqual(responseWithPagination)
```
(do this for all existing stake pool tests that have successful response)

packages/cardano-services/src/util/polling/types.ts
mkazlauskas: No, I don't think it should send an actual query, as that will affect the tests. I think it just needs to be renamed as per [comment below](https://github.com/input-output-hk/cardano-js-sdk/pull/370#discussion_r942383745)

packages/cardano-services/test/util/polling/DbSyncEpochPollService.test.ts
IvayloAndonov: All the feedback is applied under https://github.com/input-output-hk/cardano-js-sdk/pull/370/commits/9c2bc2f0db9a36677df711c00f6dd5daf3a90a1a fixup commit @mkazlauskas

mkazlauskas (dismissed):
Nice work @IvayloAndonov :tada: Please squash and rebase on master, then request a review from @iadmytro and @iccicci

IvayloAndonov:
Squashed and rebased on master, tagged the relevant people as reviewers.

mkazlauskas (approved):

iccicci (approved):
I'm not so expert on the topic of this change... but I found nothing not respecting our standards or general good practices.
GJ!

iadmytro (approved):

PR #371, opened 2022-08-02
feat(e2e): add artillery to perform stress tests

iccicci:
# Context

We need to integrate artillery.io into our SDK to perform stress tests.

# Proposed Solution

Added a e2e test mainly with the purpose of artillery.io POC.

From [this line](https://github.com/input-output-hk/cardano-js-sdk/blob/449e892322407b49cb00b44f2f7889f87f6d3145/packages/e2e/package.json#L8):

- `WORKERS=1` starts only one worker: details will follow;
- `node -r ts-node/register ../../node_modules/.bin/artillery` running `artillery` this way `node-ts` does the magic: we can `require` a `.ts` file which imports the SDK without the need to rebuild nothing.

In [`artillery.ts`](https://github.com/input-output-hk/cardano-js-sdk/blob/449e892322407b49cb00b44f2f7889f87f6d3145/packages/e2e/artillery/artillery.ts) are defined some types to make more comfortable the custom hooks development.

Probably this is not a best practice, but I think this is something important to know.
In the custom hooks file I used a [global variable](https://github.com/input-output-hk/cardano-js-sdk/blob/449e892322407b49cb00b44f2f7889f87f6d3145/packages/e2e/artillery/StakePoolSearch.ts#L11), since artillery.io runs more workers, the global variables are not shared between the workers, so neither between hooks executions. In order to make the global variable shared between all the hooks executions I had to specify to run only one worker.


packages/e2e/artillery/StakePoolSearch.yml
```diff
@@ -0,0 +1,25 @@
+config:
+  target: 'http://localhost:4000'
```
rhyslbw, line 2: For other reviewers, there's a CLI override argument: https://www.artillery.io/docs/guides/guides/command-line#override-the-target-url

packages/e2e/artillery/artillery.ts
```diff
@@ -0,0 +1,16 @@
+export interface Context<T> {
+  _uid: string;
+  vars: T;
+}
+
+export type AfterResponse<T> = (
+  params: unknown,
```
rhyslbw, line 7: Have a look at how the `@types/artillery` package for inspiration:
https://www.npmjs.com/package/@types/artillery

packages/e2e/package.json
```diff
@@ -5,6 +5,7 @@
   "description": "End to end tests for the cardano-js-sdk packages.",
   "main": "index.js",
   "scripts": {
+    "artillery:pool": "WORKERS=1 node -r ts-node/register ../../node_modules/.bin/artillery run artillery/StakePoolSearch.yml",
```
rhyslbw, line 8: :sunglasses: We may also need to compile this to run distributed tests, but it's a nice solution for the current scope.

rhyslbw (commented):
I'd like to run this myself before providing a more detailed review, once we have sorted the performance issue making this a bit impractical to run at any real scale. I've left some thoughts to consider.

packages/e2e/artillery/StakePoolSearch.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  if (!options) {
+    options = { pagination: { limit: PAGE_SIZE, startAt: 0 } };
+    ctx.vars.args = [options];
+  }
+
+  if (ctx.vars.pageResults) options.pagination!.startAt += PAGE_SIZE;
+  else if (!isFirstUser) {
+    // TODO randomization of parameters
```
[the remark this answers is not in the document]
mkazlauskas: Problem with this is that then we won't be measuring how long the 1st fetch takes (no cache).

Ideally we would use a freshly started server for running artillery, so yeah seems like it'd be best to do outside of artillery.

packages/e2e/artillery/artillery.ts
mkazlauskas: Why don't we use @types/artillery directly?

packages/e2e/package.json
iccicci: When we need to compile, no problem: we are free to use a `js` file, it is supported.

The _only_ reason for this is for development: while writing a new test, the developer don't need to build the SDK **at each** run attempt of its test.

This will save a lot of team time while writing artillery tests.

packages/e2e/artillery/artillery.ts
iccicci: It is useless (it actually defines nothing about the callbacks) and inefficient (the only _half useful_ defined type is `ScenarioContext`, but it do not includes `_uid` and `vars` is of an hard coded type).

Please take a look to [its source code](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/artillery/index.d.ts) and you'll agree this file defines much more useful and correct stuff.

packages/e2e/artillery/StakePoolSearch.yml
iccicci: For me too! I didn't noticed it, thank you @rhyslbw

packages/e2e/artillery/StakePoolSearch.ts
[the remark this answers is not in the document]
iccicci: I think we need to keep separated the two topics:

1. the logic of a (each) test
2. to demonstrate the POC is able to do all what we need

actually, the ability to run on any network is a requirement I didn't thought, thank you @rhyslbw .

I was thinking something like extracting a set of pool names and other attributes (in test development stage), insert them somehow in the test and use them to generate randomized search parameters; by this way (as @mkazlauskas said) we preserve the ability to run the test against a freshly started server.

But I think this is part of test logic, that's the reason why didn't that yet.

While about demonstrating the POC is able to run on any network; for those tests which need a subset of data from the network in order to generate randomized inputs I think a good solution could be to extract the subset of data from mainnet and testnet and making the test to choose which set to use depending on a simple configuration environment variable. I'm going to write something on this...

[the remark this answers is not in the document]
rhyslbw: > I think a good solution could be to extract the subset of data from mainnet and testnet and making the test to choose which set to use depending on a simple configuration environment variable

I can see the attraction for taking this approach, but it's not a complete full demonstration of how this prototype will run on **any** network, which is a critical aspect here. Please remember the _testnet_ will be destroyed soon, so we're actively working toward removing all dependency on it explicitly, however short-lived this inclusion will be.

packages/e2e/artillery/artillery.ts
rhyslbw: It's also [outdated](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/9f9e20f594027285c597a96af85f913c61ba915b/types/artillery/index.d.ts#L1) (we're using v2), but my comment was more related to seeing how the request and response are typed using the Node.js built-in libs, which could be an improvement over the `unknown` types. Also, the `params` could be typed with a generic, i.e `AfterResponse<T, P>`

packages/e2e/artillery/artillery.ts
```diff
@@ -0,0 +1,16 @@
...
+  response: unknown,
+  ctx: Context<T>,
+  ee: unknown,
```
rhyslbw, line 10: What's `ee`?

iccicci: If read correctly the artillery source code should be some event emitter... but I didn't found what it actually is.
The callbacks receive it as an argument, probably I didn't investigated its meaning/purpose as I was able to catch the POC target without using it.

packages/e2e/artillery/artillery.ts
iccicci: Yep, I agree; the point is that (as far I was able to find on internet) there is no native built-in types for artillery callbacks.

About changing `unknown` into something specific; I think that spending time now to do this, is not a good time investment: we risk to spend time to discover something we will never use; what we have now let us to intercept calls and to control the flow.

In the case (in my opinion unlikely) in which we need to write an artillery tests which needs something else: we will need to study the _new feature_ and that will be the right occasion to enrich the types I wrote till now.

packages/e2e/artillery/StakePoolSearch.ts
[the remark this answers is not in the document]
iccicci: The only way coming to my mind which simultaneously respect:

- do not produces cache data while preparing the run against a freshly started server
- can run against any network

is something similar to:

- starting with an empty set of data
- populate the requests with randomized data from the set (empty at the beginning of the test)
- gather useful data from the calls response to populate the set of data

... but "defining the data set", "randomizing data from the set to perform new requests" and "gathering data from the responses" is much more close to test logic than to demonstrate the POC capabilities.

If you agree I can write a _basic_ algorithm (i.e. which gather only some data, not all possible data) mainly with the purpose to demonstrate this is a viable solution.

packages/e2e/artillery/artillery.ts
mkazlauskas: I think that's reasonable as long as we list all fields, so devs can see that they exist and can then improve the types if needed

packages/e2e/artillery/artillery.ts
iccicci: I confirm, it is the `EventEmitter` and it's used to collect custom metrics (POC added).

packages/e2e/artillery/artillery.ts
iccicci: Added all the fields with a description (where I was able to find... unknown otherwise)

packages/e2e/artillery/StakePoolSearch.ts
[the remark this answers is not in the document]
iccicci: Hi @rhyslbw and @mkazlauskas ;

I changed the POC to start with empty query parameters, then it will start adding parameters when it has available data.

rhyslbw (changes requested):

packages/e2e/artillery/StakePoolSearch.ts
```diff
@@ -0,0 +1,94 @@
...
+    ctx.vars.args = [options];
+  }
+
+  // If we have a pageResults, this is not the first call of the session: request next page
+  if (ctx.vars.pageResults) options.pagination!.startAt += PAGE_SIZE;
+  // If this is the first call of the session: randomize query parameters
+  else {
+    const namesList = Object.keys(names);
+
+    // If this is the first virtual user, we need to make a query that will fully prime the API
```
rhyslbw, line 51:
```suggestion
    // If this is the first virtual user, we need to make a query that will fully prime the API cache
```

image

rhyslbw (changes requested):
Looks good @iccicci, just a minor issue with the previously suggested change implementation. You can commit that, rebase, and squash prior to additional reviews.

iadmytro (approved):

packages/e2e/test/artillery/StakePoolSearch.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  // Send custom metrics
+  ee.emit('histogram', `${getStatName(ctx.vars)}.time`, Date.now() - ctx.vars.startedAt);
+
+  done();
+};
+
+export const before: BeforeRequest<StakePoolSearchVars> = async (_params, ctx, ee, done) => {
+  let options = ctx.vars.args?.[0];
```
IvayloAndonov, line 40: The arguments structure of our providers is changed from `args:[arg1, arg2]` to `{ arg1, arg2}`
PR for a reference: https://github.com/input-output-hk/cardano-js-sdk/pull/393 @iccicci

iccicci:
Hi @rhyslbw , @mkazlauskas , @iadmytro and @IvayloAndonov ;

this should be ready to be reviewed.
Since the test is run through **artillery** rather than **jest**,  I wasn't able to load the environment from the `.env` file, so to run the tests currently following command is required:

```
STAKE_POOL_PROVIDER_URL="http://localhost:4000/stake-pool" yarn artillery:stake-pool-query
```

or change the target to perform the tests against a different system from our local network.

packages/e2e/test/artillery/StakePoolSearch.yml
```diff
@@ -0,0 +1,20 @@
+config:
+  target: 'http://localhost:4000'
```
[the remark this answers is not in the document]
iccicci: Thank you @IvayloAndonov .

About the `target` attribute:
- it is actually a refuse, after the today refactoring it is no longer used (need to test if artillery works without it or it is a required field)
- I agree: if required, any solution is better than an hard coded value!

About the 2nd proposal, the initial check: I'll do that.

packages/e2e/artillery/StakePoolSearch.ts
iccicci: f104f574b should fix this

packages/e2e/test/artillery/StakePoolSearch.ts
iccicci: 10e1a9b97 should fix this

packages/e2e/test/artillery/StakePoolSearch.yml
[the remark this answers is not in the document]
iccicci: Hi @IvayloAndonov , @rhyslbw and @iadmytro ,

441952f06 should solve this, but please let me give you some clarifications.

`target` is a required field in artillery configuration: I simply used `http://localhost` as a placeholder.

I thought the most appropriate method to perform the check you suggested is actually using the `healthCheck` method of the `provider`. I searched a lot but it seems artillery doesn't offer a _phase level hook_, it offers only _scenario level hooks_. Please look at [this request](https://github.com/artilleryio/artillery/issues/131).

In artillery, phases are executed only once and generates several virtual users; scenarios are executed once for each generated virtual user.

I had two options:
1. to write a script to perform the check externally from artillery and running this one just in case
2. perform the `healthCheck` at the beginning of the scenario

I chosen the second one for more than one reason: robustness, cheaper to be written, makes the test flow more compliant with our sdk use pattern...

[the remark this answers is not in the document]
IvayloAndonov: The second option sounds good to me as well. Just some extra explanation about the scenario definition in [StakePoolSearch.yml](https://github.com/input-output-hk/cardano-js-sdk/pull/371/files#diff-2a2837d6babf0035796a40acb10cad57a8145407840d9870c9799294fd055a27) would be good (how the nested 2 loops work) but this is not a requested change by my side. Thanks for the clarifications! @iccicci

IvayloAndonov (approved):
Great work!

packages/e2e/test/artillery/StakePoolSearch.yml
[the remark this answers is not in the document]
iccicci: Hi @IvayloAndonov ,
defaf32ae should answer to your question.

rhyslbw:
Please rebase and squash in preparation for final review @iccicci

iccicci:
Hi @rhyslbw and @iadmytro ;

I rebased this with artillery bump to v2

IvayloAndonov (dismissed):

packages/e2e/test/artillery/StakePoolSearch.ts
```diff
@@ -0,0 +1,155 @@
...
+
+const env = envalid.cleanEnv(process.env, { STAKE_POOL_PROVIDER_URL: envalid.str() });
+const provider = stakePoolHttpProvider({ baseUrl: env.STAKE_POOL_PROVIDER_URL, logger });
+
+const PAGE_SIZE = 20;
+
+/**
+ * The set of pools id found while getting results
+ */
+const poolsId: Cardano.PoolId[] = [];
```
mkazlauskas, line 38:
```suggestion
const poolIds: Cardano.PoolId[] = [];
```

packages/e2e/test/artillery/StakePoolSearch.ts
```diff
@@ -0,0 +1,155 @@
...
+    args = { filters: {}, pagination: { limit: PAGE_SIZE, startAt: 0 } };
+    vars.args = args;
+  }
+
+  // If we have a pageResults, this is not the first call of the session: request next page
+  if (pageResults) args.pagination!.startAt += PAGE_SIZE;
+  // This is the first call of the first virtual user, we need to make a query that will fully prime the API cache
+  else if (firstUser === '') firstUser = _uid;
+  // This is the first call of the session for all users but the first one: randomize query parameters
+  else randomizeFilter(args);
```
mkazlauskas, line 124:
```suggestion
  else {
    args.filter = randomizeFilter(args);
  }
```

packages/e2e/test/artillery/StakePoolSearch.ts
```diff
@@ -0,0 +1,155 @@
...
+  // If not already initialized: prepare the new request starting from the first page
+  if (!args) {
+    args = { filters: {}, pagination: { limit: PAGE_SIZE, startAt: 0 } };
+    vars.args = args;
+  }
+
+  // If we have a pageResults, this is not the first call of the session: request next page
+  if (pageResults) args.pagination!.startAt += PAGE_SIZE;
+  // This is the first call of the first virtual user, we need to make a query that will fully prime the API cache
+  else if (firstUser === '') firstUser = _uid;
```
mkazlauskas, line 122:
```suggestion
  else if (!firstUser) firstUser = _uid;
```

packages/e2e/test/artillery/StakePoolSearch.ts
```diff
@@ -0,0 +1,155 @@
...
+   * `performQuery()` stores here if it gets an error from the query to let `moreResults()` knows when exit on error
+   */
+  gotError?: boolean;
+
+  /**
+   * Tricky array which contains `1` element when the healthCheck is ok, `0` elements otherwise.
+   *
+   * The most _artillery native_ found way to implement logic `if(healthCheck is ok) perform the test;`
+   */
+  healthCheckArray: number[];
```
mkazlauskas, line 27:
```suggestion
   * The most _artillery native_ found way to implement logic `if(healthCheck is ok) perform the test;`
   */
  emptyWhenOver: number[];
```

This is quite odd. Are there alternatives?

mkazlauskas (dismissed):
Looking good! Just a few suggestions on style

packages/e2e/test/artillery/StakePoolSearch.ts
rhyslbw: Our lint rules should probably reflect this

mkazlauskas: It's about procedural vs functional, not {}

packages/e2e/test/artillery/StakePoolSearch.ts
iccicci: 418332fb7 should solve this

packages/e2e/test/artillery/StakePoolSearch.ts
iccicci: 418332fb7 should solve this

packages/e2e/test/artillery/StakePoolSearch.ts
iccicci: Sorry @mkazlauskas ,

if this is for a matter of readability, simply adding a `return args.filter` at the end of `randomizeFilter` (which results in a useless assignment) would be enough.
If the core of the request is to change `randomizeFilter` in order to make a copy of `args.filter`, modifying it rather than the original, and returning it; I think the amount of the additional resources cost from the change would not worth.

mkazlauskas: It's about modifying the argument object. In general treating arguments as read-only makes it easier to reason about a function: arguments=inputs, return=outputs. I think it is a good practice to use this functional style instead of procedural.

> I think the amount of the additional resources cost from the change would not worth.
Why would it use significantly more resources? It can simply return `typeof args.filters` (there's probably an interface for that). You could then just do `{ filters: randomizeFilter(poolIds) }` [here](https://github.com/input-output-hk/cardano-js-sdk/pull/371/files#diff-f69af2c64d445d76a60faebdd225486f0997596e8ac518387fa57d2e237ad898R115)

packages/e2e/test/artillery/StakePoolSearch.ts
iccicci: Hi @mkazlauskas ,

this was mainly an attempt to make a most _artillery native implementation_ as possible.

A possible alternative could be:
- to store the result of the health check in a `boolean` (rather than in a `number[]`)
- to check the result of the health check at the beginning of `performQuery` and completely skipping the query in case of provider unhealthy

This would make the artillery yaml pseudocode more readable

from:
```
scenarios:
  - name: 'Browse'
    flow:
      - function: 'healthCheck'
      - loop:
          - loop:
              - function: 'performQuery'
            whileTrue: 'moreResults'
        over: 'healthCheckArray'
```

to:
```
scenarios:
  - name: 'Browse'
    flow:
      - function: 'healthCheck'
      - loop:
          - function: 'performQuery'
        whileTrue: 'moreResults'
```

(more or less...)

mkazlauskas: If we could still use `over:` with
> to store the result of the health check in a boolean (rather than in a number[])

then I think that would be easier to understand and also to name the property :sweat_smile:

packages/e2e/test/artillery/StakePoolSearch.ts
iccicci: 1185fc810 should solve this

packages/e2e/test/artillery/StakePoolSearch.ts
iccicci: 1185fc810 should solve this

packages/e2e/test/artillery/StakePoolSearch.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  const { identifier, status } = vars.args.filters!;
+  const nameStatId = identifier ? 'id' : '__'; // ATM id is the only identifier used by the test
+  const statusStatName = status && status.length > 0 ? 'status' : '______';
+
+  return `query.${nameStatId}_${statusStatName}`;
+};
+
+const randomizeFilter = () => {
+  const filters: Exclude<QueryStakePoolsArgs['filters'], undefined> = {};
```
mkazlauskas, line 63: There is a type utility for that:
```suggestion
  const filters: NonNullable<QueryStakePoolsArgs['filters']> = {};
```

mkazlauskas (dismissed):
Nice work! :rocket:

packages/e2e/test/artillery/StakePoolSearch.ts
iccicci: Thanks for the suggestion @mkazlauskas ! Applied with fc5eda72e

rhyslbw (approved):
Fantastic work @iccicci :muscle:

PR #374, opened 2022-08-04
test(e2e): fixed cardano services load test compilation error

AngelCastilloB:
# Context

Currently Cardano services load tests are broken on master.

# Proposed Solution

Apply minor fixed to make the module compile again.

None


rhyslbw (approved):
I can confirm the test is passing when running locally with Blockfrost providers for the supplemental data

PR #375, opened 2022-08-04
Refactor/adp 1818 context aware logger

mirceahasegan:
# Context

Allow adding context to log messages.
Nested contexts should be possible too (e.g. logs in a function, from a module, from a package).

# Proposed Solution

/**
 * Prepends context to each log entry
 *
 * @param {Logger} logger To which logger we are adding a context. Could be a simple logger or
 * another context logger, in which case the new context is added along with the parent one.
 * @param {string} context string that is prepended to the log messages. It is concatenated
 * with the parent context if any.
 * @param contextStringify Optional function to override how the context is stringified.
 * It takes as parameter an array of contexts and returns a concatenated string.
 * Default contextStringify example: ['ctx1', 'ctx'2] becomes "[ctx1|ctx2]".
 * @returns A logger object that prepends the context when logging
 *
 * Example: single context, default formatting
 *   - const oneContextLogger = contextLogger('ctx1', rawLogger);
 *   - oneContextLogger.debug('hi') calls `rawLogger('[ctx1]', hi)
 *
 * Example: nested context, default formatting
 *   - const twoContextLogger = contextLogger('ctx2', oneContextLogger);
 *   - twoContextLogger.debug('hi') calls `rawLogger('[ctx1|ctx2]', hi)
 *
 * Example: nested context with custom formatter
 *   - const mergeLogger = contextLogger('ctx2', oneContextLogger, (ctx) => `(${ctx.join('&')}) =>`));
 *   - mergeLogger.debug('hi') calls `rawLogger('(ctx1&ctx2) =>', 'hi')
 */


packages/util/src/logging.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+export const contextLogger = (logger: Logger, context: string): Logger => ({
+  context: [...(logger.context || []), context],
+  ...(<Logger>(<unknown>Object.fromEntries(
+    loggerMethodNames.map<[keyof Logger, Logger['info']]>((method) => [
+      method,
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (message?: any, ...optionalParams: any[]) => {
+        const contextArg = { context: [...(logger.context || []), context] };
+        if (typeof message === 'object') {
```
iccicci, line 29: In case of structured log, this is interesting! But in case of human readable log this will change the line layout depending on the type of the first logged argument making the whole log confused and harder to read.

I would remove this condition and I would only use the `logger[method](contextArg, message, ...optionalParams)` form (the `return` statement is not required here).

iccicci (commented):
Great!

Just a couple of considerations...

packages/util/src/logging.ts
```diff
@@ -0,0 +1,36 @@
...
+ *   - mergeLogger.debug({info: 'Superb'}, 'hi') calls `rawLogger({context: ['ctx1', 'ctx2'], moreInfo: 'Superb'}, hi)
+ */
+export const contextLogger = (logger: Logger, context: string): Logger => ({
+  context: [...(logger.context || []), context],
+  ...(<Logger>(<unknown>Object.fromEntries(
+    loggerMethodNames.map<[keyof Logger, Logger['info']]>((method) => [
+      method,
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (message?: any, ...optionalParams: any[]) => {
+        const contextArg = { context: [...(logger.context || []), context] };
```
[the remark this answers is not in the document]
mirceahasegan: I think you're right about the formatted form.
I am trying to figure out how it will work with [bunyan APi](https://github.com/trentm/node-bunyan#log-method-api), which allows providing a "fields" object as first arg, which is merged into the log record (e.g. see Error special case).
If we add a formatted context string as first arg, we shadow this functionality to users wanting to derive from `bunyan`

packages/util/src/logging.ts
mirceahasegan: Same comment here, regarding the special meaning of object type first arg in `bunyan`

iccicci: Didn't thought to that feature... probably because I don't like it... it's just a choice...

Anyway the current implementation replaces the first object with something else loosing both the type and the not own properties of the logged object.
That will prevent the ability to distinguish between an `Error` a `TypeError` or a `ProviderError` and moreover the `message` and `stack` properties.

This should not happen regardless from the choice.

mirceahasegan: Actually it doesn't lose it. It merges the context object with the first arg, when the first arg is an object. But I do agree to not liking it :)

mirceahasegan:
@iccicci, I have added a second implementation (and unit tests), that formats the context.
Let me know if it looks better

packages/util/src/logging.ts
mkazlauskas: > But in case of human readable log this will change the line layout depending on the type of the first logged argument making the whole log confused and harder to read.

Good point - would be most convenient to have context readable in a single line.

> If we add a formatted context string as first arg, we shadow this functionality to users wanting to derive from bunyan

I think it's fine, unless we want to use this bunyan feature internally. If we do, then we can extend `contextLogger` function to support it.

mkazlauskas (approved):
Nice work @mirceahasegan . I'm leaning towards keeping it simple and not supporting bunyan context object, but let's discuss it with the team

packages/util/src/logging.ts
```diff
@@ -0,0 +1,60 @@
...
+        if (typeof message === 'object') {
+          return logger[method]({ ...contextArg, ...message }, ...optionalParams);
+        }
+        return logger[method](contextArg, message, ...optionalParams);
+      }
+    ])
+  )))
+});
+
+export const contextLogger2 = (
```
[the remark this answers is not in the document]
mkazlauskas: I think this is a variant to consider, we need to decide which one we want to use. I like this one better, but it's not compatible with some `bunyan` functionality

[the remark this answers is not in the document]
mirceahasegan: You are correct @mkazlauskas, the `2` is just to show it's a variant. Only one of them will stay.
I am 100% sure we should go with the second one proposed by @iccicci . It's cleaner, makes the logs more readable. The first one just helps in exposing a rather obscure feature of `bunyan`, so it's sacrificing a lot.
I don't think there's the need for a team debate, as the second options is much better.

packages/util/src/logging.ts
iccicci: Even if the first implementation was really nice as well, I think the #2 version is closer to what we need.
GJ @mirceahasegan !

packages/util/src/logging.ts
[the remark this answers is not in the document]
mirceahasegan: Code cleanup done. Added docs and examples ([5d6de03](https://github.com/input-output-hk/cardano-js-sdk/pull/375/commits/5d6de03fe81c2d214b7659777d8a2f5a7b30ce2d)).
After review, I will squash everything in a single commit.

mkazlauskas (dismissed):
:rocket:

iccicci (approved):

rhyslbw (dismissed):
Ready for squashing

rhyslbw (approved):

PR #376, opened 2022-08-04
Replace live stake DB query with Local State Query 

jvieiro:
# Context

Live stake is the current stake that will be used to determine the block leadership schedule of the epoch following the next, and can potentially change with each new chain tip.  Active stake is the stake snapshotted two epochs prior and is now playing a role in block production. Querying the cardano-db-sync schema for the network's live stake is an extremely inefficient operation requiring a multitude of joins and complexity that is very hard to grok, plus it cannot be practically offered with complete precision, only as a cached response. Querying active stake is very efficient, as the database contains a table for the snapshot results, and it also is static for the entire epoch.

# Proposed Solution

Ogmios has a[ LocalStateQuery](https://ogmios.dev/mini-protocols/local-state-query/) to return the current stake distribution in the form of a map that contains the fraction of stake each pool controls.  This can be reduced into a single value in around 270ms on testnet (as of time of writing).

- [x] Extend the [CardanoNode interface](https://github.com/input-output-hk/cardano-js-sdk/blob/5cf462d045301d664c7d1227f8c3067524dc8eef/packages/core/src/CardanoNode/types/CardanoNode.ts#L16-L41)
- [x] Extend the [OgmiosCardanoNode](https://github.com/input-output-hk/cardano-js-sdk/blob/5cf462d045301d664c7d1227f8c3067524dc8eef/packages/ogmios/src/CardanoNode/OgmiosCardanoNode.ts) class to implement the new method
- [x] Extend [this](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services/src/Program/services/ogmios.ts#L13-L14)
- [x] Replace the [call to the DB ](https://github.com/input-output-hk/cardano-js-sdk/blob/5cf462d045301d664c7d1227f8c3067524dc8eef/packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts#L95)with this.#cardanoNode  invocation
- [x] Remove the DB queries and other fragments
- [x] Extend [Ogmios mock](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/ogmios/test/mocks/mockOgmiosServer.ts)
- [x] Add `CardanoNodeUtil` to transform `stakeDistribution` to `liveStake`

- Fixes a bug recently introduced in an e2e test - 592a2805ef879311fe98fa9345201e76ca3bc181

### DIFF
--- a/packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
+++ b/packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
@@ -1,6 +1,7 @@
 import {
   Cardano,
   CardanoNode,
+  CardanoNodeUtil,
   NetworkInfoProvider,
   ProtocolParametersRequiredByWallet,
   StakeSummary,
@@ -20,7 +21,6 @@ import {
   loadGenesisData,
   toGenesisParams,
   toLedgerTip,
-  toStake,
   toSupply,
   toTimeSettings,
   toWalletProtocolParams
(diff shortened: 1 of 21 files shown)


mkazlauskas (approved):
Looks good! :rocket:

ssong-iohk (dismissed):

PR #377, opened 2022-08-05
Feat/adp 2116 create native script

AngelCastilloB:
# Context

We must allow users who wishes to mint & burn (and perform other functionality like multisig) to construct a native script that guards the minting of a new native asset.

# Proposed Solution

Define a core type for a native script and create mapping functions to the CSL.

Adds a new native script core type and mapping functions from and to the CLS.

### DIFF
--- a/packages/core/src/CSL/coreToCsl/coreToCsl.ts
+++ b/packages/core/src/CSL/coreToCsl/coreToCsl.ts
@@ -17,10 +17,18 @@ import {
   Mint,
   MintAssets,
   MultiAsset,
+  NativeScript,
+  NativeScripts,
   PublicKey,
   RewardAddress,
+  ScriptAll,
+  ScriptAny,
   ScriptDataHash,
   ScriptHash,
+  ScriptNOfK,
+  ScriptPubkey,
+  TimelockExpiry,
+  TimelockStart,
   Transaction,
   TransactionBody,
   TransactionHash,
@@ -40,8 +48,7 @@ import {
 } from '@emurgo/cardano-serialization-lib-nodejs';
 
 import * as certificate from './certificate';
-import { SerializationError } from '../../errors';
-import { SerializationFailure } from '../..';
+import { SerializationError, SerializationFailure } from '../..';
 import { parseCslAddress } from '../parseCslAddress';
 
 export const tokenMap = (map: Cardano.TokenMap) => {
@@ -282,4 +289,61 @@ export const tx = ({ body, witness, auxiliaryData }: Cardano.NewTxAlonzo): Trans
   const txWitnessSet = witnessSet(witness.signatures);
   // Possible optimization: only convert auxiliary data once
   return Transaction.new(txBody(body, auxiliaryData), txWitnessSet, txAuxiliaryData(auxiliaryData));
+};
+
+export const nativeScript = (script: Cardano.NativeScript): NativeScript => {
+  let cslScript: NativeScript;
+  const type = script.__type;
+
+  switch (type) {
+    case Cardano.NativeScriptType.RequireSignature: {
+      cslScript = NativeScript.new_script_pubkey(
+        ScriptPubkey.new(Ed25519KeyHash.from_bytes(Buffer.from(script.keyHash, 'hex')))
+      );
+      break;
+    }
+    case Cardano.NativeScriptType.RequireAllOf: {
+      const cslScripts = NativeScripts.new();
+      for (const subscript of script.scripts) {
+        cslScripts.add(nativeScript(subscript));
+      }
+      cslScript = NativeScript.new_script_all(ScriptAll.new(cslScripts));
+      break;
+    }
+    case Cardano.NativeScriptType.RequireAnyOf: {
+      const cslScripts2 = NativeScripts.new();
+      for (const subscript of script.scripts) {
+        cslScripts2.add(nativeScript(subscript));
+      }
+      cslScript = NativeScript.new_script_any(ScriptAny.new(cslScripts2));
+      break;
+    }
+    case Cardano.NativeScriptType.RequireMOf: {
+      const cslScripts3 = NativeScripts.new();
+      for (const subscript of script.scripts) {
+        cslScripts3.add(nativeScript(subscript));
+      }
+      cslScript = NativeScript.new_script_n_of_k(ScriptNOfK.new(script.required, cslScripts3));
+      break;
+    }
+    case Cardano.NativeScriptType.RequireTimeBefore: {
+      cslScript = NativeScript.new_timelock_expiry(
+        TimelockExpiry.new_timelockexpiry(BigNum.from_str(script.slot.toString()))
+      );
+      break;
+    }
+    case Cardano.NativeScriptType.RequireTimeAfter: {
+      cslScript = NativeScript.new_timelock_start(
+        TimelockStart.new_timelockstart(BigNum.from_str(script.slot.toString()))
+      );
+      break;
+    }
+    default:
+      throw new SerializationError(
+        SerializationFailure.InvalidNativeScriptType,
+        `Native Script Type value '${type}' is not supported.`
+      );
+  }
+
+  return cslScript;
 };
--- a/packages/core/src/Cardano/types/NativeScript.ts
+++ b/packages/core/src/Cardano/types/NativeScript.ts
@@ -1,0 +1,142 @@
+/* eslint-disable no-use-before-define */
+
+import { Ed25519KeyHash } from './Key';
+import { Slot } from '@cardano-ogmios/schema';
+
+/**
+ * Native script type.
+ */
+export enum NativeScriptType {
+  RequireAllOf = 'all',
+  RequireAnyOf = 'any',
+  RequireMOf = 'atLeast',
+  RequireSignature = 'sig',
+  RequireTimeAfter = 'after',
+  RequireTimeBefore = 'before'
+}
+
+/**
+ * This script evaluates to true if the transaction also includes a valid key witness
+ * where the witness verification key hashes to the given hash.
+ *
+ * In other words, this checks that the transaction is signed by a particular key, identified by its verification
+ * key hash.
+ */
+export interface RequireSignatureScript {
+  /**
+   * The hash of a verification key.
+   */
+  keyHash: Ed25519KeyHash;
+
+  /**
+   * Script type.
+   */
+  __type: NativeScriptType.RequireSignature;
+}
+
+/**
+ * This script evaluates to true if  all the sub-scripts evaluate to true.
+ *
+ * If the list of sub-scripts is empty, this script evaluates to true.
+ */
+export interface RequireAllOfScript {
+  /**
+   * The list of sub-scripts.
+   */
+  scripts: NativeScript[];
+
+  /**
+   * Script type.
+   */
+  __type: NativeScriptType.RequireAllOf;
+}
+
+/**
+ * This script evaluates to true if any the sub-scripts evaluate to true. That is, if one
+ * or more evaluate to true.
+ *
+ * If the list of sub-scripts is empty, this script evaluates to false.
+ */
+export interface RequireAnyOfScript {
+  /**
+   * The list of sub-scripts.
+   */
+  scripts: NativeScript[];
+
+  /**
+   * Script type.
+   */
+  __type: NativeScriptType.RequireAnyOf;
+}
+
+/**
+ * This script evaluates to true if at least M (required field) of the sub-scripts evaluate to true.
+ */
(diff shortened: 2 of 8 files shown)


packages/core/src/Cardano/types/NativeScript.ts
```diff
@@ -0,0 +1,127 @@
...
+ * where the witness verification key hashes to the given hash.
+ *
+ * In other words, this checks that the transaction is signed by a particular key, identified by its verification
+ * key hash.
+ */
+export interface RequireSignatureScript {
+  /**
+   * The hash of a verification key.
+   */
+  keyHash: string;
```
rhyslbw, line 13:
```suggestion
  keyHash: Ed25519KeyHash;
```

rhyslbw (changes requested):
Looking good so far @AngelCastilloB.

Another improvement here is to add an `enum` keyed with abstract syntax, using the [JSON script syntax](https://github.com/input-output-hk/cardano-node/blob/master/doc/reference/simple-scripts.md#json-script-syntax) for values.

``` ts
export enum NativeScriptType {
  RequireAllOf = 'all',
  RequireAnyOf = 'any',
  RequireMOf = 'atLeast',
  RequireSignature = 'sig',
  RequireTimeAfter = 'after',
  RequireTimeBefore = 'before'
}
```

Also, this PR should also refactor the codebase to remove the Ogmios https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/core/src/Cardano/types/NativeScriptType.ts

packages/core/src/Cardano/types/NativeScript.ts
```diff
@@ -0,0 +1,127 @@
...
+ * slot number Y, and Y <= X.
+ *
+ * This condition guarantees that the actual slot number in which the transaction is included
+ * is greater than or equal to slot number X.
+ */
+export interface RequireTimeAfterScript {
+  /**
+   * The slot number specifying the lower bound of the validity interval.
+   */
+  slot: number;
```
mkazlauskas, line 106:
```suggestion
  slot: Slot;
```

mkazlauskas (changes requested):

packages/core/src/Cardano/types/NativeScript.ts
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/377/commits/c4c598f4dec66535b09e7477a38ab60756ae3209

packages/core/src/Cardano/types/NativeScript.ts
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/377/commits/c4c598f4dec66535b09e7477a38ab60756ae3209

rhyslbw (changes requested):
I've done a refactor in 661d371da773deb07108882720e7fa069ba54589, which I did to validate what I wanted to "suggest". It can be dropped if tests can proved it's not valid :smile:

packages/core/test/Cardano/types/NativeScript.test.ts
```diff
@@ -0,0 +1,30 @@
+import { Ed25519KeyHash, NativeScript, NativeScriptType } from '../../../src/Cardano';
+
+describe('Cardano/types/NativeScript', () => {
+  it('can validate nested script', () => {
```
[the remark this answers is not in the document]
AngelCastilloB: Good point, that test will actually never fails, if the format is invalid it wont compile instead, so I think we can remove it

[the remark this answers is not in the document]
rhyslbw: Yeah, I agree, just remove it. The linked test overlaps with existing coverage of `Ed25519KeyHash`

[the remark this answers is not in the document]
AngelCastilloB: addressed here https://github.com/input-output-hk/cardano-js-sdk/pull/377/commits/bf8eb295bdbdf32c9d8790074ff3da1c4f1c57a1

packages/core/src/errors.ts
```diff
@@ -48,3 +49,9 @@ export class InvalidStringError extends CustomError {
     super(`Invalid string: "${expectation}"`);
   }
 }
+
+export class InvalidNativeScript extends CustomError {
```
mkazlauskas, line 53: This is really a variant of SerializationError since it's thrown when mapping to CSL. It's a common error type in coreToCsl, so I think we want to `InvalidNativeScript instanceof SerializationError === true`

Although this error might also be useful in a different context in the future (like mapping from cardano-cli json), so not 100% sure about this one.

```suggestion
export class InvalidNativeScript extends SerializationError {
```

packages/core/src/CSL/coreToCsl/coreToCsl.ts
```diff
@@ -283,3 +291,56 @@ export const tx = ({ body, witness, auxiliaryData }: Cardano.NewTxAlonzo): Trans
...
+      break;
+    }
+    case Cardano.NativeScriptType.RequireTimeAfter: {
+      cslScript = NativeScript.new_timelock_start(
+        TimelockStart.new_timelockstart(BigNum.from_str(script.slot.toString()))
+      );
+      break;
+    }
+    default:
+      throw new InvalidNativeScript('Unknown script type');
```
mkazlauskas, line 342: I think this error should be aligned with error in [cslToCore](https://github.com/input-output-hk/cardano-js-sdk/pull/377/files#diff-d44f980f3fe3aa875421d19871eb64f3c3481069f3fe6b499ee8891692364728R347)

mkazlauskas (changes requested):
Looks good, just a small request on errors

packages/core/src/CSL/coreToCsl/coreToCsl.ts
AngelCastilloB: I think you are right, we can stick with SerializationError for consistency (and because it is literally a serialization error) and drop this custom error. We can always added it back when legitimately needed.

packages/core/src/errors.ts
AngelCastilloB: I will remove it and use the SerializationError instead, I think we can add it back later down the line if needed.

packages/core/src/CSL/coreToCsl/coreToCsl.ts
AngelCastilloB: addressed here https://github.com/input-output-hk/cardano-js-sdk/pull/377/commits/b5a80d971f77d88933f07c51622f45ec4028b7b4

packages/core/src/errors.ts
AngelCastilloB: addressed here https://github.com/input-output-hk/cardano-js-sdk/pull/377/commits/b5a80d971f77d88933f07c51622f45ec4028b7b4

mkazlauskas (approved):

rhyslbw (dismissed):
Great work @AngelCastilloB :tada: 
Please squash and then request a review from @james-iohk

mkazlauskas (approved):

packages/core/src/errors.ts
```diff
@@ -22,7 +22,8 @@ export enum SerializationFailure {
   InvalidType = 'INVALID_TYPE',
   Overflow = 'OVERFLOW',
   InvalidAddress = 'INVALID_ADDRESS',
-  MaxLengthLimit = 'MAX_LENGTH_LIMIT'
+  MaxLengthLimit = 'MAX_LENGTH_LIMIT',
+  InvalidScriptType = 'INVALID_SCRIPT_TYPE'
```
james-iohk, line 26: Seeing as there'll be support of other (non-simple/native) scripts types in future, it would make sense to make this more descriptive. E.g. `INVALID_SIMPLE_SCRIPT_TYPE`

james-iohk (changes requested):
Something to consider when referring to scripts: cardano-node exclusively refers to these at "simple scripts", the ledger does a bit of both, and they are used interchangeably across components but I hear/see "simple" more often..

Although "non-native" does refer to plutus for now, there is the possibility of other non-native scripts to exist in future, so imo "simple" and "plutus" are better names.

packages/core/src/errors.ts
AngelCastilloB: Thanks for the input @james-iohk I was totally oblivious to the fact these are called Simple Scripts, I have been referring to those as Native Scripts since forever (the Cardano serialization lib calls them as such (https://input-output-hk.github.io/cardano-js-sdk/classes/_cardano_sdk_core.CSL.NativeScript.html), however looking at the reference documentation (https://github.com/input-output-hk/cardano-node/blob/master/doc/reference/simple-scripts.md), they are indeed referred as simple scripts. I will refactor the name to SimpleScript instead of NativeScript. What do you think @mkazlauskas?

mkazlauskas: Agree, let's rename :+1: Thanks @james-iohk

AngelCastilloB: addressed here https://github.com/input-output-hk/cardano-js-sdk/pull/377/commits/b5ac8f5732130918aabc58ba3243252df838b4a4

mkazlauskas (approved):
:rocket:

mkazlauskas (approved):

packages/core/src/errors.ts
AngelCastilloB: @james-iohk @mkazlauskas I already made the refactor; however, I was checking the CDDL schema of transactions for another ticket I am working on, and I see that in the Alonzo CDDL schema this type of scripts are referred to as Native Scripts formally: https://github.com/input-output-hk/cardano-ledger/blob/b44a3b8959c00e8243c3004c01adf412d1b2d317/eras/alonzo/test-suite/cddl-files/alonzo.cddl#L351

I think the schema should be followed over anything else no?, le me know your thoughts.

mkazlauskas: cc @rhyslbw

rhyslbw: Yes, and it makes sense to call these scripts "native", given Plutus is an explicit type: https://github.com/input-output-hk/cardano-ledger/blob/b44a3b8959c00e8243c3004c01adf412d1b2d317/eras/alonzo/test-suite/cddl-files/alonzo.cddl#L274

rhyslbw: Let's base on the Ogmios [representation](https://github.com/CardanoSolutions/ogmios/blob/e2a6ff63e06be3b16065c82a9c0e3be7d51dfca1/docs/static/ogmios.wsp.json#L3910-L3937), where they can be either `Native`, `PlutusV1`, or `PlutusV2`: 

``` json
  , "Script":
    { "oneOf":
      [ { "type": "object"
        , "title": "Native"
        , "additionalProperties": false
        , "required": [ "native" ]
        , "properties":
          { "native": { "$ref": "#/definitions/Script[Native]" }
          }
        }
      , { "type": "object"
        , "title": "PlutusV1"
        , "additionalProperties": false
        , "required": [ "plutus:v1" ]
        , "properties":
          { "plutus:v1": { "$ref": "#/definitions/Script[Plutus]" }
          }
        }
      , { "type": "object"
        , "title": "PlutusV2"
        , "additionalProperties": false
        , "required": [ "plutus:v2" ]
        , "properties":
          { "plutus:v2": { "$ref": "#/definitions/Script[Plutus]" }
          }
        }
      ]
    }
```

AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/377/commits/70cbbbcdb8d9a3cb1af9edd313274d5e1ace0dff

packages/core/src/Cardano/types/NativeScript.ts
```diff
@@ -0,0 +1,142 @@
...
+export interface RequireTimeAfterScript {
+  /**
+   * The slot number specifying the lower bound of the validity interval.
+   */
+  slot: Slot;
+
+  /**
+   * Script type.
+   */
+  type: NativeScriptType.RequireTimeAfter;
```
mkazlauskas, line 127: We've been using `__type` as discriminator field, let's just change to that to keep it consistent

```suggestion
  __type: NativeScriptType.RequireTimeAfter;
```

mkazlauskas (changes requested):
Looks good, just 1 small change request. You can just squash without a new fixup once you make this change. Also, it is technically a breaking change as we were using ogmios types that are different - please use ! in commit msg

james-iohk (dismissed):

mkazlauskas (approved):
Nice work! :tada:

rhyslbw (approved):
:rocket:

PR #378, opened 2022-08-05
Refacter/adp 1933 make logger a required dependendcy

iccicci:
# Context

Sometimes it is hard to investigate the source of problems due to lack of logs.

# Proposed Solution

- To make `logger` a required parameter/dependency for all the packages.
- Removed any reference to `dummyLogger` from the `src` directory of each package.

Many interface signatures was impacted.


mkazlauskas (approved):
Nice work! :rocket:

rhyslbw (changes requested):
The change to `golden-test-generator` actually introduces `bunyan`, and there are also some `console.log` statements still in the code, so it's a distinct change better committed separately as a scoped refactor.

It would also be better to squash all the `*logger is now required` commits into one, as this is a single change we're making against the codebase in one step.

rhyslbw (approved):
:rocket:

mkazlauskas (approved):

PR #379, opened 2022-08-05
Chore/adp 1890 single address wallet logging

mirceahasegan:
# Context

SingleAddressWallet issues can be difficult to troubleshoot as it’s doing a lot of things internally. Add logger.debug statements to impove this.

# Proposed Solution
- Add logger parameter to trackers constructed by SingleAddressWallet.
- Context is added by SingleAddressWallet before passing logger to tracker builders (e.g. [SingleAddressWallet|SyncStatus]), so that context is the one in which SingleAddressWallet is using the tracker.
-Since most of the logic is in rxjs observable streams, use tap() to add debug logs: e.g. statsReady$...pipe(tap(() => logger.debug('Stats are ready').

Add `logger` parameter to many tracker builders.

### DIFF
--- a/packages/wallet/src/SingleAddressWallet.ts
+++ b/packages/wallet/src/SingleAddressWallet.ts
@@ -78,7 +78,7 @@ import { Cip30DataSignature } from '@cardano-sdk/cip30';
 import { InputSelector, defaultSelectionConstraints, roundRobinRandomImprove } from '@cardano-sdk/cip2';
 import { Logger } from 'ts-log';
 import { RetryBackoffConfig } from 'backoff-rxjs';
-import { Shutdown } from '@cardano-sdk/util';
+import { Shutdown, contextLogger } from '@cardano-sdk/util';
 import { TrackedUtxoProvider } from './services/ProviderTracker/TrackedUtxoProvider';
 import { TxInternals, createTransactionInternals, ensureValidityInterval } from './Transaction';
 import { WalletStores, createInMemoryWalletStores } from './persistence';
@@ -170,8 +170,10 @@ export class SingleAddressWallet implements ObservableWallet {
       connectionStatusTracker$ = createSimpleConnectionStatusTracker()
     }: SingleAddressWalletDependencies
   ) {
-    this.#logger = logger;
+    this.#logger = contextLogger(logger, this.name);
+
     this.#inputSelector = inputSelector;
+
     this.txSubmitProvider = new TrackedTxSubmitProvider(txSubmitProvider);
     this.utxoProvider = new TrackedUtxoProvider(utxoProvider);
     this.networkInfoProvider = new TrackedWalletNetworkInfoProvider(networkInfoProvider);
@@ -179,10 +181,12 @@ export class SingleAddressWallet implements ObservableWallet {
     this.assetProvider = new TrackedAssetProvider(assetProvider);
     this.chainHistoryProvider = new TrackedChainHistoryProvider(chainHistoryProvider);
     this.rewardsProvider = new TrackedRewardsProvider(rewardsProvider);
+
     this.syncStatus = createProviderStatusTracker(
       {
         assetProvider: this.assetProvider,
         chainHistoryProvider: this.chainHistoryProvider,
+        logger: contextLogger(this.#logger, 'syncStatus'),
         networkInfoProvider: this.networkInfoProvider,
         rewardsProvider: this.rewardsProvider,
         stakePoolProvider: this.stakePoolProvider,
@@ -191,6 +195,7 @@ export class SingleAddressWallet implements ObservableWallet {
       },
       { consideredOutOfSyncAfter }
     );
+
     this.keyAgent = keyAgent;
     this.addresses$ = new TrackerSubject<GroupedAddress[]>(
       concat(
@@ -199,21 +204,29 @@ export class SingleAddressWallet implements ObservableWallet {
           distinctUntilChanged(groupedAddressesEquals),
           tap(
             // derive an address if none available
-            (addresses) =>
-              addresses.length === 0 &&
-              void keyAgent
-                .deriveAddress({ index: 0, type: AddressType.External })
-                .catch(() => logger.error('SingleAddressWallet failed to derive address'))
+            (addresses) => {
+              if (addresses.length === 0) {
+                this.#logger.debug('No addresses available; deriving one');
+                void keyAgent
+                  .deriveAddress({ index: 0, type: AddressType.External })
+                  .catch(() => this.#logger.error('Failed to derive address'));
+              }
+            }
           ),
           filter((addresses) => addresses.length > 0),
           tap(stores.addresses.set.bind(stores.addresses))
         )
       )
     );
+
     this.name = name;
-    const cancel$ = connectionStatusTracker$.pipe(filter((status) => status === ConnectionStatus.down));
+    const cancel$ = connectionStatusTracker$.pipe(
+      tap((status) => (status === ConnectionStatus.up ? 'Connection UP' : 'Connection DOWN')),
+      filter((status) => status === ConnectionStatus.down)
+    );
     this.#tip$ = this.tip$ = new TipTracker({
       connectionStatus$: connectionStatusTracker$,
+      logger: contextLogger(this.#logger, 'tip$'),
       maxPollInterval: maxInterval,
       minPollInterval: pollInterval,
       provider$: coldObservableProvider({
@@ -234,16 +247,22 @@ export class SingleAddressWallet implements ObservableWallet {
         equals: deepEquals,
         provider: this.networkInfoProvider.eraSummaries,
         retryBackoffConfig,
-        trigger$: eraSummariesTrigger
+        trigger$: eraSummariesTrigger.pipe(tap(() => 'Trigger request era summaries'))
       }),
       stores.eraSummaries
     );
 
     // Epoch tracker triggers the first eraSummaries fetch from eraSummariesTrigger
     // Epoch changes also trigger refetch of eraSummaries
-    this.currentEpoch$ = currentEpochTracker(this.tip$, this.eraSummaries$);
+    this.currentEpoch$ = currentEpochTracker(
+      this.tip$,
+      this.eraSummaries$.pipe(tap((es) => this.#logger.debug('Era summaries are', es)))
+    );
     this.currentEpoch$.pipe(map(() => void 0)).subscribe(eraSummariesTrigger);
-    const epoch$ = this.currentEpoch$.pipe(map((epoch) => epoch.epochNo));
+    const epoch$ = this.currentEpoch$.pipe(
+      map((epoch) => epoch.epochNo),
+      tap((epoch) => this.#logger.debug(`Current epoch is ${epoch}`))
+    );
     this.protocolParameters$ = new PersistentDocumentTrackerSubject(
       coldObservableProvider({
         cancel$,
@@ -272,6 +291,7 @@ export class SingleAddressWallet implements ObservableWallet {
       addresses$,
       chainHistoryProvider: this.chainHistoryProvider,
       inFlightTransactionsStore: stores.inFlightTransactions,
+      logger: contextLogger(this.#logger, 'transactions'),
       newTransactions: this.#newTransactions,
       retryBackoffConfig,
       tip$: this.tip$,
@@ -280,7 +300,7 @@ export class SingleAddressWallet implements ObservableWallet {
 
     this.#resubmitSubscription = createTransactionReemitter({
       confirmed$: this.transactions.outgoing.confirmed$,
-      logger,
+      logger: contextLogger(this.#logger, 'transactionsReemiter'),
       rollback$: this.transactions.rollback$,
       store: stores.volatileTransactions,
       submitting$: this.transactions.outgoing.submitting$,
@@ -297,6 +317,7 @@ export class SingleAddressWallet implements ObservableWallet {
 
     this.utxo = createUtxoTracker({
       addresses$,
+      logger: contextLogger(logger, 'utxo'),
       retryBackoffConfig,
       stores,
       tipBlockHeight$,
@@ -307,6 +328,7 @@ export class SingleAddressWallet implements ObservableWallet {
     this.delegation = createDelegationTracker({
       epoch$,
       eraSummaries$,
+      logger: contextLogger(this.#logger, 'delegation'),
       retryBackoffConfig,
       rewardAccountAddresses$: this.addresses$.pipe(
         map((addresses) => addresses.map((groupedAddress) => groupedAddress.rewardAccount))
@@ -316,16 +338,19 @@ export class SingleAddressWallet implements ObservableWallet {
       stores,
       transactionsTracker: this.transactions
     });
+
     this.balance = createBalanceTracker(this.protocolParameters$, this.utxo, this.delegation);
     this.assets$ = new PersistentDocumentTrackerSubject(
       createAssetsTracker({
         assetProvider: this.assetProvider,
         balanceTracker: this.balance,
+        logger: contextLogger(logger, 'assets$'),
         retryBackoffConfig
       }),
       stores.assets
     );
     this.util = createWalletUtil(this);
+    this.#logger.debug('Created');
   }
 
   async getName(): Promise<string> {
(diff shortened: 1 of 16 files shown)


packages/wallet/src/services/ProviderTracker/ProviderStatusTracker.ts
```diff
@@ -74,6 +76,7 @@ export interface ProviderStatusTrackerInternals {
 export const createProviderStatusTracker = (
   dependencies: ProviderStatusTrackerDependencies,
   { consideredOutOfSyncAfter }: ProviderStatusTrackerProps,
+  logger: Logger, // REVIEW: maybe we should transform this into object params instead of positional
```
mirceahasegan, line 79: REVIEW comment here.
We should probably stop using positional params here. If the debugger logs solution is ok, I can update this code to use named params.

rhyslbw: Yes, it could be part of `ProviderStatusTrackerDependencies`

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -170,15 +170,19 @@ export class SingleAddressWallet implements ObservableWallet {
       connectionStatusTracker$ = createSimpleConnectionStatusTracker()
     }: SingleAddressWalletDependencies
   ) {
-    this.#logger = logger;
+    this.#logger = contextLogger(logger, 'SingleAddressWallet');
     this.#inputSelector = inputSelector;
+
+    this.#logger.debug('Creating providers needed for syncStatus tracking');
```
mkazlauskas, line 176: Constructor is all synchronous, if something fails then `new SingleAddressWallet` would throw. I would minimize those logs to just 'initializing' and 'initialized'

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -170,15 +170,19 @@ export class SingleAddressWallet implements ObservableWallet {
       connectionStatusTracker$ = createSimpleConnectionStatusTracker()
     }: SingleAddressWalletDependencies
   ) {
-    this.#logger = logger;
+    this.#logger = contextLogger(logger, 'SingleAddressWallet');
```
mkazlauskas, line 173: I think we should keep it consistent. Right now,
1. SingleAddressWallet derives contextLogger by itself
2. createProviderStatusTracker expects to get a logger that is already derived

I think strategy 2. is a little cleaner in terms of code. How about we change this to:

```this.#logger = contextLogger(logger, `(${this.name})`);```

So that it's a different reason to derive the logger here.

File: packages/wallet/src/SingleAddressWallet.ts [hunk 4]
```diff
@@ -189,8 +193,11 @@ export class SingleAddressWallet implements ObservableWallet {
         txSubmitProvider: this.txSubmitProvider,
         utxoProvider: this.utxoProvider
       },
-      { consideredOutOfSyncAfter }
+      { consideredOutOfSyncAfter },
+      contextLogger(this.#logger, 'SyncStatus')
```
mkazlauskas, line 197: Maybe let's have those correspond to exact prop name of SingleAddressWallet?
```suggestion
      contextLogger(this.#logger, 'syncStatus')
```

mkazlauskas (changes requested):

packages/wallet/src/SingleAddressWallet.ts
mirceahasegan: Agree.
Why do you want to wrap it in parentheses? It will end up looking like this: [(My cool wallet)]

mkazlauskas: No need to wrap it then

mirceahasegan:
Rebased on top of master and addressed comments

packages/wallet/src/services/ProviderTracker/ProviderStatusTracker.ts
mirceahasegan: Done in [3eeba2a](https://github.com/input-output-hk/cardano-js-sdk/pull/379/commits/3eeba2a461134843d202116d39c1edec2c118116)

packages/wallet/src/SingleAddressWallet.ts
mirceahasegan: Done in [3eeba2a](https://github.com/input-output-hk/cardano-js-sdk/pull/379/commits/3eeba2a461134843d202116d39c1edec2c118116)

packages/wallet/src/SingleAddressWallet.ts
mirceahasegan: Done in [3eeba2a](https://github.com/input-output-hk/cardano-js-sdk/pull/379/commits/3eeba2a461134843d202116d39c1edec2c118116)

packages/wallet/src/SingleAddressWallet.ts
mirceahasegan: Done in [3eeba2a](https://github.com/input-output-hk/cardano-js-sdk/pull/379/commits/3eeba2a461134843d202116d39c1edec2c118116)

packages/wallet/src/SingleAddressWallet.ts
mkazlauskas: Maybe also add an 'Initialized' message, because right now just reading the logs it's not clear that it ever initialized given the 'Initializing...' message

packages/wallet/src/services/TipTracker.ts
```diff
@@ -62,20 +66,34 @@ export class TipTracker extends PersistentDocumentTrackerSubject<Cardano.Tip> {
           connectionStatus$
         ]).pipe(
           // Throttle syncing by interval, cancel ongoing request on external trigger
+          tap(([isSettled, connectionStatus]) => {
```
mirceahasegan, line 69: Some of these logs could also be added inline in the existing operators, but I think it's cleaner to keep them separate when possible

packages/wallet/src/services/TipTracker.ts
```diff
@@ -62,20 +66,34 @@ export class TipTracker extends PersistentDocumentTrackerSubject<Cardano.Tip> {
...
+            if (connectionStatus === ConnectionStatus.down) {
+              // REVIEW: slightly redundant since connection debug messages are already recorded separately
```
mirceahasegan, line 71: When looking at the overall debug log, user will see a 'Connection down' message anyway.
Thus this one is redundant, but I think it's nice to see it here, in context too

packages/wallet/src/services/TipTracker.ts
```diff
@@ -62,20 +66,34 @@ export class TipTracker extends PersistentDocumentTrackerSubject<Cardano.Tip> {
...
+            } else {
+              logger.debug(isSettled === null ? 'Initial request' : 'Start request');
+            }
+          }),
           exhaustMap(([_, connectionStatus]) =>
             connectionStatus === ConnectionStatus.down ? EMPTY : provider$.pipe(takeUntil(externalTrigger$))
           ),
-          distinctUntilChanged(tipEquals)
+          distinctUntilChanged(tipEquals),
+          tap((tip) => logger.debug('Fetched new tip', tip))
```
mirceahasegan, line 81: Is it ok to log the whole tip object? Alternatively I could log only it's hash

mirceahasegan (commented):
A chore with breaking changes... maybe it's better to make it a refactor, since it modifies the interface too?

packages/wallet/src/services/TipTracker.ts
```diff
@@ -62,20 +66,34 @@ export class TipTracker extends PersistentDocumentTrackerSubject<Cardano.Tip> {
           connectionStatus$
         ]).pipe(
           // Throttle syncing by interval, cancel ongoing request on external trigger
+          tap(([isSettled, connectionStatus]) => {
+            if (connectionStatus === ConnectionStatus.down) {
+              // REVIEW: slightly redundant since connection debug messages are already recorded separately
+              logger.debug('Connection down. Skipping request');
+            } else {
+              logger.debug(isSettled === null ? 'Initial request' : 'Start request');
```
mkazlauskas, line 74:
```suggestion
              logger.debug('Fetching tip...');
```

packages/wallet/src/services/TipTracker.ts
```diff
@@ -62,20 +66,34 @@ export class TipTracker extends PersistentDocumentTrackerSubject<Cardano.Tip> {
...
+            }
+          }),
           exhaustMap(([_, connectionStatus]) =>
             connectionStatus === ConnectionStatus.down ? EMPTY : provider$.pipe(takeUntil(externalTrigger$))
```
mkazlauskas, line 78:
```suggestion
            connectionStatus === ConnectionStatus.down ? EMPTY : provider$.pipe(takeUntil(externalTrigger$.pipe(tap() => logger.debug('Tip fetch canceled'))))
```

packages/wallet/src/services/TipTracker.ts
mkazlauskas: I think it's ok

packages/wallet/src/services/TransactionReemitter.ts
```diff
@@ -49,6 +49,8 @@ export const createTransactionReemitter = ({
 }: TransactionReemitterProps): Observable<Cardano.NewTxAlonzo> => {
   const volatileTransactions$ = merge(
     store.get().pipe(
+      // REVIEW: is it possible for the store to return undefined?
```
mkazlauskas, line 52: No, it should never do that. When there's nothing stored it should complete without emitting anything

packages/wallet/src/services/TransactionReemitter.ts
```diff
@@ -81,6 +86,7 @@ export const createTransactionReemitter = ({
   );
 
   return rollback$.pipe(
+    tap((tx) => logger.debug(`Transaction ${tx.id} was rolled back`)),
```
mkazlauskas, line 89: I think it's better to log at source (where rollback$ is created in TransactionsTracker)

packages/wallet/src/services/TipTracker.ts
mkazlauskas: I'm wondering if we could measure the time between every `isSettled === true` (and from startup to the 1st one), without `delay(minPollInterval)` so we'd know how long it takes the wallet to make all requests. Would be a great log :thinking:

mkazlauskas:
I'm wondering if we could measure the time between every `isSettled` (and from startup to the 1st one), measuring from when it switches `true->false` until `true`. So we'd know how long it takes the wallet to make all requests. Would be a great log :thinking:

packages/wallet/src/services/TipTracker.ts
rhyslbw: The whole tip is beneficial as the combination of slot number and hash can detect a fork.

packages/wallet/src/services/TipTracker.ts
rhyslbw: How about just logging `Skipping request` then?

rhyslbw (changes requested):
> A chore with breaking changes... maybe it's better to make it a refactor, since it modifies the interface too?

~~Refactors can not be breaking either. This is a feature~~

packages/wallet/src/services/TipTracker.ts
mirceahasegan: Yeah, I think we can keep it simple for now and assume it will look informative enough. We can always add more info later

rhyslbw:
> Refactors can not be breaking either. This is a feature

Ignore this comment @mirceahasegan. I was making this judgement on another tool previously used that didn't include `refactor` in changelogs (which I should have detailed in my previous blanket statement :sweat_smile: ). `refactor` can be breaking, as it's defined as "not adding a feature or fixing a bug". Since you're clearly referring to a supporting commit that's not adding the logging, then it's certainly not a feature either!

mirceahasegan:
> I'm wondering if we could measure the time between every `isSettled` (and from startup to the 1st one), measuring from when it switches `true->false` (don't care how long it stays in 'true', cause that's just the configurable interval) until `true`. So we'd know how long it takes the wallet to make all requests. Would be a great log 🤔

Let's add a separate task for this 😄

packages/wallet/src/services/TipTracker.ts
mirceahasegan: Done in [bd1f8c0](https://github.com/input-output-hk/cardano-js-sdk/pull/379/commits/bd1f8c0a0a499a2060572ee911828d2848edbb9d)

packages/wallet/src/services/TipTracker.ts
mirceahasegan: Done in [bd1f8c0](https://github.com/input-output-hk/cardano-js-sdk/pull/379/commits/bd1f8c0a0a499a2060572ee911828d2848edbb9d)

packages/wallet/src/services/TipTracker.ts
mirceahasegan: Done in [bd1f8c0](https://github.com/input-output-hk/cardano-js-sdk/pull/379/commits/bd1f8c0a0a499a2060572ee911828d2848edbb9d)

packages/wallet/src/services/TransactionReemitter.ts
mirceahasegan: Done in [bd1f8c0](https://github.com/input-output-hk/cardano-js-sdk/pull/379/commits/bd1f8c0a0a499a2060572ee911828d2848edbb9d)

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -316,6 +316,7 @@ export class SingleAddressWallet implements ObservableWallet {
       )
       .subscribe();
 
+    // REVIEW: Any ideeas of useful debug info to log in UtxoTracker?
```
mirceahasegan, line 319: Don't worry about this one. I've already added logs. I will push the new code soon

packages/wallet/src/services/TransactionsTracker.ts
```diff
@@ -107,10 +107,10 @@ export const createAddressTransactionsProvider = ({
...
                     ({ blockHeader: { blockNo } }) => blockNo >= sinceBlock
                   );
-                  logger.debug(
-                    `Chain history has diverged. ${rollbackTransactions.length} transactions have been rolled back`
-                  );
-                  from(rollbackTransactions).subscribe(rollback$);
+
+                  from(rollbackTransactions)
+                    .pipe(tap((tx) => logger.debug(`Transaction ${tx.id} was rolled back`)))
+                    .subscribe(rollback$);
```
mkazlauskas, line 113: @mirceahasegan not related to changes in this PR, but this might be a bug: `from()` emits all items and *completes* which I think will also call `complete()` on the subject. Could you please verify?

mirceahasegan: Good catch @mkazlauskas . It was indeed completing the rollback subject.
I've updated a unit test to catch this and fixed it. Thank you.
[8754449](https://github.com/input-output-hk/cardano-js-sdk/pull/379/commits/8754449574b70d31374e913c8728a1063b1a0376)

packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
```diff
@@ -118,6 +127,7 @@ export const createDelegationTracker = ({
     shutdown: () => {
       rewardAccounts$.complete();
       rewardsHistory$.complete();
+      logger.debug('Shutdown complete');
```
mkazlauskas, line 130: Maybe just 'Shutdown' since this is synchronous?

```suggestion
      logger.debug('Shutdown');
```

mkazlauskas (dismissed):
Nice work @mirceahasegan :tada:, we can tweak the logs as we go in the future

mirceahasegan:
> 

[ADP-2133](https://input-output.atlassian.net/browse/ADP-2133)

mirceahasegan:
> me between every `isSettled` (and from startup to the 1st one), measuring from when it switches `true->false` (don't care how long it stays in 'true', cause that's just the configurable interval) until `true`. So we'd know how long it takes the wa

[ADP-2133 SingleAddressWallet: measure and log how long it takes the wallet to make all requests](https://input-output.atlassian.net/browse/ADP-2133)

mkazlauskas (approved):

IvayloAndonov:
The changes related to adding a logger in SingleAddressWallet look good to me, however, I'm not so familiar with the `rxjs` specific updates in some places, for which I trust you.
Nice work! @mirceahasegan

IvayloAndonov (approved):

mirceahasegan:
> reaking eithe

done

PR #382, opened 2022-08-08
fix(wallet): stop querying the `StakePoolProvider` for all pools when no current delegation

rhyslbw:
# Context
Undesirable behavior was observed during testing of a remote deployment of the HTTP server with restricted resources, where a wallet without a delegation was requesting all pools via an unbounded query, resulting in resource exhaustion. The fix is two-fold, one contained in this branch, and the other to reject such query as a `400` in the HTTP service (ADP-2120).

# Proposed Solution
This fix stops the observable returned from `createQueryStakePoolsProvider` querying the `StakePoolProvider` if an empty array of pool IDs is passed, as this indicates no data is required, rather than all.

### DIFF
--- a/packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
+++ b/packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
@@ -1,11 +1,4 @@
-import {
-  Cardano,
-  ChainHistoryProvider,
-  SlotEpochCalc,
-  StakePoolProvider,
-  TimeSettings,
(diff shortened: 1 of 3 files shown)


mkazlauskas (dismissed):
:rocket:

mkazlauskas (dismissed):

mirceahasegan (approved):

mirceahasegan (approved):

mkazlauskas (approved):

PR #383, opened 2022-08-08
Test/e2e wallet NFT fix

rhyslbw:
# Context
I discovered a few issues in the NFT e2e test (currently not enforced in CI).

# Proposed Solution
Please see each commit message

### DIFF
--- a/packages/e2e/test/wallet/SingleAddressWallet/nft.test.ts
+++ b/packages/e2e/test/wallet/SingleAddressWallet/nft.test.ts
@@ -94,21 +94,16 @@ describe('SingleAddressWallet.assets/nft', () => {
     expect(nfts.find((nft) => nft.nftMetadata!.name === 'One')).toEqual({
       assetId: 'd1ac67dcebc491ce17635d3d9c8775eb739325ce522f6eac733489aa4e4654303031',
       fingerprint: 'asset15wsawsn72dw07m33fqp42suze636mv2k4agxvs',
-      history: undefined,
       mintOrBurnCount: 1,
       name: '4e4654303031',
       nftMetadata: {
-        description: undefined,
-        files: undefined,
         image: ['ipfs://some_hash1'],
-        mediaType: undefined,
         name: 'One',
         version: '1.0'
       },
-      otherProperties: undefined,
       policyId: 'd1ac67dcebc491ce17635d3d9c8775eb739325ce522f6eac733489aa',
       quantity: 1n,
-      tokenMetadata: { desc: undefined, icon: 'ipfs://some_hash1', name: 'One' }
+      tokenMetadata: null
     });
     expect(nfts.find((nft) => nft.nftMetadata!.name === 'Two')).toBeDefined();
   });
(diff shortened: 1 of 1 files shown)


packages/e2e/test/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "baseUrl": "."
   },
+  "rootDir": ".",
```
iccicci, line 6: I removed it in b5551f7d0 exactly to resolve compilation problems.

I did lots of checks and I found something (in our repos) which needs to be discussed and something else (in `tsc`) so strange...

Let's start from the strange thing: [rootDir](https://www.typescriptlang.org/tsconfig#rootDir) is a compiler option, not a root field.
The fact that in this position affects both your and my builds is the evidence that `tsc`, under some circumstances (which I neither investigated nor researched), accepts it also as a root field.
... but I would completely ignore this point!

In my opinion, the only place where `rootDir` should appare is the `compilerOptions` filed of the `tsconfig.json` file in the root of the repository as that actually is the root directory of our project. if I checked correctly all other `tsconfig.json` files extend it: the option will be correct in all other directories.
If I did a bad check and there is some `tsconfig.json` file which do not extends the root one, probably that's something else to be fixed.

When I saw 7180e2c I asked "how is it possible if I run `yarn test:build:verify` from the root of the repos before pushing?".
This brought me to check the `include` directive (or root field) in this file.
The effect of it is: when we run `test:build:verify` in the `e2e` package we verify the build of the `test` directory of the other four packages **and not** the `test` directory of the package itself. To have evidence of this: if we remove the `noEmit` compiler option from [test/tsconfig.json](https://github.com/input-output-hk/cardano-js-sdk/blob/0e1656bcf5095da4d96b81a7af1e5db3ea560eda/test/tsconfig.json) file and run the command we will find the transpiled files in the `test` directory of the other four packages **and not** in the `test` directory of the `e2e` package.
I think this should be fixed

1. removing the `include` root field; It's default value is `**`: this will make the command to verify only the right directory;
2. adding the other four `test` directories as `references` since we want to `import` files from there.

Last: the `references` root field seems to lack `{ "path": "../../blockfrost/src" }`.

rhyslbw: Thanks for the detailed report @iccicci. I'll do some digging today

packages/e2e/test/wallet/SingleAddressWallet/nft.test.ts
```diff
@@ -145,29 +140,9 @@ describe('SingleAddressWallet.assets/nft', () => {
...
-            name: 'some name',
-            src: ['file://some_audio_file', 'file://another_audio_file']
-          }
-        ],
-        icon: 'ipfs://somehash',
-        id: '1',
-        mediaType: 'image/png',
-        name: 'NFT with files'
-      }
+      tokenMetadata: null
```
AngelCastilloB, line 145: I remember this assertion use to work, what changed that made the tokenMetadata now null?

rhyslbw: It was a bug that was fixed in https://github.com/input-output-hk/cardano-js-sdk/pull/345, but this e2e test assertion wasn't updated.
image

I've checked there is no result when manually querying the off-chain token metadata server, so `null` is the expected result.

iadmytro (dismissed):

AngelCastilloB (approved):

packages/e2e/test/tsconfig.json
rhyslbw: I ran out of time to look at this further, so have cherry-picked the contentious commit to a new branch to be handled separately.

mkazlauskas (approved):

PR #385, opened 2022-08-09
feat!: replace `NetworkInfoProvider.timeSettings` with `eraSummaries`

rhyslbw:
# Context
The `timeSettings` concept was named before becoming aware of the almost equivalent Ouroboros Local State Query that describes this as `eraSummaries`. It was [noted](https://github.com/input-output-hk/cardano-js-sdk/pull/298#pullrequestreview-1021022252) in a previous PR to perform this additional refactor when an opportunity arose based on other open branches, but since it's a breaking change, is best done sooner rather than later.

# Proposed Solution
This change replaces the widespread interface and removes the now redundant mapping of the results when fetched from `cardano-node`.

This is a refactor that touches many files, however, it should be a fairly clean rebase target for other open branches, so I suggest to merge it ASAP.

### DIFF
--- a/packages/ogmios/test/CardanoNode/mappers.test.ts
+++ b/packages/ogmios/test/CardanoNode/mappers.test.ts
@@ -9,7 +9,7 @@ describe('cardano node mappers', () => {
       parameters: { epochLength: 432_000, safeZone: 129_600, slotLength: 1 },
       start: { epoch: 74, slot: 1_598_400, time: 31_968_000 }
     };
-    test('map ogmios EraSummary to TimeSettings', () => {
+    it('maps ogmios EraSummary to core EraSummary', () => {
       const result = mappers.mapEraSummary(eraSummary, new Date(1_506_203_091_000));
       expect(result).toEqual<EraSummary>({
         parameters: {
--- a/packages/blockfrost/src/blockfrostNetworkInfoProvider.ts
+++ b/packages/blockfrost/src/blockfrostNetworkInfoProvider.ts
@@ -1,7 +1,7 @@
 import { BlockFrostAPI } from '@blockfrost/blockfrost-js';
 import { BlockfrostToCore } from './BlockfrostToCore';
 import { NetworkInfoProvider, ProviderError, ProviderFailure } from '@cardano-sdk/core';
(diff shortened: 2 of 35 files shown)


AngelCastilloB (dismissed):

packages/ogmios/test/CardanoNode/mappers.test.ts
```diff
@@ -1,26 +0,0 @@
-import * as mappers from '../../src/CardanoNode/mappers';
-import { EraSummary } from '@cardano-sdk/core';
-import { Schema } from '@cardano-ogmios/client';
-
-describe('cardano node mappers', () => {
-  describe('mapEraSummary', () => {
```
mkazlauskas, line 6: This test is still useful as it transforms `time` into `Date`

mkazlauskas: Brought it back in 036ff826f617b8d0d6f937ef2e729028061dd081

rhyslbw: Good catch. I misinterpreted the test intent. Will suggest a change to fix the name

packages/ogmios/test/CardanoNode/mappers.test.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Schema } from '@cardano-ogmios/client';
+
+describe('cardano node mappers', () => {
+  describe('mapEraSummary', () => {
+    const eraSummary: Schema.EraSummary = {
+      end: { epoch: 102, slot: 13_694_400, time: 44_064_000 },
+      parameters: { epochLength: 432_000, safeZone: 129_600, slotLength: 1 },
+      start: { epoch: 74, slot: 1_598_400, time: 31_968_000 }
+    };
+    test('map ogmios EraSummary to TimeSettings', () => {
```
rhyslbw, line 12: `map ogmios EraSummary to core EraSummary`

mkazlauskas: Done 12e43c9d5d30b08a49deffa127a776b82d0a5a03

mkazlauskas (approved):

PR #386, opened 2022-08-09
Refactor/live stake query stake pool service

jvieiro:
# Context

`StakePool` http service is performing a live stake query, which takes a long time to resolve in mainnet. In order to reduce time, the info should be queried against ogmios.

# Proposed Solution

- [x] Add `CardanoNode` dependency to `DbSyncStakePool` and perform the live_stake query

- [x] Add mappers to build pool metrics based on db query and ogmios retrieved info.

### DIFF
--- a/packages/cardano-services/src/Program/loadHttpServer.ts
+++ b/packages/cardano-services/src/Program/loadHttpServer.ts
@@ -88,9 +88,11 @@ const serviceMapFactory = (args: ProgramArgs, logger: Logger, dnsResolver: DnsRe
 
       return new AssetHttpService({ assetProvider, logger });
     }, ServiceNames.Asset),
-    [ServiceNames.StakePool]: withDb((db) => {
+    [ServiceNames.StakePool]: withDb(async (db) => {
+      const cardanoNode = await getOgmiosCardanoNode(dnsResolver, logger, args.options);
       const stakePoolProvider = new DbSyncStakePoolProvider({
         cache: new InMemoryCache(args.options!.dbCacheTtl!),
+        cardanoNode,
         db,
         epochMonitor: getEpochMonitor(db),
         logger
@@ -116,9 +118,6 @@ const serviceMapFactory = (args: ProgramArgs, logger: Logger, dnsResolver: DnsRe
     [ServiceNames.NetworkInfo]: withDb(async (db) => {
       if (args.options?.cardanoNodeConfigPath === undefined)
         throw new MissingProgramOption(ServiceNames.NetworkInfo, ProgramOptionDescriptions.CardanoNodeConfigPath);
-      if (args.options?.ogmiosUrl === undefined)
-        throw new MissingProgramOption(ServiceNames.NetworkInfo, ProgramOptionDescriptions.OgmiosUrl);
-
       const cardanoNode = await getOgmiosCardanoNode(dnsResolver, logger, args.options);
       const networkInfoProvider = new DbSyncNetworkInfoProvider(
         { cardanoNodeConfigPath: args.options.cardanoNodeConfigPath },
--- a/packages/cardano-services/src/Program/services/ogmios.ts
+++ b/packages/cardano-services/src/Program/services/ogmios.ts
@@ -4,11 +4,11 @@
 import { CommonOptionDescriptions, CommonProgramOptions } from '../../ProgramsCommon';
 import { DnsResolver } from '../utils';
 import { Logger } from 'ts-log';
-import { MissingProgramOption } from '../errors';
+import { MissingCardanoNodeOption } from '../errors';
 import { OgmiosCardanoNode, ogmiosTxSubmitProvider, urlToConnectionConfig } from '@cardano-sdk/ogmios';
-import { ServiceNames } from '../ServiceNames';
 import { TxSubmitProvider } from '@cardano-sdk/core';
 import { isConnectionError } from '@cardano-sdk/util';
+import memoize from 'lodash/memoize';
 
 const isCardanoNodeOperation = (prop: string | symbol): prop is 'eraSummaries' | 'systemStart' | 'stakeDistribution' =>
   ['eraSummaries', 'systemStart', 'stakeDistribution'].includes(prop as string);
@@ -82,7 +82,7 @@ export const getOgmiosTxSubmitProvider = async (
   if (options?.ogmiosSrvServiceName)
     return ogmiosTxSubmitProviderWithDiscovery(dnsResolver, logger, options.ogmiosSrvServiceName);
   if (options?.ogmiosUrl) return ogmiosTxSubmitProvider(urlToConnectionConfig(options?.ogmiosUrl), logger);
-  throw new MissingProgramOption(ServiceNames.TxSubmit, [
+  throw new MissingCardanoNodeOption([
     CommonOptionDescriptions.OgmiosUrl,
     CommonOptionDescriptions.OgmiosSrvServiceName
   ]);
@@ -140,16 +140,14 @@ export const ogmiosCardanoNodeWithDiscovery = async (
   });
 };
 
-export const getOgmiosCardanoNode = async (
-  dnsResolver: DnsResolver,
-  logger: Logger,
-  options?: CommonProgramOptions
-): Promise<OgmiosCardanoNode> => {
-  if (options?.ogmiosSrvServiceName)
-    return ogmiosCardanoNodeWithDiscovery(dnsResolver, logger, options.ogmiosSrvServiceName);
-  if (options?.ogmiosUrl) return new OgmiosCardanoNode(urlToConnectionConfig(options.ogmiosUrl), logger);
-  throw new MissingProgramOption(ServiceNames.TxSubmit, [
-    CommonOptionDescriptions.OgmiosUrl,
-    CommonOptionDescriptions.OgmiosSrvServiceName
-  ]);
-};
+export const getOgmiosCardanoNode = memoize(
+  async (dnsResolver: DnsResolver, logger: Logger, options?: CommonProgramOptions): Promise<OgmiosCardanoNode> => {
+    if (options?.ogmiosSrvServiceName)
+      return ogmiosCardanoNodeWithDiscovery(dnsResolver, logger, options.ogmiosSrvServiceName);
+    if (options?.ogmiosUrl) return new OgmiosCardanoNode(urlToConnectionConfig(options.ogmiosUrl), logger);
+    throw new MissingCardanoNodeOption([
+      CommonOptionDescriptions.OgmiosUrl,
+      CommonOptionDescriptions.OgmiosSrvServiceName
+    ]);
+  }
+);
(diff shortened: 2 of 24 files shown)


mkazlauskas (changes requested):

packages/cardano-services/src/Program/loadHttpServer.ts
```diff
@@ -82,9 +82,13 @@ const serviceMapFactory = (args: ProgramArgs, logger: Logger, dnsResolver: DnsRe
 
       return new AssetHttpService({ assetProvider, logger });
     }),
-    [ServiceNames.StakePool]: withDb((db) => {
+    [ServiceNames.StakePool]: withDb(async (db) => {
+      if (args.options?.ogmiosUrl === undefined)
+        throw new MissingProgramOption(ServiceNames.StakePool, ProgramOptionDescriptions.OgmiosUrl);
+      const cardanoNode = await getOgmiosCardanoNode(dnsResolver, logger, ServiceNames.StakePool, args.options);
```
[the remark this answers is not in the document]
IvayloAndonov: First, I think that this check at line 86 is redundant because it is already handled by [getOgmiosCardanoNode](https://github.com/input-output-hk/cardano-js-sdk/blob/cc82bc27539e3ff07f7c2d5816fa7e70c32d06ac/packages/cardano-services/src/Program/services/ogmios.ts#L143-L155) internally.

Second, in theory it never could be `undefined` as we always [default](https://github.com/input-output-hk/cardano-js-sdk/blob/3b65972b74d410312005c9c60d1b48c269b81aaa/packages/cardano-services/src/cli.ts#L72-L77) to a [value](https://github.com/input-output-hk/cardano-js-sdk/blob/cb741368bb312e55c8285be1bbb3ae13c25cee4a/packages/cardano-services/src/ProgramsCommon/defaults.ts#L3-L6)

About the IoC container - I totally agree that once integrated with our code base it will simplify the future dependency resolution a lot as it will allow us to inject instances depending on their types easily as singletons or transient, so we could prototype it in PoC.

This will resolve the issues with the same instances, but not sure how will skip the `MissingProgramOption` check if some of the arguments could be `undefined` by definition, for instance, `dbConnection` @mkazlauskas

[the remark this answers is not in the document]
rhyslbw: > You could just memoize like [here](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services/src/Program/loadHttpServer.ts#L71).

Let's proceed with this option, then we can handle the IOC refactor separately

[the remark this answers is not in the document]
mkazlauskas: > This will resolve the issues with the same instances, but not sure how will skip the MissingProgramOption check if some of the arguments could be undefined by definition, for instance, dbConnection @mkazlauskas

You could simply register services based on program options, then rely on IoC library to give meaningful errors that we can parse, e.g. 'Service A depends on B but it was not provided' => 'Missing --some-b-option, required by Service A'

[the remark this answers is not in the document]
IvayloAndonov: so yeah, in the instances registration step will determine whether to register or not a given instance to the container depends on the program opts, then the package should handle the errors accordingly what's missing to what as you said 👍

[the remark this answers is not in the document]
jvieiro: Implemented at [9a2f1c4](https://github.com/input-output-hk/cardano-js-sdk/pull/386/commits/9a2f1c4a2d916974622a13696ab23949a55a31e0)

packages/cardano-services/test/Program/services/ogmios.test.ts
```diff
@@ -255,8 +255,8 @@ describe('Service dependency abstractions', () => {
           await serverClosePromise(ogmiosServer);
         });
 
-        it('ogmiosCardanoNode should not be a instance of Proxy ', () => {
-          expect(types.isProxy(ogmiosCardanoNode)).toEqual(false);
+        it('ogmiosCardanoNode should be a instance of Proxy ', () => {
+          expect(types.isProxy(ogmiosCardanoNode)).toEqual(true);
```
IvayloAndonov, line 259: Why `ogmiosCardanoNode` becomes an instance of proxy now since `ogmiosUrl` is used for the connection?

packages/cardano-services/src/Program/services/ogmios.ts
```diff
@@ -140,16 +141,19 @@ export const ogmiosCardanoNodeWithDiscovery = async (
...
-): Promise<OgmiosCardanoNode> => {
-  if (options?.ogmiosSrvServiceName)
-    return ogmiosCardanoNodeWithDiscovery(dnsResolver, logger, options.ogmiosSrvServiceName);
-  if (options?.ogmiosUrl) return new OgmiosCardanoNode(urlToConnectionConfig(options.ogmiosUrl), logger);
-  throw new MissingProgramOption(ServiceNames.TxSubmit, [
-    CommonOptionDescriptions.OgmiosUrl,
-    CommonOptionDescriptions.OgmiosSrvServiceName
-  ]);
-};
+export const getOgmiosCardanoNode = memoize(
```
IvayloAndonov, line 144: I think it never will memoize one and same instance of `ogmiosCardanoNode` since we use `await getOgmiosCardanoNode()` with different arguments for both services [StakePool](https://github.com/input-output-hk/cardano-js-sdk/pull/386/files#diff-cfc78f77666af1469e2e670c258fecabed5a923995563490c52370f7a6b4888cR86) and [NetworkInfo](https://github.com/input-output-hk/cardano-js-sdk/pull/386/files#diff-cfc78f77666af1469e2e670c258fecabed5a923995563490c52370f7a6b4888cR112) as it caches the callbacks by given a set of arguments :thinking_face:? @mkazlauskas  @jvieiro

mkazlauskas: By default `memoize` only considers the 1st argument. If it doesn't match then you can add a 2nd `resolver` argument to memoize as `() => true`.

It should be refactored to **not** take in `serviceName` because the goal of memoize here is to use a single instance of OgmiosCardanoNode for **multiple** different services.

IvayloAndonov: After a fast look into it, I see `serviceName` is passed only to be used by `throw new MissingProgramOption` to expose which is the exact service with missing program options in the error. This could be changed if we don't care about it and throw a different error (even if we never go in that case because have defaulted to Ogmios URL always)

packages/cardano-services/test/Program/services/ogmios.test.ts
jvieiro: Solved!

IvayloAndonov (approved):
Nice work

rhyslbw (changes requested):

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePool.ts
```diff
@@ -160,6 +165,9 @@ export class DbSyncStakePoolProvider extends DbSyncProvider implements StakePool
             this.#builder.queryPoolAPY(hashesIds, { rewardsHistoryLimit: options?.rewardsHistoryLimit })
           );
 
+    // Get pools' stake distribution
+    const stakeDistribution = await this.#cardanoNode.stakeDistribution();
```
rhyslbw, line 169: This should also be cached independently right @IvayloAndonov? Otherwise, it will be called for every pool. This is how it's used in the `NetworkInfoProvider`
https://github.com/input-output-hk/cardano-js-sdk/blob/f5f657789421d0fb186142e6494f0a190d26b4f9/packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts#L87-L92

rhyslbw (changes requested):

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePool.ts
IvayloAndonov: Right, should be cached with the default TTL since it's provide live data, good catch.

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/mappers.ts
```diff
@@ -1,5 +1,8 @@
-import { Cardano, StakePoolStats } from '@cardano-sdk/core';
+import { BigNumber } from 'bignumber.js';
```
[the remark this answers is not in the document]
jvieiro: I've implemented the divide new method which returns a floating-point number as a string: https://github.com/input-output-hk/cardano-js-sdk/pull/386/files#diff-a0babbdbd7404c2e0a27f11cabb68ffa69bb38957121aaad95e465c3d7dda23cR7

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePool.ts
jvieiro: Implemented here https://github.com/input-output-hk/cardano-js-sdk/pull/386/files#diff-5ad8b26faabe50605214aec922ed96428a741e5e07da541676062cec7a693a17R176

packages/util/src/BigIntMath.ts
```diff
@@ -1,7 +1,12 @@
+import { BigNumber } from 'bignumber.js';
+
 export const BigIntMath = {
   abs(x: bigint): bigint {
     return x < 0n ? -x : x;
   },
+  divide(numerator: bigint, denominator: bigint): string {
```
mkazlauskas, line 7: I'm actually not sure if it belongs here. `bigint` is supposed to be an integer (as the name implies) and integer math division is a little different. @rhyslbw do you still think it's a good idea to have it here? I'm leaning towards rolling rolling this back to put usage of `bignumber.js` where it's needed directly. 

Otherwise, we should come up with a better name for this function. Maybe `floatDivision`?

rhyslbw: Yeah, my request wasn't considering that this module was originally intended to be based on `Math`, so probably just revert and move the `bignumber.js` dependency back to `cardano-services`

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/util.ts
```diff
@@ -42,3 +44,8 @@ export const emptyPoolsExtraInfo = {
   poolRetirements: [],
   poolRewards: []
 };
+
+export const toHexString = (bytes: Buffer) => bytes.toString('hex');
```
mkazlauskas, line 48: This belongs in [src/util/hexString](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services/src/util/hexString.ts)

mkazlauskas (changes requested):
Looks good, just 1 minor request

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/util.ts
rhyslbw: Done in aa0a867

mkazlauskas (dismissed):

PR #387, opened 2022-08-10
Adp 2113 test logger

iccicci:
# Context

Sometimes it is hard to investigate unit tests failure because we often use `dummyLogger`.
To investigate problem we often need to edit the test file to substitute `dummyLogger` with a real logger.
This makes the test output _not clean_, so at the end of the investigation it is a good practice to restore `dummyLogger`...
This process is often a heavy time consumer.

# Proposed Solution

- Added `TestLogger`, a simple unit tests dedicated logger.
- Used `TestLogger` in `@cardano-sdk/rabbitmq` as the latter uses the `record` feature of the new logger.
- Used `TestLogger` in `@cardano-sdk/cardano-services-client` to produce a _clean_ test output.


packages/util-dev/src/TestLogger.ts
```diff
@@ -0,0 +1,140 @@
...
+ * Options for createLogger
+ */
+export interface TestLoggerOptions {
+  /**
+   * The environment variables to use. Default: process.env - Used to test TestLogger
+   */
+  env?: NodeJS.ProcessEnv;
+
+  /**
+   * If true, the logger will records all the logged values
```
mirceahasegan, line 47: Is the default 'false'?

packages/util-dev/test/TestLogger.test.ts
```diff
@@ -87,7 +87,7 @@ describe('TestLogger', () => {
     });
 
     describe('TL_BREAK', () => {
-      const loggedData = Object.fromEntries(Array.from({ length: 10 }, (v, i) => [i.toString(), `v${i}`]));
+      const loggedData = Object.fromEntries(Array.from({ length: 10 }, (_v, i) => [i.toString(), `v${i}`]));
```
mirceahasegan, line 90: Why `_v` here?

iccicci: Because it is unused. Without the underscore `eslint` raises an error (or warning...)

mirceahasegan: Oh. I see. Didn't notice it's not used. I've seen the following conventions until now.
```
(_, i) => [i.toString(), `v${i}`];
// or
(, i) => [i.toString(), `v${i}`];

IvayloAndonov: I'm using that one `(_, i) => [i.toString(), `v${i}`];`

packages/util-dev/src/TestLogger.ts
iccicci: b6849e8a6 should make the question disappear.

packages/util-dev/test/TestLogger.test.ts
iccicci: b6849e8a6 should solve this

mirceahasegan:
Nice work @iccicci! 🎉

mirceahasegan (approved):

AngelCastilloB (dismissed):
LGTM Awesome job @iccicci ! (I really liked the explanation on the README file :)

mkazlauskas (dismissed):
Nice work @iccicci, good to squash :rocket:

IvayloAndonov (approved):
Nice work!

mirceahasegan (dismissed):

mkazlauskas (dismissed):

mirceahasegan (approved):

mkazlauskas (approved):

PR #388, opened 2022-08-11
[WIP]refactor(core): removed Ogmios schema package dependency

AngelCastilloB:
# Context

We must removed dependency on the Ogmios schema package and have our own types in the core package so we can have more typescript-friendly types.

# Proposed Solution

Port over the required types from the Ogmios schema package and do the necessary adjustments.

Dependency with @ogmios/schema removed.


packages/core/src/Cardano/util/numeric.ts
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable no-use-before-define */
+
+export declare type UInt64 = number;
+export declare type UInt32 = number;
+export declare type Int64 = number;
```
mkazlauskas, line 5: int64 > Number.MAX_SAFE_INTEGER
```suggestion
export declare type UInt64 = bigint;
export declare type UInt32 = number;
export declare type Int64 = bigint;
```

packages/core/src/Cardano/types/Transaction.ts
```diff
@@ -82,9 +81,29 @@ export interface Redeemer {
 
 export type Signatures = Map<Ed25519PublicKey, Ed25519Signature>;
 
-export type Witness = Omit<Partial<Ogmios.TxAlonzo['witness']>, 'redeemers' | 'signatures'> & {
+export declare type Signature = string;
+export declare type ChainCode = string;
+export declare type AddressAttributes = string;
+export declare type VerificationKey = string;
+export declare type Datum = string;
```
mkazlauskas, line 88:
```suggestion
export declare type Datum = util.HexBlob;
```

packages/core/src/Cardano/types/Transaction.ts
```diff
@@ -82,9 +81,29 @@ export interface Redeemer {
 
 export type Signatures = Map<Ed25519PublicKey, Ed25519Signature>;
 
-export type Witness = Omit<Partial<Ogmios.TxAlonzo['witness']>, 'redeemers' | 'signatures'> & {
+export declare type Signature = string;
+export declare type ChainCode = string;
+export declare type AddressAttributes = string;
+export declare type VerificationKey = string;
```
mkazlauskas, line 87: Not 100% sure about this one
```suggestion
export declare type VerificationKey = Ed25519PublicKey;
```

packages/core/src/Cardano/types/Transaction.ts
```diff
@@ -82,9 +81,29 @@ export interface Redeemer {
 
 export type Signatures = Map<Ed25519PublicKey, Ed25519Signature>;
 
-export type Witness = Omit<Partial<Ogmios.TxAlonzo['witness']>, 'redeemers' | 'signatures'> & {
+export declare type Signature = string;
```
mkazlauskas, line 84:
```suggestion
export declare type Signature = Ed25519Signature;
```

packages/core/src/Cardano/types/Transaction.ts
```diff
@@ -82,9 +81,29 @@ export interface Redeemer {
...
+export declare type ChainCode = string;
```
mkazlauskas, line 85: Let's find what this is exactly, maybe we can better type it

packages/core/src/Cardano/types/Script.ts
```diff
@@ -0,0 +1,40 @@
...
+  expiresAt: Slot;
+}
+export interface StartsAt {
+  startsAt: Slot;
+}
+
+export declare type DigestBlake2BVerificationKey = string;
+
+export declare type Script = Native | PlutusV1 | PlutusV2;
+export declare type ScriptNative = DigestBlake2BVerificationKey | Any | All | NOf | ExpiresAt | StartsAt;
```
mkazlauskas, line 24: Let's make sure to use consistent terminology
```suggestion
export declare type ScriptSimple = DigestBlake2BVerificationKey | Any | All | NOf | ExpiresAt | StartsAt;
```

packages/core/src/Cardano/types/Script.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export interface ExpiresAt {
+  expiresAt: Slot;
+}
+export interface StartsAt {
+  startsAt: Slot;
+}
+
+export declare type DigestBlake2BVerificationKey = string;
```
mkazlauskas, line 21: This could probably be better typed

packages/core/src/Cardano/types/Transaction.ts
```diff
@@ -82,9 +81,29 @@ export interface Redeemer {
...
+  chainCode?: ChainCode | undefined;
+  addressAttributes?: AddressAttributes | undefined;
+  key?: VerificationKey;
+}
+
+export type Witness = {
   redeemers?: Redeemer[];
   signatures: Signatures;
+  scripts?: {
+    [k: string]: Cardano.Script;
```
mkazlauskas, line 101: Let's use type aliases for indexer types so that it's clear what it is

mkazlauskas (commented):
So far so good, some suggestions below

packages/core/src/Cardano/types/Block.ts
```diff
@@ -1,9 +1,26 @@
...
+ */
+export declare type BlockSize = number;
+
+/**
+ * The block position in the blockchain
+ */
+export declare type BlockNo = number;
+
+/**
+ * Time period measure in slots
```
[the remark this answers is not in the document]
AngelCastilloB: How would you define epoch vs epoch number?

packages/core/src/Cardano/types/Script.ts
AngelCastilloB: I will update it once we merge the other branch where we are introducing the new type

packages/core/src/Cardano/types/Block.ts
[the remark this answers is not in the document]
mkazlauskas: Yeah, it's used as an epochNo :+1:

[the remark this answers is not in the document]
AngelCastilloB: addressed here https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/f93bd1a2bac1896cc1289c422db283dd6dfd92dd

packages/core/src/Cardano/util/numeric.ts
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/f93bd1a2bac1896cc1289c422db283dd6dfd92dd

This implied a bigger refactor than I was expecting. Let me know if it was done correctly

packages/core/src/Cardano/types/Transaction.ts
AngelCastilloB: addressed here https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/f93bd1a2bac1896cc1289c422db283dd6dfd92dd

packages/core/src/Cardano/types/Transaction.ts
AngelCastilloB: I made the change here: https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/f93bd1a2bac1896cc1289c422db283dd6dfd92dd

packages/core/src/Cardano/types/Transaction.ts
AngelCastilloB: done here: https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/f93bd1a2bac1896cc1289c422db283dd6dfd92dd

packages/core/src/Cardano/types/Transaction.ts
AngelCastilloB: I couldn't find much information about it, other than it is a byte array, I will try tomorrow again.

packages/core/src/Cardano/types/Script.ts
AngelCastilloB: remove and replaced with the proper kind of native script, this will change once we merge the native/simple script branch and when eventually we add support for plutus v1 and v2

packages/core/src/Cardano/types/Transaction.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/f93bd1a2bac1896cc1289c422db283dd6dfd92dd

packages/core/src/Cardano/util/numeric.ts
mkazlauskas: Yep, looking good :+1:

packages/core/src/Cardano/types/Block.ts
```diff
@@ -1,9 +1,26 @@
...
-import { Epoch, Lovelace, PoolId } from '.';
 import { Hash28ByteBase16, Hash32ByteBase16, OpaqueString, typedBech32 } from '../util';
 import { InvalidStringError } from '../..';
+import { Lovelace, PoolId } from '.';
 
-export { BlockNo } from '@cardano-ogmios/schema';
+/**
+ * The block size in bytes
+ */
+export declare type BlockSize = number;
```
mkazlauskas, line 8: I think `declare` is not needed here (and in other changes in this PR). This keyword is used to tell tsc that some variable is available as constant during runtime, typically used for global variables

```suggestion
export type BlockSize = number;
```

packages/core/src/Cardano/types/Transaction.ts
mkazlauskas: Could you please just make a list of types to be narrowed down and create a separate ticket for it?

packages/core/src/Cardano/types/Transaction.ts
mkazlauskas: Sorry this is not what I meant here. More like:

```
[k: WhateverThisIs]: Cardano.Script
```

Do you think having separate fields for script types is nicer? I guess it depends on how it would be used. Consider something like this:

```ts
type ScriptPlutusV2 = { __type: 'plutusv2', bytes: Cardano.HexBlob; }
type Script = ScriptNative | ScriptPlutusV2 | ...`.
```

AngelCastilloB: I am not sure why they are typified like that in Ogmios:

`[k: string]: Cardano.Script;`

Scripts are stored in the witness set in plain arrays (that's why I changed that type to just Script[]), each type of script; Native, PlutusV1 and PlutusV2 are stored in three different arrays inside the witness set:

https://github.com/input-output-hk/cardano-ledger/blob/b24fa846e960ff1e78d73b4a2d21e7c6a59a20dc/eras/babbage/test-suite/cddl-files/babbage.cddl#L275

I split them to keep our Witness type consistent with the type defined in the CDDL; however; I think merging them in a Script type also works, I will refactor this.

packages/core/src/Cardano/types/Block.ts
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/34fce0e418d63211ce69197f26e5ce198c2f5600

packages/core/src/Cardano/types/Transaction.ts
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/e5cd2cc64cabb6553678c4661aab7639650ca9be

packages/core/src/Cardano/types/Transaction.ts
AngelCastilloB: ticket created (ADP-2146), as far as I can see only those two types remaining unidentified.

packages/core/src/Cardano/types/Script.ts
rhyslbw: This is now a non-issue right?

yarn-project.nix
```diff
@@ -142,14 +142,13 @@ cacheEntries = {
 "webextension-polyfill@npm:0.8.0" = { filename = "webextension-polyfill-npm-0.8.0-eab1744c30-56286e6216.zip"; sha512 = "56286e6216e15834fdcdc3dcf611000d96b49c9d925de3d657eb5f1e294074180dd3efe78a0966223eb311ac3a0ac5fdf71d324b836bf05724455ffd6c7c98c3"; };
 "buffer@npm:5.7.1" = { filename = "buffer-npm-5.7.1-513ef8259e-e2cf8429e1.zip"; sha512 = "e2cf8429e1c4c7b8cbd30834ac09bd61da46ce35f5c22a78e6c2f04497d6d25541b16881e30a019c6fd3154150650ccee27a308eff3e26229d788bbdeb08ab84"; };
 "@types/jest@npm:26.0.24" = { filename = "@types-jest-npm-26.0.24-47c570294f-ae39675412.zip"; sha512 = "ae39675412f08d884926254e9b12bfd2b5a4e4d204c94d3148cb942174a474930d0c60540133c968f22241d4712b7940c96cbc883096eb326a4d5b206fb78bd0"; };
+"@emurgo/cardano-serialization-lib-browser@npm:11.0.0-rc.6" = { filename = "@emurgo-cardano-serialization-lib-browser-npm-11.0.0-rc.6-f8fcee633c-a5f154735a.zip"; sha512 = "a5f154735a0a4bc9a608bd7dcd431050819dda200ad8824bffb2a49342a83abb0048b5b453ea7090e603df0f883692015eec274921702e4c10e2714f06db363c"; };
```
rhyslbw, line 145: Flagging for review of the nix lock file issue. What determines the order of this file?

image

packages/core/src/Cardano/types/Script.ts
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/946ca05a728386d06962959c2997dd07059329dd

rhyslbw (changes requested):
It's now clear to me we need to look at removing the `Cardano` module scoping, but it will be covered by future refactorings, as determined by SPIKE `ADP-2004`.

packages/core/src/Cardano/types/Script.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export interface All {
+  __type: ScriptNativeType.All;
+  scripts: ScriptNative[];
+}
+
+export interface NOf {
+  __type: ScriptNativeType.NOf;
+  scripts: {
+    // TODO: type alias for 'k'
```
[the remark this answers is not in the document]
AngelCastilloB: This was already taken care of in another PR that is already merged (I just rebased this branch to have that change)

packages/core/src/Cardano/util/numeric.ts
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable no-use-before-define */
+
+export declare type Ratio = number;
```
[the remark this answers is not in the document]
AngelCastilloB: not really, I just kept it from Ogmios types, actually I am a bit confused by the over aliasing I am seeing all over the place, but I am not sure what should be the good practice regarding this or what the guideline so I am just keeping it as it is, this is the first language where I see this is done, usually the field name plus the documentation is enough, in some cases aliasing makes sense, but most of the time to me look superfluous ( stuff like { epochLengh: EpochLenght } vs simply just { epochLengh: number}), I would lean to just use primitive types and document properly.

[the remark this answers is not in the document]
rhyslbw: Domain-type aliases are valuable since they guard against passing invalid values that may happen to match primitive types, and it also improves documentation since a single type alias' documentation is then inherited by all consumers.

packages/e2e/test/web-extension/extension/background/config.ts
```diff
@@ -91,7 +91,7 @@ export const utxoProvider = (async () => {
       return blockfrostUtxoProvider(await blockfrostApi!);
     }
     case 'http': {
-      return utxoHttpProvider({ adapter: axiosFetchAdapter, baseUrl: env.UTXO_PROVIDER_PARAMS.url, logger });
+      return utxoHttpProvider({ adapter: axiosFetchAdapter, baseUrl: env.UTXO_PROVIDER_PARAMS.baseUrl, logger });
```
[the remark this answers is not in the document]
AngelCastilloB: I will split in a different commit

packages/e2e/test/wallet/SingleAddressWallet/delegation.test.ts
```diff
@@ -87,9 +87,11 @@ describe('SingleAddressWallet/delegation', () => {
       filters: { status: [Cardano.StakePoolStatus.Active] },
       pagination: { limit: 2, startAt: 0 }
     });
-    return activePools.pageResults.filter(({ id }) => id !== delegateeBefore1stTx)[
-      Math.floor(Math.random() * activePools.pageResults.length) - 1
-    ].id;
+
+    const filteredPools = activePools.pageResults.filter(({ id }) => id !== delegateeBefore1stTx);
```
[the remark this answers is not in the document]
AngelCastilloB: I will split in a different commit

packages/core/src/Cardano/types/Block.ts
```diff
@@ -1,9 +1,26 @@
...
-import { Hash28ByteBase16, Hash32ByteBase16, OpaqueString, typedBech32 } from '../util';
+import { Hash28ByteBase16, Hash32ByteBase16, OpaqueString, UInt32, typedBech32 } from '../util';
 import { InvalidStringError } from '../..';
+import { Lovelace, PoolId } from '.';
 
-export { BlockNo } from '@cardano-ogmios/schema';
+/**
+ * The block size in bytes
+ */
+export type BlockSize = number;
```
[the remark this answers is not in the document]
mkazlauskas: I find `Bytes` a little confusing, it could be a byte array or a hex string. How about `NumBytes`?

packages/core/src/Cardano/util/numeric.ts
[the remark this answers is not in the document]
mkazlauskas: Types that use OpaqueString guards against passing invalid values. Type aliases don't, do they :thinking:? I find it helpful for documentation, but doc comments are even more helpful as they show up in 'intellisense' popup

packages/core/src/Cardano/types/Script.ts
```diff
@@ -1,18 +1,26 @@
...
+export enum ScriptType {
+  Native = 'native',
+  Plutus = 'plutus'
+}
+
+/**
+ * native script kind.
+ */
+export enum NativeScriptKind {
+  RequireSignature = 0,
```
mkazlauskas, line 18: Do these numbers correspond to something in cbor? If not I think strings is nicer

```suggestion
  RequireSignature = 'sig',
```

packages/core/src/Cardano/types/Script.ts
```diff
@@ -140,3 +178,51 @@ export type NativeScript =
...
+ * is generally no requirement that they be similar or compatible in any way.
+ */
+export enum PlutusLanguageVersion {
+  /**
+   * PlutusV1 was the initial version of Plutus, introduced in the Alonzo hard fork.
+   */
+  PlutusV1 = 1,
+
+  /**
+   * PlutusV2 was introduced in the Vasil hard fork.
```
mkazlauskas, line 196: Nice docs :+1:

packages/blockfrost/src/BlockfrostToCore.ts
```diff
@@ -26,16 +26,16 @@ export const BlockfrostToCore = {
...
   ): ProtocolParametersRequiredByWallet => ({
-    coinsPerUtxoByte: Number(blockfrost.coins_per_utxo_word),
-    maxCollateralInputs: Number(blockfrost.max_collateral_inputs),
-    maxTxSize: Number(blockfrost.max_tx_size),
-    maxValueSize: Number(blockfrost.max_val_size),
-    minFeeCoefficient: blockfrost.min_fee_a,
-    minFeeConstant: blockfrost.min_fee_b,
-    minPoolCost: Number(blockfrost.min_pool_cost),
-    poolDeposit: Number(blockfrost.pool_deposit),
+    coinsPerUtxoByte: BigInt(blockfrost.coins_per_utxo_word ?? 0),
```
mkazlauskas, line 29: This can lead to weird bugs. It would be better to throw an error (BigInt already does that, and blockfrost impl already wraps it in ProviderError)

```suggestion
    coinsPerUtxoByte: BigInt(blockfrost.coins_per_utxo_word!),
```

packages/core/src/Cardano/util/numeric.ts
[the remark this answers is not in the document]
rhyslbw: > Type aliases don't, do they ?

Yikes, that's surprising:
https://www.typescriptlang.org/play?#code/C4TwDgpgBAYg9nKBeKA7ArgWwEYQE4DcAsAFCiRQBCAhnsmlroaaQMZyoDOwUY1nneIhQAKAGYIAXLAQBKZAD4o7LnAA2EAHRq4Ac3FziJNh25RstaTTooAjEb4ChIi3lkEgA

OK, scrap that aspect. I still think it's worth doing for centralising documentation. Can we not configure IDEs to access type doc blocks?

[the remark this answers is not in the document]
mkazlauskas: > Can we not configure IDEs to access type doc blocks?

Maybe, I think it comes from TypeScript tooling that IDE then utilizes

packages/cardano-services/test/NetworkInfo/NetworkInfoHttpService.test.ts
```diff
@@ -339,7 +339,7 @@ describe('NetworkInfoHttpService', () => {
 
       it('successful request', async () => {
         const response = await provider.currentWalletProtocolParameters();
-        expect(response.maxTxSize).toBeDefined();
+        expect(response?.maxTxSize).toBeDefined();
```
mkazlauskas, line 342: This looks suspicious, is this needed? :thinking:

packages/cardano-services/src/ChainHistory/openApi.json
```diff
@@ -204,7 +204,13 @@
         "type": "object",
         "properties": {
           "blob": { "anyOf": [{ "$ref": "#/components/schemas/Map" }, { "$ref": "#/components/schemas/Undefined" }] },
-          "scripts": {
+          "nativeScripts": {
```
mkazlauskas, line 207: Is this right? I think it's still just `scripts` array

mkazlauskas (changes requested):

packages/core/src/Cardano/types/Script.ts
AngelCastilloB: These are the ids of the script kinds in CBOR, previously I had two enums, one with the numbers (needed to map from CBOR to core types) and one with the strings, but in the end I merged them in a single enum and used it in both places (mapping to and from CBOR)

packages/core/src/Cardano/util/numeric.ts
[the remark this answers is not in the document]
AngelCastilloB: Also IMHO I think the documentation should also be available where the type, structure or function prototype is defined, some ppl uses simple dev environments (IE vi or emacs), and I think even if I open a source file with a text editor the relevant documentation should be available without that need to go hunting for the documentation in different source files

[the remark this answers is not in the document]
AngelCastilloB: I.E I would spect all the relevant documentation regarding a function or class is located at where that function or class is defined

packages/cardano-services/src/ChainHistory/openApi.json
AngelCastilloB: Good catch, I forgot to update it

packages/core/src/Cardano/util/numeric.ts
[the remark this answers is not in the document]
mkazlauskas: Yep, everything that is exported should have doc comments

> some ppl uses simple dev environments (IE vi or emacs)

Those have plugins that provide tooling integration too, but I know what you're saying.

packages/cardano-services/src/ChainHistory/openApi.json
mkazlauskas: So it seems openapi doesn't validate the responses, only requests? I'm wondering if we can configure it to validate responses in test env

AngelCastilloB: I think it does, it just happens that those fields are optional and right now we are no retrieving them in any test it seems

packages/core/src/Cardano/util/numeric.ts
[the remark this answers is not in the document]
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/13a4d6b921fa14faf0c75a591a3805de1d3c13a5

packages/core/src/Cardano/types/Block.ts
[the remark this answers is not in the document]
AngelCastilloB: I agree with @mkazlauskas. But this is one of those cases where I think we are overdoing it a bit (type aliasing the type alias), this:

```
/**
 * The block size in bytes
 */
export type BlockSize = number;
```

To me is clear. In fact, the BlockSize type alias is probably also not needed, is used here:

```
export interface Block {
  ...
  size: BlockSize;
  ...
}
```

Which I think could be changed to:

```
export interface Block {
  ...
  /**
   * The block size in bytes
   */
  size: number;
  ...
}
```

But if we do need the alias, I would go with NumBytes as well

packages/core/src/Cardano/types/Script.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export interface NOf {
+  __type: ScriptNativeType.NOf;
+  scripts: {
+    // TODO: type alias for 'k'
+    [k: string]: ScriptNative[];
+  };
+}
+
+// TODO: delete or move this somewhere
```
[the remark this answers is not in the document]
AngelCastilloB: addressed here https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/946ca05a728386d06962959c2997dd07059329dd

packages/core/src/Cardano/types/Script.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export interface ScriptPlutus {
+  __type: ScriptType.Plutus;
+  bytes: util.HexBlob;
+  version: ScriptPlutusVersion;
+}
+
+export interface ScriptNative {
+  __type: ScriptType.Native;
+  // eslint-disable-next-line no-use-before-define
```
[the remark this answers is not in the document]
AngelCastilloB: addressed here https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/946ca05a728386d06962959c2997dd07059329dd. However, we are still suppressing that warning in the Script.ts, since native scripts are a recursive data structure (tree).

packages/cardano-services/test/ChainHistory/DbSyncChainHistoryProvider/__snapshots__/ChainHistoryBuilder.test.ts.snap
```diff
@@ -99,19 +99,19 @@ Map {
 
 exports[`ChainHistoryBuilder queryProtocolParams query protocol parameters from last epoch 1`] = `
 Object {
-  "coinsPerUtxoByte": NaN,
```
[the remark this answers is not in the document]
AngelCastilloB: This happened because I replaced the type from number to bigint, number return NaN when undefined, however, bigint doesn't have NaN so I am defaulting to 0 when the field is missing. This in fact could be a problem I am looking into this.

packages/core/src/Cardano/types/AuxiliaryData.ts
```diff
@@ -12,7 +9,7 @@ export type TxMetadata = Map<bigint, Metadatum>;
 
 export interface AuxiliaryDataBody {
   blob?: TxMetadata;
-  scripts?: Script[];
+  scripts?: Cardano.Script[] | undefined;
```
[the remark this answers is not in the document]
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/fbfd2e81d4cb3a8d033a7fffd34e1067dd8db39b

packages/cardano-services/test/NetworkInfo/NetworkInfoHttpService.test.ts
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/723a72ecc4fe7bc976472a733eed0c47e0fbaf38

packages/cardano-services/src/ChainHistory/openApi.json
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/de5acdc64e9567ec43e6142e650a2607ba6a428c

packages/blockfrost/src/BlockfrostToCore.ts
AngelCastilloB: @mkazlauskas you are correct. The thing is that we shouldn't throw an error here (according to the interface); these fields are optional, previously the Number constructor was being used and it was converting undefined values to 'NaN'; however; bigint has no NaN equivalent so I defaulted it to 0 (which is incorrect as you pointed out), this si what I did to correct this:

https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/c6d7dc4b26a5a050931020e910b6af6ea9730397

mkazlauskas: I still think it should throw since our `ProtocolParametersRequiredByWallet` has every field as required. The fix you made in c6d7dc4b26a5a050931020e910b6af6ea9730397 can return an object that does not conform to the return type - we should try avoid that if we can.

It should only be undefined in earlier eras. For a wallet, requiring them to be always defined should be fine.

AngelCastilloB: Ok I see it now, some of the test are omitting some of the fields on the mocks and I saw before we had this:

```
export type ProtocolParametersBabbage = OptionalUndefined<
  RecursivelyReplaceNullWithUndefined<Ogmios.ProtocolParametersBabbage>
>;
```

So I translated that to this:

```
export interface ProtocolParametersBabbage {
  minFeeCoefficient?: bigint;
  minFeeConstant?: bigint;
  maxBlockBodySize?: bigint;
  maxBlockHeaderSize?: bigint;
  maxTxSize?: bigint;
  stakeKeyDeposit?: bigint;
  poolDeposit?: bigint;
  poolRetirementEpochBound?: bigint;
  desiredNumberOfPools?: bigint;
  poolInfluence?: number;
  monetaryExpansion?: number;
  treasuryExpansion?: number;
  minPoolCost?: bigint;
  coinsPerUtxoByte?: bigint;
  maxValueSize?: bigint;
  collateralPercentage?: bigint;
  maxCollateralInputs?: bigint;
  protocolVersion?: ProtocolVersion;
  costModels?: CostModels;
  prices?: Prices;
  maxExecutionUnitsPerTransaction?: ExUnits;
  maxExecutionUnitsPerBlock?: ExUnits;
}
```

I alos missed the `Required<T>`, I will make the fields required them and fix the tests

mkazlauskas: Right, omitting unused fields in tests is fine.

> export type ProtocolParametersBabbage = OptionalUndefined<

This should probably be updated to have required fields, but let's just leave it as-is for now.

AngelCastilloB: Okay I think now it should be appropriately fixed: https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/a5b17a79f1044221c5a3b13d7c76fe74210089c1

The mock cant ommit any value now since we are making all fields required (it was a mock blockfrost response).

packages/cardano-services/test/ChainHistory/DbSyncChainHistoryProvider/__snapshots__/ChainHistoryBuilder.test.ts.snap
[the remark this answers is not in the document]
AngelCastilloB: ok addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/388#discussion_r951056969

packages/core/src/Cardano/types/Block.ts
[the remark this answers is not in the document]
rhyslbw: OK. Probably now just leaning towards `number`. Will retract request

rhyslbw (changes requested):
OK, ready for squashing and rescoping of the commits mention in open conversations.

packages/e2e/test/web-extension/extension/background/config.ts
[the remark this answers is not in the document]
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/3aec1a5a15885aebc2a248b9ab6785a78b88f418

packages/e2e/test/wallet/SingleAddressWallet/delegation.test.ts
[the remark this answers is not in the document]
AngelCastilloB: addressed here https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/cb6488821f25efb6f2c79fe06a81b670905ca519

packages/core/src/Cardano/types/Genesis.ts
```diff
@@ -1,22 +1,19 @@
...
   networkId: NetworkId;
-  activeSlotsCoefficient: Ratio;
-  securityParameter: UInt64;
-  epochLength: Epoch;
-  slotsPerKesPeriod: UInt64;
-  maxKesEvolutions: UInt64;
-  slotLength: Int64;
-  updateQuorum: UInt64;
+  activeSlotsCoefficient: number;
+  securityParameter: bigint;
```
rhyslbw, line 12: Some of these properties being typed as `bigint` caught my attention this time around. I now recall the reason Ogmios WSP schema has them as `UInt64`, however it's not specified to be `bigint` in the generated [TypeScript definition](https://github.com/CardanoSolutions/ogmios/blob/d16e4ede0d40d0b4d40f2fb47f265513942d845f/clients/TypeScript/packages/schema/src/index.ts#L102). 

https://github.com/CardanoSolutions/ogmios/blob/c45d1a1a65d7f2fa35a0d60be25fe2baf0f58172/architectural-decisions/accepted/013-typescript-client_bigint-parsing.md

 https://github.com/CardanoSolutions/ogmios/commit/0a91b15e6a761819e78f7e74d603104b6ea3a174

rhyslbw (changes requested):
Some of the properties being typed as `bigint` caught my attention this time around. I now recall the reason Ogmios WSP schema has them as `UInt64`, however, it's not specified to be `bigint` in the generated [TypeScript definition](https://github.com/CardanoSolutions/ogmios/blob/d16e4ede0d40d0b4d40f2fb47f265513942d845f/clients/TypeScript/packages/schema/src/index.ts#L102). 

https://github.com/CardanoSolutions/ogmios/blob/c45d1a1a65d7f2fa35a0d60be25fe2baf0f58172/architectural-decisions/accepted/013-typescript-client_bigint-parsing.md

 https://github.com/CardanoSolutions/ogmios/commit/0a91b15e6a761819e78f7e74d603104b6ea3a174

https://github.com/CardanoSolutions/ogmios/blob/d16e4ede0d40d0b4d40f2fb47f265513942d845f/clients/TypeScript/packages/client/src/util.ts#L27-L67

Given Ogmios has a valid understanding of legitimate value needing to be `bigint`, I think the simplest approach is to revert any changes from `number` to `bigint`

packages/core/src/CSL/coreToCsl/coreToCsl.ts
```diff
@@ -293,55 +293,55 @@ export const tx = ({ body, witness, auxiliaryData }: Cardano.NewTxAlonzo): Trans
 
 export const nativeScript = (script: Cardano.NativeScript): NativeScript => {
   let cslScript: NativeScript;
-  const type = script.__type;
+  const kind = script.kind;
```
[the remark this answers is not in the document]
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/388/commits/14fdefefa29108f67c0b6e0148ded660b91fb05d

AngelCastilloB:
> Some of the properties being typed as `bigint` caught my attention this time around. I now recall the reason Ogmios WSP schema has them as `UInt64`, however, it's not specified to be `bigint` in the generated [TypeScript definition](https://github.com/CardanoSolutions/ogmios/blob/d16e4ede0d40d0b4d40f2fb47f265513942d845f/clients/TypeScript/packages/schema/src/index.ts#L102).
> 
> https://github.com/CardanoSolutions/ogmios/blob/c45d1a1a65d7f2fa35a0d60be25fe2baf0f58172/architectural-decisions/accepted/013-typescript-client_bigint-parsing.md
> 
> [CardanoSolutions/ogmios@0a91b15](https://github.com/CardanoSolutions/ogmios/commit/0a91b15e6a761819e78f7e74d603104b6ea3a174)
> 
> https://github.com/CardanoSolutions/ogmios/blob/d16e4ede0d40d0b4d40f2fb47f265513942d845f/clients/TypeScript/packages/client/src/util.ts#L27-L67
> 
> Given Ogmios has a valid understanding of legitimate value needing to be `bigint`, I think the simplest approach is to revert any changes from `number` to `bigint`

The revert on the changes is done :)

packages/core/src/CSL/coreToCsl/coreToCsl.ts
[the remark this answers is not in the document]
rhyslbw: image

These changes here are not part of `refactor(core)!: renamed NativeScriptType to NativeScriptKind`

packages/core/src/Cardano/types/Genesis.ts
```diff
@@ -1,22 +1,19 @@
-import { Epoch, Int64, NetworkMagic, UInt64 } from '@cardano-ogmios/schema';
 import { Lovelace } from '.';
-import { NetworkId } from '../NetworkId';
-
-export type Ratio = number;
+import { NetworkId, NetworkMagic } from '../NetworkId';
 
 /**
- * A compact (without genesis UTxO) representation of the genesis configuration.
+ * A compact (without genesis UTxO) representation of the genesis configunumbern.
```
rhyslbw, line 5: :laughing: - Find replace fail

packages/core/src/Cardano/types/Script.ts
```diff
@@ -23,15 +31,20 @@ export enum NativeScriptType {
...
    * The hash of a verification key.
    */
   keyHash: Ed25519KeyHash;
 
   /**
-   * Script type.
+   * The native script kind.
    */
-  __type: NativeScriptType.RequireSignature;
+  kind: NativeScriptType.RequireSignature;
```
rhyslbw: These changes are in the wrong commit too

rhyslbw (changes requested):
`test(e2e): fix index out of bounds issue in delegationt test` also has a spelling mistake `delegationt`

packages/core/src/CSL/coreToCsl/coreToCsl.ts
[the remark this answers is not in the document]
AngelCastilloB: changes moved to appropriate commit

packages/core/src/Cardano/types/Genesis.ts
AngelCastilloB: fixed (I am squashing as I make this final changes)

packages/core/src/Cardano/types/Script.ts
AngelCastilloB: changes moved to appropriate commit

rhyslbw: It wasn't a clean separation, so I've just squashed them again with a more detailed commit message
05a83b9bba28b0d7fc958d5b30703ad3121749c2

rhyslbw (dismissed):

packages/core/src/CSL/cslToCore/cslToCore.ts
```diff
@@ -141,16 +140,16 @@ export const txBody = (body: CSL.TransactionBody): Cardano.NewTxBodyAlonzo => {
...
-  const result: BootstrapWitness[] = [];
+  const result: Cardano.BootstrapWitness[] = [];
   for (let i = 0; i < bootstraps.len(); i++) {
     const bootstrap = bootstraps.get(i);
     result.push({
       addressAttributes: bootstrap.attributes().toString(),
       chainCode: bootstrap.chain_code().toString(),
-      key: bootstrap.vkey.toString(),
-      signature: bootstrap.signature.toString()
+      key: typedHex(bootstrap.vkey.toString()),
```
mkazlauskas, line 151:
```suggestion
      key: Cardano.Ed25519PublicKey(bootstrap.vkey.toString()),
```

packages/core/src/CSL/cslToCore/cslToCore.ts
```diff
@@ -141,16 +140,16 @@ export const txBody = (body: CSL.TransactionBody): Cardano.NewTxBodyAlonzo => {
...
+      signature: typedHex(bootstrap.signature.toString())
```
mkazlauskas, line 152:
```suggestion
      signature: Cardano.Ed25519Signature(bootstrap.signature.toString())
```

packages/core/src/CSL/cslToCore/cslToCore.ts
```diff
@@ -1,6 +1,5 @@
 import { Asset, CSL, Cardano, SerializationFailure, util } from '../..';
-import { BootstrapWitness } from '@cardano-ogmios/schema';
-import { Ed25519KeyHash, NativeScriptType } from '../../Cardano';
+import { Ed25519KeyHash, NativeScriptKind, ScriptType, typedHex } from '../../Cardano';
```
mkazlauskas, line 2: Not an issue with this PR as it was already there, but we already have an import of the entire Cardano namespace
```suggestion
```

packages/core/src/Cardano/types/ProtocolParameters.ts
```diff
@@ -0,0 +1,50 @@
...
+  memory: number;
+  steps: number;
+}
+
+export interface ValidityInterval {
+  invalidBefore?: Slot;
+  invalidHereafter?: Slot;
+}
+
+export interface ProtocolParametersBabbage {
```
mkazlauskas, line 27: Don't think all of those parameters have to be nullable. Since this has 'Babbage' in it's name, it should probably have all parameters as non-nullable. Alternatively, we could have a `ProtocolParameters` type that has original parameters as non-nullable and parameters added in later eras as nullable.

Changes in this PR didn't change this (was already like that), so this could be addressed as a separate task.

mkazlauskas (changes requested):

packages/core/src/CSL/cslToCore/cslToCore.ts
rhyslbw: Done in 5f8f046abf806f189c96be63ca12934907da27bc

packages/core/src/CSL/cslToCore/cslToCore.ts
rhyslbw: Done in 96f35a2cd2c61a2e4295b77d7757d2a68bb9c4cf

packages/core/src/CSL/cslToCore/cslToCore.ts
rhyslbw: Done in 96f35a2cd2c61a2e4295b77d7757d2a68bb9c4cf

packages/core/src/Cardano/types/ProtocolParameters.ts
rhyslbw: I agree we should be reworking this. Ogmios types the properties as nullable since it's in context of a protocol update, which can include one or many alterations to the current blockchain params. We should have a base that's non-nullable then a version to model the update type.

mkazlauskas (approved):
Looks good!

packages/core/src/Cardano/types/ProtocolParameters.ts
rhyslbw: ADP-2165

rhyslbw (approved):

mkazlauskas (approved):

PR #389, opened 2022-08-11
build: fix interdependencies between tsconfig.json files

iccicci:
# Context

- `yarn test:build:verify` is not detecting all compile errors. The reason is: in some of our unit tests we are including files from `test` directories of other packages and we achieve this through the `include` root field of `packages/*/test/tsconfig.json` files with a list of `test` directories of other packages from which we need to import files. 
- We experienced some build inconsistencies between build environments. `rootDir` seems to take part in the reason (AFAIK still not perfectly identified) of these inconsistencies.

While investigating the problems I checked the relationships between following `tsconfig.json` options.

- `include`: this is the list of **all the files which compose the project**.  When it appears in our `test` directories it do not includes the files from the directory itself: this makes `yarn test:build:verify` to not check the same directory.
- `compilerOptions.composite`: allows the project to be target of `references.path` from another project. We have this set to `true` in all the projects of our repos, as it is set in the root `tsconfig.json` files and never modified in other files extending it directly or indirectly.
- `references.path`: allows the project to import files from other packages whose files are available on our disk and using these files rather than those present in the `npm` registry (i.e. other `@cardano-sdk` packages). The target package needs following configuration:
  ```json
  {
    "compilerOptions": {
      "composite": true,
      "noEmit": true
    }
  }
  ```
  Being the meaning of `"composite": true` "I want to makes this project target of some `references.path`" and being this not possible when `"noEmit": true` following configuration makes no sense:
  ```json
  {
    "compilerOptions": {
      "composite": true,
      "noEmit": true
    }
  }
  ```
  but it is exactly the configuration of our `test/tsconfig.json` file which is extended by all `packages/*/test/**/tsconfig.json` files.
- `compilerOptions.rootDir`: specifies the root directory within which all the files composing the project must be. Its default value is interesting:
  - when `"composite": true`: the directory of the `packege.json` file itself; i.e. `tsc` set it correctly;
  - when `"composite": false`: the longest path common to all the files `include`d in the project; i.e. `tsc` computes it correctly;

# Proposed Solution

- Added `"./**/*.ts"` to the `include` root filed of all `packages/*/test/tsconfig.json` files which defines it to solve the `yarn test:build:verify` problem.
- Set `"composite": false` in `test/tsconfig.json`:
  - it is useless since `"noEmit": true`;
  - removing it makes `tsc` to compute the right `rootDir` value (the default value) in all `packages/*/test/**/tsconfig.json` files.
- Simplified a bit the configuration of our project removing all the occurrences of `rootDir` as now its default value is correct both in `packages/*/src/tsconfig.json` files (`"composite": true`) and in `packages/*/test/**/tsconfig.json` files (`"composite": false`).

# Still missing

With this configuration, the `yarn test:build:verify` do not checks only the files in the right directory, but also in all the other `include`d `test` directories.
Simply a small waste of resources? No, a possible source of problems as well.
Let's take in exam the `packages/rabbitmq/test/tsconfig.json` file, it `include`s `packages/ogmios/test/mocks`.
This means that files in `packages/ogmios/test/mocks` directory are checked by `yarn test:build:verify` in `ogmios` workspace correctly using the `packages/ogmios/test/tsconfig.json` file **and** by `yarn test:build:verify` in `rabbitmq` workspace but using the `packages/rabbitmq/test/tsconfig.json` file.
So far it hasn't been a problem because all `packages/*/test/tsconfig.json` files have the same `compilerOptions`, but if we need to `include` `packages/e2e/test/` directory from another test directory we will simultaneously need at least to `exclude` `packages/e2e/test/web-extension` as its files requires different `compilerOptions`; or if we need to change `compilerOptions` in some `packages/*/test/**/tsconfig.json` files we need to take care whether this directory is `include`d or not in other tests...

To solve this I see two options:

1. moving all the features common to at least tests of two workspaces into the `@cardano-sdk/util-dev` package;
2. changing the `test/tsconfig.json` to make all `packages/*/test/` valid target of `references.path`

### DIFF
--- a/packages/blockfrost/src/tsconfig.json
+++ b/packages/blockfrost/src/tsconfig.json
@@ -1,8 +1,7 @@
 {
   "extends": "../../../tsconfig.json",
   "compilerOptions": {
-    "outDir": "../dist/cjs",
-    "rootDir": "."
+    "outDir": "../dist/cjs"
   },
   "references": [
     {
--- a/packages/cardano-services-client/src/tsconfig.json
+++ b/packages/cardano-services-client/src/tsconfig.json
@@ -1,8 +1,7 @@
 {
   "extends": "../../../tsconfig.json",
   "compilerOptions": {
-    "outDir": "../dist/cjs",
-    "rootDir": "."
+    "outDir": "../dist/cjs"
   },
   "references": [
     {
--- a/packages/cardano-services/src/tsconfig.json
+++ b/packages/cardano-services/src/tsconfig.json
@@ -1,8 +1,7 @@
 {
   "extends": "../../../tsconfig.json",
   "compilerOptions": {
-    "outDir": "../dist/cjs",
-    "rootDir": "."
+    "outDir": "../dist/cjs"
   },
   "references": [
     {
--- a/packages/cardano-services/test/entrypoints.txWorker.test.ts
+++ b/packages/cardano-services/test/entrypoints.txWorker.test.ts
@@ -6,6 +6,7 @@ import { RabbitMQContainer } from '../../rabbitmq/test/docker';
 import { createHealthyMockOgmiosServer, ogmiosServerReady } from './util';
 import { getRandomPort } from 'get-port-please';
 import { listenPromise, serverClosePromise } from '../src/util';
+import { logger } from '@cardano-sdk/util-dev';
 import http from 'http';
 import path from 'path';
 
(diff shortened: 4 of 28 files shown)


AngelCastilloB (approved):
LGTM

mkazlauskas (approved):
Looks good! I noticed `ts-jest` doesn't care about project references and does not build referenced projects, but it also seems to be the case in `master`. We should look if we can do something about it in a separate task.

rhyslbw (approved):
Thanks for the detailed investigation and resolution @iccicci :mechanical_arm:

PR #390, opened 2022-08-12
heat: Root `/health` endpoint for cardano-services HTTP server + HealthCheckResponse evolution

IvayloAndonov:
# Context
As an SRE Resident deploying `cardano-services`
I want to access the `/health` endpoint of each service
so that I only have to make a single request and not multiple.

# Proposed Solution
Screenshot 2022-08-12 at 10 50 40


packages/ogmios/src/CardanoNode/OgmiosCardanoNode.ts
```diff
@@ -93,4 +108,25 @@ export class OgmiosCardanoNode implements CardanoNode {
       throw CardanoNodeUtil.asCardanoNodeError(error) || new CardanoNodeErrors.UnknownCardanoNodeError(error);
     }
   }
+
+  async healthCheck() {
```
mirceahasegan, line 112: Maybe also add the return type? Not all the methods have it, but I think they should (even if CardanoNode has the return type)

mirceahasegan (approved):
Nice and clear @IvayloAndonov 👍 . The code looks good to me.

packages/core/src/Provider/Provider.ts
```diff
@@ -1,5 +1,17 @@
+import { Percent } from '../Cardano';
+import { Tip } from '@cardano-ogmios/schema';
+
 export type HealthCheckResponse = {
   ok: boolean;
+  localNode?: {
```
mkazlauskas, line 6: Shouldn't this only be on `CardanoNodeHealthCheckResponse`, like
```ts
interface HealthCheckResponse { ok: boolean }
interface CardanoNodeHealthCheckResponse extends HealthCheckResponse {
  localNode: { ... }
}
```

packages/core/src/Provider/Provider.ts
```diff
@@ -1,5 +1,17 @@
...
+    ledgerTip?: Tip;
+    networkSync?: Percent;
+  };
+};
+
+export type ServicesHealthCheckResponse = {
```
mkazlauskas, line 12: I think a more appropriate place for this is in `cardano-services` package as it's not per-provider, but per the http server.

packages/core/src/Provider/Provider.ts
```diff
@@ -1,5 +1,17 @@
...
+  ok: boolean;
+  services: HealthCheckResponse[];
```
mkazlauskas, line 14: This doesn't match the screenshots, shouldn't it be `{name, ok}`?

packages/cardano-services/src/Http/HttpServer.ts
```diff
@@ -65,6 +65,32 @@ export class HttpServer extends RunnableModule {
...
+    const servicesHealthCheckHandler = async (req: express.Request, res: express.Response) => {
+      this.logger.debug('/health', { ip: req.ip });
+      let body: ServicesHealthCheckResponse | Error['message'];
+      try {
+        const servicesHealth = await Promise.all(
+          this.#dependencies.services.map((service) =>
+            service.healthCheck().then(({ ok }) => ({
+              name: service.name,
+              ok
+            }))
```
mkazlauskas, line 78: Better be defensive, some bug in 1 healthcheck could mess up the entire `Promise.all`

```suggestion
            })).catch((err) => {
              logger.error(err);
              return { name: service.name, ok: false };
            })
```

mkazlauskas (changes requested):

packages/core/src/Provider/Provider.ts
IvayloAndonov: I followed the task requirements to extend the type as follows:
Screenshot 2022-08-16 at 15 32 30

mkazlauskas: I see, ok let's leave it as is then. There is value for all services to include this information.

packages/core/src/Provider/Provider.ts
IvayloAndonov: Not sure I understood the exact type mismatch? 
From the screenshots, root health check response type: `{ok, services}`, individual service health check type: `{name, ok}`

mkazlauskas: [HealthCheckResponse](https://github.com/input-output-hk/cardano-js-sdk/pull/390/files#diff-5913afa26cddc3441a4bc0483565c14043739d6ec4c32fa357ea50345892ccffR4) seems to be `{ok, localNode}` instead of `{ok, name}`

IvayloAndonov: Well, it depends on whether the HTTP server has started with dependent services on `cardano node` (`tx-submit` or `network-info`).
But you are right, the `services: HealthCheckResponse[]` type from the root should be different than the extended `HealthCheckResponse` with `localNode`.
So the question is do we want to show optional `localNode` data from the root health check endpoint or to be visible only on individual service health check endpoint?

mkazlauskas: Probably on individual services only. Tip on the node is not necessarily consistent with tip of the service data. For example, service dependent on ogmios is likely to be slightly ahead of the service dependent on db-sync. The benefit of having those on service level is that users can then reason about data consistency. Thinking about this, I think only 'tip' is necessary for service-level healthcheck and `networkSync?: Percent;` is probably only relevant on CardanoNode. Thoughts?

IvayloAndonov: Got it, about your thought, based on the requirements both `tip` and `networkSync` are grouped under 
`localNode?: {
          ledgerTip: lastKnownTip,
          networkSync: networkSynchronization
        }` so they will be populated only from `ogmios` dependent service `/health` as we have built-in [method](https://github.com/input-output-hk/cardano-js-sdk/pull/390/files#diff-f39a864229b1800d755f14614cd8496fe47244069e7b95d5ae851156616cf34bR114-R116) for it.

So you're thinking that `tip` should be mandatory on service level `/health` regardless the dependency type? So if it depends on db-sync only to query and map tip data from db rather than just `SELECT 1` as health check query ?

mkazlauskas: > So you're thinking that tip should be mandatory on service level /health regardless the dependency type?

Yeah, it's out of scope of this ticket though. Let's leave them grouped and optional for now and think about it when working on ADP-2128

packages/ogmios/src/CardanoNode/OgmiosCardanoNode.ts
IvayloAndonov: 753ce1beadd3e717fa08387b303b9209f456c66f

packages/core/src/Provider/Provider.ts
IvayloAndonov: be86382cbc1aadd303a9902ff3be8ba07273ccd6

packages/core/src/Provider/Provider.ts
IvayloAndonov: 41fa605dd9cf00bedd1371e10ba716464835ed66

packages/cardano-services/src/Http/HttpServer.ts
IvayloAndonov: 41fa605dd9cf00bedd1371e10ba716464835ed66

mkazlauskas (dismissed):
Nice work! :tada:

mirceahasegan (approved):

packages/core/src/Provider/Provider.ts
rhyslbw: `localNode` can only be the response from querying the node directly. We should probably make `CardanoNode` a mandatory dependency for all Providers for this consistency check alone.

rhyslbw (changes requested):
@IvayloAndonov please squash commits so I can review from a clean set.

packages/cardano-services/docker-compose.yml
```diff
@@ -114,6 +114,15 @@ services:
       - LOGGER_MIN_SEVERITY=${LOGGER_MIN_SEVERITY:-info}
       - SERVICE_NAMES=${SERVICE_NAMES:-asset,chain-history,network-info,rewards,stake-pool,tx-submit,utxo}
       - USE_QUEUE=${USE_QUEUE:-false}
+    healthcheck:
```
rhyslbw, line 117: This is valid when not in control of the `Dockerfile`, but it's better to define it as a build definition where possible so it's always available when using the image standalone. https://docs.docker.com/engine/reference/builder/#healthcheck

rhyslbw (changes requested):
This looks great @IvayloAndonov, just some minor issues:

The Ogmios implementation of `CardanoNode` will break in `feat(core): extend HealthCheckResponse`, so best to combine it with 25d81d143bf4b65f8a8f9350f1d230c53905aa92 as an unscoped commit.

packages/cardano-services/docker-compose.yml
IvayloAndonov: Change is applied here [964c1e2](https://github.com/input-output-hk/cardano-js-sdk/pull/390/commits/964c1e2c0b68edec9b24042ec4e0e927e4a49cd9)

IvayloAndonov:
> This looks great @IvayloAndonov, just some minor issues:
> 
> The Ogmios implementation of `CardanoNode` will break in `feat(core): extend HealthCheckResponse`, so best to combine it with [25d81d1](https://github.com/input-output-hk/cardano-js-sdk/commit/25d81d143bf4b65f8a8f9350f1d230c53905aa92) as an unscoped commit.

Both commits `feat(core): extend HealthCheckResponse` and `feat(ogmios): implement CardanoNode health check` are combined as an unscoped commit as suggested: https://github.com/input-output-hk/cardano-js-sdk/pull/390/commits/a2b8e2e9c5568ed4caed108c012e9e1cf2a98579

packages/cardano-services/src/Http/HttpServer.ts
IvayloAndonov: After the applied change here (nested catch block) it turned out we never go into `unhealthy` state of Docker's `healthcheck`  request we send each 30s `curl --fail http://0.0.0.0:3000/health || exit 1`  never would fail as the catch shallows the actual error and failed response. 
This [catch](https://github.com/input-output-hk/cardano-js-sdk/pull/390/files#diff-db63771378e2c7af74fab64645861ab1f0556c3320ce46fabf60eedf796ab08eR73-R76) block will handle the failures. @mkazlauskas

rhyslbw (approved):

packages/cardano-services/src/Http/HttpServer.ts
mkazlauskas: Shouldn't /health respond with 200 status regardless if it's healthy or not? cc @rhyslbw

rhyslbw: Yes. It's a successful HTTP request

IvayloAndonov: Addin the catch block https://github.com/input-output-hk/cardano-js-sdk/pull/390/commits/b3258cf016f6f82bcaf4e8ac8563e19b2c4772c2 @mkazlauskas

IvayloAndonov:
With the latest PR state, we do not terminate `http-server` container if some of the service dependencies are not healthy, instead, it is a running container with `unhealthy` state in Docker and if the dependency becomes alive again, then the server container state is healthy as well. Each service `/health` respond with 200 status regardless if it's healthy or not.

The screenshots shown in the PR desc show the exact Docker health check behavior.

mkazlauskas (dismissed):
Nice work! :tada:

mirceahasegan (dismissed):

rhyslbw (dismissed):
Great work @IvayloAndonov

rhyslbw (approved):

PR #396, opened 2022-08-18
Feat/tx builder prototype

mirceahasegan:
# Context

Currently, the wallet app uses intializeTx from SingleAddressWallet to either get the transaction data or any corresponding error that specifies the particular reason for failure. Problem: possible to validate an invalid transaction

# Proposed Solution

TxBuilder util separates the processes of transaction construction into individual components of OutputBuilder, SignedTx, ValidTx, TxBuilder

Todo:

- [x] Documentation
- [x] TxBuilder with outputs
- [x] OutputBuilder
- [x] Delegate
- [x] Metadata
- [x] OutputValidationError with full context and output
- [x] e2e tests
- [x] rebase and add `extraWitness` logic


packages/wallet/src/TxBuilder/buildTx.ts
```diff
@@ -0,0 +1,142 @@
...
+  auxiliaryData?: Cardano.AuxiliaryData;
+}
+
+const createSignedTx = async ({
+  wallet,
+  initializeTxResult,
+  props,
+  auxiliaryData
+}: CreateSignedTxProps): Promise<SignedTx> => {
+  // REVIEW: we have to update the ObservableWallet interface `finalizeTx` method
```
mkazlauskas, line 37: :+1:

packages/wallet/src/TxBuilder/buildTx.ts
```diff
@@ -0,0 +1,142 @@
...
+    } catch (error) {
+      const errors = Array.isArray(error) ? (error as TxBodyValidationError[]) : [error as TxBodyValidationError];
+      return {
+        errors,
+        isValid: false
+      };
+    }
+  }
+
+  // REVIEW: we could make this public and allow users to run validations at any time.
```
mkazlauskas, line 111: Not sure about this, users could just use `WalletUtil` directly. @rhyslbw what do you think?

packages/wallet/src/TxBuilder/buildTx.ts
```diff
@@ -0,0 +1,142 @@
...
+  async #validateOutputs(): Promise<OutputValidationError[]> {
+    const outputValidations =
+      this.partialTxBody.outputs && (await this.#util.validateOutputs(this.partialTxBody.outputs));
+
+    const errors = [...(outputValidations?.values() || [])]
+      .map((validation) => {
+        // REVIEW: Should we add the whole validation data, for more context
```
mkazlauskas, line 118: Yep, and the invalid output obj as well. Should probably be part of OutputValidationError. Might need to split it into 2 error types then?

packages/wallet/src/TxBuilder/OutputBuilder.ts
```diff
@@ -0,0 +1,106 @@
...
+  MaybeValidTxOut,
+  OutputBuilder,
+  OutputValidationMinimumCoin,
+  OutputValidationMissingRequired,
+  OutputValidationTokenBundleSize
+} from './types';
+import { OutputValidation } from '../types';
+import { OutputValidator } from '../services';
+
+// Typescript is complaining about DeepPartial<Cardano.TokenMap> methods being optional
```
mkazlauskas, line 14: Maybe best to not use DeepPartial then, but define a type that explicitly specifies  the Partial props

packages/wallet/src/TxBuilder/types.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+  : T;
+
+export enum TxOutputFailure {
+  minimumCoin = 'Minimum coin not met',
+  tokenBundleSizeExceedsLimit = 'Token Bundle Exceeds Limit',
+  missingRequiredFields = 'Mandatory fields address or coin are missing'
+}
+
+export class OutputValidationMissingRequired extends CustomError {
```
mkazlauskas, line 19: All other error types in the SDK have an `Error` suffix. Let's follow that for consistency.

```suggestion
export class OutputValidationMissingRequiredError extends CustomError {
```

packages/wallet/src/TxBuilder/OutputBuilder.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  /**
+   * @param assetId id
+   * @param quantity To remove an asset, set quantity to 0
+   */
+  asset(assetId: Cardano.AssetId, quantity: bigint): OutputBuilder {
+    const existingAssets = this.partialOutput?.value?.assets;
+    const assets: Cardano.TokenMap = isTokenMap(existingAssets)
+      ? existingAssets.set(assetId, quantity)
```
mkazlauskas, line 67: Other methods do not mutate `partialOutput`. Let's keep it consistent and not mutate the `Map`. Instead, create a new map:
```suggestion
      ? new Map(existingAssets.entries()).set(assetId, quantity)
```

packages/wallet/src/TxBuilder/buildTx.ts
```diff
@@ -0,0 +1,171 @@
...
+import { OutputBuilderSaw, toOutputValidationError } from './OutputBuilder';
+import { OutputValidator, StakeKeyStatus, createWalletUtil, deepEquals } from '../services';
+import { SignTransactionOptions } from '../KeyManagement';
+
+/**
+ * Minimal sub-type of ObservableWallet that is used by TxBuilder
+ */
+export type TxBuilderObservableWallet = Pick<
+  ObservableWallet,
+  'utxo' | 'protocolParameters$' | 'finalizeTx' | 'initializeTx' | 'submitTx' | 'delegation'
```
mkazlauskas, line 25: A potential improvement here is to narrow down the types of some of the picked properties too

packages/wallet/src/TxBuilder/buildTx.ts
```diff
@@ -0,0 +1,171 @@
...
+  #observableWallet: TxBuilderObservableWallet;
+  #util: OutputValidator;
+
+  constructor(observableWallet: TxBuilderObservableWallet, util: OutputValidator) {
+    this.#observableWallet = observableWallet;
+    this.#util = util;
+  }
+
+  addOutput(txOut: Cardano.TxOut): TxBuilder {
+    this.partialTxBody.outputs = [...(this.partialTxBody.outputs || []), txOut];
```
mkazlauskas, line 62: Maybe let's make it never mutate the `partialTxBody` object too but always recreate as in OutputBuilder?

packages/wallet/src/TxBuilder/buildTx.ts
```diff
@@ -0,0 +1,171 @@
...
+    return this;
+  }
+
+  buildOutput(txOut?: DeepPartial<Cardano.TxOut>): OutputBuilder {
+    return new OutputBuilderSaw(this.#util, txOut);
+  }
+
+  async delegate(poolId: Cardano.PoolId): Promise<TxBuilder> {
+    // REVIEW: should we have an error if no reward account is found. Not sure it's valid...
+    //   Should I be more defensive here and handle an error or empty rewardAccounts?
```
mkazlauskas, line 77: It would have to be a bug for this to happen. I suppose you could have some sort of "InternalError" type to generalize cases like this.

packages/wallet/src/TxBuilder/buildTx.ts
```diff
@@ -0,0 +1,171 @@
...
+    const rewardAccounts = await firstValueFrom(this.#observableWallet.delegation.rewardAccounts$);
+
+    // Discard previous delegation and prepare for new one
+    this.partialTxBody.certificates = [];
+
+    for (const rewardAccount of rewardAccounts) {
+      // REVIEW: Is it ok to create certificates for all the reward accounts?
```
mkazlauskas, line 84: Yes, OK for now. Make sure to add a doc comment for this method to highlight this behavior

packages/wallet/src/types.ts
```diff
@@ -29,10 +29,18 @@ export type InitializeTxProps = {
   requiredExtraSignatures?: Cardano.Ed25519KeyHash[];
 };
 
+// REVIEW: this interface is not used anywhere in the SDK. Was it requested by another team? Can we remove it?
```
mkazlauskas, line 32: Probably just trash, good to remove

mkazlauskas (changes requested):

packages/wallet/src/TxBuilder/buildTx.ts
rhyslbw: I would leave to keep the interface strictly for building at this point

packages/wallet/src/TxBuilder/buildTx.ts
```diff
@@ -0,0 +1,142 @@
...
+  // to include options?: SignTransactionOptions. Otherwise there's no way to pass the `props` to finalizeTx
+  props; // TODO mh: remove me
+  const tx = await wallet.finalizeTx(initializeTxResult, auxiliaryData);
+  return {
+    submit: () => wallet.submitTx(tx),
+    tx
+  };
+};
+
+class TxBuilderSaw implements TxBuilder {
```
rhyslbw, line 48:
```suggestion
class ObserveableWalletTxBuilder implements TxBuilder {
```
?

packages/wallet/src/TxBuilder/types.ts
```diff
@@ -0,0 +1,106 @@
+import { Cardano } from '@cardano-sdk/core';
+import { CustomError } from 'ts-custom-error';
+import { InputSelectionError } from '@cardano-sdk/cip2';
+import { SignTransactionOptions } from '../KeyManagement';
+
+export type DeepPartial<T> = T extends object
```
rhyslbw, line 7: Maybe source this in `@cardano-sdk/util`

packages/wallet/src/TxBuilder/types.ts
```diff
@@ -0,0 +1,106 @@
...
+   */
+  buildOutput(txOut?: DeepPartial<Cardano.TxOut>): OutputBuilder;
+  /**
+   * Add StakeDelegation and (if needed) StakeKeyRegistration certificate
+   */
+  delegate(poolId: Cardano.PoolId): TxBuilder;
+  setMetadata(metadata: Cardano.TxMetadata): TxBuilder;
+  build(): Promise<MaybeValidTx>;
+
+  // REVIEW: assuming fields below are not needed for Lace right now, so out of scope of MVP:
```
rhyslbw, line 116:
```suggestion
  // TODO:
      - mint
      - burn
```

packages/wallet/test/integration/buildTx.test.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  it('can remove outputs and build', async () => {
+    const txBuilder = buildTx(wallet);
+    const txTwoOutputs = txBuilder.addOutput(mocks.utxo[0][1]).addOutput(mocks.utxo[1][1]);
+    const tx = await txTwoOutputs.removeOutput(mocks.utxo[0][1]).build();
+    assertTxIsValid(tx);
+    expect(tx.body.outputs.length).toBe(2);
+  });
+
+  it('builds transactions that are not modified by subsequent builder changes', async () => {
```
rhyslbw, line 66: I'm not following based on the test content. Is there maybe a missing assertion between line 63 and 64?

packages/wallet/test/integration/buildTx.test.ts
```diff
@@ -0,0 +1,143 @@
...
+      expect(error1.failure === TxOutputFailure.coinMissing);
+    }
+
+    expect(error2 instanceof OutputValidationError).toBeTruthy();
+    if (error2 instanceof OutputValidationError) {
+      expect(error2.failure === TxOutputFailure.tokenBundleSizeExceedsLimit);
+    }
+  });
+
+  describe('buildOutput', () => {
```
rhyslbw, line 124: It would be good to have a `describe` block for each method up the top of the suite, to ensure the precise behavior for each is covered, as this test suite will become quite large. Tests to demonstrate sequences can then be positioned below

packages/wallet/test/integration/buildTx.test.ts
```diff
@@ -0,0 +1,143 @@
...
+      const assetQuantity = 100n;
+      const assets = new Map([[assetId, assetQuantity]]);
+      const address = Cardano.Address('addr_test1vr8nl4u0u6fmtfnawx2rxfz95dy7m46t6dhzdftp2uha87syeufdg');
+      const datum = Cardano.Hash32ByteBase16('3e33018e8293d319ef5b3ac72366dd28006bd315b715f7e7cfcbd3004129b80d');
+      const output1Coin = 10_000_000n;
+      const output2Base = mocks.utxo[0][1];
+
+      const txBuilder = buildTx(wallet);
+
+      const output1 = await txBuilder
```
rhyslbw, line 147: Maybe it would be better communicated as:

``` ts
const output1 = await txBuilder.buildOutput();
      output1
        .address(address)
        .coin(output1Coin)
        .asset(assetId, assetQuantity)
        .build();
```

... since I read it the first time as "adding an address to a transaction"

mkazlauskas: `build()` has to be awaited (while `buildOutput()` does not)

packages/wallet/src/TxBuilder/buildTx.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    if (errors.length > 0) {
+      throw errors;
+    }
+    return [];
+  }
+}
+
+/**
+ * REVIEW: `ObservableWallet.buildTx()` would be nice, but it adds quite a lot of complexity
```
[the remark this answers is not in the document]
mirceahasegan: [2c333cc](https://github.com/input-output-hk/cardano-js-sdk/pull/396/commits/2c333ccbeb6420d7ec1e79fb36955cc7b79cb982)

packages/wallet/test/integration/buildTx.test.ts
mirceahasegan: [1d5d8f7](https://github.com/input-output-hk/cardano-js-sdk/pull/396/commits/1d5d8f785403e6238a2206d8be8673b8f3220f7e)

packages/wallet/src/TxBuilder/buildTx.ts
```diff
@@ -79,20 +96,47 @@ class TxBuilderSaw implements TxBuilder {
...
+      const errors = Array.isArray(error) ? (error as TxBodyValidationError[]) : [error as TxBodyValidationError];
       return {
-        errors: [error as TxBodyValidationError],
+        errors,
         isValid: false
       };
     }
   }
+
+  // REVIEW: we could make this public and allow users to run validations at any time.
```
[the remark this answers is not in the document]
mkazlauskas: 1. Agree, the error should have a reference to the output
2. Agree
3. Not at this point. Users can just use WalletUtil directly.

packages/wallet/src/TxBuilder/types.ts
```diff
@@ -0,0 +1,126 @@
...
+
+import { OutputValidation } from '../types';
+import { SignTransactionOptions } from '../KeyManagement';
+
+// Does not transform TokenMap into Partial<TokenMap> because it would also make it's methods and map types optional
+// and that forces users to do many checks before being able to use any of it's properties and methods
+export type PartialTxOut = DeepPartial<Cardano.TxOut, Cardano.TokenMap>;
+
+export enum TxOutputFailure {
+  minimumCoin = 'Minimum coin not met',
```
mkazlauskas, line 14: Let's follow the convention that other errors are already using
```suggestion
  MinimumCoin = 'Minimum coin not met',
```

packages/wallet/src/TxBuilder/buildTx.ts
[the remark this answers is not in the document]
mirceahasegan: In the meantime I added both the output reference and the original output error, so 2 & 3 are already fixed.
This was an old comment, not visible because I missed the fact that I have to "submit the review". So there is lots of commits and comments that appeared only just now, even if they are old 🤦

packages/wallet/test/integration/buildTx.test.ts
```diff
@@ -0,0 +1,415 @@
...
+
+    it('can remove output without mutating partialTxBody', () => {
+      const initialPartialTxBody = txBuilder.partialTxBody;
+      txBuilder.removeOutput(output);
+      expect(txBuilder.partialTxBody.outputs).toEqual([output2]);
+      // check that original partialTxBody was not mutated
+      assertObjectRefsAreDifferent(txBuilder.partialTxBody, initialPartialTxBody);
+    });
+
+    it('can add remove outputs one by one', () => {
```
rhyslbw, line 83:
```suggestion
    it('can remove outputs one by one', () => {
```

packages/wallet/test/integration/buildTx.test.ts
rhyslbw: Huge improvement, well done!

rhyslbw (changes requested):
Looks great! Just a typo found in the tests

packages/wallet/test/integration/buildTx.test.ts
mirceahasegan: [231a55c](https://github.com/input-output-hk/cardano-js-sdk/pull/396/commits/231a55c8ae512493304f6585d6bd230b5c65ced3)

packages/wallet/src/TxBuilder/types.ts
mirceahasegan: [231a55c](https://github.com/input-output-hk/cardano-js-sdk/pull/396/commits/231a55c8ae512493304f6585d6bd230b5c65ced3)

packages/util/package.json
```diff
@@ -63,7 +63,8 @@
   },
   "dependencies": {
     "lodash": "^4.17.21",
-    "serialize-error": "^8"
+    "serialize-error": "^8",
+    "type-fest": "^2.19.0"
```
mkazlauskas, line 67: Shouldn't this be in `devDependencies`? Seems like it only has types

packages/wallet/src/TxBuilder/types.ts
```diff
@@ -0,0 +1,126 @@
...
+import { CustomError } from 'ts-custom-error';
+import { DeepPartial } from '@cardano-sdk/util';
+import { InputSelectionError } from '@cardano-sdk/cip2';
+
+import { OutputValidation } from '../types';
+import { SignTransactionOptions } from '../KeyManagement';
+
+// Does not transform TokenMap into Partial<TokenMap> because it would also make it's methods and map types optional
+// and that forces users to do many checks before being able to use any of it's properties and methods
+export type PartialTxOut = DeepPartial<Cardano.TxOut, Cardano.TokenMap>;
```
mkazlauskas, line 11: This type doesn't look right
image

packages/wallet/src/TxBuilder/types.ts
```diff
@@ -0,0 +1,126 @@
...
+
+export interface ValidTxOutData {
+  readonly txOut: Cardano.TxOut;
+}
+
+export type ValidTxOut = Valid<ValidTxOutData>;
+export type InvalidTxOut = Invalid<TxOutValidationError>;
+export type MaybeValidTxOut = ValidTxOut | InvalidTxOut;
+
+export interface OutputBuilder {
```
mkazlauskas, line 62: More doc comments on this type would be great

packages/wallet/src/TxBuilder/types.ts
```diff
@@ -0,0 +1,126 @@
...
+  readonly tx: Cardano.NewTxAlonzo;
+  submit(): Promise<void>;
+}
+
+export interface ValidTxBody {
+  readonly body: Cardano.NewTxBodyAlonzo;
+  readonly auxiliaryData?: Cardano.AuxiliaryData;
+  readonly extraWitness?: Partial<Cardano.Witness>;
+
+  sign(props?: SignTransactionOptions): Promise<SignedTx>;
```
mkazlauskas, line 86: Options should come from previous step. `initializeTx` has to know the exact number of signatures that the transaction is going to have, so there's no point of having this argument, as it should also be supplied when building the tx.

```suggestion
  sign(): Promise<SignedTx>;
```

packages/wallet/src/TxBuilder/types.ts
```diff
@@ -0,0 +1,126 @@
...
+}
+
+export type ValidTx = Valid<ValidTxBody>;
+export type InvalidTx = Invalid<TxBodyValidationError>;
+export type MaybeValidTx = ValidTx | InvalidTx;
+
+export interface TxBuilder {
+  partialTxBody: Partial<Cardano.NewTxBodyAlonzo>;
+  auxiliaryData?: Cardano.AuxiliaryData;
+  extraWitness?: Partial<Cardano.Witness>;
```
mkazlauskas, line 96: This doesn't seem to be used anywhere. It should probably need to be changed into [these 2](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/wallet/src/types.ts#L30-L31) props (same for `ValidTxBody` interface) that should then be passed to both `initializeTx` and `finalizeTx`

packages/wallet/src/TxBuilder/types.ts
```diff
@@ -0,0 +1,126 @@
...
+   * Add StakeDelegation and (if needed) StakeKeyRegistration certificate
+   *
+   * @throws IncompatibleWalletError when no reward accounts are found
+   */
+  delegate(poolId: Cardano.PoolId): Promise<TxBuilder>;
+  setMetadata(metadata: Cardano.TxMetadata): TxBuilder;
+  build(): Promise<MaybeValidTx>;
+
+  // TODO:
+  // - setMint
```
mkazlauskas, line 117: This could also be useful and should be straightforward to implement

```suggestion
  // - setMint
  // - addMetadata(label: bigint, metadatum: Cardano.Metadatum);
  // - removeMetadata(label: bigint);
```

mkazlauskas (changes requested):

packages/util/package.json
mirceahasegan: I'm a bit confused too. Yes, it has only types.
At the same time, would a user of the `util` library, importing DeepPartial, be able to see what `Primitive` type is? (since it comes from type-fest)

mkazlauskas: Good question, don't know, could you please try and see how it works?

packages/wallet/src/TxBuilder/types.ts
mirceahasegan: It's because DeepPartial dived into Cardano.Address type and made all it's properties optional. It's not necessarily bad, but I can add it to the Omit types

mkazlauskas: Maybe something like: `Partial<Pick<TxOut, 'address' | 'datum'>> & {value: Partial<Value>}`

mirceahasegan: Yeap, this works fine. I'll update it

packages/wallet/src/TxBuilder/types.ts
mirceahasegan: I've added doc comments in the concrete implementations. Should I hoist them to the interfaces?

mkazlauskas: Yes please, I think concrete implementations inherit docs from the interface. Seems to be working with vscode intellisense, although not sure if it gets propagated to the generated docs of concrete implementations. If it doesn't then we might want to have the doc comments duplicated.

mkazlauskas (commented):
Also please commit changes to `yarn-project.nix`. It should change as `yarn.lock` changes

packages/wallet/src/TxBuilder/types.ts
mirceahasegan: Fixed in [d092381](https://github.com/input-output-hk/cardano-js-sdk/pull/396/commits/d092381a0921bb8258d6ebe3a1baca6f50a0f540)

packages/util/package.json
mirceahasegan: I have tried on a test project and it shows the types correctly.
[cb6c0bc](https://github.com/input-output-hk/cardano-js-sdk/pull/396/commits/cb6c0bc3fbf52e3131186a505b45f2d23879a324)

packages/wallet/src/TxBuilder/types.ts
mirceahasegan: [53ef574](https://github.com/input-output-hk/cardano-js-sdk/pull/396/commits/53ef574ab31271196145f491bc145e55c4774848) & [c4037b8](https://github.com/input-output-hk/cardano-js-sdk/pull/396/commits/c4037b828fc9dc3b4b1fad0446101aaa8f63f4cf)

packages/wallet/src/TxBuilder/types.ts
mirceahasegan: [53ef574](https://github.com/input-output-hk/cardano-js-sdk/pull/396/commits/53ef574ab31271196145f491bc145e55c4774848) & [c4037b8](https://github.com/input-output-hk/cardano-js-sdk/pull/396/commits/c4037b828fc9dc3b4b1fad0446101aaa8f63f4cf)

packages/wallet/src/TxBuilder/types.ts
mirceahasegan: Do you mean both `setMint` and `setMetadatum`, or just `setMetadatum`?

packages/wallet/src/TxBuilder/buildTx.ts
```diff
@@ -0,0 +1,262 @@
...
+ *
+ * @returns Promise with `SignedTx` object that has a {@link SignedTx.submit} method that submits the transaction.
+ */
+const createSignedTx = async ({
+  wallet,
+  initializeTxResult,
+  extraSigners,
+  signingOptions,
+  auxiliaryData
+}: CreateSignedTxProps): Promise<SignedTx> => {
```
mkazlauskas, line 62: Props seem to be all the same, so I think CreateSignedTxProps type is redundant
```suggestion
}: FinalizeTxProps): Promise<SignedTx> => {
```

mkazlauskas (changes requested):
Looking good. Would be great to refactor 1 of the e2e tests to use tx builder before merging.

packages/wallet/src/TxBuilder/types.ts
mirceahasegan: [9b5e7cb](https://github.com/input-output-hk/cardano-js-sdk/pull/396/commits/9b5e7cbbb5fcb0d2de9dea91ff28c8ce4ae02511)

packages/wallet/src/TxBuilder/buildTx.ts
mirceahasegan: [9b5e7cb](https://github.com/input-output-hk/cardano-js-sdk/pull/396/commits/9b5e7cbbb5fcb0d2de9dea91ff28c8ce4ae02511)

packages/wallet/src/TxBuilder/types.ts
mirceahasegan: I have hoisted the docs to the interface here: [58a49fb](https://github.com/input-output-hk/cardano-js-sdk/pull/396/commits/58a49fb87bfb7d17aa07f9c26147f9ffcff70cd7).
The generated documentation propagates nicely to the concrete classes too.

yarn-project.nix
```diff
@@ -6,8 +6,8 @@
 
 let
 
-  yarnPath = ./.yarn/releases/yarn-3.2.1.cjs;
-  lockfile = ./yarn.lock;
+  yarnPath = ./packages/util/.yarn/releases/yarn-3.2.1.cjs;
```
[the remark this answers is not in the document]
mirceahasegan: I will take care of this one after the e2e tests. I will have to rebase on master before doing it because there it contains nix updates too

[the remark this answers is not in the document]
mirceahasegan: > Looking good. Would be great to refactor 1 of the e2e tests to use tx builder before merging.

[4525f30](https://github.com/input-output-hk/cardano-js-sdk/pull/396/commits/4525f30384f6957da64cd9ba99944e88e143616d)

packages/e2e/test/wallet/SingleAddressWallet/delegation.test.ts
```diff
@@ -58,6 +47,16 @@ const waitForTx = async (wallet: ObservableWallet, { hash }: Transaction.TxInter
   await waitForWalletStateSettle(wallet);
 };
 
+// eslint-disable-next-line func-style, jsdoc/require-jsdoc
+function assertTxIsValid(tx: MaybeValidTx): asserts tx is ValidTx {
```
mkazlauskas, line 51: Could export those from `wallet/test/util`, [add 'includes'](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/wallet/test/tsconfig.json#L8) in tsconfig and [import with relative path](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/wallet/test/Transaction/createTransactionInternals.test.ts#L4)

mkazlauskas (approved):
Great work @mirceahasegan :rocket: Please squash and request a final review from @rhyslbw and @AngelCastilloB 

Just 1 small change request below (no need to do a fixup commit for this one)

packages/e2e/test/wallet/SingleAddressWallet/delegation.test.ts
mirceahasegan: Updated as part of the rebase+squash : [bbc0574](https://github.com/input-output-hk/cardano-js-sdk/pull/396/commits/bbc0574c21c69e324f351edefad84e317c7f46f7)

PR #397, opened 2022-08-19
Feat/adp 2119 stake pool avarage

iccicci:
# Context

We need to provide average margin and average reward for all active stake pools.

# Proposed Solution

- Added a dedicated method to the interfaces
- Added a `DbSyncStakePoolProvider` sup-provider (`StakePoolAverages`) which actually uses the first one to perform the required queries to compute the data.

### DIFF
--- a/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/StakePoolAverages.ts
+++ b/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/StakePoolAverages.ts
@@ -1,0 +1,183 @@
+import {
+  Cardano,
+  Provider,
+  ProviderError,
+  ProviderFailure,
+  StakePoolAveragesResponse,
+  StakePoolProvider,
+  StakePoolQueryOptions
+} from '@cardano-sdk/core';
+import { Logger } from 'ts-log';
+
+/**
+ * The required dependencies to create a StakePoolAverages provider.
+ */
+export interface StakePoolAveragesDependencies {
+  logger: Logger;
+  stakePoolProvider: StakePoolProvider;
+}
+
+/**
+ * The sub-provider which takes care of computing and caching average values for stake pools.
+ */
+export class StakePoolAverages implements Provider {
+  /**
+   * The cached result. It is a Promise to reflect the status of the compute job:
+   * - if resolved: the compute job exited correctly and the Promise stores the result
+   * - if rejected: the compute job exited with error and the Promise stores the error
+   * - if still pending: the compute job is still working
+   *
+   * Therefore it can be used (and it is) to await the end of the running compute job
+   */
+  #cachedResult?: Promise<StakePoolAveragesResponse>;
+
+  /**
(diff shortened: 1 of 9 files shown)


packages/cardano-services/test/StakePool/StakePoolHttpService.test.ts
```diff
@@ -248,7 +249,7 @@ describe('StakePoolHttpService', () => {
           expect(clearCacheSpy).toHaveBeenCalled();
 
           expect(await epochMonitor.getLastKnownEpoch()).toEqual(greaterEpoch);
-          expect(cache.keys().length).toEqual(0);
+          // expect(cache.keys().length).toEqual(0);
```
iccicci, line 252: This implementation invalidates this assertion. ATM just commented: the final implementation could not invalidate this.

iccicci:
Hi @rhyslbw , @mkazlauskas and @IvayloAndonov ;

I opened this draft pull request to check if I'm going in the right direction.
Thank you

packages/core/src/Provider/StakePoolProvider/types/StakePoolProvider.ts
```diff
@@ -66,6 +86,10 @@ export interface StakePoolStats {
 }
 
 export interface StakePoolProvider extends Provider {
+  /**
+   * @returns {StakePoolAveragesResponse} Stake pool averages
+   */
+  getAverages: () => Promise<StakePoolAveragesResponse>;
```
mkazlauskas, line 92: I think it should be part of StakePoolStats instead of a separate endpoint

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/StakePoolAverages.ts
```diff
@@ -0,0 +1,183 @@
...
+      marginPartialTotal++;
+
+      if (metrics.apy !== undefined) {
+        rewardPartialSum += metrics.apy;
+        rewardPartialTotal++;
+      }
+    };
+
+    do {
+      const { pageResults, totalResultCount } = await this.#stakePoolProvider.queryStakePools(queryOptions);
```
mkazlauskas, line 132: `queryStakePools` is quite heavy and not fully cached. I suggest

- replace StakePoolProvider dependency with Pool (db) dependency
- write ligher queries for just the data that StakePoolAverages needs

mkazlauskas (changes requested):

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/StakePoolAverages.ts
iccicci: > write ligher queries for just the data that StakePoolAverages needs

I'll try to dig into the `queries.ts` file, but I do not exclude I'll need some support on this: I still don't know the schema very well.

iccicci:
This is a wrong approach

PR #398, opened 2022-08-19
fix(e2e): add missing logger param

mirceahasegan:
# Context

Blockfrost e2e tests update after logger mandatory param was added

### DIFF
--- a/packages/e2e/test/blockfrost/getAsset.test.ts
+++ b/packages/e2e/test/blockfrost/getAsset.test.ts
@@ -1,6 +1,7 @@
 import * as envalid from 'envalid';
 import { Cardano } from '@cardano-sdk/core';
 import { assetProviderFactory } from '../../src/factories';
+import { logger } from '@cardano-sdk/util-dev';
 
 // Verify environment.
 export const env = envalid.cleanEnv(process.env, {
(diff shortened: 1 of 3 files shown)


packages/e2e/test/blockfrost/getAsset.test.ts
```diff
@@ -1,6 +1,7 @@
 import * as envalid from 'envalid';
 import { Cardano } from '@cardano-sdk/core';
 import { assetProviderFactory } from '../../src/factories';
+import { dummyLogger } from 'ts-log';
```
mkazlauskas, line 4:
```suggestion
import { TestLogger } from 'util-dev';
```

mkazlauskas (changes requested):
I think [this file](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/e2e/test/web-extension/extension/background/config.ts) is also missing some updates for logger params

packages/e2e/test/blockfrost/getAsset.test.ts
mirceahasegan: I force pushed the changes because they are contained and simple

mirceahasegan: We agreed offline that webextension logs are ok to go to the browser console because that is where they will execute

mkazlauskas (approved):
Thanks @mirceahasegan

rhyslbw (approved):

rhyslbw (changes requested):
This is a fix, but in the scope of tests only, so should therefore be typed as `test`

rhyslbw (approved):

mkazlauskas (approved):

PR #399, opened 2022-08-22
fix: messenger connect race

mkazlauskas:
# Context

web-extension messaging sometimes works unreliably (observable doesn't emit)

# Proposed Solution

Problem seems to be that `message$` emits before it is subscribed to (probably synchronously). Replacing `Subject` with `ReplaySubject(1)` seems to fix the issue. 2095fc944c6d1eea47b400d9eaaf507f232aa704

- [fix(wallet): create contextLogger with non-undefined 'name'](https://github.com/input-output-hk/cardano-js-sdk/commit/89a93bd417fa0940fa9b09028d71a8da7900bc72)
- [feat(web-extension): slightly improve messengers log output](https://github.com/input-output-hk/cardano-js-sdk/commit/6d3b82b380959d6d0f5047059a0e1539e4f71e0a)
- Consolidate 2 `pipe` calls into 1 (not part of the main fix in 2095fc944c6d1eea47b400d9eaaf507f232aa704)

### DIFF
--- a/packages/web-extension/src/messaging/BackgroundMessenger.ts
+++ b/packages/web-extension/src/messaging/BackgroundMessenger.ts
@@ -32,7 +32,7 @@ export const createBackgroundMessenger = ({ logger, runtime }: MessengerDependen
     return channel;
   };
   const onPortMessage = (data: unknown, port: MessengerPort) => {
-    logger.debug('[BackgroundMessenger] Port message', data, port);
+    logger.debug(`[BackgroundMessenger(${port.name})] message`, data);
     const { message$ } = channels.get(port.name)!;
     message$.next({ data, port });
   };
@@ -43,7 +43,7 @@ export const createBackgroundMessenger = ({ logger, runtime }: MessengerDependen
     const newPorts = new Set(ports$.value);
     newPorts.delete(port);
     ports$.next(newPorts);
-    logger.debug('[BackgroundMessenger] Port disconnected', port);
+    logger.debug(`[BackgroundMessenger(${port.name})] disconnected`, port);
   };
   const onConnect = (port: MessengerPort) => {
     const { ports$ } = getChannel(port.name);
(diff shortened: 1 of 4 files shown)


mirceahasegan (approved):

packages/web-extension/src/messaging/BackgroundMessenger.ts
```diff
@@ -22,7 +22,7 @@ export const createBackgroundMessenger = ({ logger, runtime }: MessengerDependen
   const getChannel = (channelName: ChannelName) => {
     let channel = channels.get(channelName);
     if (!channel) {
-      channels.set(channelName, (channel = { message$: new Subject(), ports$: new BehaviorSubject(new Set()) }));
+      channels.set(channelName, (channel = { message$: new ReplaySubject(1), ports$: new BehaviorSubject(new Set()) }));
```
rhyslbw, line 25: Given this is a fix we're not completely satisfied with, please add a comment inline to ensure there's some context to an otherwise curious need to use this special `Subject`

packages/web-extension/src/messaging/NonBackgroundMessenger.ts
```diff
@@ -26,7 +26,7 @@ export const createNonBackgroundMessenger = (
   let delay = initialDelay;
   let isDestroyed = false;
   const port$ = new BehaviorSubject<MessengerPort | null | 'destroyed'>(null);
-  const message$ = new Subject<PortMessage>();
+  const message$ = new ReplaySubject<PortMessage>(1);
```
rhyslbw, line 29: Can use the same comment here too.

rhyslbw (changes requested):
The fix is fine given it's satisfying the requirements, just requesting inline comments for context.

mirceahasegan (approved):

rhyslbw (approved):
Great work @mkazlauskas :sunflower:

PR #403, opened 2022-08-25
feat: Enhance /metrics with custom HTTP server wide HealthCheckResponse

IvayloAndonov:
# Context

The current metrics available are not enough for alerting and it is required that more suitable metrics be exposed.
We need to include a Prometheus gauge with the same object returned.

# Proposed Solution

The exposed Prometheus Gauge interface allows values only with the type `number`.Due to that reason, we expose the server healthiness with the values 0 or 1 with Prometheus-scrapable format under `/metrics`

With a healthy state:
Screenshot 2022-08-30 at 10 17 20

With an unhealthy state:

Screenshot 2022-08-29 at 20 45 49

- add a health check to the `/metrics` endpoint using [Prometheus Gauge](https://github.com/siimon/prom-client#gauge)

### DIFF
--- a/packages/cardano-services/src/Http/HttpServer.ts
+++ b/packages/cardano-services/src/Http/HttpServer.ts
@@ -1,3 +1,4 @@
+import { Gauge, Registry } from 'prom-client';
 import { HttpServerConfig, ServiceHealth, ServicesHealthCheckResponse } from './types';
 import { HttpService } from './HttpService';
 import { Logger } from 'ts-log';
@@ -9,7 +10,6 @@ import bodyParser from 'body-parser';
 import express from 'express';
 import expressPromBundle from 'express-prom-bundle';
 import http from 'http';
-import promClient from 'prom-client';
 
 export const CONTENT_TYPE = 'Content-Type';
 export const APPLICATION_JSON = 'application/json';
@@ -24,12 +24,36 @@ export class HttpServer extends RunnableModule {
   public server: http.Server;
   #config: HttpServerConfig;
   #dependencies: HttpServerDependencies;
+  #healthGauge: Gauge<string>;
 
   constructor(config: HttpServerConfig, { logger, ...rest }: HttpServerDependencies) {
     super(config.name || 'HttpServer', logger);
     this.#config = config;
     this.#dependencies = { logger, ...rest };
   }
+
+  async #getServicesHealth() {
+    const servicesHealth: ServiceHealth[] = await Promise.all(
+      this.#dependencies.services.map((service) =>
+        service
+          .healthCheck()
+          .then(({ ok }) => ({
+            name: service.name,
+            ok
+          }))
+          .catch((error) => {
+            this.logger.error(error);
+            return { name: service.name, ok: false };
+          })
+      )
+    );
+
+    return {
+      ok: servicesHealth.every((service) => service.ok),
+      services: servicesHealth
+    };
+  }
+
   async initializeImpl(): Promise<void> {
     this.app = express();
     this.app.use(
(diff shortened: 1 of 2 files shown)


rhyslbw (changes requested):
A simple test of the response could be `res.data.split('\n').includes('healthcheck 1');` (just guessing at the assertion here)

packages/cardano-services/src/Http/HttpServer.ts
```diff
@@ -38,16 +61,38 @@ export class HttpServer extends RunnableModule {
         reviver: (key, value) => (key === '' ? fromSerializableObject(value) : value)
       })
     );
+
     if (this.#config?.metrics?.enabled) {
+      const promRegistry = new Registry();
+      const getServicesHealth = async () => await this.#getServicesHealth();
+      void ((registry) =>
```
rhyslbw, line 68: I think it would be better to assign this to a private class field, then call `reset` on it within `shutdownImpl`

packages/cardano-services/src/Http/HttpServer.ts
```diff
@@ -38,16 +61,38 @@ export class HttpServer extends RunnableModule {
...
           includeMethod: true,
-          promRegistry: new promClient.Registry(),
+          promRegistry,
           ...this.#config.metrics.options
         })
       );
-      this.logger.info(`Prometheus metrics configured at ${this.#config.metrics.options?.metricsPath || '/metrics'}`);
+
+      this.logger.info(
+        `Prometheus metrics: ${(await promRegistry.getMetricsAsArray()).map(({ name }) => name)} configured at ${
```
rhyslbw, line 90: Nice :+1:

rhyslbw (changes requested):

packages/cardano-services/src/Http/HttpServer.ts
IvayloAndonov: Actually, I already did it initially in this way you mentioned (assigning the registry to the private class field, registering the gauge metric to the registry with `this.#registry.registerMetric(metric)` and then calling `this.#registry.clear()` within `shutdownImpl`  which should to `Remove all metrics from the registry`  based on the exposed prom client interface but it turned out didn't work the and metric reference still remained uncleared and the error popped up between test cases: `A metric with the name ${metric.name} has already been registered.` even we have a new instance of HTTP service and reinitialize it in each unit test.

Due to that issue, I dig it into that package and saw there is another way of registering so I chose to approach it this way with `registers: [registry]`.

I could apply back changes with another approach with `registry.registerMetric(metric)` and `clear` it in`shutdownImpl` if needed in order to debug it together but already spent much time on it.

@rhyslbw

packages/cardano-services/src/Http/HttpServer.ts
```diff
@@ -30,6 +30,29 @@ export class HttpServer extends RunnableModule {
     this.#config = config;
     this.#dependencies = { logger, ...rest };
   }
+
+  async #getServicesHealth() {
+    const servicesHealth: ServiceHealth[] = await Promise.all(
+      this.#dependencies.services.map((service) =>
```
gytis-ivaskevicius, line 36: Maybe it would be nice to expose each service separately?

gytis-ivaskevicius (approved):

packages/cardano-services/src/Http/HttpServer.ts
rhyslbw: > assigning the registry to the private class field

Sorry, I mean to assign the _Gauge_ instance to a private field. Rather than call `reset` on shutdown though, I think `set`ting it to `0` is more appropriate

packages/cardano-services/src/Http/HttpServer.ts
rhyslbw: We decided to go with a single reduced server health metric as the goal here is to just signal when the `http-server` should be restarted. There's a more appropriate HTTP endpoint `/health` that returns a structured JSON response with granular information.

packages/cardano-services/src/Http/HttpServer.ts
IvayloAndonov: Usage is updated in https://github.com/input-output-hk/cardano-js-sdk/pull/403/commits/c79eb0bfd75947dae19a1f1db0a6b9e341fa4b6b, thanks for the help! @rhyslbw

rhyslbw (dismissed):
Nice work @IvayloAndonov. Please squash

iadmytro (approved):

rhyslbw (approved):

PR #404, opened 2022-08-25
feat: Allow signing transactions with arbitrary keys

AngelCastilloB:
# Context

Currently if you initializeTx and then add additional signatures manually before submitting tx, the transaction will fail for having too low fee. Input selection has to account for entire transaction size, so it has to be aware of # of extra signatures.

# Proposed Solution

Create mock of witnesses witnesses when we calculate the fee.

Things to keep in mind:

1. Mock witnesses have different lengths depending on the type which changes the expecting fee
2. Witnesses are a set so we need to get rid of duplicates to avoid over-estimating the fee


packages/wallet/src/KeyManagement/types.ts
```diff
@@ -138,8 +138,15 @@ export interface Ed25519KeyPair {
  */
 export type GetPassword = (noCache?: true) => Promise<Uint8Array>;
 
+/**
+ * @returns A Ed25519 transaction signature.
```
AngelCastilloB, line 142: Inaccurate docs

packages/wallet/src/KeyManagement/types.ts
```diff
@@ -138,8 +138,15 @@ export interface Ed25519KeyPair {
...
+ */
+// eslint-disable-next-line no-use-before-define
+export type SignTx = (txInternals: TxInternals, options?: SignTransactionOptions) => Promise<Cardano.Signatures>;
```
mkazlauskas, line 145: If the goal is for this to be compatible with KeyAgent's `signTransaction`, then this might be slightly better:
```suggestion
export type SignTx = KeyAgent['signTransaction'];
```

packages/wallet/src/KeyManagement/util/stubSignTransaction.ts
```diff
@@ -9,14 +9,19 @@ const randomPublicKey = () => Cardano.Ed25519PublicKey(Array.from({ length: 64 }
...
+  inputResolver: InputResolver,
+  signOptions?: SignTransactionOptions
+): Promise<Cardano.Signatures> => {
+  const mockSignature = Cardano.Ed25519Signature(
+    // eslint-disable-next-line max-len
+    'ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff'
   );
+
+  const signatureKeyPaths = await ownSignatureKeyPaths(txBody, knownAddresses, inputResolver);
+  const totalSignature = signatureKeyPaths.length + (signOptions?.extraSigners?.length || 0);
```
mkazlauskas, line 21: This is a problem because extraSigners can return more than 1 signature, but we also don't want to call it here so there's no way of knowing how many signatures we'll have. Perhaps let's change `SignTx` to return just 1 signature? Then if needed you could create a util like `keyAgentAsSigner(keyAgent: KeyAgent): SignTx`

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -446,6 +449,19 @@ export class SingleAddressWallet implements ObservableWallet {
     this.#logger.debug('Shutdown');
   }
 
+  async #getSignatures(txInternals: TxInternals, options?: SignTransactionOptions) {
+    let signatures: Cardano.Signatures = await this.keyAgent.signTransaction(txInternals, options);
+
+    if (options?.extraSigners) {
```
mkazlauskas, line 455: Since KeyAgent takes in options that include the extraSigners, it would make sense for KeyAgents to call the extraSigners, or else KeyAgent.signTransaction has an option that's never used. Alternatively we could remove it from options used in KeyAgent.signTransaction and leave it a sign option only in scope of ObservableWallet/SingleAddressWallet

mkazlauskas (changes requested):
Looking good, just a few change requests. Also, please [pass through](https://github.com/input-output-hk/cardano-js-sdk/commit/70e9b41c27d98d668af4c13ff9df39055078c536#diff-a97e442b4ed0aab61c11e7c2950105185c135e6870439d107887939005dbb2f5R370) 'mint' to input selection as a separate commit

packages/wallet/src/KeyManagement/util/stubSignTransaction.ts
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/404/commits/1745bf74c984767cac4ca69730a1da8f2ee40a90

packages/wallet/src/SingleAddressWallet.ts
AngelCastilloB: addressed here https://github.com/input-output-hk/cardano-js-sdk/pull/404/commits/1745bf74c984767cac4ca69730a1da8f2ee40a90

packages/core/src/Cardano/types/Asset.ts
```diff
@@ -14,7 +14,7 @@ export const AssetName = (value: string): AssetName => {
       throw new InvalidStringError('too long');
     }
   }
-  return value as unknown as AssetName;
+  return value.toLowerCase() as unknown as AssetName;
```
AngelCastilloB, line 17: These opaque string types encapsulated a byte array as a hex string, however, we need to be careful when comparing with these since a HEX string is case INsensitive, and can have valid representations in upper or lower case characters. We must either make sure they are in the same case all the time (byte converting them to lower or upper case) or better yet, store this as byte arrays (since this is essentially what they are) and make byte to byte comparisons. I understand the convenience of having those as strings instead of byte arrays, however, I think when we use them, we always need to convert them to byte arrays (pass down to serialization lib).

I think we may as well just keep them as byte arrays internally from the start (keeping the string constructor) and just add a toString() method that converts back to its hexadecimal representation.

rhyslbw: > I think we may as well just keep them as byte arrays internally from the start

Agree

packages/e2e/test/local-network/local-network.test.ts
```diff
@@ -108,4 +109,190 @@ describe('Local Network', () => {
...
+          __type: Cardano.ScriptType.Native,
+          keyHash: Cardano.Ed25519KeyHash(keyHash),
+          kind: Cardano.NativeScriptKind.RequireSignature
+        }
+      ]
+    };
+
+    const policyId = Cardano.nativeScriptPolicyId(script);
+    const assetName = Cardano.AssetName('BABABABABABABA');
+    const assetId = Cardano.AssetId(policyId.toString() + assetName.toString());
```
AngelCastilloB, line 254: I think we should introduce here the higher abstraction **AssetClass** domain concept:

In haskell we have:

``` haskell
newtype AssetClass = AssetClass { unAssetClass :: (CurrencySymbol, TokenName) }
```


So TokenMap, instead of taking the asset id directly, could take an asset class:

``` javascript
let assetClass = {
  tokenName: "XXXX", // This is the asset name
  currencySymbol: "YYYY" // This is the policy ID
}
```
https://github.com/input-output-hk/plutus/blob/4b40b31893a8a31456946670af981b06d71be6d5/plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs#L138

We could then get the asset ID under the hood to add to the transaction.

@rhyslbw @mkazlauskas

AngelCastilloB: Btw using the asset class concept, the lovelace coins should also be part of the token map as:

``` javascript
let assetClass = {
  tokenName: 'lovelace',
  currencySymbol: '' // The lovelace asset currency symbol is an empty string
}
```

packages/wallet/src/KeyManagement/types.ts
mkazlauskas: There's a problem with the refactor: it will now call `signTransaction` for every extra key path, which will result in bad UX (e.g. for InMemoryKeyAgent it will call `getPassword` multiple times). Original implementation with `additionalKeyPaths` only needed to decrypt root private key once to sign with all key paths.

Maybe let's consider `extraSigners` only for external keys, and if the goal is to use the same wallet's key agent then user is supposed to use 'additionalKeyPaths'?

packages/e2e/test/local-network/local-network.test.ts
rhyslbw: Noted. Deferring to a future ticket.

packages/core/src/Cardano/types/Asset.ts
mkazlauskas: Nice catch. We should probably apply this change to all opaque strings that are hex strings. Byte array does seem like a nice way to store it internally, but I'm not 100% sold on it because it would make it more difficult to debug: byte strings are not very human readable. Maybe it could work if if the type had a hex string getter.

packages/wallet/src/KeyManagement/types.ts
AngelCastilloB: The extra signatures will probably come from different key agents, but I do understand what you mean, if the same key agent is performing all the signatures it will have to authorize several times.

I will add the SignTransactionOptions so the user can choose how to add the extra signatures, via adding additional key paths for the signing operation or by adding extra signers

AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/404/commits/4ec4aadd1582e7a836d93599be9e2ddad4f556c3

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -446,6 +455,19 @@ export class SingleAddressWallet implements ObservableWallet {
     this.#logger.debug('Shutdown');
   }
 
+  async #getSignatures(txInternals: TxInternals, extraSigners?: TransactionSigner[]) {
+    const signatures: Cardano.Signatures = await this.keyAgent.signTransaction(txInternals);
```
mkazlauskas, line 459:
```suggestion
    const signatures: Cardano.Signatures = await this.keyAgent.signTransaction(txInternals, signingOptions);
```

packages/e2e/test/local-network/local-network.test.ts
```diff
@@ -108,4 +109,190 @@ describe('Local Network', () => {
     wallet1.shutdown();
     wallet2.shutdown();
   });
+
+  it('will mint an NFT using native scripts.', async () => {
```
mkazlauskas, line 113: Would be best to make this test network-agnostic. Shouldn't be too hard?

packages/e2e/test/local-network/local-network.test.ts
```diff
@@ -108,4 +109,190 @@ describe('Local Network', () => {
...
+
+    const signedTx = await wallet.finalizeTx(finalizeProps);
+    await wallet.submitTx(signedTx);
+
+    // Wait until wallet is aware of the minted token.
+    await firstValueFrom(
+      wallet.balance.utxo.total$.pipe(filter(({ assets }) => (assets ? assets.has(assetId) : false)))
+    );
+
+    // Wait until wallet is aware of the burned token.
```
mkazlauskas, line 195: This `await` seems redundant though
```suggestion
    // Wait until wallet is aware of the minted token.
```

mkazlauskas (changes requested):

packages/wallet/src/SingleAddressWallet.ts
AngelCastilloB: Addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/404/commits/5a045e1d9b2a4c6faf5dbb042f038da9fb2730a7 (and was properly linted here https://github.com/input-output-hk/cardano-js-sdk/pull/404/commits/480ab9087c32cf72254f2609d04d91aa49d5c3e6)

packages/e2e/test/local-network/local-network.test.ts
AngelCastilloB: This was particular test was removed, the nft test we already have will cover this uses cases, that test was just made network agnostic :)

packages/e2e/test/local-network/local-network.test.ts
AngelCastilloB: Agree, and also this test was just removed since our nft e2e test now covers this case

packages/e2e/test/wallet/SingleAddressWallet/nft.test.ts
```diff
@@ -1,83 +1,163 @@
...
+      tx: unsignedTx
+    };
+
+    const signedTx = await wallet.finalizeTx(finalizeProps);
+    await wallet.submitTx(signedTx);
+
+    // Wait until wallet is aware of the minted tokens.
+    await firstValueFrom(
+      wallet.balance.utxo.total$.pipe(
+        filter(({ assets }) => (assets ? assetIds.every((element) => assets.has(element)) : false))
```
mkazlauskas, line 158: Maybe instead of awaiting for balance await for a matching NFT
```
const nfts = await firstValueFrom(
      combineLatest([wallet.assets$, wallet.balance.utxo.total$]).pipe(
        filter(([assets, balance]) => assets.size === balance.assets?.size),
        map(([assets]) => [...assets.values()].filter((asset) => !!asset.nftMetadata))
      )
    );
```
and assert that nft props match to the ones created in metadata

mkazlauskas (changes requested):

packages/core/src/Cardano/types/Script.ts
```diff
@@ -226,3 +228,26 @@ export interface PlutusScript {
...
+  Cardano.PolicyId(Buffer.from(nativeScript(script).hash().to_bytes()).toString('hex'));
+
+/**
+ * Gets the public key hash from the given Bech32 address.
+ *
+ * @param address The address to get the public key hash from.
+ * @returns the key hash.
+ */
+// TODO: Use opaque strings
+export const policyKeyHash = (address: Cardano.Address): string => {
```
mkazlauskas, line 248: Isn't this just [Ed25519KeyHash](https://github.com/input-output-hk/cardano-js-sdk/blob/e4197cdae87991abc7565946f50a9d82189ab7ec/packages/core/src/Cardano/types/Key.ts#L46)?

Also, this util is likely not needed. See comment below.

mkazlauskas (changes requested):
Looking good, just 1 more issue I noticed and need to move the utils from `/types` to `/util`

packages/e2e/test/wallet/SingleAddressWallet/nft.test.ts
```diff
@@ -1,83 +1,163 @@
...
-            utxoProvider: await utxoProviderFactory.create(env.UTXO_PROVIDER, env.UTXO_PROVIDER_PARAMS, logger)
+      { inputResolver: { resolveInputAddress: async () => null } }
+    );
+
+    const derivedAddress = await keyAgent.deriveAddress({
+      index: 0,
+      type: KeyManagement.AddressType.External
+    });
+
+    const keyHash = Cardano.policyKeyHash(derivedAddress.address);
```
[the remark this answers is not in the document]
AngelCastilloB: We cant derive the key hash just from the public key as it doesnt have enough information; the key hash needs the stake credential, the network id, and the pub key, which are all encoded in the address. If you notice the RewardAccount which is used in the other function is also an address (the staking address). I will, however, move that utility function to the place you suggest.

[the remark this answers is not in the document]
mkazlauskas: From `Cardano.policyKeyHash`:

> BaseAddress.from_address(cslAddress)!.payment_cred().to_keyhash()

Do we really need stake credential and network ID here? `payment_cred().to_keyhash()` should equal to just payment key hash as far as I understand

packages/e2e/test/wallet/SingleAddressWallet/nft.test.ts
```diff
@@ -115,34 +200,82 @@ describe('SingleAddressWallet.assets/nft', () => {
...
       mintOrBurnCount: 1,
-      name: '4e46542d66696c6573',
+      name: assetNames[TOKEN_METADATA_1_INDEX],
       nftMetadata: {
         description: ['NFT with different types of files'],
         files: [
           {
             mediaType: 'video/mp4',
             name: 'some name',
+            otherProperties: undefined,
```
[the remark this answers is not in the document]
AngelCastilloB: It was falling on me and marking those as differences. but you are right, is strange, I will look into it with more care and see whats going on

packages/e2e/test/wallet/SingleAddressWallet/nft.test.ts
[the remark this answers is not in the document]
mkazlauskas: I'm wondering why `payment` is of type `StakeCredential` :thinking: is this from CSL?

Anyway, my point was that you don't need to create a grouped address just to get the key hash of payment key. `derivePublicKey` should be sufficient. Or maybe I'm missing something here, let's chat about this tomorrow.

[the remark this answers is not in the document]
AngelCastilloB: Yes that's from the CSL; you are right, otherwise, the key hash would change between networks and it doesn't; I will dig into the CSL code and see how the payment_cred is actually being calculated.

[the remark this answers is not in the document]
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/404/commits/96640f3b9db0271f290550eeb86a2f9809ee9358

packages/e2e/test/wallet/SingleAddressWallet/nft.test.ts
AngelCastilloB: This test was moved to a different branch, but this comment was addressed there: https://github.com/input-output-hk/cardano-js-sdk/pull/414/commits/0ddd06266b0067738c01545215acfb2366a5e4e2

packages/e2e/test/wallet/SingleAddressWallet/nft.test.ts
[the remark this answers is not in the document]
AngelCastilloB: Addressed here (test moved to a different PR): https://github.com/input-output-hk/cardano-js-sdk/pull/414/commits/0ddd06266b0067738c01545215acfb2366a5e4e2

packages/core/src/Cardano/types/Script.ts
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/commit/96640f3b9db0271f290550eeb86a2f9809ee9358

mkazlauskas (approved):
Great work @AngelCastilloB! :tada:

rhyslbw (approved):
:rocket:

rhyslbw (approved):

PR #406, opened 2022-08-26
RabbitMQ load test strategy fix

iccicci:
# Context

The load test have a bad design: tries to submit multiple simultaneous transactions from the same address, this is causing a lot of double spending error.

# Proposed Solution

Changed the strategy of the test to send simultaneous transactions from distinct `SingleAddressWallet` accounts.

- Changed the `getWallet` to fix a bug preventing it to be parallelized and making required its `logger` parameter.
- Removed some occurrences of `dummyLogger`.

# Test report

This is the test result with standard configuration (10 simultaneous transactions).

```
  Test result
directly to ogmios              - submission   -> mempool:   113ms - mempool -> blockchain:  4973ms - total:  5086ms
without running serial   worker - start worker -> mempool:   205ms - mempool -> blockchain:  4018ms - total:  4628ms
with    running serial   worker - submission   -> mempool:   750ms - mempool -> blockchain:  3936ms - total:  4686ms
without running parallel worker - start worker -> mempool:   154ms - mempool -> blockchain:  4515ms - total:  5062ms
with    running parallel worker - submission   -> mempool:   245ms - mempool -> blockchain:  4425ms - total:  4670ms
```

This is the test result with 200 simultaneous transactions. I also tried with some more simultaneous transactions, but the test was opening too many connections... Last: in this test output there is no reference to the _directly to ogmios_ test: it seems ogmios server is not able to handle too many simultaneous transactions, so I excluded that test case when testing with more than 30 simultaneous transactions.

```
  Test result
without running serial   worker - start worker -> mempool:  5285ms - mempool -> blockchain:  5740ms - total: 11689ms
with    running serial   worker - submission   -> mempool:  6142ms - mempool -> blockchain:  5349ms - total: 11491ms
without running parallel worker - start worker -> mempool:  3663ms - mempool -> blockchain:  5518ms - total:  9963ms
with    running parallel worker - submission   -> mempool:  4853ms - mempool -> blockchain:  5320ms - total: 10173ms
```

# Side note

It seems to ogmios (the package, not the server) arrives some bad logger as I'm receiving following error each time the test starts a tx worker:

```
bunyan usage error: /home/cicci/iohk/cardano-js-sdk/node_modules/@cardano-ogmios/client/node_modules/ws/lib/websocket.js:246: attempt to log with an unbound log method: `this` is: <ref *1> WebSocket
```

I haven't investigated if the bad logger is actually built in ogmios package or in our packages and passed to some ogmios function... I would consider this a separated task.


rhyslbw (approved):
Extension to the existing test looks good @iccicci

package.json
```diff
@@ -91,6 +91,7 @@
     "jest": "^28.1.3",
     "lerna": "^5.2.0",
     "lint-staged": "^13.0.1",
+    "node-gyp": "^9.0.0",
```
mkazlauskas, line 94: Why is this a new dependency? :thinking:

mkazlauskas (approved):
Looks good, just not sure about the node-gyp dependency

package.json
iccicci: I added it as without it the command `yarn local-network:up --build` was failing to install `ssh` as it needs `node-gyp` but is not in its dependencies.
This is strange as it seems this is happening only on my local machine...

PR #408, opened 2022-08-29
Uniform ttl parameter of InMemoryCache methods to be expressed in seconds

iccicci:
# Context

Some methods of `InMemoryCache` class accept a `ttl` parameter; sometimes they expect it expressed in seconds sometimes in minutes.

# Proposed Solution

Changed the interface of `InMemoryCache` class to make all the methods to accept the `ttl` parameter expressed in seconds; 

# Changes Introduced

- Since the `cli` script accepts TTLs in minutes, to not make this breaking change for all the `cli` script users: the parameter validator converts the TTLs from minutes to seconds.

### DIFF
--- a/packages/cardano-services/src/InMemoryCache/InMemoryCache.ts
+++ b/packages/cardano-services/src/InMemoryCache/InMemoryCache.ts
@@ -7,17 +7,22 @@ export class InMemoryCache {
   #cache: NodeCache;
   #ttlDefault: number;
 
-  constructor(cacheTtlInMins: number, cache: NodeCache = new NodeCache()) {
-    this.#ttlDefault = cacheTtlInMins * 60;
+  /**
+   *
+   * @param ttl The default time to live in seconds
+   * @param cache The cache engine. It must extend NodeCache
+   */
+  constructor(ttl: number, cache: NodeCache = new NodeCache()) {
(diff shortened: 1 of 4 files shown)


packages/cardano-services/src/InMemoryCache/InMemoryCache.ts
```diff
@@ -50,7 +55,8 @@ export class InMemoryCache {
    *
    * @param key Cache key
    * @param value A value to cache
-   * @param ttl The time to live in seconds.
+   * @param ttl Cache duration in seconds
```
rhyslbw, line 58: "cache duration" was the original terminology that was replaced with TTL. Let's revert this and change the other docblocks to match

rhyslbw (changes requested):

packages/cardano-services/src/InMemoryCache/InMemoryCache.ts
iccicci: I lost this info. Thank you for the repetition.

797d581da should solve this.

rhyslbw (dismissed):

IvayloAndonov (dismissed):

rhyslbw (approved):

IvayloAndonov (approved):

PR #412, opened 2022-08-30
fix: Gracefully handle `CardanoTokenRegistry` errors during `getAsset` request

IvayloAndonov:
# Context
`AssetInfo.tokenMetatdata` is stored in an off-chain registry, which cannot be depended on for returning a match for the given assetIds, nor it's service availability. Currently we’re rethrowing request errors as a Provider error, which is not ideal since it then couples our own service health, as opposed to the desired behaviour of omitting the metadata from the response. 
[cardano-js-sdk/CardanoTokenRegistry.ts at 8fe1cdd7683a7802bebd081616abead292488953 · input-output-hk/cardano-js-sdk](https://github.com/input-output-hk/cardano-js-sdk/blob/8fe1cdd7683a7802bebd081616abead292488953/packages/cardano-services/src/Asset/CardanoTokenRegistry.ts#L151) 

This issue does raise an important question of how can a client differentiate between a response that included a `404` from the registry vs a `5xx` with our current shape of from  AssetInfo.tokenMetadata

# Proposed Solution

**Approach A)**
Introduce the suggested approach from the task description with minor diffs as follows:
```
interface TokenMetadataSource {
  type: MetadataSourceType; \\ enum to support: 'registry', 'on-chain' or 'referenced'  
  requestStatus: 'success' | 'fail'; 
}

export interface AssetInfo<Source> {
  ...
  tokenMetadata?: { data?: TokenMetadata | null, source?: Source } ;
  ...
}

// Provider interface use AssetInfo<unknown>
// cardano-services use AssetInfo<TokenMetadataSource>
```

Notes:
- `cached: boolean` is committed because it's not so related to the `Source` scope, however, it could be added easily with correct values within `CardanoTokenRegistry` easily
- `source?` property is nullable to keep the compatibility with the other providers which implement same`AssetProvider` such as `blockfrost`. Not sure what `source.type` in `blockfrost` should be since the tokenMetadata is [fetched](https://github.com/input-output-hk/cardano-js-sdk/blob/d08353bd3e8bc1d1a17db79757de007e82ddede7/packages/blockfrost/src/blockfrostAssetProvider.ts#L64) as part of the whole `blockfrost.assetsById()` response.
- in `cardano-services` is used `AssetInfo<TokenMetadataSource>`, everywhere else is `AssetInfo<unknown>`
- token metadata responses are adjusted based on the interface changes within the tests (wrapper into `data` obj and added `source` obj in `cardano-services`)

**Approach B)** Omitting the metadata from the response in case of 5xx error thrown by the external source (in our current situation - token metadata server)

Following the [CIP-0035](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/core/src/Asset/types/AssetInfo.ts#L25-L28) comment, the second approach could be simply to avoid the `AssetInfo` interface changes and to keep it as it is:
`tokenMetadata?: TokenMetadata | null;` with the only difference is to shallow the `ProviderError` in catch block and to omit `tokenMetadata` in case of network error by the external token server by returning simply `null`.

The downside here is losing the details of the source/origin of that data.

### DIFF
--- a/packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
+++ b/packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
@@ -148,7 +148,15 @@ export class CardanoTokenRegistry implements TokenMetadataService {
         }
       }
     } catch (error) {
-      throw toProviderError(error, 'while fetching metadata from token registry');
+      if (axios.isAxiosError(error)) {
+        throw new ProviderError(
+          ProviderFailure.ConnectionFailure,
+          error,
+          'CardanoTokenRegistry failed to fetch asset metatada from the token registry server'
+        );
+      }
+
+      throw error;
     }
 
     return tokenMetadata;
--- a/packages/cardano-services/src/Asset/DbSyncAssetProvider.ts
+++ b/packages/cardano-services/src/Asset/DbSyncAssetProvider.ts
@@ -66,8 +66,17 @@ export class DbSyncAssetProvider extends DbSyncProvider implements AssetProvider
     if (extraData?.history) await this.loadHistory(assetInfo);
     if (extraData?.nftMetadata)
       assetInfo.nftMetadata = await this.#dependencies.ntfMetadataService.getNftMetadata(assetInfo);
-    if (extraData?.tokenMetadata)
-      assetInfo.tokenMetadata = (await this.#dependencies.tokenMetadataService.getTokenMetadata([assetId]))[0];
+    if (extraData?.tokenMetadata) {
+      try {
+        assetInfo.tokenMetadata = (await this.#dependencies.tokenMetadataService.getTokenMetadata([assetId]))[0];
+      } catch (error) {
+        if (error instanceof ProviderError && error.reason === ProviderFailure.ConnectionFailure) {
+          assetInfo.tokenMetadata = undefined;
+        } else {
+          throw error;
+        }
+      }
+    }
 
     return assetInfo;
   }
--- a/packages/cardano-services/src/Asset/openApi.json
+++ b/packages/cardano-services/src/Asset/openApi.json
@@ -95,7 +95,9 @@
         }
       },
       "GetAssetRequest": {
-        "required": ["assetId"],
+        "required": [
+          "assetId"
+        ],
         "type": "object",
         "properties": {
           "assetId": {
--- a/packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
+++ b/packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
@@ -1,4 +1,4 @@
-import { Asset, Cardano, ProviderError } from '@cardano-sdk/core';
+import { Cardano, ProviderError, ProviderFailure } from '@cardano-sdk/core';
 import { CardanoTokenRegistry, toCoreTokenMetadata } from '../../src/Asset';
 import { InMemoryCache, Key } from '../../src/InMemoryCache';
 import { IncomingMessage, createServer } from 'http';
@@ -135,11 +135,11 @@ describe('CardanoTokenRegistry', () => {
     });
 
     it('returns metadata when subject exists', async () => {
-      const [metadata] = await tokenRegistry.getTokenMetadata([validAssetId]);
+      const metadata = await tokenRegistry.getTokenMetadata([validAssetId]);
 
-      expect(metadata).not.toBeNull();
+      expect(metadata![0]).not.toBeNull();
 
-      const result: Asset.TokenMetadata = {
+      const result = {
         decimals: 8,
         desc: 'SingularityNET',
         icon: 'testLogo',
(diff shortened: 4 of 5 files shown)


packages/cardano-services/src/Asset/types.ts
```diff
@@ -1,10 +1,21 @@
 import { Asset, Cardano } from '@cardano-sdk/core';
 import { Shutdown } from '@cardano-sdk/util';
 
+export enum MetadataSourceType {
+  Registry = 'registry',
+  OnChain = 'on-chain',
```
rhyslbw, line 6: Given we've not yet implemented `on-chain` or `referenced` support, let's just have a single option in this enum. Also, we'll probably be looking to align with a type registry that [this CIP](https://github.com/cardano-foundation/CIPs/pull/298) introduces.

rhyslbw (changes requested):
Looks good so far @IvayloAndonov, just a minor change request

packages/cardano-services/src/Asset/types.ts
IvayloAndonov: Updated in 050fc2bd1a1b3204fee85152928253535131d24c

iadmytro (dismissed):

rhyslbw (changes requested):
Given there's currently only a single source for this metadata, let's revert the breaking `AssetInfo` type change given `tokenMetadata` === `undefined` can be used to infer the success of the registry request.

IvayloAndonov:
Rebased with master and resolved conflicts.

Revert the breaking `AssetInfo` type change: [039e0f0](https://github.com/input-output-hk/cardano-js-sdk/pull/412/commits/039e0f0069d493cf226c0b37aeec34464a5d96ed) (it eliminates most of the previously introduced changes)

packages/cardano-services/src/Asset/DbSyncAssetProvider.ts
```diff
@@ -66,8 +66,10 @@ export class DbSyncAssetProvider extends DbSyncProvider implements AssetProvider
     if (extraData?.history) await this.loadHistory(assetInfo);
     if (extraData?.nftMetadata)
       assetInfo.nftMetadata = await this.#dependencies.ntfMetadataService.getNftMetadata(assetInfo);
-    if (extraData?.tokenMetadata)
-      assetInfo.tokenMetadata = (await this.#dependencies.tokenMetadataService.getTokenMetadata([assetId]))[0];
+    if (extraData?.tokenMetadata) {
+      const tokenMetadata = await this.#dependencies.tokenMetadataService.getTokenMetadata([assetId]);
+      assetInfo.tokenMetadata = tokenMetadata && tokenMetadata.length > 0 ? tokenMetadata[0] : undefined;
```
rhyslbw, line 71:
```suggestion
      assetInfo.tokenMetadata = tokenMetadata[0];
```
With a try/catch around this, the value can then be assigned as so

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
```diff
@@ -148,7 +148,18 @@ export class CardanoTokenRegistry implements TokenMetadataService {
...
       }
     } catch (error) {
-      throw toProviderError(error, 'while fetching metadata from token registry');
+      // Propagete only the Provider errors
+      // We are interested only from network errors while fetching from external token metadata server
+      if (error instanceof ProviderError) {
+        throw error;
+      }
+
+      // Return `undefined` in case of HTTP error while fetching from Token Metadata server
```
rhyslbw, line 157: On second look at this, I think the handling should be done within `DbSyncAssetProvider`, as that's the more appropriate context to assign the value of `undefined`. Here I would just ensure the errors being rethrown are suitable for interpretation.

rhyslbw (changes requested):

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
IvayloAndonov: Agree that the error handling could be lifted one level up in `DbSyncAssetProvider`. The necessary distinction of errors in our case is already suitable for interpretation since we have this [try/catch](https://github.com/input-output-hk/cardano-js-sdk/blob/039e0f0069d493cf226c0b37aeec34464a5d96ed/packages/cardano-services/src/Asset/CardanoTokenRegistry.ts#L131-L148)  for all **provider** related errors and interpret all others as **HTTP** related ones thrown by [axios client](https://github.com/input-output-hk/cardano-js-sdk/blob/039e0f0069d493cf226c0b37aeec34464a5d96ed/packages/cardano-services/src/Asset/CardanoTokenRegistry.ts#L125-L128)

IvayloAndonov: b856d33aa8738e7d6cd6c7685c7a9689cea2de6d

rhyslbw (changes requested):
`DbSyncAssetProvider` test is missing the assertion that demonstrates the `assetInfo.tokenMetadata` value is `undefined` if the asset is not found in the token registry.

mkazlauskas (changes requested):

packages/core/src/Asset/types/AssetInfo.ts
```diff
@@ -10,7 +10,6 @@ export interface AssetMintOrBurn {
    */
   quantity: bigint;
 }
-
```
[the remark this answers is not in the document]
rhyslbw: Good catch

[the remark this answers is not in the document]
IvayloAndonov: 61db0faa7b3b898b78436739c832c38c359cf307

rhyslbw (dismissed):

mkazlauskas (dismissed):
Nice work! :rocket:

rhyslbw (approved):

mkazlauskas (approved):

PR #413, opened 2022-08-30
Fix --use-queue arg parser

gytis-ivaskevicius:
_Edited by @rhyslbw_
# Context
There is a bug with the CLI parsing of intended boolean values, which was detected when one of the ENVs was set to `false` and treated as a truthy value. There were no tests covering this scenario.

# Proposed Solution
- Add missing test cases
- Make the options expect either 'true' or 'false'
- Attach the `stringToBoolean` arg parser where required.

- `--use-queue`, `--enable-metrics`, and `--parallel` now require a value of `true` to be provided.
- I took the opportunity to consolidate the CLI tests into a single suite, so it's best to review d14bd9d8aeec64f04aab094e0aceb8dc5b803926 for the diff


packages/cardano-services/src/cli.ts
```diff
@@ -246,7 +252,18 @@ commonOptions(
...
+      .env('USE_QUEUE')
+      .default(USE_QUEUE_DEFAULT)
+      .argParser((useQueue) => {
+        stringToBoolean(
+          useQueue === undefined ? 'true' : useQueue,
+          Programs.HttpServer,
+          ProgramOptionDescriptions.UseQueue
+        );
+      })
+  )
```
gytis-ivaskevicius, line 266: Are you sure this will work? I tried something similar but it requires a boolean

rhyslbw: This was quite a nasty bug, but I've resolved it in 8f92b7de7e37e651c2fb782c1d48f30bbc02c647

rhyslbw (approved):

mkazlauskas (approved):
Looks good, maybe we should look if we can fix commander.js to support our old interface

PR #416, opened 2022-08-31
fix: Remove nullability of ProtocolParametersBabbage properties

IvayloAndonov:
# Context

The type we’ve been inheriting from Ogmios as described by it’s name is invalid:

[refactor: removed Ogmios schema package dependency by AngelCastilloB · Pull Request #388 · input-output-hk/cardano-js-sdk](https://github.com/input-output-hk/cardano-js-sdk/pull/388#discussion_r953396868) 

The reason for this loose nullability is due to context within Ogmios. It’s the type specified for both protocol update objects and the current parameters, which, while it works, is not an accurate model.

We should make each parameter non-nullable, remove the now unnecessary [casting](https://github.com/input-output-hk/cardano-js-sdk/blob/141f2caa1b845d73616437bc9b6ddbd33268f3d0/packages/core/src/Provider/NetworkInfoProvider/types.ts#L3), and fix any tests that are only checking for a limited set of properties. In the future, if needing to model the update type,  we can define it as ProtocolParametersBabbageUpdate = Partial<ProtocolParametersBabbage>
# Proposed Solution

To introduce a more declarative way of all parameter updates in time between eras so we can benefit by following the protocol parameter updates in time easily

The goal with this approach is to have a type that represents the new params being introduced, which then represents the (current) protocol params when composed with the prior era, considering deprecations

- (core): compose new `ProtocolParameters` type composed by prev eras
- (cardano-services): extend `db-sync` query with current protocol params
- (cardano-services): align DB model type and mappings
- (cardano-services): update snapshot tests with all protocol params
- (blockfrost): update `currentWalletProtocolParameters` mapping

# Note:
Keep an eye on that almost all protocol parameters are marked with `Nullable...` with is equal to `type` | `null` in Ogmios schema, for instance [ProtocolParametersBabbage](https://github.com/CardanoSolutions/ogmios/blob/master/clients/TypeScript/packages/schema/src/index.ts#L1244-L1267) which makes me think that potentially some of them could be fetched as `null` in theory which is against our ProtocolParameters type with all required fields on it. 
This is about only the ts types, but not sure if some of the data sources (`blockfrost` or `db-sync`) return `null` as value for some of the protocol parameters.

### DIFF
--- a/packages/cip2/src/types.ts
+++ b/packages/cip2/src/types.ts
@@ -110,7 +110,7 @@ export interface InputSelector {
 }
 
 export type ProtocolParametersForInputSelection = Pick<
-  Cardano.ProtocolParametersBabbage,
+  Cardano.ProtocolParameters,
   'coinsPerUtxoByte' | 'maxTxSize' | 'maxValueSize' | 'minFeeCoefficient' | 'minFeeConstant'
 >;
 
--- a/packages/core/src/Cardano/types/ProtocolParameters.ts
+++ b/packages/core/src/Cardano/types/ProtocolParameters.ts
@@ -6,9 +6,14 @@ export interface ProtocolVersion {
   minor: number;
   patch?: number;
 }
-export type CostModels = CostModel[];
 
-export type CostModel = number[];
+export type ModelKey = string;
+
+export type CostModel = {
+  [k: ModelKey]: number;
+};
+
+export type CostModels = CostModel[];
 
 export interface Prices {
   memory: number;
@@ -24,27 +29,73 @@ export interface ValidityInterval {
   invalidHereafter?: Slot;
 }
 
-export interface ProtocolParametersBabbage {
-  minFeeCoefficient?: number;
-  minFeeConstant?: number;
-  maxBlockBodySize?: number;
-  maxBlockHeaderSize?: number;
-  maxTxSize?: number;
-  stakeKeyDeposit?: number;
-  poolDeposit?: number;
-  poolRetirementEpochBound?: number;
-  desiredNumberOfPools?: number;
-  poolInfluence?: number;
-  monetaryExpansion?: number;
-  treasuryExpansion?: number;
-  minPoolCost?: number;
-  coinsPerUtxoByte?: number;
-  maxValueSize?: number;
-  collateralPercentage?: number;
-  maxCollateralInputs?: number;
-  protocolVersion?: ProtocolVersion;
-  costModels?: CostModels;
-  prices?: Prices;
-  maxExecutionUnitsPerTransaction?: ExUnits;
-  maxExecutionUnitsPerBlock?: ExUnits;
+export interface TxFeePolicy {
+  coefficient: string;
+  constant: number;
 }
+
+export interface SoftforkRule {
+  initThreshold: string;
+  minThreshold: string;
+  decrementThreshold: string;
+}
+
+type ProtocolParametersByron = {
+  heavyDlgThreshold: string;
+  maxBlockSize: number;
+  maxHeaderSize: number;
+  maxProposalSize: number;
+  maxTxSize: number;
+  mpcThreshold: string;
+  scriptVersion: number;
+  slotDuration: number;
+  unlockStakeEpoch: number;
+  updateProposalThreshold: string;
+  updateProposalTimeToLive: number;
+  updateVoteThreshold: string;
+  txFeePolicy: TxFeePolicy;
+  softforkRule: SoftforkRule;
+};
+
+type NewProtocolParamsInShelley = {
+  minFeeCoefficient: number;
+  minFeeConstant: number;
+  maxBlockBodySize: number;
+  maxBlockHeaderSize: number;
+  stakeKeyDeposit: number;
+  poolDeposit: number | null;
+  poolRetirementEpochBound: number;
+  desiredNumberOfPools: number;
+  poolInfluence: string;
+  monetaryExpansion: string;
+  treasuryExpansion: string;
+  decentralizationParameter: string;
+  minUtxoValue: number;
+  minPoolCost: number;
+  extraEntropy: 'neutral' | string;
+  protocolVersion: ProtocolVersion;
+};
+
+type ShelleyProtocolParams = Pick<ProtocolParametersByron, 'maxTxSize'> & NewProtocolParamsInShelley;
+
+type NewProtocolParamsInAlonzo = {
+  coinsPerUtxoWord: number;
+  maxValueSize: number;
+  collateralPercentage: number;
+  maxCollateralInputs: number;
+  costModels: CostModels;
+  prices: Prices;
+  maxExecutionUnitsPerTransaction: ExUnits;
+  maxExecutionUnitsPerBlock: ExUnits;
+};
+
+type AlonzoProtocolParams = Omit<ShelleyProtocolParams, 'minUtxoValue'> & NewProtocolParamsInAlonzo;
+
+type NewProtocolParamsInBabbage = {
+  coinsPerUtxoByte: number;
+};
+
+type BabbageProtocolParameters = Omit<AlonzoProtocolParams, 'coinsPerUtxoWord' | 'extraEntropy'> &
+  NewProtocolParamsInBabbage;
+
+export type ProtocolParameters = BabbageProtocolParameters;
--- a/packages/core/src/Provider/NetworkInfoProvider/types.ts
+++ b/packages/core/src/Provider/NetworkInfoProvider/types.ts
@@ -1,20 +1,6 @@
 import { Cardano, EraSummary, Provider } from '../..';
 
-export type ProtocolParametersRequiredByWallet = Required<
-  Pick<
-    Cardano.ProtocolParametersBabbage,
-    | 'coinsPerUtxoByte'
-    | 'maxTxSize'
-    | 'maxValueSize'
-    | 'stakeKeyDeposit'
-    | 'maxCollateralInputs'
-    | 'minFeeCoefficient'
-    | 'minFeeConstant'
-    | 'minPoolCost'
-    | 'poolDeposit'
-    | 'protocolVersion'
-  >
->;
+export type ProtocolParametersRequiredByWallet = Omit<Cardano.ProtocolParameters, 'costModels'>;
 
 export type SupplySummary = {
   circulating: Cardano.Lovelace;
--- a/packages/blockfrost/src/BlockfrostToCore.ts
+++ b/packages/blockfrost/src/BlockfrostToCore.ts
@@ -27,15 +27,36 @@ export const BlockfrostToCore = {
     blockfrost: Responses['epoch_param_content']
   ): ProtocolParametersRequiredByWallet => ({
     coinsPerUtxoByte: Number(blockfrost.coins_per_utxo_word),
+    collateralPercentage: Number(blockfrost.collateral_percent),
+    decentralizationParameter: String(blockfrost.decentralisation_param),
+    desiredNumberOfPools: blockfrost.n_opt,
+    maxBlockBodySize: blockfrost.max_block_size,
+    maxBlockHeaderSize: blockfrost.max_block_header_size,
     maxCollateralInputs: Number(blockfrost.max_collateral_inputs),
+    maxExecutionUnitsPerBlock: {
+      memory: Number(blockfrost.max_block_ex_mem),
+      steps: Number(blockfrost.max_block_ex_mem)
+    },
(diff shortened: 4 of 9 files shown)


packages/core/src/Cardano/types/ProtocolParameters.ts
```diff
@@ -6,9 +6,12 @@ export interface ProtocolVersion {
   minor: number;
   patch?: number;
 }
-export type CostModels = CostModel[];
-
-export type CostModel = number[];
+export interface CostModels {
```
IvayloAndonov, line 9: Updated based on Ogmios schema [CostModels](https://github.com/CardanoSolutions/ogmios/blob/master/clients/TypeScript/packages/schema/src/index.ts#L1273-L1278)

packages/core/src/Provider/NetworkInfoProvider/types.ts
```diff
@@ -1,20 +1,6 @@
...
-    | 'stakeKeyDeposit'
-    | 'maxCollateralInputs'
-    | 'minFeeCoefficient'
-    | 'minFeeConstant'
-    | 'minPoolCost'
-    | 'poolDeposit'
-    | 'protocolVersion'
-  >
->;
+export type ProtocolParametersRequiredByWallet = Omit<Cardano.ProtocolParameters, 'costModels'>;
```
IvayloAndonov, line 3: The reason I omit `costModels` is that it is the only properly added from `Alonzo` era now that does not exist in Blockfrost API [epoch_param_content](epoch_param_content) section. It's needed for the [mapping](https://github.com/input-output-hk/cardano-js-sdk/pull/416/files#diff-88e0e444f82c6720c69afd174e9e6a88e9300309d629a851c8571a42926f76c8R26-R60) of protocol parameters since now we return non-nullable composed `ProtocolParameters` type from `currentWalletProtocolParameters` within `cardano-services` and `blockfrost`.

With the latest db-sync schema, there is one column called `cost_model_id` in `epoch_param` but again it's not clear how it could be mapped to `CostModels` type exposed by `ProtocolParametersAlonzo` in Ogmios schema
@rhyslbw

packages/core/src/Cardano/types/ProtocolParameters.ts
```diff
@@ -6,9 +6,12 @@ export interface ProtocolVersion {
...
+  [k: string]: CostModel;
```
mkazlauskas, line 10: Please add a type aliases for all primitive types (string & number). It is not clear what they are.

packages/core/src/Provider/NetworkInfoProvider/types.ts
mkazlauskas: It's ok as wallet is not using this parameter. We could probably further narrow down this type to only includue parameters that are actually required by wallet.

packages/cip2/src/types.ts
```diff
@@ -110,7 +110,7 @@ export interface InputSelector {
 }
 
 export type ProtocolParametersForInputSelection = Pick<
-  Cardano.ProtocolParametersBabbage,
+  Cardano.ProtocolParameters,
```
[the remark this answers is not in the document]
mkazlauskas: Yes, let's keep it. Best to keep dependencies to the minimum required type.

mkazlauskas (changes requested):

packages/core/src/Cardano/types/ProtocolParameters.ts
rhyslbw: There are currently two values this could be `PlutusV1` or `PlutusV2`

packages/core/src/Provider/NetworkInfoProvider/types.ts
rhyslbw: Let's try to include these now, as they are relevant to wallet, even if ours isn't using it yet:
image

> On the db-sync side with the latest schema, there is one column called cost_model_id in epoch_param but again it's not clear how it could be mapped to CostModels type exposed by ProtocolParametersAlonzo in Ogmios schema

https://github.com/input-output-hk/cardano-db-sync/blob/master/doc/schema.md#cost_model

We'll have to infer the Plutus version based on sort order of `block_id`

packages/core/src/Cardano/types/ProtocolParameters.ts
```diff
@@ -24,27 +27,87 @@ export interface ValidityInterval {
...
+  monetaryExpansion: string;
+  treasuryExpansion: string;
+  decentralizationParameter: string;
+  minUtxoValue: number;
+  minPoolCost: number;
+  extraEntropy: 'neutral' | string;
+  protocolVersion: ProtocolVersion;
+};
+
+export type ShelleyProtocolParams = Omit<
```
rhyslbw, line 77: Any reason to not `Pick` instead?

packages/core/src/Cardano/types/ProtocolParameters.ts
```diff
@@ -24,27 +27,87 @@ export interface ValidityInterval {
...
+  maxExecutionUnitsPerBlock: ExUnits;
+};
+
+export type AlonzoProtocolParams = Omit<ShelleyProtocolParams, 'minUtxoValue'> & NewProtocolParamsInAlonzo;
+
+type NewProtocolParamsInBabbage = {
+  coinsPerUtxoByte: number;
+};
+
+export type ProtocolParameters = Omit<AlonzoProtocolParams, 'coinsPerUtxoWord' | 'extraEntropy'> &
```
rhyslbw, line 112:
```suggestion
export type BabbageProtocolParameters = Omit<AlonzoProtocolParams, 'coinsPerUtxoWord' | 'extraEntropy'> &
export type ProtocolParameters = BabbageProtocolParameters;
```
Better if `ProtocolParameters` is just an alias to the current era

rhyslbw (changes requested):

packages/core/src/Cardano/types/ProtocolParameters.ts
mkazlauskas: I suggest refactoring [this enum](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/core/src/Cardano/types/Script.ts#L193) to have string values and doing this:
```suggestion
  [k in PlutusLanguageVersion]: CostModel | undefined;
```

@AngelCastilloB would that be ok or are number values of that enum important?

packages/core/src/Cardano/types/ProtocolParameters.ts
IvayloAndonov: Just following the consistency of how we define all next protocol params: 
1. omit the obsolete ones
2. add the new ones

Like [here](https://github.com/input-output-hk/cardano-js-sdk/pull/416/files#diff-2fd64a4659e387924f9375a568cbeb7bb58ff9f4baa9d03d35402c27d81d65adR106), and [here](https://github.com/input-output-hk/cardano-js-sdk/pull/416/files#diff-2fd64a4659e387924f9375a568cbeb7bb58ff9f4baa9d03d35402c27d81d65adR112)

packages/core/src/Cardano/types/ProtocolParameters.ts
AngelCastilloB: The values are not important, we can change it to strings without problem. However I think this:

` [k: string]: CostModel;`

Should be typed as this:

`export type CostModels = CostModel[]`

The `[k: string]: Value` is the way Ogmios represent arrays, since arrays in the CDDL are represented as maps where the keys is the numerical position (index) in the array.

AngelCastilloB: From CDDL:
```
costmdls =
  { ? 0 : [ 166*166 int ] ; Plutus v1
  , ? 1 : [ 175*175 int ] ; Plutus v2
  }
```

something like:

```
models[0] = plutusV1costmdls;
models[1] = plutusV2costmdls;
```

mkazlauskas: Thanks @AngelCastilloB. Agree, let's go with `CostModel[]`. Perhaps it would then make sense to update the enum to correspond to cost model index, so that you could do: `costModels[PlutusLanguageVersion.PlutusV1]`?

AngelCastilloB: Yes @mkazlauskas :), I am not sure why I initialized the enum with 1 instead of 0, that was pretty random

packages/core/src/Cardano/types/ProtocolParameters.ts
rhyslbw: I think it'd be better to use the most appropriate utility class to be more DRY, however, I can see the appeal of the semantics. I'm happy with either, so let's see what the others think

packages/core/src/Cardano/types/ProtocolParameters.ts
rhyslbw: How about?:
```ts
export enum PlutusLanguageVersion {
  V1 = 'PlutusV1',
  V2 = 'PlutusV2'
}
```

packages/core/src/Cardano/types/ProtocolParameters.ts
IvayloAndonov: As I remember from our discussion (a week ago) we agreed on this more declarative approach of defining protocol parameters type for each era which represents the composition of the prior era parameterts, considering deprecations + the new ones added by the current era. 
So we can keep track of all protocol parameter updates in time quickly, correct me if I'm missing something 🤔

packages/core/src/Cardano/types/ProtocolParameters.ts
IvayloAndonov: Following all suggestions, do we all agree with:

1. Fix [PlutusLanguageVersion](https://github.com/input-output-hk/cardano-js-sdk/blob/3b757ba77e163b44136227ae2a8fa59600b4474f/packages/core/src/Cardano/types/Script.ts#L189-L207) enum to
`export enum PlutusLanguageVersion {
  V1 = 0,
  V2 = 1
}`

2. Moving cost model types to array:

   `export type CostModel = {
  [k: string]: number;
}`

    `export type CostModels = CostModel[]`

3. And then would have:

```
models: CostModels;

models[PlutusLanguageVersion.V1] = plutusV1costModel;
models[PlutusLanguageVersion.V2] = plutusV2costModel;
```

@rhyslbw @mkazlauskas @AngelCastilloB

packages/core/src/Provider/NetworkInfoProvider/types.ts
IvayloAndonov: What is the link of this swagger doc? 👀

rhyslbw: https://docs.blockfrost.io/

IvayloAndonov: ah didn't notice it's the blockfrost from the screenshot, thanks

packages/core/src/Cardano/types/ProtocolParameters.ts
IvayloAndonov: Applied here 004f780300900891ba24d193125644e6c31f8d30

packages/core/src/Cardano/types/ProtocolParameters.ts
IvayloAndonov: The changes are applied here 004f780300900891ba24d193125644e6c31f8d30

Only to note, regarding the introduced `PlutusLanguageVersion` enum fix and the keys renaming to `V1` and `V2`, I checked that in `cardano-services` config genesis files remain as:
```
"costModels": {
        "PlutusV1": {
            "sha2_256-memory-arguments": 4,
            ...
            }
         }
```

rhyslbw: We'll just need a translation between the object and array representation.

packages/core/src/Cardano/types/ProtocolParameters.ts
rhyslbw: The idea was to maintain the interfaces using `ProtocolParameters`, not change to `BabbageProtocolParameters`, by aliasing like this:

```ts
export type ProtocolParameters = BabbageProtocolParameters;
```

IvayloAndonov: 6d49527a6a5de79d9e8350a785f02acfdf560fb0

rhyslbw (dismissed):

mkazlauskas (approved):
Looks good!

rhyslbw (approved):
:rocket:

mkazlauskas (approved):

PR #417, opened 2022-09-01
feat(cardano-services): adds log of incoming requests

iccicci:
# Context

We need to increase the verbosity of our log

# Proposed Solution

Now incoming requests are logged.

### DIFF
--- a/packages/cardano-services/src/Http/HttpServer.ts
+++ b/packages/cardano-services/src/Http/HttpServer.ts
@@ -4,7 +4,7 @@ import { HttpService } from './HttpService';
 import { Logger } from 'ts-log';
 import { ProviderError, ProviderFailure } from '@cardano-sdk/core';
 import { RunnableModule } from '../RunnableModule';
-import { fromSerializableObject, toSerializableObject } from '@cardano-sdk/util';
+import { contextLogger, fromSerializableObject, toSerializableObject } from '@cardano-sdk/util';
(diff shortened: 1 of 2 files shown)


packages/cardano-services/src/Http/HttpServer.ts
```diff
@@ -93,6 +93,15 @@ export class HttpServer extends RunnableModule {
       );
     }
 
+    const requestLogger = contextLogger(this.logger, 'request');
+
+    this.app.use((req, _res, next) => {
+      const { body, method, path, query } = req;
+
+      requestLogger.debug({ body, method, path, query, time: Date.now() });
```
mirceahasegan, line 101: I am not sure if the `time` part is needed. I think it should/is added by the logger implementation (by bunyan for example).

mkazlauskas (approved):
LGTM

packages/cardano-services/src/Http/HttpServer.ts
iccicci: Yes... loggers _usually_ add time by themselves...
Removed in 60c13c5fa

mirceahasegan (dismissed):

rhyslbw (dismissed):
Ready for squashing

rhyslbw (approved):

mirceahasegan (approved):

PR #418, opened 2022-09-01
fix: adds node-gyp to docker base build image

iccicci:
# Context

**Sometimes** `yarn local-network:up --build` fails to start due to problems installing `npm-ssh` package.

```
#18 28.52 ➤ YN0007: │ ssh2@npm:1.9.0 must be built because it never has been before or the last one failed
#18 29.15 ➤ YN0000: │ ssh2@npm:1.9.0 STDOUT Usage Error: Couldn't find a script name "node-gyp" in the top-level (used by ssh2@npm:1.9.0). This typically happens because some package depends on "node-gyp" to build itself, but didn't list it in their dependencies. To fix that, please run "yarn add node-gyp" into your top-level workspace. You also can open an issue on the repository of the specified package to suggest them to use an optional peer dependency.
```

The worst aspect is that the problem can't be certainly reproduced... but it also a time waster as sometimes it happens breaking my workflow, so I would like to solve it once forever.

# Proposed Solution

- Removed `node-gyp` from `devDependencies` as I verified it doesn't fixes the problem
- Installed `node-gyp` globally in the docker base build image


mirceahasegan (approved):

mirceahasegan (approved):
🤞 🚀

rhyslbw (approved):
Agree with the reasoning for including this, and @mirceahasegan's feedback. Please squash

AngelCastilloB (approved):
LGTM 🚀

rhyslbw (approved):

PR #419, opened 2022-09-01
fix(web-extension): un-exposing an object in background process doesn't destroy the entire messaging

mkazlauskas:
# Context

```ts
const obj1 = exposeApi(...);
const obj2 = exposeApi(...);
obj1.shutdown(); // this destroys ports for both 'obj1' and 'obj2'
```

# Proposed Solution

87de6e126f9846f5a376bd120bd89c25c5cc985a

@lgobbi-atix confirmed that this fix resolves the error he was having

### DIFF
--- a/packages/web-extension/src/messaging/BackgroundMessenger.ts
+++ b/packages/web-extension/src/messaging/BackgroundMessenger.ts
@@ -60,14 +60,21 @@ export const createBackgroundMessenger = ({ logger, runtime }: MessengerDependen
      * Disconnect all existing ports and stop listening for new ones.
      */
     destroy() {
-      for (const { message$, ports$ } of channels.values()) {
-        message$.complete();
-        for (const port of ports$.value) {
-          port.disconnect();
-        }
+      // eslint-disable-next-line unicorn/no-useless-spread
+      for (const channelName of [...channels.keys()]) {
+        this.destroyChannel(channelName);
       }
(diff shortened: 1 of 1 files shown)


mirceahasegan (dismissed):
If it's confirmed, it's confirmed! 🚀 
LGTM

rhyslbw (approved):

mirceahasegan (approved):

packages/web-extension/src/messaging/BackgroundMessenger.ts
```diff
@@ -60,15 +60,22 @@ export const createBackgroundMessenger = ({ logger, runtime }: MessengerDependen
...
       logger.warn('[BackgroundMessenger] destroyed');
     },
+    destroyChannel(channelName: string) {
+      const channel = channels.get(channelName);
+      if (!channel) return;
+      channel.message$.complete();
+      for (const port of channel.ports$.value) {
+        port.disconnect();
+      }
+      channels.delete(channelName);
```
mirceahasegan, line 77: This is the only part that is new, right?
And calling destroyChannel from destroy()

mkazlauskas: Yes, now instead of destroying all channels, it will only destroy channels with relevant name

mirceahasegan (approved):

PR #420, opened 2022-09-02
chore(cardano-services): remove support for starting a testnet stack

rhyslbw:
# Context
Our integration tests DB snapshot is still of _testnet_, but as the e2e tests are now network agnostic and the Docker Hub registry no longer has the required base images, removing support for starting the services makes sense.

# Proposed Solution
- Remove the npm script that simplifies starting the _testnet_ stack
- Update docs to use _preprod_ instead


AngelCastilloB (approved):
LGTM

iadmytro (approved):

mkazlauskas (approved):

PR #421, opened 2022-09-05
fix: invalid epoch number calculations when using testnet auto-upgrading configuration

rhyslbw:
# Context
It's possible to configure when particular eras are upgraded, without an upgrade proposal, in testnet `cardano-node` configuration, including the specification of multiple eras in the same epoch.

Both of these configs are valid:
```json
  "TestAllegraHardForkAtEpoch": 0,
  "TestAlonzoHardForkAtEpoch": 0,
  "TestEnableDevelopmentHardForkEras": true,
  "TestMaryHardForkAtEpoch": 0,
  "TestShelleyHardForkAtEpoch": 0,
```

```json
  "TestAllegraHardForkAtEpoch": 2,
  "TestAlonzoHardForkAtEpoch": 4,
  "TestEnableDevelopmentHardForkEras": true,
  "TestMaryHardForkAtEpoch": 3,
  "TestShelleyHardForkAtEpoch": 1,
```

# Proposed Solution
Filter the era summaries that represent eras effectively being skipped, which is evidenced by a later era starting in the same slot.

### DIFF
--- a/packages/core/src/util/slotCalc.ts
+++ b/packages/core/src/util/slotCalc.ts
@@ -1,6 +1,8 @@
 import { CardanoNetworkMagic, EpochNo, Slot } from '../Cardano';
 import { CustomError } from 'ts-custom-error';
 import { EraSummary } from '../CardanoNode';
+import groupBy from 'lodash/groupBy';
+import last from 'lodash/last';
 import orderBy from 'lodash/orderBy';
 
 export interface SlotDate {
@@ -17,7 +19,7 @@ export interface EpochInfo {
 export class EraSummaryError extends CustomError {}
 
 /**
- * Were valid at 2022-05-28
+ * Valid at 2022-05-28
  */
 export const mainnetEraSummaries: EraSummary[] = [
   { parameters: { epochLength: 21_600, slotLength: 20_000 }, start: { slot: 0, time: new Date(1_506_192_291_000) } },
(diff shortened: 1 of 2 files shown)


AngelCastilloB (dismissed):
LGTM

mkazlauskas (dismissed):
Looks good!

iadmytro (approved):

mkazlauskas (approved):

packages/core/src/util/slotCalc.ts
```diff
@@ -43,7 +45,12 @@ export const eraSummariesConfig: EraSummariesMap = {
 };
 
 const createSlotEpochCalcImpl = (eraSummaries: EraSummary[]) => {
-  const eraSummariesAsc = orderBy(eraSummaries, ({ start }) => start.slot);
+  // It's possible to configure when particular eras are upgraded, without an upgrade proposal, in
+  // testnet cardano-node configuration, including the specification of multiple eras in the same
+  // epoch. Era summaries therefore need to be filtered to remove eras that are being skipped,
+  // which is evidenced by a later era starting in the same slot.
+  const eraSummariesWithoutSkippedEras = Object.values(groupBy(eraSummaries, 'start.slot')).map(last) as EraSummary[];
```
mirceahasegan, line 52: Is there any chance the eras in eraSummaries are not ordered by start.slot?
If there is, groupBy does not order them, so slot 5 will be before slot 0 if the era was start.slot 5 was before the one with start.slot 0

mirceahasegan (approved):

PR #424, opened 2022-09-06
Feat/add e2e tests to CI

AngelCastilloB:
# Context

Currently our CI environment is not running the end to end test, so there is no way to automatically detect when some commit breaks them.

# Proposed Solution

Add the e2e test to our CI environment.


rhyslbw (changes requested):
Looking at this now, it would be better to separate the e2e test run in the workflow into its own step, for better isolation with the ENV configuration.

``` yaml
- name: 🔬 Test - unit and integration
      run: |
        yarn test:build:verify
        yarn test
      
- name: 🔬 Test - e2e
      run: |
        yarn workspace @cardano-sdk/e2e test:wallet
        yarn workspace @cardano-sdk/e2e test:cardano-services
      env:
```

.github/workflows/continuous-integration.yaml
```diff
@@ -36,7 +36,36 @@ jobs:
...
+        OGMIOS_PORT: '1340'
+        OGMIOS_URL: 'ws://cardano-node-ogmios:1340'
+
     - name: 🔬 Test
       run: |
         yarn test:build:verify
         yarn test
+        yarn workspace @cardano-sdk/e2e test:wallet
+      env:
+        NODE_AUTH_TOKEN: ${{ secrets.NPM_REGISTRY_TOKEN }}
```
rhyslbw, line 54:
```suggestion
```

.github/workflows/continuous-integration.yaml
```diff
@@ -36,7 +36,36 @@ jobs:
...
+        BLOCKFROST_API_KEY: ''
```
rhyslbw, line 55: Is this a required ENV? We should adjust our config validation

rhyslbw (changes requested):

.github/workflows/continuous-integration.yaml
AngelCastilloB: I think is required now, I wil fix this as part of this PR

packages/e2e/test/wallet/environment.ts
```diff
@@ -3,7 +3,6 @@ import * as envalid from 'envalid';
 export const env = envalid.cleanEnv(process.env, {
   ASSET_PROVIDER: envalid.str(),
   ASSET_PROVIDER_PARAMS: envalid.json({ default: {} }),
-  BLOCKFROST_API_KEY: envalid.str(),
   CHAIN_HISTORY_PROVIDER: envalid.str(),
```
AngelCastilloB, line 6: The API key presence validation is already done by the factories, so this is not needed here

```javascript
/**
 * Gets the singleton blockfrost API instance.
 *
 * @returns The blockfrost API instance, this function always returns the same instance.
 */
const getBlockfrostApi = async () => {
  if (blockfrostApi !== undefined) return blockfrostApi;

  if (process.env.BLOCKFROST_API_KEY === undefined) throw new Error(BLOCKFROST_MISSING_PROJECT_ID);

  return new BlockFrostAPI({ isTestnet: true, projectId: process.env.BLOCKFROST_API_KEY });
};
```

.github/workflows/continuous-integration.yaml
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/424/commits/4dbc20b6cd320dbc026cad4a926523cc5db605b4

.github/workflows/continuous-integration.yaml
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/424/commits/4dbc20b6cd320dbc026cad4a926523cc5db605b4 also see https://github.com/input-output-hk/cardano-js-sdk/pull/424#discussion_r964356053

.github/workflows/continuous-integration.yaml
```diff
@@ -36,7 +36,43 @@ jobs:
...
+
     - name: 🔬 Test
       run: |
         yarn test:build:verify
         yarn test
+
+    - name: 🔬 Test - e2e
+      run: |
+        yarn workspace @cardano-sdk/e2e test:wallet
+        TX_SUBMIT_PROVIDER_PARAMS='{"baseUrl":"http://localhost:5001/tx-submit"}' yarn workspace @cardano-sdk/e2e test:cardano-services
```
mirceahasegan, line 56: Out of curiosity, why does this change for cardano-services?

packages/e2e/test/wallet/environment.ts
```diff
@@ -3,7 +3,6 @@ import * as envalid from 'envalid';
 export const env = envalid.cleanEnv(process.env, {
   ASSET_PROVIDER: envalid.str(),
   ASSET_PROVIDER_PARAMS: envalid.json({ default: {} }),
-  BLOCKFROST_API_KEY: envalid.str(),
```
mirceahasegan, line 6: 👍

mirceahasegan (dismissed):

.github/workflows/continuous-integration.yaml
rhyslbw: It would be better to change the test config to use a single server if possible

AngelCastilloB: @mirceahasegan The load test inside the cardano-services starts its own version of the tx-submit service with a specific configuration, so we need to point it to that service. Maybe @iccicci can you us more context.

@rhyslbw I agree, as it is right now this will clash with the env needed by future tests added to the cardano-services e2e test suite. I think we either need to change the name of the envs on that test so it doesn't clash with other tests, or we need to somehow make it work with the tx-submit endpoint we already have, but I am not sure what would be required to do so. Do you want me to add this refactor to the scope of this PR?

AngelCastilloB: Another option would be to give it default values that would work without needing to pass the envs

rhyslbw: Oh, we don't need to run the load test in CI. Let's separate it out, and not run it in this context.

AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/424/commits/8ecd12ac41af88dc67315efd42430158d17df340 (I am squashing since its a small change to the file overall)

mirceahasegan (approved):
It will be a great improvement having these e2e tests in the CI.
Thank you @AngelCastilloB! 🚀

packages/e2e/test/wallet/SingleAddressWallet/delegation.test.ts
```diff
@@ -151,13 +151,6 @@ describe('SingleAddressWallet/delegation', () => {
     expect(tx1PendingState.balance.deposit).toEqual(stakeKeyDeposit);
 
     expect(tx1ConfirmedState.rewardAccount.delegatee?.nextNextEpoch!.id).toEqual(poolId);
-    // nothing changes for 2 epochs
-    expect(tx1ConfirmedState.rewardAccount.delegatee?.nextEpoch).toEqual(
-      initialState.rewardAccount?.delegatee?.nextEpoch
-    );
-    expect(tx1ConfirmedState.rewardAccount.delegatee?.currentEpoch).toEqual(
-      initialState.rewardAccount?.delegatee?.currentEpoch
-    );
```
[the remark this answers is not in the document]
rhyslbw: > will fail since tx1ConfirmedState currentEpoch and nextEpoch will be one greater than initialState currentEpoch and nextEpoch.

`delegatee` is typed as this:
https://github.com/input-output-hk/cardano-js-sdk/blob/2a4ccb0abead34481e817f807850d29e77d7340a/packages/wallet/src/services/types.ts#L75-L83

... so the assertion could be refactored to check if the epoch has already rolled over, and use the `currentEpoch` to compare with the `nextEpoch` for example.

> Also, this current assert checks that nothing changes for two epochs, but isn't this guaranteed by the rules of the network? (seems to be an assert on the network behavior)

The intent of these assertions is to check that our state reflects the domain rules, being that delegations are not active for two epochs.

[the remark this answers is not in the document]
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/424/commits/ef3462834c092294ee4f15176be6e2dc72b953f4

rhyslbw (approved):
Nice work @AngelCastilloB :muscle:

PR #427, opened 2022-09-07
Feat/tx builder stake key dereg

mirceahasegan:
# Context

1. TxBuilder `delegate(...)` call is async only because it fetches rewardAccounts in order to generate certificates.
This forces users to await for delegate call and check for errors.
1. User cannot deregister stake key using TxBuilder.
1. User can use OutputBuilder txOut snapshot in `addOutput`.

# Proposed Solution
1. Postpone fetching rewards accounts and adding certificates until `TxBuilder.build()`
`delegate(...)` is stored as an intent configuration in TxBuilder, which is inspected and applied at `build()` time.
1. Update api to allow calling `delegate()` with undefined `poolId`. Doing this is considered a `deregister` request, thus adding the Deregister certificate at build time. 
1. OutputBuilder `toTxOut()` can be used to get a snapshot of the transaction output built so far. This snapshot can be used in `txBuilder.addOutput(...)`. TxBuilder will validate the provided output(s) on `build` by calling the `OutputBuilder.build()` method on all outputs.

- `delegate(...)` is no longer async.
- `OutputBuilder.partialOutput` is private.

### DIFF
--- a/packages/wallet/src/TxBuilder/OutputBuilder.ts
+++ b/packages/wallet/src/TxBuilder/OutputBuilder.ts
@@ -11,6 +11,14 @@ import {
 import { OutputValidation } from '../types';
 import { OutputValidator } from '../services';
 
+/** Properties needed to construct an {@link ObservableWalletTxOutputBuilder} */
+export interface OutputBuilderProps {
+  /** This validator is normally created and passed as an arg here by the {@link TxBuilder.buildOutput} method */
+  outputValidator: OutputValidator;
+  /** Optional partial transaction output to use for initialization. */
+  txOut?: PartialTxOut;
+}
+
 /** Determines if the `PartialTxOut` arg have at least an address and coins. */
 const isViableTxOut = (txOut: PartialTxOut): txOut is Cardano.TxOut => !!(txOut?.address && txOut?.value?.coins);
 
@@ -31,22 +39,17 @@ const toOutputValidationError = (
 };
 
 /**
- * `OutputBuilder` implementation based on the minimal wallet type.
+ * `OutputBuilder` implementation based on the minimal wallet type: {@link ObservableWalletTxBuilderDependencies}.
  */
 export class ObservableWalletTxOutputBuilder implements OutputBuilder {
   /**
-   * Transaction output that is updated by `OutputBuilder` methods.
+   * Transaction output that is updated by `ObservableWalletTxOutputBuilder` methods.
    * Every method call recreates the `partialOutput`, thus updating it immutably.
    */
   #partialOutput: PartialTxOut;
   #outputValidator: OutputValidator;
 
-  /**
-   *
-   * @param outputValidator this validator is normally created and passed as an arg here, by the TxBuilder
-   * @param txOut optional partial transaction output to use for initialization.
-   */
-  constructor(outputValidator: OutputValidator, txOut?: PartialTxOut) {
+  constructor({ outputValidator, txOut }: OutputBuilderProps) {
     this.#partialOutput = { ...txOut };
     this.#outputValidator = outputValidator;
   }
--- a/packages/wallet/src/TxBuilder/OutputBuilder.ts
+++ b/packages/wallet/src/TxBuilder/OutputBuilder.ts
@@ -18,7 +18,7 @@ const isViableTxOut = (txOut: PartialTxOut): txOut is Cardano.TxOut => !!(txOut?
  * Transforms from `OutputValidation` type emitted by `OutputValidator`, to
  * `OutputValidationMinimumCoinError` | `OutputValidationTokenBundleSizeError`
  */
-export const toOutputValidationError = (
+const toOutputValidationError = (
   txOut: Cardano.TxOut,
   validation: OutputValidation
 ): OutputValidationMinimumCoinError | OutputValidationTokenBundleSizeError | undefined => {
@@ -34,8 +34,11 @@ export const toOutputValidationError = (
  * `OutputBuilder` implementation based on the minimal wallet type.
  */
 export class ObservableWalletTxOutputBuilder implements OutputBuilder {
-  partialOutput: PartialTxOut;
-
+  /**
+   * Transaction output that is updated by `OutputBuilder` methods.
+   * Every method call recreates the `partialOutput`, thus updating it immutably.
+   */
+  #partialOutput: PartialTxOut;
   #outputValidator: OutputValidator;
 
   /**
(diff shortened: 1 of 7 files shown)


packages/wallet/src/TxBuilder/buildTx.ts
```diff
@@ -189,6 +199,46 @@ export class ObservableWalletTxBuilder implements TxBuilder {
...
+      // This shouldn't happen
+      throw new IncompatibleWalletError();
+    }
+
+    // Discard previous delegation and prepare for new one
+    this.partialTxBody = { ...this.partialTxBody, certificates: [] };
+
+    for (const rewardAccount of rewardAccounts) {
+      const stakeKeyHash = Cardano.Ed25519KeyHash.fromRewardAccount(rewardAccount.address);
+      if (this.#delegateConfig.type === 'deregister' && rewardAccount.keyStatus !== StakeKeyStatus.Unregistered) {
```
mkazlauskas, line 220: Would be great to `logger.warn` here if `rewardAccount.keyStatus === StakeKeyStatus.Unregistered`

packages/wallet/src/TxBuilder/OutputBuilder.ts
```diff
@@ -83,20 +53,12 @@ export class ObservableWalletTxOutputBuilder implements OutputBuilder {
     return this;
   }
 
-  async build(): Promise<MaybeValidTxOut> {
+  build(): MaybeValidTxOut {
```
mkazlauskas, line 56: `MaybeValidTxOut` here isn't  great since we're not doing full validation. How about rolling back changes to `OutputBuilder` and doing sth like:

```
interface TxBuilder {
  addOutput(TxOut | OutputBuilder);
}
```

Then we can call the async `build` internally when `TxBuilder.build` is called.

mkazlauskas (changes requested):

packages/wallet/src/TxBuilder/OutputBuilder.ts
mirceahasegan: So users could either run async `OutputBuilder.build()` themselves if they want to check the validity, or pass the output directly into TxBuilder, which will call `OutputBuilder.build()` when building the transaction?

mkazlauskas: > So users could either run async OutputBuilder.build() themselves if they want to check the validity, or pass the output directly into TxBuilder, which will call OutputBuilder.build() when building the transaction?

Yes, that's the idea

> If so, we could rename OutputBuilder.build() to OutputBuilder.validate()

Not sure about this - it can still be used to build an output (use case isn't limited to just validating)

mirceahasegan: Done in [9165582](https://github.com/input-output-hk/cardano-js-sdk/pull/427/commits/9165582d01e8eccd409d731d88a40e39905f4570)

packages/wallet/src/TxBuilder/buildTx.ts
mirceahasegan: [8e14559](https://github.com/input-output-hk/cardano-js-sdk/pull/427/commits/8e145591f64f1a8596d23642196267409543cae3)

mirceahasegan:
For the moment I am not squashing so that the review is easier to do.
I will squash to a smaller number of commits after the final review.

mirceahasegan:
[f5a779e](https://github.com/input-output-hk/cardano-js-sdk/pull/427/commits/f5a779ed6065c3c89f55a945a3b856ceaeb2dfb9) contains only some improvements for the generated docs

packages/wallet/src/TxBuilder/buildTx.ts
```diff
@@ -170,19 +173,70 @@ export class ObservableWalletTxBuilder implements TxBuilder {
...
+    // Discard previous delegation and prepare for new one
+    this.partialTxBody = { ...this.partialTxBody, certificates: [] };
+
+    for (const rewardAccount of rewardAccounts) {
+      const stakeKeyHash = Cardano.Ed25519KeyHash.fromRewardAccount(rewardAccount.address);
+      if (this.#delegateConfig.type === 'deregister') {
+        // Deregister scenario
+        if (rewardAccount.keyStatus === StakeKeyStatus.Unregistered) {
+          this.#logger.warn(
+            'Skipping stake key deregister. Reward account not registered',
```
rhyslbw, line 208:
```suggestion
            'Skipping stake key deregister. Stake key not registered',
```
Domain terminology will always refer to the stake key being registered, rather than "reward account"

packages/wallet/src/TxBuilder/buildTx.ts
```diff
@@ -170,19 +173,70 @@ export class ObservableWalletTxBuilder implements TxBuilder {
...
+          this.partialTxBody.certificates!.push({
+            __typename: Cardano.CertificateType.StakeKeyDeregistration,
+            stakeKeyHash
+          });
+        }
+      } else if (this.#delegateConfig.type === 'delegate') {
+        // Register and delegate scenario
+        if (rewardAccount.keyStatus !== StakeKeyStatus.Unregistered) {
+          this.#logger.warn(
+            'Skipping stake key register. Reward account already registered',
```
rhyslbw, line 222:
```suggestion
            'Skipping stake key register. Stake key already registered',
```

packages/wallet/src/TxBuilder/buildTx.ts
```diff
@@ -170,19 +173,70 @@ export class ObservableWalletTxBuilder implements TxBuilder {
...
+        } else {
+          this.partialTxBody.certificates!.push({
+            __typename: Cardano.CertificateType.StakeKeyDeregistration,
+            stakeKeyHash
+          });
+        }
+      } else if (this.#delegateConfig.type === 'delegate') {
+        // Register and delegate scenario
+        if (rewardAccount.keyStatus !== StakeKeyStatus.Unregistered) {
+          this.#logger.warn(
```
mkazlauskas, line 221: De-register deserves a warning because it's user's action that doesn't make sense. In this case re-delegating is a normal flow.

```suggestion
          this.#logger.debug(
```

mkazlauskas (changes requested):
Looks great, just 1 minor change request

packages/wallet/src/TxBuilder/buildTx.ts
mirceahasegan: [ccdd790](https://github.com/input-output-hk/cardano-js-sdk/pull/427/commits/ccdd7900f9104273db3cfc67f1ca133dd7f6d9de)

packages/wallet/src/TxBuilder/buildTx.ts
mirceahasegan: [ccdd790](https://github.com/input-output-hk/cardano-js-sdk/pull/427/commits/ccdd7900f9104273db3cfc67f1ca133dd7f6d9de)

packages/wallet/src/TxBuilder/buildTx.ts
mirceahasegan: [ccdd790](https://github.com/input-output-hk/cardano-js-sdk/pull/427/commits/ccdd7900f9104273db3cfc67f1ca133dd7f6d9de)

packages/wallet/test/integration/buildTx.test.ts
```diff
@@ -346,7 +346,7 @@ describe('buildTx', () => {
       expect(txBuilt.body.certificates?.length).toBe(2);
     });
 
-    it('adds both stake key and delegation certificates when reward account was not registered', async () => {
+    it('adds both stake key and delegation certificates when stake key was not registered', async () => {
```
rhyslbw, line 349: :+1:

mkazlauskas (approved):
Nice work! :rocket:

rhyslbw (dismissed):

mirceahasegan:
> Nice work! 🚀

Thank you. I have squashed and force pushed.

rhyslbw (approved):

mkazlauskas (approved):

PR #428, opened 2022-09-07
Refactor/adp 1894 runnable provider

iccicci:
# Context

Recently we faced issues with modifying interfaces from `core` package and their reusability across different packages such as `cardano-services`, `rabbitmq` and `ogmios`.

# Proposed Solution

- Removed the `start` and `close` methods from the wrong place: the `interface Provider`;
- Introduced the `interface RunnableProvider` in `cardano-services` (which `extends` the latter) and implemented it in all the providers requiring it.


packages/cardano-services/src/TxSubmit/rabbitmq/RabbitMqTxSubmitProvider.ts
```diff
@@ -63,14 +64,14 @@ export class RabbitMqTxSubmitProvider implements TxSubmitProvider {
   /**
    * Connects to the RabbitMQ server and create the channel
    */
-  async #connectAndCreateChannel() {
+  async start() {
```
IvayloAndonov, line 67: 👍🏻

IvayloAndonov (dismissed):
LGTM! Much cleaner now

packages/cardano-services/src/util/provider.ts
```diff
@@ -29,3 +29,8 @@ export const providerHandler =
     }
     handler(req.body, req, res, next, handlerFn);
   };
+
+export interface RunnableProvider extends Provider {
```
rhyslbw, line 33: Better to put this in `./src/types.ts`

rhyslbw (changes requested):
The discussion we had during planning was to look at a solution that incorperated [RunnableModule](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services/src/RunnableModule.ts) too. Have you considered making the interface `Runnable`, then either using this in the providers that have runnable traits, or defining a `RunnableProvider` as a composition?

mkazlauskas (changes requested):
Looking good! `RunnableModule` should probably explicitly implement `Runnable`.

PR #429, opened 2022-09-07
Refactor/key management governance packages

rhyslbw:
# Context
ADP-2031

# Proposed Solution
Make the smallest number of changes to extract these two concepts into their own packages.

- A number of common types, and util was lifted to common packages, which I've kept to a minimum to not bloat this PR. We can consider lifting more to keep the design consistent, but not in this scope.
- 6c940ed16258bcd3ab97ed882cce28907c1edf8e was needed to keep `tsc` happy


packages/key-management/src/errors/index.ts
```diff
@@ -1,6 +1,6 @@
 export * from './AuthenticationError';
 export * from './InvalidMnemonicError';
-export * from './InvalidSerializableDataError ';
```
rhyslbw, line 3: :eyes:

packages/key-management/src/util/cachedGetPassword.ts
```diff
@@ -1,5 +1,6 @@
 import { GetPassword } from '../types';
-import { Milliseconds } from '../../services';
+
+type Milliseconds = number;
```
rhyslbw, line 3: I just defined this alias again for simplicity. Could also lift to `util` package

mkazlauskas: I thought we deleted this util, but apparently not. I think we should delete it as it's a security hazard.

packages/key-management/test/KeyAgentBase.test.ts
```diff
@@ -1,13 +1,20 @@
...
 
 const NETWORK_ID = Cardano.NetworkId.testnet;
 const ACCOUNT_INDEX = 1;
 
-class MockKeyAgent extends KeyManagement.KeyAgentBase {
-  constructor(data: KeyManagement.SerializableInMemoryKeyAgentData) {
-    super(data, { inputResolver: createLazyWalletUtil() });
+class MockKeyAgent extends KeyAgentBase {
+  constructor(data: SerializableInMemoryKeyAgentData) {
+    super(data, { inputResolver: { resolveInputAddress: () => Promise.resolve(null) } });
```
rhyslbw, line 17: Basic object stubbed to remove the circular dependency

packages/wallet/test/SingleAddressWallet/methods.test.ts
```diff
@@ -1,16 +1,15 @@
...
 import { Cardano } from '@cardano-sdk/core';
-import { InitializeTxProps, KeyManagement, SingleAddressWallet, setupWallet } from '../../src';
+import { InitializeTxProps, SingleAddressWallet, setupWallet } from '../../src';
 import { firstValueFrom, skip } from 'rxjs';
+import { getPassword, testAsyncKeyAgent } from '../../../key-management/test/mocks';
 import { dummyLogger as logger } from 'ts-log';
 import { mockChainHistoryProvider, mockRewardsProvider, utxo } from '../mocks';
 import { waitForWalletStateSettle } from '../util';
 
-jest.mock('../../src/KeyManagement/cip8/cip30signData');
```
rhyslbw, line 11: Jest, with our existing config, wasn't happy referring to a sibling package, so I just substituted the assertion to inspect the return value. This is easier than managing tedious path references, although I'd be happy to change it back if you can see why it's failing

packages/key-management/src/util/cachedGetPassword.ts
rhyslbw: 067191f8f4f036cd38d64bd1e4520cd8ba3ae3a7

packages/core/src/Cardano/types/Transaction.ts
```diff
@@ -100,3 +100,8 @@ export interface TxAlonzo extends NewTxAlonzo<TxBodyAlonzo> {
   body: TxBodyAlonzo;
   txSize: number;
 }
+
+export type TxInternals = {
```
mkazlauskas, line 104: I think this name isn't great now that it's no longer in scope of the wallet. Let's try to come up with a better one

mkazlauskas (approved):
Looks good! :rocket: we can probably move the entire `equals.ts` to `util` as you've indicated in commit msg, but doesn't need to happen in this PR

packages/wallet/test/SingleAddressWallet/methods.test.ts
mkazlauskas: I think this is fine

packages/core/src/Cardano/types/Transaction.ts
rhyslbw: `TxBodyWithHash`?

mkazlauskas: Yeah that's a decent one, definitely better than TxInternals

rhyslbw: [b946473](https://github.com/input-output-hk/cardano-js-sdk/pull/429/commits/b94647391a315ba376c91cf345f44119410ea8c1)

mirceahasegan (approved):
🚀

PR #430, opened 2022-09-08
feat(core): narrow down BootstrapWitness types

mkazlauskas:
# Context

It is not clear what format/encoding the following strings use:

```ts
export type ChainCode = string;
export type AddressAttributes = string;
```

# Proposed Solution

See notes in commit message a217e00161e3e8b86fd1df5cc780fed20d213708

### DIFF
--- a/packages/core/test/Cardano/util/primitives.test.ts
+++ b/packages/core/test/Cardano/util/primitives.test.ts
@@ -1,5 +1,6 @@
 /* eslint-disable sonarjs/no-duplicate-string */
 import {
+  Base64Blob,
   Hash28ByteBase16,
   Hash32ByteBase16,
   HexBlob,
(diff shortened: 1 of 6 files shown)


packages/core/test/Cardano/util/primitives.test.ts
```diff
@@ -85,6 +86,30 @@ describe('Cardano.util/primitives', () => {
...
+    });
+  });
+
+  describe('Base64Blob', () => {
+    it('allows an empty string', () => {
+      expect(() => Base64Blob('')).not.toThrow();
+    });
+
+    it('does not throw when asserting a valid base64 encoded string', () => {
+      // from ogmios.wsp.json
```
rhyslbw, line 101: What's from `ogmios.wsp.json`?

rhyslbw (changes requested):
Great work with the commit message, but I think it's more relevant to be typed as a `fix` if separating the feature is not practical.

packages/core/test/Cardano/util/primitives.test.ts
mkazlauskas: The example base64 encoded string. Not a very useful comment in this context, should I remove it?

rhyslbw: Yeah

mkazlauskas: Amended the commit (didn't do fixup since it's only 1 commit and only 1 change request)

rhyslbw (approved):

AngelCastilloB (approved):
LGTM! :rocket:

rhyslbw (approved):

PR #431, opened 2022-09-09
Refactor runnable provider alternative

iccicci:
# Context

Recently we faced issues with modifying interfaces from `core` package and their reusability across different packages such as `cardano-services`, `rabbitmq` and `ogmios`.

# Proposed Solution

- Removed the `start` and `close` methods from the wrong place: the `interface Provider`;
- Introduced the `interface RunnableProvider` in `cardano-services` (which `extends` the latter) and implemented it in all the providers requiring it.


packages/cardano-services/src/DbSyncProvider.ts
```diff
@@ -1,12 +1,15 @@
 import { HealthCheckResponse, Provider } from '@cardano-sdk/core';
+import { Logger } from 'ts-log';
 import { Pool } from 'pg';
+import { RunnableModule } from './RunnableModule';
 
 const HEALTH_CHECK_QUERY = 'SELECT 1';
 
-export abstract class DbSyncProvider implements Provider {
+export class DbSyncProvider extends RunnableModule implements Provider {
```
rhyslbw, line 8: This doesn't need to be runnable

packages/cardano-services/test/Http/HttpServer.test.ts
```diff
@@ -74,7 +75,9 @@ describe('HttpServer', () => {
   beforeEach(async () => {
     port = await getRandomPort();
     apiUrlBase = `http://localhost:${port}`;
-    provider = new DbSyncUtxoProvider(db, logger);
+    const providerInternal = new DbSyncUtxoProvider(db, logger);
```
rhyslbw, line 78: These test changes are out of scope, so it would be best to commit them separately.

rhyslbw (changes requested):
If taking this approach we should consider hoisting the logger setup out of `RunnableModule`, and maybe also the `initialize` method.   

17ecae4ed2d16073bb072c7ca8cfed657348a8ae is not an atomic commit, as it breaks the implementations, so will need to be committed with the refactoring.

mkazlauskas (changes requested):
Agree with @rhyslbw about hoisting contextLogger setup out of `RunnableModule`

packages/cardano-services/src/Http/HttpService.ts
```diff
@@ -33,17 +33,17 @@ export abstract class HttpService extends RunnableModule {
...
-      throw new ProviderError(ProviderFailure.Unhealthy);
-    }
+    if (this.provider instanceof RunnableModule) await this.provider.initialize();
+
+    if (!(await this.healthCheck()).ok) throw new ProviderError(ProviderFailure.Unhealthy);
   }
 
   async startImpl(): Promise<void> {
-    await this.provider.start?.();
+    if (this.provider instanceof RunnableModule) await this.provider.start();
```
mkazlauskas, line 42: I think the 2 approaches can be combined. `Runnable` as an interface is still a good idea here:
```suggestion
    function isRunnable(obj: unknown): obj is Runnable { return 'start' in obj && 'shutdown' in obj }
    if (isRunnable(this.provider)) await this.provider.start();
```

iccicci:
Hi all,

I just pushed a new proposal based on @mirceahasegan suggestion.
Tests still hang, but at least all the syntax error are solved I could do the preview push.
Probably I've missed a `start` call... still working on it.

If you want to take a look too @rhyslbw , @mkazlauskas and @IvayloAndonov , you are welcome.

packages/cardano-services/src/DbSyncProvider.ts
iccicci: Completely refactored in last proposal

packages/cardano-services/test/Http/HttpServer.test.ts
iccicci: Rolled back with last proposal

packages/cardano-services/src/DbSyncProvider.ts
```diff
@@ -3,19 +3,113 @@ import { Pool } from 'pg';
 
 const HEALTH_CHECK_QUERY = 'SELECT 1';
 
-export abstract class DbSyncProvider implements Provider {
-  public db: Pool;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type AnyArgs = any[];
```
mkazlauskas, line 7: In general we're following a pattern of 2 args: `(props, dependencies)` so it could probably be typed with generic arguments <TProps, TDependencies>

mkazlauskas (commented):
This looks like a good approach. I'm wondering if it would make sense to use "DbSyncProvider" as a mixin (the other way around) and have all the mixin logic within RunnableModule

packages/cardano-services/src/DbSyncProvider.ts
iccicci: You hit two distinct topics here, @mkazlauskas .

`type AnyArgs = any[]` is more or less mandatory as TypeScript requires the `constructor` of a _mixin_ to be of type `new(...args: any[]) => T`; but this is just an implementation detail. We can (and IMHO we should) achieve what you are describing working on the tuples inferred from the arguments of the two constructors.

While about the core of the suggestion, I 100% agree we should do that. We just need to simplify the way to achieve it, please let me explain in details.

Now the code simply concatenates the two tuples inferred from the constructors: two arbitrary tuples. trying to merge the types we have as elements of arbitrary tuples risk to increase a lot the complexity of the code. As an example, the `RunnableModule.constructor` itself doesn't respect the `(props, deps)` pattern.
I thought that in order to achieve this more standardization would be helpful to standardize all or `Provider` implementations (and probably `Service`s as well) to something similar to:

```typescript
type ProviderConfiguration = { [k: string]?: unknown };

interface ProviderDependencies {
  logger: Logger;
  db?: Pool;
  cache?: MemCache;
  ...
}

interface ProviderImplementation {
  new(config: ProviderConfiguration, dependencies: ProviderDependencies): Provider;
}
```

This will make the code required much more easy and clean than the one required now.
On the other side to implement this on all our `Provider`s and `Service`s (which means do the same for all the configuration and dependencies interfaces) would require a dedicated epic split in several tasks.

What do you think about?

mkazlauskas: > working on the tuples inferred from the arguments of the two constructors.

> trying to merge the types we have as elements of arbitrary tuples risk to increase a lot the complexity of the code

Sounds good, but if there's no obvious/easy way to do that then I'm fine with any[] as long as it's contained within a single class (any[] doesn't bleed to other classes).

> I thought that in order to achieve this more standardization would be helpful to standardize all or Provider implementations

Agree - we've been using a similar pattern for the most part. I'm sure there are some different signatures, but it should be an easy task to refactor to make all services compliant. Note: I think this is more of a "Service" thing than a "Provider" thing - provider is just an interface, where services within `cardano-services` package implements it.

What are your thoughts on this?:

> I'm wondering if it would make sense to use "DbSyncProvider" as a mixin (the other way around) and have all the mixin logic within RunnableModule

iccicci: > Note: I think this is more of a "Service" thing than a "Provider" thing - provider is just an interface, where services within cardano-services package implements it.

I agree this is not a `Provider` related topic, (sorry if I was not clear enough); in the example on the fly I called it `ProviderImplementation` but I was referring strictly to the domain of `cardano.services` domain.

> I'm wondering if it would make sense to use "DbSyncProvider" as a mixin (the other way around) and have all the mixin logic within RunnableModule

I'm not sure I got the point; if you are proposing to leave `DbSyncProvider` a standard class and implementing the `RunnableModule` as the mixin; this is not a viable solution due to a limitation of my implementation: it doesn't let us the overload the methods of the mixin, and the design of `RunnableModule` requires subclasses do that.

If I answered to another question than your one, please help me better understanding.

mkazlauskas: Yes, I thought maybe DbSyncProvider can be a mixin, so maybe it could look like `extends HttpService(RunnableModule, DbSyncProvider)`

mirceahasegan (approved):
Great work, @iccicci 🚀

packages/cardano-services/src/DbSyncProvider.ts
```diff
@@ -3,19 +3,113 @@ import { Pool } from 'pg';
...
 
-  protected constructor(db: Pool) {
-    this.db = db;
+export const DbSyncProvider = <
```
IvayloAndonov, line 9: I was out of this refactoring topic and I need first to read the whole discussion above in order to understand each suggestion in detail, but at first look, my personal opinion is that having all these types and abstractions is a bit "overengineered". The reason is: it's initially hard for me to understand the usage of it without having the demo usage lines or any other readme doc

Also not sure how flexible for extensions would be, considering we need to add new functionality/deps to the base `DbSyncProvider` class (like `CardanoNode`)

mkazlauskas: Do you like the original implementation in #428 better?

packages/cardano-services/src/DbSyncProvider.ts
```diff
@@ -3,19 +3,113 @@ import { Pool } from 'pg';
...
+  return Mixin as unknown as (T extends new (...baseArgs: AnyArgs) => {}
+    ? new (db: Pool, ...args: BaseArgs) => ReturnedType
+    : abstract new (db: Pool, ...args: BaseArgs) => ReturnedType) & {
+    prototype: { healthCheck: () => Promise<HealthCheckResponse> };
+  };
+};
+
+export type DbSyncProvider<T> = T extends new (...args: AnyArgs) => infer O ? O : never;
+
+// TODO: delete following
```
mkazlauskas, line 52: Please do the TODO

packages/cardano-services/src/DbSyncProvider.ts
```diff
@@ -3,19 +3,113 @@ import { Pool } from 'pg';
...
+    }
+  }
+
+  const testProvider3 = new TestProvider3(db, 23, 'test');
+  const test3 = [await testProvider3.healthCheck(), testProvider3.test('test'), testProvider3.check(23)] as const;
+
+  const BaseTestProvider4 = DbSyncProvider(AbstractClass);
+
+  // Uncomment this to check
+  // const baseTestProvider4 = new BaseTestProvider4();
```
mkazlauskas, line 101: Why is this commented out?

mkazlauskas (changes requested):
Looks good, just need to clean up a few inline comments

packages/cardano-services/src/DbSyncProvider.ts
IvayloAndonov: Yep, so once I read all discussion and check the other PR will be able to express opinion, agree about the Mixin class and the need of docs/usage.

Also to keep in mind the new implementation should be opened for extensions easily, I refer to that `DbSyncProvider` will be probably extended in near future, even now with another dependency of `CardanoNode`

IvayloAndonov: The #428 approach looks much more intuitive and simplified, so my question is what was the key advantage of this #431 variant of implementation over the #428?

packages/cardano-services/src/DbSyncProvider.ts
iccicci: Yes, I left there as a sort of preview of documentation

packages/cardano-services/src/DbSyncProvider.ts
iccicci: It is a sort of preview test about the constructor produced by the mixin: uncommenting it, it will rise a compile error about missing parameters

packages/cardano-services/src/DbSyncProvider.ts
iccicci: Hi @IvayloAndonov , this implementation is more robust as we can trust the properties of _classes inheritance_, as an example this let us to check if a given `DbSyncProvider` is a `RunnableModule` as well through an `instanceof` check rather than checking by hands for the presence of methods.

yarn-project.nix
```diff
@@ -1246,6 +1246,7 @@ cacheEntries = {
 "@lerna/symlink-dependencies@npm:5.2.0" = { filename = "@lerna-symlink-dependencies-npm-5.2.0-a6aa998578-4a00e85ac8.zip"; sha512 = "4a00e85ac8bd63896834978cd6d380fecb4093098f09e7d53112b4543ea3ab70b563c24467748c6a4d1e77a25cb0a463f2a3284339bf7366eacff9f1d2ca6593"; };
 "slice-ansi@npm:3.0.0" = { filename = "slice-ansi-npm-3.0.0-d9999864af-5ec6d022d1.zip"; sha512 = "5ec6d022d12e016347e9e3e98a7eb2a592213a43a65f1b61b74d2c78288da0aded781f665807a9f3876b9daa9ad94f64f77d7633a0458876c3a4fdc4eb223f24"; };
 "minipass-fetch@npm:2.1.0" = { filename = "minipass-fetch-npm-2.1.0-300ce55188-1334732859.zip"; sha512 = "1334732859a3f7959ed22589bafd9c40384b885aebb5932328071c33f86b3eb181d54c86919675d1825ab5f1c8e4f328878c863873258d113c29d79a4b0c9c9f"; };
+"fsevents@patch:fsevents@npm%3A2.3.2#~builtin<compat/fsevents>::version=2.3.2&hash=18f3a7" = { filename = "fsevents-patch-3340e2eb10-8.zip"; sha512 = "edbd0fd80be379c14409605f77e52fdc78a119e17f875e8b90a220c3e5b29e54a1477c21d91fd30b957ea4866406dc3ff87b61432d2840ff8866b309e5866140"; };
```
mirceahasegan, line 1249: I wonder why this change is here? 🤔

mirceahasegan (approved):

packages/cardano-services/src/DbSyncProvider.ts
iccicci: As agreed in the standup, I'm resolving this

packages/cardano-services/src/DbSyncProvider.ts
iccicci: As agreed in the standup, I'm resolving this

packages/cardano-services/src/DbSyncProvider.ts
iccicci: @mkazlauskas , 0c1b5d5bc should resolve this.

yarn-project.nix
iccicci: thanks @mirceahasegan , a refuse... 0c1b5d5bc should resolve this.

packages/cardano-services/src/DbSyncProvider.ts
iccicci: sorry @mirceahasegan , 2f2c8e7d4 should really solve

mirceahasegan (approved):
Fantastic work, @iccicci 👍

mkazlauskas (approved):
Nice work! Please squash and rebase

mkazlauskas (dismissed):

mirceahasegan (approved):

iadmytro (approved):

mkazlauskas (approved):

IvayloAndonov (approved):

PR #432, opened 2022-09-09
feat(cardano-services): stake pools apy and margin averages

jvieiro:
# Context

Provide average APY and margin of all active stake pools for latest epoch at StakePoolProvider

### DIFF
--- a/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/queries.ts
+++ b/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/queries.ts
@@ -766,7 +766,7 @@ SELECT
 FROM (${query}) as query
 `;
 
-export const findPoolStats = `
+export const findPoolQty = `
 WITH current_epoch AS (
 	SELECT max(epoch_no) AS epoch_no 
 	FROM block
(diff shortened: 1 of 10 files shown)


packages/core/src/Provider/StakePoolProvider/types/StakePoolProvider.ts
```diff
@@ -58,6 +58,10 @@ export interface StakePoolSearchResults {
 }
 
 export interface StakePoolStats {
+  averages: {
+    margin: number;
```
mkazlauskas, line 62: Please document what the new values are, e.g.

```suggestion
    /** 
     * Within range [0; 1]
     */
    margin: Percent;
```

packages/cardano-services/test/StakePool/DbSyncStakePoolProvider/__snapshots__/StakePoolBuilder.test.ts.snap
```diff
@@ -1518,6 +1518,10 @@ Array [
 
 exports[`StakePoolBuilder queryPoolStats returns active, retired and retiring pools count 1`] = `
 Object {
+  "averages": Object {
+    "apy": 0.0000158041676290942,
+    "margin": 0.000201057770545159,
```
mkazlauskas, line 1523: This value seems too low, but could be due to the data snapshot. We need to verify that it is a realistic number. cc @iadmytro

mkazlauskas (changes requested):

packages/cardano-services/test/StakePool/DbSyncStakePoolProvider/__snapshots__/StakePoolBuilder.test.ts.snap
jvieiro: Yep, this is due to the db snapshot, there are many missing rows of `pool_update` table. So, a lot of pools have margin = 0

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/queries.ts
```diff
@@ -821,6 +821,114 @@ FROM last_pool_update AS pool_update
 LEFT JOIN last_pool_retire AS pool_retire 
 	ON pool_update.hash_id = pool_retire.hash_id`;
 
+const findActivePoolsAverages = `
```
iadmytro, line 824: we definitely need tests for this query

iadmytro (changes requested):

PR #433, opened 2022-09-12
feat: calculate stability window slots count in wallet

IvayloAndonov:
# Context
The stability window must be calculated to determine which confirmed transactions might be subject to rollback.

[The formula to calculate this window is 3k/f (where k is the security parameter in genesis, and f is the active slot co-efficient parameter in genesis that determines the probability of amount of blocks created in an epoch.)](https://docs.cardano.org/learn/chain-confirmation-versus-transaction-confirmation)

Once NetworkInfo includes the parameters, it should be used to calculate the stability window and use it instead of the hardcoded value.

# Proposed Solution
Apply the stability window formula on the client (wallet) and provide the value as observable where it is needed.

- extend `createTransactionReemitter` with `stabilityWindowSlotsCount$`
- apply the formula to calculate `stabilityWindowSlotsCount` value
- update marble tests

### DIFF
--- a/packages/core/src/util/calcStabilityWindow.ts
+++ b/packages/core/src/util/calcStabilityWindow.ts
@@ -1,0 +1,11 @@
+import { Cardano } from '..';
+
+/**
+ * 3k/f (where k is the security parameter in genesis, and f is the active slot co-efficient parameter
+ * in genesis that determines the probability for amount of blocks created in an epoch.)
+ */
+export const calculateStabilityWindowSlotsCount = ({
+  securityParameter,
+  activeSlotsCoefficient
+}: Pick<Cardano.CompactGenesis, 'securityParameter' | 'activeSlotsCoefficient'>): number =>
+  (3 * securityParameter) / activeSlotsCoefficient;
--- a/packages/wallet/src/SingleAddressWallet.ts
+++ b/packages/wallet/src/SingleAddressWallet.ts
@@ -307,6 +307,7 @@ export class SingleAddressWallet implements ObservableWallet {
 
     this.#resubmitSubscription = createTransactionReemitter({
       confirmed$: this.transactions.outgoing.confirmed$,
+      genesisParameters$: this.genesisParameters$,
       logger: contextLogger(this.#logger, 'transactionsReemitter'),
       rollback$: this.transactions.rollback$,
       store: stores.volatileTransactions,
--- a/packages/wallet/src/services/TransactionReemitter.ts
+++ b/packages/wallet/src/services/TransactionReemitter.ts
@@ -1,4 +1,4 @@
-import { Cardano } from '@cardano-sdk/core';
+import { Cardano, calculateStabilityWindowSlotsCount } from '@cardano-sdk/core';
 import { Logger } from 'ts-log';
 import { Observable, filter, from, map, merge, mergeMap, scan, tap, withLatestFrom } from 'rxjs';
 
@@ -24,7 +24,7 @@ interface TransactionReemitterProps {
   submitting$: Observable<Cardano.NewTxAlonzo>;
   store: DocumentStore<NewTxAlonzoWithSlot[]>;
   tipSlot$: Observable<Cardano.Slot>;
-  stabilityWindowSlotsCount?: number;
+  genesisParameters$: Observable<Pick<Cardano.CompactGenesis, 'securityParameter' | 'activeSlotsCoefficient'>>;
   logger: Logger;
 }
 
@@ -34,17 +34,13 @@ enum txSource {
   submitting
 }
 
-// 3k/f (where k is the security parameter in genesis, and f is the active slot co-efficient parameter
-// in genesis that determines the probability for amount of blocks created in an epoch.)
-const kStabilityWindowSlotsCount = 129_600; // 3k/f on current mainnet
-
 export const createTransactionReemitter = ({
   rollback$,
   confirmed$,
   submitting$,
   store,
   tipSlot$,
-  stabilityWindowSlotsCount = kStabilityWindowSlotsCount,
+  genesisParameters$,
   logger
 }: TransactionReemitterProps): Observable<Cardano.NewTxAlonzo> => {
   const volatileTransactions$ = merge(
@@ -56,7 +52,15 @@ export const createTransactionReemitter = ({
     confirmed$.pipe(map((tx) => ({ source: txSource.confirmed, tx }))),
     submitting$.pipe(map((tx) => ({ source: txSource.submitting, tx: { ...tx, slot: null! } })))
   ).pipe(
-    scan((volatiles, { tx, source }) => {
+    mergeMap((vt) =>
+      genesisParameters$.pipe(
+        map(({ securityParameter, activeSlotsCoefficient }) =>
+          calculateStabilityWindowSlotsCount({ activeSlotsCoefficient, securityParameter })
+        ),
+        map((sw) => ({ sw, vt }))
+      )
+    ),
+    scan((volatiles, { vt: { tx, source }, sw: stabilityWindowSlotsCount }) => {
       switch (source) {
         case txSource.store: {
           // Do not calculate stability window for old transactions coming from the store
--- a/packages/core/src/util/index.ts
+++ b/packages/core/src/util/index.ts
@@ -2,3 +2,4 @@ export * as util from './misc';
 export * from './slotCalc';
 export * from './txInspector';
 export * from './utxo';
+export * from './calcStabilityWindow';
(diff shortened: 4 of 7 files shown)


packages/wallet/src/services/TransactionReemitter.ts
```diff
@@ -24,7 +24,7 @@ interface TransactionReemitterProps {
   submitting$: Observable<Cardano.NewTxAlonzo>;
   store: DocumentStore<NewTxAlonzoWithSlot[]>;
   tipSlot$: Observable<Cardano.Slot>;
-  stabilityWindowSlotsCount?: number;
```
mirceahasegan, line 27: Please mark the commit as breaking change.

packages/wallet/src/util.ts
```diff
@@ -0,0 +1,9 @@
+// 3k/f (where k is the security parameter in genesis, and f is the active slot co-efficient parameter
+// in genesis that determines the probability for amount of blocks created in an epoch.)
+
+import { Cardano } from '@cardano-sdk/core';
+
+export const calculateStabilityWindowSlotsCount = ({
+  securityParameter,
+  activeSlotsCoefficient
+}: Cardano.CompactGenesis) => (3 * securityParameter) / activeSlotsCoefficient;
```
mirceahasegan, line 9: Please add the return type (number?)

packages/wallet/test/services/util.test.ts
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { calculateStabilityWindowSlotsCount } from '../../src/util';
+
+describe('calculateStabilityWindowSlotsCount', () => {
+  it('computes stability window slots count', () => {
+    const securityParameter = 2160;
+    const activeSlotsCoefficient = 0.05;
+    const expectedStabilityWindowValue = 129_600;
+    expect(calculateStabilityWindowSlotsCount({ activeSlotsCoefficient, securityParameter } as any)).toEqual(
```
mirceahasegan, line 9: cast to any should not be necessary once the return type for `calculateStabilityWindowSlotsCount ` is specified

mirceahasegan (changes requested):

packages/wallet/src/util.ts
```diff
@@ -0,0 +1,9 @@
+// 3k/f (where k is the security parameter in genesis, and f is the active slot co-efficient parameter
+// in genesis that determines the probability for amount of blocks created in an epoch.)
+
+import { Cardano } from '@cardano-sdk/core';
+
+export const calculateStabilityWindowSlotsCount = ({
```
rhyslbw, line 6: This function would be better in `core` or `util` package, as it's not only relevant to our wallet package

rhyslbw (changes requested):

packages/wallet/test/services/util.test.ts
IvayloAndonov: Here I cast the function arguments (`genesis: CompactGenesis` type) to any, not the return type. Even with the return type for `calculateStabilityWindowSlotsCount` (`number`), the genesis type consists of many other props which we do not need in the test. @mirceahasegan

packages/wallet/src/services/TransactionReemitter.ts
IvayloAndonov: Yep, will reword the commit name once the PR is ready for merge, to avoid force pushing during the review process.

packages/wallet/test/services/util.test.ts
mirceahasegan: Got it now. Can you try to cast to `Partial<CompactGenesis>`, instead of `any`? Then it would still check that the properties included are actually part of `CompactGenesis`.

IvayloAndonov: yep, exactly what was going to change :)

packages/wallet/src/util.ts
IvayloAndonov: Applied in https://github.com/input-output-hk/cardano-js-sdk/pull/433/commits/4c407e6e7521733781853cc636c544f6613584ff

packages/wallet/test/services/util.test.ts
IvayloAndonov: Applied in https://github.com/input-output-hk/cardano-js-sdk/pull/433/commits/4c407e6e7521733781853cc636c544f6613584ff

packages/wallet/src/util.ts
IvayloAndonov: Applied in https://github.com/input-output-hk/cardano-js-sdk/pull/433/commits/4c407e6e7521733781853cc636c544f6613584ff @rhyslbw

mirceahasegan (dismissed):
Nice work @IvayloAndonov  👍

rhyslbw (changes requested):
I think for now this would be better put in the upper scope as `src/util/stabilityWindowCalc.ts`, which aligns with the other calc filename `slotCalc.ts`

IvayloAndonov:
> 

Now it's located as suggested: `src/util/stabilityWindowCalc.ts` - 95f44ec92b502271f276ef2fb8792be2220519ac @rhyslbw

rhyslbw (dismissed):

packages/core/src/util/calcStabilityWindow.ts
```diff
@@ -0,0 +1,9 @@
+// 3k/f (where k is the security parameter in genesis, and f is the active slot co-efficient parameter
```
mkazlauskas, line 1: This should be a
```
/**
 * doc comment on calculateStabilityWindowSlotsCount
 */
```

packages/core/src/util/calcStabilityWindow.ts
```diff
@@ -0,0 +1,9 @@
...
+// in genesis that determines the probability for amount of blocks created in an epoch.)
+
+import { Cardano } from '..';
+
+export const calculateStabilityWindowSlotsCount = ({
+  securityParameter,
+  activeSlotsCoefficient
+}: Cardano.CompactGenesis): number => (3 * securityParameter) / activeSlotsCoefficient;
```
mkazlauskas, line 9: The less constrained the utilities are the better
```suggestion
}: Pick<Cardano.CompactGenesis, 'securityParameter' | 'activeSlotsCoefficient'>): number => (3 * securityParameter) / activeSlotsCoefficient;
```

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -309,6 +310,9 @@ export class SingleAddressWallet implements ObservableWallet {
       confirmed$: this.transactions.outgoing.confirmed$,
       logger: contextLogger(this.#logger, 'transactionsReemitter'),
       rollback$: this.transactions.rollback$,
+      stabilityWindowSlotsCount$: this.genesisParameters$.pipe(
+        map((params) => calculateStabilityWindowSlotsCount(params))
```
mkazlauskas, line 314: Since it's not used anywhere else in SingleAddressWallet, I think it would be slightly better to move this .pipe(map) to createTransactionReemitter (passing just the necessary genesisParameters$ to it).

mkazlauskas (changes requested):

packages/wallet/src/SingleAddressWallet.ts
mirceahasegan: On the other hand, doing this passes unnecessary data to TransactionReemiter, instead of just the bare minimum it needs.

packages/core/src/util/calcStabilityWindow.ts
IvayloAndonov: c66f5797594c56d155aeb62836e4228622916a55

packages/wallet/src/SingleAddressWallet.ts
IvayloAndonov: c66f5797594c56d155aeb62836e4228622916a55

packages/core/src/util/calcStabilityWindow.ts
IvayloAndonov: c66f5797594c56d155aeb62836e4228622916a55

mirceahasegan (approved):
🚀

mkazlauskas (dismissed):
Nice work! :rocket:

rhyslbw (approved):
Ready for squashing

IvayloAndonov:
Rebased with _master,_ commits are squashed and the commit is marked as breaking change. (raised by @mirceahasegan [here](https://github.com/input-output-hk/cardano-js-sdk/pull/433#discussion_r968753483))

rhyslbw (approved):

PR #434, opened 2022-09-12
feat(wallet): txBuilder disable after submit

mirceahasegan:
# Context
ADP-2192
I want to know if the TxBuilder instance was already used to build and submit a transaction, so that I can avoid building and sending another transaction using the same instance, that still has the configurations (outputs/delegations/etc.) from the previously sent transaction.

# Proposed Solution
TxBuilder `build()` method throws `TxBuilderDisabled` exception, if a transaction built with the same TxBuilder instance was already successfully submitted.
TxBuilder can be reenabled by calling it's `forceEnable()` method, to allow reusing it's configuration if needed.

### DIFF
--- a/packages/wallet/src/TxBuilder/buildTx.ts
+++ b/packages/wallet/src/TxBuilder/buildTx.ts
@@ -1,16 +1,18 @@
 import { Cardano } from '@cardano-sdk/core';
 import { Observable, firstValueFrom } from 'rxjs';
 
-import { FinalizeTxProps, InitializeTxProps, ObservableWallet } from '../types';
+import { FinalizeTxProps, InitializeTxProps, InitializeTxResult, ObservableWallet } from '../types';
 import {
   IncompatibleWalletError,
   MaybeValidTx,
   OutputBuilder,
   PartialTxOut,
   SignedTx,
+  TxAlreadySubmittedError,
   TxBodyValidationError,
   TxBuilder,
-  TxOutValidationError
+  TxOutValidationError,
+  ValidTx
 } from './types';
 import { Logger } from 'ts-log';
 import { ObservableWalletTxOutputBuilder } from './OutputBuilder';
@@ -67,11 +69,15 @@ const createSignedTx = async ({
   tx,
   extraSigners,
   signingOptions,
-  auxiliaryData
-}: FinalizeTxProps & { wallet: ObservableWalletTxBuilderDependencies }): Promise<SignedTx> => {
+  auxiliaryData,
+  afterSubmitCb
+}: FinalizeTxProps & {
+  wallet: ObservableWalletTxBuilderDependencies;
+  afterSubmitCb: () => void;
+}): Promise<SignedTx> => {
   const finalizedTx = await wallet.finalizeTx({ auxiliaryData, extraSigners, signingOptions, tx });
   return {
-    submit: () => wallet.submitTx(finalizedTx),
+    submit: () => wallet.submitTx(finalizedTx).then(() => afterSubmitCb()),
     tx: finalizedTx
   };
 };
@@ -91,11 +97,16 @@ export class ObservableWalletTxBuilder implements TxBuilder {
   #outputValidator: OutputValidator;
   #delegateConfig: DelegateConfig;
   #logger: Logger;
+  #isSubmitted = false; // Do not allow building if a transaction built by this builder was already submitted
 
   constructor({ observableWallet, outputValidator = createWalletUtil(observableWallet), logger }: BuildTxProps) {
     this.#observableWallet = observableWallet;
     this.#outputValidator = outputValidator;
     this.#logger = logger;
+  }
+
+  isSubmitted(): boolean {
+    return this.#isSubmitted;
   }
 
   addOutput(txOut: Cardano.TxOut): TxBuilder {
@@ -137,6 +148,7 @@ export class ObservableWalletTxBuilder implements TxBuilder {
 
   async build(): Promise<MaybeValidTx> {
     try {
+      if (this.isSubmitted()) throw new TxAlreadySubmittedError();
       await this.#addDelegationCertificates();
       await this.#validateOutputs();
 
@@ -147,21 +159,7 @@ export class ObservableWalletTxBuilder implements TxBuilder {
         outputs: new Set(this.partialTxBody.outputs || []),
         signingOptions: this.signingOptions
       });
-      return {
-        auxiliaryData: this.auxiliaryData && { ...this.auxiliaryData },
-        body: tx.body,
-        extraSigners: this.extraSigners && [...this.extraSigners],
-        isValid: true,
-        sign: () =>
-          createSignedTx({
-            auxiliaryData: this.auxiliaryData && { ...this.auxiliaryData },
-            extraSigners: this.extraSigners && [...this.extraSigners],
-            signingOptions: this.signingOptions && { ...this.signingOptions },
-            tx,
-            wallet: this.#observableWallet
-          }),
-        signingOptions: this.signingOptions && { ...this.signingOptions }
-      };
+      return this.#createValidTx(tx);
     } catch (error) {
       const errors = Array.isArray(error) ? (error as TxBodyValidationError[]) : [error as TxBodyValidationError];
       this.#logger.debug('Build errors', errors);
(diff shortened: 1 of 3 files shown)


packages/wallet/src/TxBuilder/types.ts
```diff
@@ -32,13 +32,23 @@ export class OutputValidationTokenBundleSizeError extends CustomError {
   }
 }
 
+export class TxBuilderDisabled extends CustomError {
```
rhyslbw, line 35:
```suggestion
export class TxAlreadySubmitted extends CustomError {
```

packages/wallet/src/TxBuilder/types.ts
```diff
@@ -173,12 +188,18 @@ export interface TxBuilder {
...
+   *   This is a snapshot of transaction. Further changes done via TxBuilder, will not update this snapshot.
+   *   Once a transaction is sucessfully submitted, the TxBuilder cannot be reused, to avoid.
+   * - In case of it is an `InvalidTx`, it embeds a TxBodyValidationError with more details.
+   * - `InvalidTx` embeds a {@link TxBuilderDisabled} error if a transaction built by this TxBuilder
+   *   was already successfully submitted.
    */
   build(): Promise<MaybeValidTx>;
 
+  /** Force enable TxBuilder to be reused, even if a transaction with this configuration was successfully submitted. */
+  forceEnable(): TxBuilder;
```
rhyslbw, line 201: How about offering a method that removes the no longer relevant parts of the transaction. Maybe `copy()`?

rhyslbw (changes requested):

packages/wallet/src/TxBuilder/types.ts
mkazlauskas, line 201: Let's try to find a better term for this. 1 suggestion below, but not sure about this. Thoughts @rhyslbw ?
```suggestion
  isDestroyed: boolean;
  restore(): TxBuilder;
```

mkazlauskas (changes requested):

packages/wallet/src/TxBuilder/types.ts
rhyslbw: Suggestion here:
https://github.com/input-output-hk/cardano-js-sdk/pull/434#discussion_r969158433

mkazlauskas: I think the idea here is to be able to reuse the parts. Although not sure if there are any uses cases outside of tests. If we want a clean state then `reset()` could work.

rhyslbw: > Although not sure if there are any uses cases outside of tests

If that's the case it'd be better to remove it from the builder and implement a test util to provide similar convenience

packages/wallet/src/TxBuilder/types.ts
mirceahasegan: Sounds much better, thanks! I will also add the `Error` suffix: `TxAlreadySubmittedError`

packages/wallet/src/TxBuilder/types.ts
mirceahasegan: Agree. It makes it unnecessarily complex. I will remove it. For the moment we don't need it in tests either 😄

mirceahasegan: It would be nice to have a getter, but `isDestroyed` is not very accurate. It's still there, you can use most of its methods, only build() is restricted.

mirceahasegan: Done in [d0bfbb2](https://github.com/input-output-hk/cardano-js-sdk/pull/434/commits/d0bfbb2dbae618fec7ec774c5aa6b12578e92f34)

packages/wallet/src/TxBuilder/types.ts
mirceahasegan: Updated in: [d0bfbb2](https://github.com/input-output-hk/cardano-js-sdk/pull/434/commits/d0bfbb2dbae618fec7ec774c5aa6b12578e92f34)

rhyslbw (approved):

packages/wallet/src/TxBuilder/buildTx.ts
```diff
@@ -91,13 +97,18 @@ export class ObservableWalletTxBuilder implements TxBuilder {
...
   #logger: Logger;
+  #enabled = true; // Do not allow building if a transaction built by this builder was already submitted
 
   constructor({ observableWallet, outputValidator = createWalletUtil(observableWallet), logger }: BuildTxProps) {
     this.#observableWallet = observableWallet;
     this.#outputValidator = outputValidator;
     this.#logger = logger;
   }
 
+  isDestroyed(): boolean {
```
mkazlauskas, line 108: Let's keep the terminology consistent, I like this one (from the error name) better 
```suggestion
  isSubmitted(): boolean {
```

mirceahasegan: Updated here: [e2a9138](https://github.com/input-output-hk/cardano-js-sdk/pull/434/commits/e2a9138bb61064f99b1d37cfbabfdffa81a77fac)

rhyslbw (approved):

mkazlauskas (approved):
Nice work! :rocket:

mirceahasegan:
@mkazlauskas , @rhyslbw , I only added a couple more negative tests for error propagation, in [88982d9](https://github.com/input-output-hk/cardano-js-sdk/pull/434/commits/88982d9820d61fe73b10dff32732f05042215cb7).
I will squash the commits after your approval.
Thank you

mirceahasegan:
Rebased on master and squashed the fixups in [4c26833](https://github.com/input-output-hk/cardano-js-sdk/pull/434/commits/4c26833b988b9bc87c577a99a671ae21a33e7088).
Also added 2 more unit tests for error handling:
- [it('can return a sign error'](https://github.com/input-output-hk/cardano-js-sdk/blob/feat/tx-builder-no-reuse-after-submit/packages/wallet/test/integration/buildTx.test.ts#L539)
- [it('can return a submit error')](https://github.com/input-output-hk/cardano-js-sdk/blob/feat/tx-builder-no-reuse-after-submit/packages/wallet/test/integration/buildTx.test.ts#L550)

packages/wallet/test/integration/buildTx.test.ts
```diff
@@ -459,6 +460,106 @@ describe('buildTx', () => {
...
+
+      expect(error2 instanceof OutputValidationTokenBundleSizeError).toBeTruthy();
+      if (error2 instanceof OutputValidationTokenBundleSizeError) {
+        expect(error2.message === TxOutputFailure.TokenBundleSizeExceedsLimit);
+        expect(error2.outputValidation).toEqual(bundleSizeValidation);
+        expect(error2.txOut).toEqual(output2);
+      }
+    });
+
+    it('can return a sign error', async () => {
```
rhyslbw, line 539:
```suggestion
    it('rejects if error is encountered during transaction finalization', async () => {
```

packages/wallet/test/integration/buildTx.test.ts
```diff
@@ -459,6 +460,106 @@ describe('buildTx', () => {
...
+
+      const builtTx = await buildTx({ logger, observableWallet }).addOutput(output).build();
+      assertTxIsValid(builtTx);
+
+      observableWallet.finalizeTx = jest.fn().mockRejectedValue(signError);
+
+      await expect(builtTx.sign()).rejects.toBe(signError);
+    });
+
+    it('can return a submit error', async () => {
```
rhyslbw, line 550:
```suggestion
    it('rejects if error is encountered during submission', async () => {
```

rhyslbw (changes requested):

packages/wallet/test/integration/buildTx.test.ts
mirceahasegan: fixed/squashed/force pushed in one go since the reviews have settled.
[169d79b](https://github.com/input-output-hk/cardano-js-sdk/pull/434/commits/169d79b1749f882aaa667c5cae5fe0fe65c2d639)

packages/wallet/test/integration/buildTx.test.ts
mirceahasegan: fixed/squashed/force pushed in one go since the reviews have settled.
[169d79b](https://github.com/input-output-hk/cardano-js-sdk/pull/434/commits/169d79b1749f882aaa667c5cae5fe0fe65c2d639)

rhyslbw (approved):

mkazlauskas (approved):

PR #435, opened 2022-09-13
test(e2e): add balance check before running e2e tests

AngelCastilloB:
# Context

There is a race condition between starting the local test network and running the e2e tests, sometimes the e2e test can start before the state of the local test network is fully indexed by dbSync, which causes the tests to fail since the wallet balance is still not available. 

# Proposed Solution

Make the e2e test wallet for balance to be available at the beginning of the test.

### DIFF
--- a/packages/e2e/test/wallet/util.ts
+++ b/packages/e2e/test/wallet/util.ts
@@ -1,10 +1,12 @@
-import { Observable, catchError, filter, firstValueFrom, throwError, timeout } from 'rxjs';
+import { Observable, catchError, combineLatest, filter, firstValueFrom, throwError, timeout } from 'rxjs';
 import { ObservableWallet } from '@cardano-sdk/wallet';
 
 const SECOND = 1000;
 const MINUTE = 60 * SECOND;
 export const TX_TIMEOUT = 7 * MINUTE;
 const SYNC_TIMEOUT = 3 * MINUTE;
+const BALANCE_TIMEOUT = 3 * MINUTE;
+
 export const FAST_OPERATION_TIMEOUT = 15 * SECOND;
 
 export const firstValueFromTimed = <T>(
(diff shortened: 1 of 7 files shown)


rhyslbw (changes requested):
It would be good to define a test util, considering it looks like every tests need:

```ts
await firstValueFrom(wallet.syncStatus.isSettled$.pipe(filter((isSettled) => isSettled)));
await firstValueFrom(wallet.balance.utxo.total$.pipe(filter(({ coins }) => coins > 0)));
```
Maybe `walletReady()`

AngelCastilloB:
> It would be good to define a test util, considering it looks like each test needs:
> 
> ```ts
> await firstValueFrom(wallet.syncStatus.isSettled$.pipe(filter((isSettled) => isSettled)));
> await firstValueFrom(wallet.balance.utxo.total$.pipe(filter(({ coins }) => coins > 0)));
> ```
> 
> Maybe `walletReady()`

addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/435/commits/0875c29000e43ce9ca04a77f259b1ce0d46f72af

rhyslbw (approved):

rhyslbw (approved):
Nice work @AngelCastilloB

packages/e2e/test/wallet/util.ts
```diff
@@ -22,6 +24,16 @@ export const firstValueFromTimed = <T>(
...
+
+export const waitForWalletBalance = (wallet: ObservableWallet) =>
+  firstValueFromTimed(
+    wallet.balance.utxo.total$.pipe(filter(({ coins }) => coins > 0)),
+    'Took too long to load balance',
+    BALANCE_TIMEOUT
+  );
+
+export const walletReady = (wallet: ObservableWallet) =>
+  Promise.all([waitForWalletStateSettle(wallet), waitForWalletBalance(wallet)]);
```
mkazlauskas, line 39: Not sure if this is an issue that would surface, but the following scenario is possible:
1. Wallet settles with 0 balance (promise resolves)
2. Wallet fetches balance >0 (promise resolves), but wallet is still fetching other data that's important for the test

```suggestion
  firstValueFrom(combineLatest([state$, balance$]).pipe(filter(([state, balance]) => state === settled && balance.coins > 0n)));
```

mkazlauskas (changes requested):

packages/e2e/test/wallet/util.ts
AngelCastilloB: addressed here: https://github.com/input-output-hk/cardano-js-sdk/pull/435/commits/25d6f7a4a630e4b456ca94930b5a44468d6368c8

mkazlauskas (dismissed):
Looks good!

rhyslbw (dismissed):
e2e test failure could be a race condition, but it's not in this PR scope to fix. Please squash

rhyslbw (approved):

PR #440, opened 2022-09-14
fix: stake pool search with applied pledge met filter

IvayloAndonov:
# Context

Run this twice:

`time curl -d '{"filters":{"pledgeMet":true,"status":["active","activating"]},"pagination":{"startAt":0,"limit":10}}' -H "Content-Type: application/json" -X POST https://testnet-dev-backend.dev.lw.iog.io/stake-pool/search`
2nd run (cache should be primed) takes ~12s
The run this twice:

`time curl -d '{"filters":{"status":["active","activating"]},"pagination":{"startAt":0,"limit":10}}' -H "Content-Type: application/json" -X POST https://testnet-dev-backend.dev.lw.iog.io/stake-pool/search`
2nd run takes ~0.3s

# Proposed Solution
After troubleshooting and reproducing the issue locally with the restored latest `db-sync` snapshot, it's found out that the bottleneck is `#builder.queryTotalCount(query, params)` which initially wasn't cached due to its relative simplicity but it turned out its performance is significantly affected by `pledgeMet` filter option which takes part in the [query builders](https://github.com/input-output-hk/cardano-js-sdk/blob/5910f552755e2a3fda75c48437f8534a913ad6c2/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts#L140-L143) at the beginning.

The proposed solution is to cache the query with the default TTL value.
With the applied fix, the stake pool search total response time with the applied arguments is reduced to a minimum as expected when we have a cache hit.

QA: should be tested in our `https://testnet-dev-backend.dev.lw.iog.io` env once it's merged and deployed.

- fix the `hash_id` and `update_id` model value types, their type is `bigint` in the database, so when we query them the values are coming as strings but the return types expose them as numbers. Updated the mappers and tests as we need to cast them as `number` to match the return types.
- cache `queryTotalCount` response to reduce the total response time (explained above in # Proposed Solution)
- fix `orderedResultUpdateIds` [sorting](https://github.com/input-output-hk/cardano-js-sdk/compare/fix/stake-pool-search-with-pledge-met?expand=1#diff-3a938d2c168d4b251612c6efabf201ac70e7585c3c1c2149b3c34e71981e6666R80-R82), while debugging I noticed that the order wasn't correct like `orderedResultHashIds` collection (both exist before caching strategy).The final ordered stake pools response is not affected by this issue at all but would be good to have the right order everywhere, even for the internal computations only.

### DIFF
--- a/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/mappers.ts
+++ b/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/mappers.ts
@@ -167,8 +167,8 @@ export const toStakePoolResults = (
 };
 
 export const mapPoolUpdate = (poolUpdateModel: PoolUpdateModel): PoolUpdate => ({
-  id: poolUpdateModel.id,
-  updateId: poolUpdateModel.update_id
+  id: Number(poolUpdateModel.id),
+  updateId: Number(poolUpdateModel.update_id)
 });
 
 const metadataKeys = new Set(['ticker', 'name', 'description', 'homepage']);
@@ -182,13 +182,13 @@ export const mapPoolData = (poolDataModel: PoolDataModel): PoolData => {
   const { n: numerator, d: denominator } = new Fraction(poolDataModel.margin);
   const toReturn: PoolData = {
     cost: BigInt(poolDataModel.fixed_cost),
-    hashId: poolDataModel.hash_id,
+    hashId: Number(poolDataModel.hash_id),
     hexId: Cardano.PoolIdHex(bufferToHexString(poolDataModel.pool_hash)),
     id: Cardano.PoolId(poolDataModel.pool_id),
     margin: { denominator, numerator },
     pledge: BigInt(poolDataModel.pledge),
     rewardAccount: Cardano.RewardAccount(poolDataModel.reward_address),
-    updateId: poolDataModel.update_id,
+    updateId: Number(poolDataModel.update_id),
     vrfKeyHash: Cardano.VrfVkHex(vrfAsHexString)
   };
   if (poolDataModel.metadata_hash) {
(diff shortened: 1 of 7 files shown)


rhyslbw (approved):
LGTM

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/mappers.ts
```diff
@@ -167,8 +167,8 @@ export const toStakePoolResults = (
 };
 
 export const mapPoolUpdate = (poolUpdateModel: PoolUpdateModel): PoolUpdate => ({
-  id: poolUpdateModel.id,
-  updateId: poolUpdateModel.update_id
+  id: Number(poolUpdateModel.id),
```
mkazlauskas, line 170: If this is `bigint` in the db, then `9223372036854775807 > Number.MAX_SAFE_INTEGER`. Although not likely to be an issue if those are incremental ID fields.

```suggestion
  id: BigInt(poolUpdateModel.id),
```

AngelCastilloB (approved):
LGTM! :rocket:

mkazlauskas (approved):
Nice work! Just 1 concern that may or may not be an issue

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/mappers.ts
IvayloAndonov: Yep, those are incremental ones.

IvayloAndonov: Currently on average ~5000 stake pools, with fewer more updates each, it shouldn't be an issue the incremental id to reach `Number.MAX_SAFE_INTEGER`.

PR #441, opened 2022-09-15
test(e2e): rewards test

iadmytro:
E2E test for rewards


packages/e2e/test/local-network/local-network.test.ts
```diff
@@ -29,6 +31,55 @@ export const env = envalid.cleanEnv(process.env, {
 
 const logger = getLogger(env.LOGGER_MIN_SEVERITY);
 
+const getWalletStateSnapshot = async (wallet: ObservableWallet) => {
```
mkazlauskas, line 34: We could export these utils from `util.ts` to be used by both this and delegation test

packages/e2e/test/local-network/local-network.test.ts
```diff
@@ -110,4 +161,67 @@ describe('Local Network', () => {
...
+    let currentEpoch = await firstValueFrom(wallet1.wallet.currentEpoch$);
+    logger.debug(`current epoch - ${currentEpoch.epochNo}`);
+    currentEpoch = await waitForEpoch(1);
+    logger.debug(`waited for epoch - ${currentEpoch.epochNo} to start txs `);
+    // workload for generating fees/rewards
+    await generateTxs(wallet1, wallet2);
+
+    // Assert
+
+    currentEpoch = await waitForEpoch(2);
```
mkazlauskas, line 217: Maybe the 2nd waitForEpoch should be relative to when the delegation was made (not since 'now' which can be further ahead)

mkazlauskas (changes requested):

AngelCastilloB (changes requested):
I think this test should be located at `test/wallet/SingleAddressWallet/` together with the rest of the SingleAddressWallet tests, maybe in `delegation.test.ts`? or in a new `staking-rewards.test.ts`.

packages/e2e/test/local-network/local-network.test.ts
```diff
@@ -110,4 +161,67 @@ describe('Local Network', () => {
...
+      const currentEpoch = await firstValueFrom(wallet1.wallet.currentEpoch$);
+      return await firstValueFrom(
+        wallet1.wallet.currentEpoch$.pipe(
+          filter(({ epochNo }) => epochNo > currentEpoch.epochNo + numberOfEpochsToWait)
+        )
+      );
+    };
+
+    logger.debug(`Address ${sendingAddress.toString()} will be funded with ${amountFromFaucet} tLovelace.`);
+    await faucetProvider.request(sendingAddress.toString(), amountFromFaucet, 1);
```
AngelCastilloB, line 187: Currently the faucet is not network agnostic since we don't have an analogous (yet) on the other networks. I think it should be fine since I don't think we can run this test on regular networks (since each epoch last 5 days), but then if we want to run this on CI we will need to also add:

```
FAUCET_PROVIDER=cardano-wallet
FAUCET_PROVIDER_PARAMS='{"baseUrl":"http://localhost:8090/v2","mnemonic":"fire method repair aware foot tray accuse brother popular olive find account sick rocket next"}'
```
In the [continuous-integration.yaml](https://github.com/input-output-hk/cardano-js-sdk/blob/ae8a3853ce794394463fd501fe95e4f9d1fadb32/.github/workflows/continuous-integration.yaml#L57) file

packages/e2e/test/local-network/local-network.test.ts
```diff
@@ -110,4 +161,67 @@ describe('Local Network', () => {
...
+    expect(poolId).toBeDefined();
+    logger.debug(`Wallet funds will be staked to pool ${poolId}.`);
+    const certificates = createDelegationCertificates(initialState, poolId);
+
+    // Act
+
+    // Create stake tx
+    const unsignedTx = await wallet1.wallet.initializeTx({ certificates });
+    const signedTx = await wallet1.wallet.finalizeTx({ tx: unsignedTx });
+    await wallet1.wallet.submitTx(signedTx);
```
mirceahasegan, line 206: A little bit off-topic, but it would be nice to use the TxBuilder 😄 . (similar to `delegate.test.ts`)
```diff
-    const unsignedTx = await wallet1.wallet.initializeTx({ certificates });
-    const signedTx = await wallet1.wallet.finalizeTx({ tx: unsignedTx });
-    await wallet1.wallet.submitTx(signedTx);
+    const tx = await buildTx({ logger, observableWallet: wallet1.wallet }).delegate(poolId).build()
+    assertTxIsValid(tx);
+    await (await tx.sign()).submit();
```

packages/e2e/test/local-network/local-network.test.ts
```diff
@@ -29,6 +31,55 @@ export const env = envalid.cleanEnv(process.env, {
...
+const generateTxs = async (wallet1, wallet2) => {
+  const tAdaToSend = 5_000_000n;
+  const [{ address: receivingAddress }] = await firstValueFrom(wallet2.wallet.addresses$);
+  for (let i = 0; i < 100; i++) {
+    const unsignedTxA = await wallet1.wallet.initializeTx({
+      outputs: new Set([{ address: receivingAddress, value: { coins: tAdaToSend } }])
+    });
+
+    const signedTxA = await wallet1.wallet.finalizeTx({ tx: unsignedTxA });
+    await wallet1.wallet.submitTx(signedTxA);
```
mirceahasegan, line 79:
```diff
-   const unsignedTxA = await wallet1.wallet.initializeTx({
-     outputs: new Set([{ address: receivingAddress, value: { coins: tAdaToSend } }])
-   });

-   const signedTxA = await wallet1.wallet.finalizeTx({ tx: unsignedTxA });
-   await wallet1.wallet.submitTx(signedTxA);
+    const txBuilder = buildTx({ logger, observableWallet: wallet1.wallet });
+    const tx = await txBuilder
+       .addOutput(txBuilder.buildOutput().address(receivingAddress).coin(tAdaToSend).toTxOut())
+       .build();
+     assertTxIsValid(tx);
+     await (await tx.sign()).submit();
```

packages/e2e/test/local-network/local-network.test.ts
iccicci: Hi @AngelCastilloB , 
as far as I understood, in CI we are sure the test will run against local-network, so using the configuration you provided should not be source of problems. Am I right?

While about the _wait for epoch_ problem; if is there a way to know it the current network doesn't have an acceptable epoch length, we could selectively choose to skip or not the test...
Do you know if is there some information which can be useful to take this decision?
Thank you

packages/e2e/test/local-network/local-network.test.ts
iccicci: Hi @mkazlauskas , I've changed `waitForEpoch` in `waitForNextEpoch` (not yet pushed); that's enough to perform the test and it makes the test much faster.
Was this the point of your suggestion or was you pointing to another problem I didn't saw?
Thank you

packages/e2e/test/local-network/local-network.test.ts
iccicci: This should be no longer required since using `TxBuilder` makes the test no longer require `getWalletStateSnapshot`.
Initially I did it, but being no longer required I rolled it back to make the change having an impact as small as possible.

iccicci:
Hi @rhyslbw , @mkazlauskas , @AngelCastilloB , @mirceahasegan ;

- the suggestions from all the conversations should be solved;
- used `TxBuilder` in `generateTxs` as well;
- I tried to make the flow as close as possible to the diagram Rhys provided us on slack.

Now I'm keeping my eyes on [this](https://github.com/input-output-hk/cardano-js-sdk/actions/runs/3099282961/jobs/5018234708) and [this](https://github.com/input-output-hk/cardano-js-sdk/actions/runs/3099377213/jobs/5018442502) CI builds to check if the test work on CI as well.

packages/e2e/test/local-network/local-network.test.ts
rhyslbw: [epochLength](https://github.com/input-output-hk/cardano-js-sdk/blob/7283619e62667fe5bb672b2903b29b06776ef3eb/packages/core/src/Cardano/types/Genesis.ts#L13) * [slotLength](https://github.com/input-output-hk/cardano-js-sdk/blob/7283619e62667fe5bb672b2903b29b06776ef3eb/packages/core/src/Cardano/types/Genesis.ts#L16) * 1000 gives you the epoch duration in seconds. [NetworkInfo.genesisParameters()](https://github.com/input-output-hk/cardano-js-sdk/blob/0cd887737cc5d4f8d920405c803f05f2c47e42f2/packages/core/src/Provider/NetworkInfoProvider/types.ts#L30)

packages/e2e/test/wallet/SingleAddressWallet/staking-rewards.test.ts
```diff
@@ -1,8 +1,8 @@
 import * as envalid from 'envalid';
```
mirceahasegan, line 1: Please also update the `describe` block. It currently says 'Local Network'

mkazlauskas: I think "delegation rewards" is a better term

packages/e2e/test/wallet/SingleAddressWallet/staking-rewards.test.ts
```diff
@@ -0,0 +1,200 @@
...
+    await signedTx.submit();
+
+    // Wait until wallet two is aware of the funds.
+    await firstValueFrom(saWallet2.balance.utxo.total$.pipe(filter(({ coins }) => coins >= tAdaToSend)));
+
+    // Search chain history to see if the transaction is there.
+    const txFoundInHistory = await firstValueFrom(
+      saWallet2.transactions.history$
+        .pipe(filter((txs) => txs.filter((tx) => tx.id === signedTx.tx.id).length === 1))
+        .pipe(map((txs) => (txs.length > 0 ? txs.find((tx) => tx.id === signedTx.tx.id) : undefined)))
```
mkazlauskas, line 106: Minor improvement: no need for 2 `.pipe`, also no need to iterate it twice. This will also result in a variable that can't be undefined, so will eliminate the need for `txFoundInHistory?`
```suggestion
        .pipe(
          map((txs) => txs.find((tx) => tx.id === signedTx.tx.id)),
          filter(isNotNil),
          take(1)
        )
```

mirceahasegan (approved):
Great work @iccicci and @iadmytro. 👏 
The test, even though complicated, is so easy to read and a good example of how to use the SDK.

mkazlauskas (changes requested):

packages/e2e/test/wallet/SingleAddressWallet/delegation-rewards.test.ts
```diff
@@ -0,0 +1,227 @@
...
+    if (!healthCheck.ok) throw new Error('Faucet provider could not be started.');
+  });
+
+  afterAll(async () => {
+    await faucetProvider.close();
+  });
+
+  beforeEach(async () => {
+    wallet1 = await getWallet({ env, logger, name: 'Sending Wallet', polling: { interval: 50 } });
+    wallet2 = await getWallet({ env, logger, name: 'Receiving Wallet', polling: { interval: 50 } });
```
mkazlauskas, line 54: Other props don't seem to be used
```suggestion
    ({ wallet: wallet1 }) = await getWallet({ env, logger, name: 'Sending Wallet', polling: { interval: 50 } });
    ({ wallet: wallet2 }) = await getWallet({ env, logger, name: 'Receiving Wallet', polling: { interval: 50 } });
```

packages/e2e/test/wallet/SingleAddressWallet/delegation-rewards.test.ts
```diff
@@ -0,0 +1,227 @@
...
+  it('will do tADA transfer between two wallets.', async () => {
+    // Arrange
+    const amountFromFaucet = 100_000_000;
+    const tAdaToSend = 50_000_000n;
+
+    const saWallet1 = wallet1.wallet;
+    const saWallet2 = wallet2.wallet;
+
+    await firstValueFrom(saWallet1.syncStatus.isSettled$.pipe(filter((isSettled) => isSettled)));
+    await firstValueFrom(saWallet2.syncStatus.isSettled$.pipe(filter((isSettled) => isSettled)));
```
mkazlauskas, line 71:
```suggestion
    await waitForWalletStateSettle(wallet1);
    await waitForWalletStateSettle(wallet2);
```
Also would be nice to make more utils like that (not required for this PR)

packages/e2e/test/wallet/SingleAddressWallet/delegation-rewards.test.ts
```diff
@@ -0,0 +1,227 @@
...
+      logger.fatal("Skipping test 'will receive rewards for delegated tADA' as it will take more than 5 minutes");
+
+      return;
+    }
+
+    // Arrange
+    const amountFromFaucet = 100_000_000_000;
+
+    await firstValueFrom(wallet1.wallet.syncStatus.isSettled$.pipe(filter((isSettled) => isSettled)));
+    await firstValueFrom(wallet2.wallet.syncStatus.isSettled$.pipe(filter((isSettled) => isSettled)));
```
mkazlauskas, line 132:
```suggestion
    await waitForWalletStateSettle(wallet2);
    await waitForWalletStateSettle(wallet2);
```

packages/e2e/test/wallet/SingleAddressWallet/delegation.test.ts
```diff
@@ -81,7 +81,7 @@ describe('SingleAddressWallet/delegation', () => {
     expect(currentEpoch.epochNo).toBeGreaterThan(0);
   });
 
-  test('balance & transaction', async () => {
+  test.skip('balance & transaction', async () => {
```
mkazlauskas, line 84:
```suggestion
  test('balance & transaction', async () => {
```

mkazlauskas (changes requested):
Looking good, just a few minor improvements suggested

iccicci:
Hi @rhyslbw ,

I'm sorry but I'm not able to go on this test.
I added the last step of the test as described in tickets AC (withdraw the reward) but I get an error trying to submit the withdraw tx; at least the error is deterministic and is shared in a [CI workflow](https://github.com/input-output-hk/cardano-js-sdk/actions/runs/3131796975/jobs/5083491009). I also tried to set the log level to `debug` as suggested by @mkazlauskas , but it seems this was generating too many logs... I had to stop CI after 3h.

This is not the only problem: once added this test to the `test:wallet` e2e suite (as suggested in this PR) many other tests in the same suite fail...

I tried to investigate the `MissingVkWitnessesError`, but I didn't understood how to handle it. I need support on this.

CC @AngelCastilloB , @mirceahasegan

packages/e2e/test/wallet/SingleAddressWallet/delegation-rewards.test.ts
iccicci: `wallet1.providers` is used a couple of times... but good suggestion! Thank you

packages/e2e/test/wallet/SingleAddressWallet/delegation-rewards.test.ts
iccicci: Yep! Didn't realized that the meaning of the line is to wait for the wallet to be settled. Thank you

packages/e2e/test/wallet/SingleAddressWallet/delegation.test.ts
iccicci: This is a temporary change: before I rebase the branch on `master` it was the only failing test.

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -525,7 +525,10 @@ export class SingleAddressWallet implements ObservableWallet {
             },
             protocolParameters
           });
-          const implicitCoin = Cardano.util.computeImplicitCoin(protocolParameters, props);
+          const implicitCoin = Cardano.util.computeImplicitCoin(protocolParameters, {
+            certificates: props.certificates,
+            withdrawals
+          });
```
iccicci, line 531: What about this change?

packages/e2e/test/wallet/util.ts
```diff
@@ -59,25 +67,100 @@ export const walletReady = (wallet: ObservableWallet) =>
 
 export const normalizeTxBody = (body: Cardano.TxBodyAlonzo | Cardano.NewTxBodyAlonzo) => {
   body.collaterals ||= [];
+  // TODO: inputs should be a Set since they're unordered.
+  // Then Jest should correctly compare it with toEqual.
+  body.inputs = sortBy(body.inputs, 'txId');
   return body;
 };
 
 export const txConfirmed = (
```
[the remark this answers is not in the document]
mkazlauskas: Done in c53c1ae6af1dcbfb347214c82adf1dcf66536865

packages/e2e/test/wallet/util.ts
```diff
@@ -59,25 +67,100 @@ export const walletReady = (wallet: ObservableWallet) =>
...
+
+  const faucetProvider = await faucetProviderFactory.create(env.FAUCET_PROVIDER, env.FAUCET_PROVIDER_PARAMS, logger);
+  await faucetProvider.start();
+  const healthCheck = await faucetProvider.healthCheck();
+  if (!healthCheck.ok) throw new Error('Faucet provider could not be started.');
+  // Request coins from faucet. This will block until the transaction is in the ledger,
+  // and has the given amount of confirmation, which means the funds can be used immediately after
+  // this call.
+  // TODO: change FaucetProvider signature to accept Cardano.Lovelace
+  const requestResult = await faucetProvider.request(address.toString(), Number.parseInt(coins.toString()), 1);
```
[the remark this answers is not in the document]
mkazlauskas: Done in fe645bcdddfe1cccfe7b78cbdc433d03c9f8f5a0

packages/e2e/test/wallet/util.ts
```diff
@@ -66,6 +67,9 @@ export const walletReady = (wallet: ObservableWallet) =>
 
 export const normalizeTxBody = (body: Cardano.TxBodyAlonzo | Cardano.NewTxBodyAlonzo) => {
   body.collaterals ||= [];
+  // TODO: inputs should be a Set since they're unordered.
```
rhyslbw, line 72: Has this been captured?

packages/e2e/test/wallet/SingleAddressWallet/delegation-rewards.test.ts
```diff
@@ -0,0 +1,156 @@
...
+import { Cardano, createSlotEpochCalc } from '@cardano-sdk/core';
+import { SignedTx, SingleAddressWallet, buildTx } from '@cardano-sdk/wallet';
+import { TestWallet, getWallet } from '../../../src';
+import { assertTxIsValid, waitForWalletStateSettle } from '../../../../wallet/test/util';
+import { filter, firstValueFrom } from 'rxjs';
+import { logger } from '@cardano-sdk/util-dev';
+import { requestCoins, runningAgainstLocalNetwork, submitAndConfirm, transferCoins, waitForEpoch } from '../util';
+
+// Verify environment.
+export const env = envalid.cleanEnv(process.env, {
```
rhyslbw, line 11: Let's turn this into a function to DRY the code, as there's quite a bit of duplication already.

something like this in the `util` package?:
```ts
const cleanEnv = (requiredProviders: Providers[], env: NodeJS.Env) => {
  const specs = {}
  if (requiredProviders.contains(Providers.Asset)) {
    specs.ASSET_PROVIDER = envalid.str(),
    specs.ASSET_PROVIDER_PARAMS = envalid.json({ default: {} }),
  }
  ...
  return envalid.cleanEnv(env, specs)
}
```

rhyslbw (changes requested):
Looks good

packages/e2e/test/wallet/SingleAddressWallet/delegation-rewards.test.ts
mkazlauskas: Captured in ADP-2334

rhyslbw (approved):

PR #443, opened 2022-09-15
fix: rollback ProtocolParametersRequiredByWallet type

mkazlauskas:
# Context

#416 introduced an unnecessary breaking change: add more properties to `ProtocolParametersRequiredByWallet`

# Proposed Solution

Rollback the change to `ProtocolParametersRequiredByWallet`

### DIFF
--- a/packages/blockfrost/src/BlockfrostToCore.ts
+++ b/packages/blockfrost/src/BlockfrostToCore.ts
@@ -27,36 +27,15 @@ export const BlockfrostToCore = {
     blockfrost: Responses['epoch_param_content']
   ): ProtocolParametersRequiredByWallet => ({
     coinsPerUtxoByte: Number(blockfrost.coins_per_utxo_word),
-    collateralPercentage: Number(blockfrost.collateral_percent),
-    decentralizationParameter: String(blockfrost.decentralisation_param),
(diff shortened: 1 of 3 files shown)


packages/blockfrost/src/BlockfrostToCore.ts
```diff
@@ -25,7 +25,7 @@ export const BlockfrostToCore = {
 
   currentWalletProtocolParameters: (
     blockfrost: Responses['epoch_param_content']
-  ): ProtocolParametersRequiredByWallet => ({
+  ): Omit<Cardano.ProtocolParameters, 'costModels'> => ({
```
rhyslbw, line 28:
```suggestion
  ): ProtocolParametersRequiredByWallet => ({
```
I think it's better to keep the return type and adjust the response, otherwise we should rename it to `NetworkInfo.protocolParameters` and return `Cardano.ProtocolParameters` (including `costModels`).

rhyslbw (changes requested):

IvayloAndonov (approved):

rhyslbw (approved):

rhyslbw (approved):

IvayloAndonov (approved):

PR #448, opened 2022-09-19
fix(core): coreToCsl converts datum in txOut function

jmagan:
# Context

I commit another change because I found that coreToCsl.txOut doesn't include the datum hash when the attibute _TxOut.datum?_ is defined. 

# Proposed Solution

I added an if statement in the function to convert the 32 bytes hex strting to a DataHash:

``` javascript
export const txOut = (core: Cardano.TxOut): TransactionOutput => {
  const cslTxOut = TransactionOutput.new(parseCslAddress(core.address.toString())!, value(core.value));

  if (core.datum !== undefined) {
    cslTxOut.set_data_hash(DataHash.from_bytes(Buffer.from(core.datum.toString(), 'hex')));
  }

  return cslTxOut;
};
```
I also added as test data _txOutWithDatum_ and the following unit test, which failed with the old implementation:

``` javascript
    it('converts datum to CSL.DataHash', () => {
      const cslTxOut = coreToCsl.txOut(txOutWithDatum);
      expect(cslTxOut.data_hash()).toBeInstanceOf(CSL.DataHash);
      expect(Buffer.from(cslTxOut.data_hash()!.to_bytes()).toString('hex')).toBe(txOutWithDatum.datum?.toString());
    });
```

None.

### DIFF
--- a/packages/core/src/CSL/coreToCsl/coreToCsl.ts
+++ b/packages/core/src/CSL/coreToCsl/coreToCsl.ts
@@ -9,6 +9,7 @@ import {
   BootstrapWitness,
   BootstrapWitnesses,
   Certificates,
+  DataHash,
   Ed25519KeyHash,
   Ed25519KeyHashes,
   Ed25519Signature,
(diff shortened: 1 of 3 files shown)


rhyslbw (approved):
Thank you again @jmagan :bow:

AngelCastilloB (approved):
LGTM!

jmagan:
You are welcome and thank you guys for this awesome library!

PR #449, opened 2022-09-20
feat!: ChainHistoryProvider pagination and constraints

ivaylo-andonov:
# Context

These queries are currently unbound and pose a DOS threat. We’re also missing an upper bound on the block range filter, which is a functional issue.

# Proposed Solution
As we discussed with @mkazlauskas, it makes sense to apply pagination only on [transactionsByAddresses](https://github.com/input-output-hk/cardano-js-sdk/blob/21334bd8185da235459424106a137763bf394bf4/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/DbSyncChainHistoryProvider.ts#L52) where the ratio between `addresses` (filter items) and `transactions` (response) is one-to-many. We also apply `addresses` count limitation up to the `PAGINATION_PAGE_SIZE_LIMIT`
For the other two provider methods, we just apply filter items (`ids`) limitation up to the `PAGINATION_PAGE_SIZE_LIMIT` value since the relation is one-to-one and the response items can not be greater than one-page size.

The pagination for `transactionsByAddresses` itself is done on the application level once we extract the transaction ids rather than at DB level across 2 subqueries due to:
- we have 2 DB queries (extract inputs/outputs for given addresses) and then compute the tx ids
- both [findTxInputsByAddresses and findTxOutputsByAddresses](https://github.com/input-output-hk/cardano-js-sdk/blob/21334bd8185da235459424106a137763bf394bf4/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/DbSyncChainHistoryProvider.ts#L82-L91) queries and [composing](https://github.com/input-output-hk/cardano-js-sdk/blob/21334bd8185da235459424106a137763bf394bf4/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/DbSyncChainHistoryProvider.ts#L96-L99) the tx ids are not expensive operations (in addition could be cached if needed)
- to apply pagination in inputs/outputs DB queries would be hard and increase the complexity which requires additional logic for the `offset` and `limit` since the tx ids are composed of 2 different subqueries

- add `PAGINATION_PAGE_SIZE_LIMIT` env var
- add `--pagination-page-size-limit` cli arg
- add a base [type](https://github.com/input-output-hk/cardano-js-sdk/pull/449/files#diff-d31d3a8147d7d60a96723b03cbd8de5e565505bf006bf230321af1d48db545c0R18-R27) to model the query as paginated
- txs by addresses response is paginated (in-app level)
- enforce `pagination` as the required [argument](https://github.com/input-output-hk/cardano-js-sdk/pull/449/files#diff-80bbe620d04cbc7ccd603f66791eb14640e13749bf2ca141296ad3ce1eaa20d2R4-R7) (Open API validation)
- add pagination page size limit constraints in provider methods
- interpret the error as a 400 code response
- make [Range](https://github.com/input-output-hk/cardano-js-sdk/pull/449/files#diff-892eb132db458bf98311f1543559dde12da2f5d11b4ead2e8186e0317f636d1bR1-R13) generic reusable interface
- refactor [TransactionsByAddressesArgs](https://github.com/input-output-hk/cardano-js-sdk/pull/449/files#diff-80bbe620d04cbc7ccd603f66791eb14640e13749bf2ca141296ad3ce1eaa20d2R7) from `sinceBlock` to `blockRange`
- in `wallet` fetch all history transactions page by page

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -48,7 +48,7 @@ jobs:
     - name: 🔬 Test
       run: |
         yarn test:build:verify
-        yarn test
+        yarn test --forceExit
 
     - name: 🔬 Test - e2e
       run: |


packages/core/src/errors.ts
```diff
@@ -10,6 +10,16 @@ export enum ProviderFailure {
   BadRequest = 'BAD_REQUEST'
 }
 
+export const failureToCodeMap: { [key in ProviderFailure]: number } = {
```
mkazlauskas, line 13:
```suggestion
export const providerFailureToStatusCodeMap: { [key in ProviderFailure]: number } = {
```

packages/cardano-services/src/Http/HttpService.ts
```diff
@@ -58,7 +58,7 @@ export abstract class HttpService extends RunnableModule {
         logger.error(error);
 
         if (error instanceof ProviderError) {
-          const code = error.reason === ProviderFailure.NotFound ? 404 : 500;
+          const code = failureToCodeMap[error.reason];
```
mkazlauskas, line 61: Nice :+1:

mkazlauskas (changes requested):
Great work! :tada: 

717bbb09a73a993050605fda7332f0b62d5b7e96 is unsigned, please recommit.

The other 2 commits should probably be squashed as adding options without them being used does not make much sense.

Are we validating stake pool search pagination constraints (or is that a separate task)?

packages/core/src/errors.ts
IvayloAndonov: f3c33f1215a0f071c3860913dce7d60013e431c8

mkazlauskas (approved):
Nice work! :rocket:

IvayloAndonov:
> 

The unsigned commit is recommitted and now is signed. The last two commits related to pagination are squashed as suggested. Force push was required due to both.

Yep, stake pool search pagination constraints are part of a separate task.
Ready for reviews. @mkazlauskas

packages/wallet/src/services/TransactionsTracker.ts
```diff
@@ -101,30 +101,31 @@ export const createAddressTransactionsProvider = ({
                     `Last stored tx: ${lastStoredTransaction?.id} block:${lastStoredTransaction?.blockHeader.blockNo}`
                   );
 
-                const sinceBlock = lastStoredTransaction?.blockHeader.blockNo;
-                const newTransactions = await chainHistoryProvider.transactionsByAddresses({
+                const lowerBound = lastStoredTransaction?.blockHeader.blockNo;
+                const response = await chainHistoryProvider.transactionsByAddresses({
```
rhyslbw, line 105: This looks like we're only fetching the first page of results, whereas, the wallet should still be fetching all results, just one page at a time.

IvayloAndonov: How the wallet suppose to change the pagination args of the chain history provider and iterate over pages?

packages/blockfrost/src/blockfrostChainHistoryProvider.ts
```diff
@@ -238,7 +239,7 @@ export const blockfrostChainHistoryProvider = (blockfrost: BlockFrostAPI, logger
 
   const transactionsByAddresses: ChainHistoryProvider['transactionsByAddresses'] = async ({
     addresses,
-    sinceBlock
+    blockRange
```
mkazlauskas, line 242: Need to also add pagination support for blockfrost
```suggestion
    blockRange,
    pagination
```

mkazlauskas (changes requested):

packages/wallet/src/services/TransactionsTracker.ts
rhyslbw: We'll need to cover this new behavior in a the wallet e2e too @IvayloAndonov , demonstrating a wallet with txs > page size fetches the complete history. Should be able to set a low page size using `CHAIN_HISTORY_PROVIDER_PARAMS`

packages/blockfrost/src/blockfrostChainHistoryProvider.ts
IvayloAndonov: See 8c1b7b2acec316796833ef6a8a0bc800204b389c commit

packages/blockfrost/src/blockfrostChainHistoryProvider.ts
```diff
@@ -238,31 +239,38 @@ export const blockfrostChainHistoryProvider = (blockfrost: BlockFrostAPI, logger
 
   const transactionsByAddresses: ChainHistoryProvider['transactionsByAddresses'] = async ({
     addresses,
-    sinceBlock
+    blockRange,
+    pagination
   }) => {
+    // TODO: add pagination support for Blockfrost
+    if (pagination) throw new ProviderError(ProviderFailure.NotImplemented);
```
IvayloAndonov, line 246: Pagination is not implemented as agreed

packages/wallet/src/services/TransactionsTracker.ts
IvayloAndonov: I'm not sure I understood how to use `CHAIN_HISTORY_PROVIDER_PARAMS` for that purpose since it's from type [HttpProviderConfig](https://github.com/input-output-hk/cardano-js-sdk/blob/dee30b52af5edc1241142a2c06708266a1ae7fa4/packages/cardano-services-client/src/HttpProvider.ts#L11-L44) in most cases to change the provider's `baseUrl`.

Currently [pagination](https://github.com/input-output-hk/cardano-js-sdk/pull/449/files#diff-b015172e50668407821a584b9a03f1c2c70ccf60bd27d3832335c85360d86d96R108) is hardcoded while fetch from the provider as `limit` is 25 ( the default `PAGINATION_PAGE_SIZE_LIMIT` in `cardano-services`)

Should we stick with that value while fetching full tx history sequentially in the loop? @rhyslbw

mkazlauskas: Hardcoded page size is fine for now.

It could also be an optional prop when creating SingleAddressWallet, but I would prefer it to be contained in provider definition. How about `ChainHistoryProvider { maxPageSize: number }`?

IvayloAndonov: We already have that [value](https://github.com/input-output-hk/cardano-js-sdk/blob/d267fa0003fbc9d9277f6dc5c226bad2e40a922f/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/DbSyncChainHistoryProvider.ts#L49) provided to the provider from outside, but the case Rhys mentioned is about asserting when txs are > `pagination.limit` , not related with max page size ?

mkazlauskas: > We already have that [value](https://github.com/input-output-hk/cardano-js-sdk/blob/d267fa0003fbc9d9277f6dc5c226bad2e40a922f/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/DbSyncChainHistoryProvider.ts#L49) provided to the provider from outside

We're currently talking about `wallet` package which is client-side, right? The provided value is in cardano-services, which is why I suggested to expose it from the provider.

> So we should somehow to change that limit value in e2e test or to be sure we already have txs greater than 25 in the address?

A unit test in `wallet` package would be sufficient to test this change.

IvayloAndonov: > A unit test in wallet package would be sufficient to test this change.

Agree

> We're currently talking about wallet package which is client-side

Aham, as a new public field in [ChainHistoryProvider](https://github.com/input-output-hk/cardano-js-sdk/blob/d267fa0003fbc9d9277f6dc5c226bad2e40a922f/packages/core/src/Provider/ChainHistoryProvider/types.ts#L12) which would be consumed by the wallet for `pagination.limit` value while fetching by pages from provider @mkazlauskas

rhyslbw: > A unit test in wallet package would be sufficient to test this change.

Yes, you're right @mkazlauskas. That's the better place for this test case

mkazlauskas: > Aham, as a new public field in [ChainHistoryProvider](https://github.com/input-output-hk/cardano-js-sdk/blob/d267fa0003fbc9d9277f6dc5c226bad2e40a922f/packages/core/src/Provider/ChainHistoryProvider/types.ts#L12) which would be consumed by the wallet for pagination.limit value while fetching by pages from provider @mkazlauskas

Yeah. It should probably be an async method like `{ getMaxPageSize(): Promise<number> }` because it can be different for each instance of cardano-services

mkazlauskas: Anyway, as I mentioned above I'd be fine with hardcoding it for this PR and then coming back to it

rhyslbw: Yes, please create a new task for this @IvayloAndonov.

IvayloAndonov: Fetching all transactions, one page at a time is implemented in dd82227870c5992c1b60c0cc491d901674ce2c4a @rhyslbw

IvayloAndonov:
Noticed today after the latest merge in master some conflicts popped up, shall I resolve them first before the final review, also to be able to see CI build triggered? @rhyslbw  @mkazlauskas

packages/wallet/src/services/TransactionsTracker.ts
```diff
@@ -101,30 +104,42 @@ export const createAddressTransactionsProvider = ({
...
+                    await chainHistoryProvider.transactionsByAddresses({
+                      addresses,
+                      blockRange: { lowerBound },
+                      pagination: { limit: PAGE_SIZE, startAt }
+                    })
+                  ).pageResults;
+
+                  startAt += PAGE_SIZE;
+                  response = [...response, ...pageResults];
+                } while (pageResults.length === PAGE_SIZE);
```
mkazlauskas, line 122: This is already a large function, please hoist this to a separate function. Also `newTransactions` is a little bit more descriptive than `response`.
```suggestion
                const newTransactions = await allTransactionsByAddresses(chainHistoryProvider, {addresses, blockRange: { lowerBound }})
```

mkazlauskas (changes requested):

mkazlauskas (approved):
Great work! :rocket:

packages/wallet/test/services/TransactionsTracker.test.ts
```diff
@@ -33,13 +45,45 @@ describe('TransactionsTracker', () => {
...
-      expect(await firstValueFrom(provider$)).toEqual(queryTransactionsResult);
+      expect(await firstValueFrom(provider$)).toEqual(queryTransactionsResult.pageResults);
       expect(store.setAll).toBeCalledTimes(1);
-      expect(store.setAll).toBeCalledWith(queryTransactionsResult);
+      expect(store.setAll).toBeCalledWith(queryTransactionsResult.pageResults);
+    });
+
+    it('emits entire transactions list resolved by ChainHistoryProvider', async () => {
+      const pageSize = PAGE_SIZE;
+      const totalTxsCount = 30;
```
mirceahasegan, line 55: It would be nice if you calculated this one, so it autoupdates if PAGE_SIZE changes:
```diff
- const totalTxsCount = 30;
+ const secondPageSize = 5;
+ const totalTxsCount = pageSize + secondPageSize;

mirceahasegan (approved):

packages/wallet/test/services/TransactionsTracker.test.ts
IvayloAndonov: Updated in 5ab9d3ab3c3777f603b9e6f8630578b796b8a06d @mirceahasegan

mirceahasegan (approved):
LGTM. Great work @IvayloAndonov 🚀

rhyslbw (dismissed):
Please rebase and squash

IvayloAndonov:
> Please rebase and squash

Rebased and resolved conflicts, commits are squashed.

mirceahasegan (dismissed):
🎉  Great work @IvayloAndonov !

mkazlauskas (approved):
:rocket:

rhyslbw (approved):

mirceahasegan (approved):

PR #451, opened 2022-09-21
refactor(input-selection)!: renamed cip2 package to input-selection

AngelCastilloB:
# Context

We must give the CIP2 package a less opaque name for the following reasons:

- cip2` is equally opaque as it is specific!
- The package provides input selection behavior, and naming it this way may help avoid a situation where a new CIP extends the current, but it supersedes it completely.

# Proposed Solution

Rename the cip2 package to input selection.

The CIP2 package was renamed input-selection.


mkazlauskas (approved):
Looks good!

mirceahasegan (approved):
👍

rhyslbw (approved):

PR #454, opened 2022-09-22
fix(cardano-services): fixed an issue that was causing pool saturation to be always 0

AngelCastilloB:
# Context

Currently pool saturation is being returned as 0. This happens due to a small bug in a `mapEpoch` in the service which maps the dbSync result to a domain type. The query returns `optimal_pool_count`  but the mapping function was expecting it as `pool_optimal_count`.

# Proposed Solution

Fix the mapping function to expect the right field name.

- Small bug was fixed in the pool saturation calculation.
- A new unit test to verify the proper function of the mapping function was added.

### DIFF
--- a/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
+++ b/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
@@ -159,7 +159,7 @@ export class DbSyncStakePoolProvider extends DbSyncProvider implements StakePool
 
     // Get last epoch data
     const lastEpoch = await this.#builder.getLastEpochWithData();
-    const { poolOptimalCount, no: lastEpochNo } = lastEpoch;
+    const { optimalPoolCount, no: lastEpochNo } = lastEpoch;
 
     // Get stake pools data cached
     const { orderedResultHashIds, orderedResultUpdateIds, orderedResult, poolDatas, hashesIds, sortType } =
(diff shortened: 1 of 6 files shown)


packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
```diff
@@ -159,7 +159,7 @@ export class DbSyncStakePoolProvider extends DbSyncProvider implements StakePool
 
     // Get last epoch data
     const lastEpoch = await this.#builder.getLastEpochWithData();
-    const { poolOptimalCount, no: lastEpochNo } = lastEpoch;
+    const { optimalPoolCount, no: lastEpochNo } = lastEpoch;
```
rhyslbw, line 162: How did this not throw tsc errors?

rhyslbw (approved):
image

:scream:

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
AngelCastilloB: Thats our domain object, so it use to exists in our Epoch type, I just renamed it to make it consistent with the field in dbSync:

`optimal_pool_count	word64type	The optimal number of stake pools.`

mkazlauskas (approved):
Nice work!

PR #455, opened 2022-09-22
Fix/Ogmios WebSocket close logging

rhyslbw:
# Context
Currently, we're logging the whole WebSocket object as an `info` message on any close event, which is cumbersome and invalid given a particular closure is actually an error. There's also some uncertainty about reconnection behaviour, however this PR is scoped just to address the logging issue.

https://www.rfc-editor.org/rfc/rfc6455#section-7.4.1

# Proposed Solution
- Add a util to `createInteractionContextWithLogger`, that includes the capability to later include a reconnect handler. 
- Refactor the invalid implementations, without introducing restart logic, which will be considered in a later PR

### DIFF
--- a/packages/ogmios/src/CardanoNode/OgmiosCardanoNode.ts
+++ b/packages/ogmios/src/CardanoNode/OgmiosCardanoNode.ts
@@ -13,11 +13,12 @@ import {
   ConnectionConfig,
   StateQuery,
   createConnectionObject,
-  createInteractionContext,
   createStateQueryClient,
   getServerHealth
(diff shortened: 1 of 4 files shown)


mkazlauskas (approved):
Looks good!

mirceahasegan (dismissed):

rhyslbw:
@mkazlauskas @mirceahasegan I just rebased as the original commits were not signed. Please review and approve again

mirceahasegan (approved):

mkazlauskas (approved):

PR #456, opened 2022-09-22
fix(core): custom errors no longer hide inner error details

iccicci:
# Context

Our custom errors hide `innerError` details making harder to investigate problems.

# Proposed Solution

Substituted the _stack trace_ part of custom errors `stack` property, with the _stack trace_ part of the same property of the `innerError`.


packages/core/src/errors.ts
```diff
@@ -12,9 +12,22 @@ export enum ProviderFailure {
...
+
+  Object.defineProperty(error, 'stack', {
+    configurable: true,
+    value: `${firstLineOfErrorStack} due to \n ${innerError.stack}`
+  });
+};
+
 export class ProviderError<InnerError = unknown> extends CustomError {
   constructor(public reason: ProviderFailure, public innerError?: InnerError, public detail?: string) {
     super(formatMessage(reason, detail));
```
mkazlauskas, line 28: Would be great to also concatenate the message

packages/core/src/errors.ts
```diff
@@ -12,9 +12,22 @@ export enum ProviderFailure {
 
 const formatMessage = (reason: string, detail?: string) => reason + (detail ? ` (${detail})` : '');
 
+const tryToUnhideInnerError = (error: Error, innerError: unknown) => {
```
mkazlauskas, line 15: How about creating a new error base class that would do this? Then we wouldn't need to call this for every single error  type.

mkazlauskas (changes requested):

packages/core/src/errors.ts
iccicci: It is @mkazlauskas : the value of the `stack` property is the concatenation of `type` + `message` + `stack trace`

mkazlauskas: Would concatenating type/message/stack trace individually work then? :thinking:

iccicci: Sorry I'm afraid I didn't got the point.

Any solution _works_... I can't understand which is the target you want to reach... which is the problem you see an you are trying to resolve.

mkazlauskas: From your comment it seems like 'stack' is a computed property. If that's the case, then overwriting the source of this property seems like a better way to do this, because then it would behave consistently with other Error objects

mkazlauskas: Anyway, current approach works, let's not overthink it. Marking this as resolved.

packages/core/src/errors.ts
mkazlauskas: Looks good @iccicci . Please search for other `extends CustomError` within sdk and update those as well

iccicci:
Hi @mkazlauskas and @rhyslbw ;
e00f9c4 should solve all the still open points.
If you could check this again...
Thank you

packages/core/src/errors.ts
```diff
@@ -12,9 +12,35 @@ export enum ProviderFailure {
 
 const formatMessage = (reason: string, detail?: string) => reason + (detail ? ` (${detail})` : '');
 
-export class ProviderError<InnerError = unknown> extends CustomError {
-  constructor(public reason: ProviderFailure, public innerError?: InnerError, public detail?: string) {
-    super(formatMessage(reason, detail));
+export class ErrorWithInnerError<InnerError = unknown> extends CustomError {
```
mkazlauskas, line 15: Just an idea, original name is ok too
```suggestion
export class ComposableError<InnerError = unknown> extends CustomError {
```

mkazlauskas (dismissed):
Looks good!

mirceahasegan (approved):

packages/core/src/errors.ts
iccicci: Hi @mkazlauskas , 18a531dea and 18a531dea should fix this

mkazlauskas (dismissed):

mkazlauskas (approved):

AngelCastilloB (approved):
LGTM! 🚀

IvayloAndonov (approved):
Nice!

rhyslbw (approved):
Nice work @iccicci

PR #457, opened 2022-09-22
fix: re-submit recoverable transactions

mkazlauskas:
# Context

Transactions should be resubmitted (TransactionReemiter) when:

- Resubmit on certain errors (probably with some timeout and maybe considering online status as well): ProviderFailure.{~~Unknown~~,Unhealthy,ConnectionFailure}
- Previously submitted transaction didn't get into history$ for too long (e.g. due to timing of rollback - when tx was submitted successfully but never got to history$)

# Proposed Solution

See notes in the commits

### DIFF
--- a/packages/wallet/src/SingleAddressWallet.ts
+++ b/packages/wallet/src/SingleAddressWallet.ts
@@ -14,6 +14,7 @@ import {
   EpochInfo,
   EraSummary,
   ProtocolParametersRequiredByWallet,
+  ProviderError,
   RewardsProvider,
   StakePoolProvider,
   TxSubmitProvider,
@@ -114,6 +115,10 @@ export interface SingleAddressWalletDependencies {
   readonly stores?: WalletStores;
   readonly logger: Logger;
   readonly connectionStatusTracker$?: ConnectionStatusTracker;
+}
+
+export interface SubmitTxOptions {
+  mightBeAlreadySubmitted?: boolean;
 }
 
 export class SingleAddressWallet implements ObservableWallet {
@@ -320,12 +325,13 @@ export class SingleAddressWallet implements ObservableWallet {
     this.#resubmitSubscription = createTransactionReemitter({
       genesisParameters$: this.genesisParameters$,
       logger: contextLogger(this.#logger, 'transactionsReemitter'),
-      store: stores.volatileTransactions,
+      maxInterval,
+      stores,
       tipSlot$: this.tip$.pipe(map((tip) => tip.slot)),
       transactions: this.transactions
     })
       .pipe(
-        mergeMap((transaction) => from(this.submitTx(transaction))),
+        mergeMap((transaction) => from(this.submitTx(transaction, { mightBeAlreadySubmitted: true }))),
         catchError((err) => {
           this.#logger.error('Failed to resubmit transaction', err);
           return EMPTY;
@@ -420,16 +426,26 @@ export class SingleAddressWallet implements ObservableWallet {
     };
   }
 
-  async submitTx(tx: Cardano.NewTxAlonzo): Promise<void> {
+  async submitTx(tx: Cardano.NewTxAlonzo, { mightBeAlreadySubmitted }: SubmitTxOptions = {}): Promise<void> {
     this.#logger.debug(`Submitting transaction ${tx.id}`);
     this.#newTransactions.submitting$.next(tx);
     try {
       await this.txSubmitProvider.submitTx({
         signedTransaction: bufferToHexString(Buffer.from(coreToCsl.tx(tx).to_bytes()))
       });
-      this.#logger.debug(`Submitted transaction ${tx.id}`);
+      const { slot: submittedAt } = await firstValueFrom(this.tip$);
+      this.#logger.debug(`Submitted transaction ${tx.id} at slot ${submittedAt}`);
       this.#newTransactions.pending$.next(tx);
     } catch (error) {
+      if (
+        mightBeAlreadySubmitted &&
+        error instanceof ProviderError &&
+        error.innerError instanceof Cardano.TxSubmissionErrors.ValueNotConservedError
+      ) {
+        this.#logger.debug(`Transaction ${tx.id} appears to be already submitted...`);
+        this.#newTransactions.pending$.next(tx);
+        return;
+      }
       this.#newTransactions.failedToSubmit$.next({
         error: error as Cardano.TxSubmissionError,
         reason: TransactionFailure.FailedToSubmit,
--- a/packages/wallet/src/SingleAddressWallet.ts
+++ b/packages/wallet/src/SingleAddressWallet.ts
@@ -38,6 +38,7 @@ import {
   FailedTx,
   PersistentDocumentTrackerSubject,
   PollingConfig,
+  SmartTxSubmitProvider,
   TipTracker,
   TrackedAssetProvider,
   TrackedChainHistoryProvider,
@@ -125,10 +126,11 @@ export class SingleAddressWallet implements ObservableWallet {
     submitting$: new Subject<Cardano.NewTxAlonzo>()
   };
   #resubmitSubscription: Subscription;
+  #trackedTxSubmitProvider: TrackedTxSubmitProvider;
 
   readonly keyAgent: AsyncKeyAgent;
   readonly currentEpoch$: TrackerSubject<EpochInfo>;
-  readonly txSubmitProvider: TrackedTxSubmitProvider;
+  readonly txSubmitProvider: TxSubmitProvider;
   readonly utxoProvider: TrackedUtxoProvider;
   readonly networkInfoProvider: TrackedWalletNetworkInfoProvider;
   readonly stakePoolProvider: TrackedStakePoolProvider;
@@ -181,8 +183,8 @@ export class SingleAddressWallet implements ObservableWallet {
     this.#logger = contextLogger(logger, name);
 
     this.#inputSelector = inputSelector;
+    this.#trackedTxSubmitProvider = new TrackedTxSubmitProvider(txSubmitProvider);
 
-    this.txSubmitProvider = new TrackedTxSubmitProvider(txSubmitProvider);
     this.utxoProvider = new TrackedUtxoProvider(utxoProvider);
     this.networkInfoProvider = new TrackedWalletNetworkInfoProvider(networkInfoProvider);
     this.stakePoolProvider = new TrackedStakePoolProvider(stakePoolProvider);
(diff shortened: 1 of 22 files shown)


mirceahasegan (changes requested):

packages/wallet/src/services/SmartTxSubmitProvider.ts
```diff
@@ -0,0 +1,92 @@
...
+  submitTx(args: SubmitTxArgs) {
+    const {
+      body: {
+        validityInterval: { invalidBefore, invalidHereafter }
+      }
+    } = cslUtil.deserializeTx(args.signedTransaction);
+    const onlineAndWithinValidityInterval$ = combineLatest([this.#connectionStatus$, this.#tip$]).pipe(
+      tap(([_, { slot }]) => {
+        if (slot >= (invalidHereafter || Number.POSITIVE_INFINITY))
+          throw new ProviderError(
```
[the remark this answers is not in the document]
mkazlauskas: Yes, I think it's ok to throw here. [This test](https://github.com/input-output-hk/cardano-js-sdk/pull/457/files#diff-1905263efb046f31a2f6cef7c33d2ceb3f12af117992780758d5204afd038a97R81) covers it.

From rxjs docs on `throwError`:

> Generally, inside of most operators when you might want to return an errored observable, this is unnecessary. In most cases, such as in the inner return of [concatMap](https://rxjs.dev/api/index/function/concatMap), [mergeMap](https://rxjs.dev/api/index/function/mergeMap), [defer](https://rxjs.dev/api/index/function/defer), and many others, you can simply throw the error, and RxJS will pick that up and notify the consumer of the error.

packages/wallet/src/services/SmartTxSubmitProvider.ts
```diff
@@ -0,0 +1,92 @@
...
+    { retryBackoffConfig }: RetryingTxSubmitProviderProps,
+    { connectionStatus$, tip$, txSubmitProvider }: RetryingTxSubmitProviderDependencies
+  ) {
+    this.#txSubmitProvider = txSubmitProvider;
+    this.#tip$ = tip$;
+    this.#connectionStatus$ = connectionStatus$;
+    this.#retryBackoffConfig$ = retryBackoffConfig;
+  }
+
+  submitTx(args: SubmitTxArgs) {
```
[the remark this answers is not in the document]
mkazlauskas: 318f6dbb26b84bc1674f22061838f67931683d73

packages/wallet/src/services/SmartTxSubmitProvider.ts
```diff
@@ -0,0 +1,92 @@
...
+ *   - Network tip is ahead of tx body `ValidityInterval.invalidBefore`.
+ *   - Connection status is 'up'.
+ * - Re-submits transactions that failed to submit due to connection or recoverable provider issue.
+ */
+// REVIEW: an alternative name is ManagedTxSubmitProvider.
+// Which one would you prefer? New name ideas are also welcome
+export class SmartTxSubmitProvider implements TxSubmitProvider {
+  readonly #txSubmitProvider: TxSubmitProvider;
+  readonly #tip$: Observable<TipSlot>;
+  readonly #retryBackoffConfig$: RetryBackoffConfig;
```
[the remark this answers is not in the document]
mkazlauskas: 318f6dbb26b84bc1674f22061838f67931683d73

packages/wallet/src/SingleAddressWallet.ts
```diff
@@ -325,6 +325,9 @@ export class SingleAddressWallet implements ObservableWallet {
     this.#resubmitSubscription = createTransactionReemitter({
       genesisParameters$: this.genesisParameters$,
       logger: contextLogger(this.#logger, 'transactionsReemitter'),
+      // REVIEW: should this be a prop to SingleAddressWallet?
+      // Should default be computed?
+      reemitUnconfirmedAfterSlots: 300,
```
mirceahasegan, line 330: I think we should reemit until we reach invalidHereafter. I can't think of any reason why we should stop reemitting if it wasn't confirmed.

mkazlauskas: This will not make it stop re-emitting, it is a duration to wait until we consider that it was likely rolled back and re-emit. Perhaps the name of this param could be improved :sweat:

mirceahasegan: 🤦 . The name is clear. I was biased by my own expectations :))

mirceahasegan: I would be for _not_ exposing it as configurable. Can we use the chainHistory polling interval as reference? If it's not there within 2 polling intervals, reemit

mkazlauskas: Great idea! I think 2 might be too low, maybe 3 (or more) polling intervals :thinking:

packages/wallet/src/services/TransactionsTracker.ts
```diff
@@ -212,27 +213,36 @@ export const createTransactionsTracker = (
       map((transactions) => transactions.filter(({ submittedAt }) => !!submittedAt)),
       mergeMap((transactions) => from(transactions))
     )
-  ).pipe(share());
+  ).pipe(
+    // Tx could be re-submitted, so we group by tx id
+    groupBy(({ tx: { id } }) => id),
```
mirceahasegan, line 218: I don't understand. Could there be multiple tx with same id, so you can group them?

mkazlauskas: Yes, this `groupBy` is needed to avoid "double tracking" the same transaction if it is re-submitted.
1. Submit tx (submitted$)
2. Re-submit tx (submitted$ same tx id) - without this change there would be 2 inner observables alive, so you would have 2 emissions on confirmed$, failed$ and inFlight$ as transaction state changes.

packages/wallet/src/SingleAddressWallet.ts
mkazlauskas: 5c3a4167e5e6e870f89bed0fda5cf59895f35678

packages/wallet/src/services/SmartTxSubmitProvider.ts
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Wraps a `TxSubmitProvider` to enchance it's `submitTx` with the following functionality:
+ * - Immediately rejects if network tip is already >= `ValidityInterval.invalidHereafter`
+ * - Awaits for the following conditions before submitting:
+ *   - Network tip is ahead of tx body `ValidityInterval.invalidBefore`.
+ *   - Connection status is 'up'.
+ * - Re-submits transactions that failed to submit due to connection or recoverable provider issue.
+ */
+// REVIEW: an alternative name is ManagedTxSubmitProvider.
```
rhyslbw, line 34: I'm happy with `SmartTxSubmitProvider` now I read the code :-)

packages/wallet/test/services/SmartTxSubmitProvider.test.ts
```diff
@@ -0,0 +1,140 @@
...
+        await provider.submitTx({ signedTransaction: txWithoutValidityIntervalHex });
+        expect(underlyingProvider.submitTx).toBeCalledTimes(1);
+      });
+
+      it('calls underlying provider when online and within validity interval', async () => {
+        await provider.submitTx({ signedTransaction: txWithValidityIntervalHex });
+        expect(underlyingProvider.submitTx).toBeCalledTimes(1);
+      });
+
+      it('re-submits transactions that failed to submit due to a connection or recoverable server issue', async () => {
```
rhyslbw, line 73:
```suggestion
      it('re-submits transactions that failed to submit due to a connection or recoverable provider issue', async () => {
```
This could be used in Node.js too :sunglasses:

rhyslbw (changes requested):
Awesome work @mkazlauskas. Just a minor suggestion

packages/wallet/src/services/TransactionReemitter.ts
```diff
@@ -139,9 +141,9 @@ export const createTransactionReemitter = ({
     mergeMap((txs) => from(txs))
   );
 
-  const reemitUnconfirmed$ = combineLatest([tipSlot$, inFlight$]).pipe(
-    mergeMap(([tip, inFlight]) => {
-      const reemitSubmittedBefore = tip - reemitUnconfirmedAfterSlots;
+  const reemitUnconfirmed$ = combineLatest([tipSlot$, inFlight$, genesisParameters$]).pipe(
```
mirceahasegan, line 144: There might be a corner case when genesisParameters$ emits twice (epoch change?).
combineLatest will make reemitUnconfirmed$ to emit the same inFlight transactions again.
```
tipSlot$ =  '         a-'
inFlight$ = '         b-'
genesisParameters$ = 'x-y--'
combineLatest  = '    (abx)-(aby)'
reemitUnconfirmed$ = 'b-b'; // same transaction emitted twice

File: packages/wallet/test/services/SmartTxSubmitProvider.test.ts (hunk 7 above)
Comment (mkazlauskas) [reply packages/wallet/test/services/SmartTxSubmitProvider.test.ts]: 3af7e3498ba6f0a20cc1fa52c410a4a7dec9ea0a

File: packages/wallet/src/services/SmartTxSubmitProvider.ts (hunk 6 above)
Comment (mkazlauskas) [reply packages/wallet/src/services/SmartTxSubmitProvider.ts]: 3af7e3498ba6f0a20cc1fa52c410a4a7dec9ea0a

File: packages/wallet/src/services/TransactionReemitter.ts (hunk 8 above)
Comment (mkazlauskas) [reply packages/wallet/src/services/TransactionReemitter.ts]: Nice catch! This has actually crossed my mind too. An alternative has a caveat too:
```
tipSlot$ =  '         a-c'
genesisParameters$ = '-b'
reemitUnconfirmed$ = '--c';
// if genesisParameters$ is not available yet when tip$ is emitted,
// it will get delayed until another emission of tip$
```

8a4c691d6520bbd7aca6de9e6827af4857c585bb I think that's slightly better than original implementation

mirceahasegan (dismissed):
Great work, @mkazlauskas! 👏

rhyslbw (dismissed):
Nice work @mkazlauskas

mkazlauskas:
Squashed and rebased on master

mirceahasegan (dismissed):

rhyslbw:
A couple of conflicts to resolve @mkazlauskas

mirceahasegan (approved):

rhyslbw (approved):

PR #459, opened 2022-09-23
fix(wallet): coldObservable retry backoff should recreate observable

mirceahasegan:
# Context

Cold observable retry backoff cannot resubscribe to the same observable.
It must recreate the provider observable when retrying.

# Proposed Solution

Wrap provider in `defer` to create an observable factory that is recreated for every retry.


mkazlauskas (approved):
Great find! :sweat_smile:

rhyslbw (approved):
:mechanical_arm:

PR #460, opened 2022-09-23
fix(cardano-services): wrong tx onput `txId` value

IvayloAndonov:
# Context
The `txId` for the inputs is different than the one provided by Blockfrost.
Cardano services have the tx id of the current transaction, Blockfrost has the tx id of the tx that had that input as output

Requests to both `cardano-services` and `blockfrost` in preview:
image (1)

image (2)

# Proposed Solution

Extend DB query to extract the correct tx input id which represents the transaction(source) id where the input was spent as an output. We can conclude that the tx input `txId` is equal to the source transaction `id`.

We extend the query because before the final mapping to `Cardano.TxIn` return type, we need both `tx_input_id` and `tx_source_id` to filter and map by.

- extract the correct `tx_source_id` from db and use it as input `txId`
- update the mappings and types
- added a new e2e test

Introduced e2e test - [txChainHistory.test.ts](https://github.com/input-output-hk/cardano-js-sdk/pull/460/files#diff-6de00ab9b73e5cc2dabcfa65c3f90552a9b8bd4ba27047d7c05e029084f5ca44) which verifies that signed transaction and the same transaction found from the history, must have the same body.

The e2e test is failing without the introduced fix, after the fix - it passes. CI build: [CI build: Test e2e step](https://github.com/input-output-hk/cardano-js-sdk/actions/runs/3128775424/jobs/5077038709)

# Notes
Due to missing enough DB test data, we observe an empty `tx.body.inputs:[]` array in the snapshots. We might need to extend our db-fixtures with more txs/blocks.

Based on the updated snapshot test [Here](https://github.com/input-output-hk/cardano-js-sdk/compare/fix/wrong-tx-id-for-tx-inputs?expand=1#diff-044edd9417ac4bfbdcfdcedd13705ade39a95cac336e0372b3b1b5544cd520bdR791), we can state that the introduced changes fix the wrong tx input `txId` value - there is no duplication between source tx id and current tx input id.

### DIFF
--- a/packages/cardano-services/test/ChainHistory/__snapshots__/ChainHistoryHttpService.test.ts.snap
+++ b/packages/cardano-services/test/ChainHistory/__snapshots__/ChainHistoryHttpService.test.ts.snap
@@ -66,20 +66,12 @@ Array [
       ],
       "collaterals": Array [],
       "fee": 197929n,
-      "inputs": Array [
-        Object {
-          "address": "addr_test1qr4m502gr9hnaxac5mxjln22jwavf7pcjmh9sw7fujdvgvj9ef6afquphwg7tj4mmm548m3t50hxfyygjuu222kx96eshcathg",
-          "index": 0,
-          "txId": "8fd14baca91c674fafae59701b7dc0eda1266202ec8445bad3244bd8669a7fb5",
-        },
-      ],
+      "inputs": Array [],
       "mint": undefined,
       "outputs": Array [
         Object {
           "address": "addr_test1qr4m502gr9hnaxac5mxjln22jwavf7pcjmh9sw7fujdvgvj9ef6afquphwg7tj4mmm548m3t50hxfyygjuu222kx96eshcathg",
           "datum": undefined,
-          "index": 0,
-          "txId": "8fd14baca91c674fafae59701b7dc0eda1266202ec8445bad3244bd8669a7fb5",
           "value": Object {
             "assets": undefined,
             "coins": 999502622402n,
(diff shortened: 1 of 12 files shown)


packages/cardano-services/test/ChainHistory/__snapshots__/ChainHistoryHttpService.test.ts.snap
```diff
@@ -1486,7 +788,7 @@ Array [
         Object {
           "address": "addr_test1qpcnmvyjmxmsm75f747u566gw7ewz4mesdw7yl278uf9r3f5l7d7dpx2ymfwlm3e56flupga8yamjr2kwdt7dw77ktyqqtx2r7",
           "index": 0,
-          "txId": "19251f57476d7af2777252270413c01383d9503110a68b4fde1a239c119c4f5d",
+          "txId": "face165bd7aa8d0d661cf1ceaa4e35d7611be3b1c7997da378c547aa2464a4fd",
```
IvayloAndonov, line 791: Updated input `txId` value

packages/e2e/test/wallet/SingleAddressWallet/txChainHistory.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  });
+
+  it('submit a transaction and find it in chain history', async () => {
+    const tAdaToSend = 10_000_000n;
+    const saWallet = wallet.wallet;
+
+    await firstValueFrom(saWallet.syncStatus.isSettled$.pipe(filter((isSettled) => isSettled)));
+
+    const [{ address: sendingAddress }] = await firstValueFrom(saWallet.addresses$);
+    const [{ address: receivingAddress }] = await firstValueFrom(saWallet.addresses$);
```
mkazlauskas, line 30:
```suggestion
    const receivingAddress = sendingAddress;
```

packages/e2e/test/wallet/SingleAddressWallet/txChainHistory.test.ts
```diff
@@ -0,0 +1,76 @@
...
+import { getWallet } from '../../../src';
+import { logger } from '@cardano-sdk/util-dev';
+
+export type TestWallet = Awaited<ReturnType<typeof getWallet>>;
+
+describe('tx chain history', () => {
+  let wallet: TestWallet;
+
+  beforeEach(async () => {
+    wallet = await getWallet({ env, logger, name: 'Sending Wallet', polling: { interval: 50 } });
```
mkazlauskas, line 16: Other props don't seem to be used
```suggestion
    ({ wallet }) = await getWallet({ env, logger, name: 'Sending Wallet', polling: { interval: 50 } });
```

packages/e2e/test/wallet/SingleAddressWallet/txChainHistory.test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    const signedTx = await unsignedTx.sign();
+    await signedTx.submit();
+
+    logger.info(
+      `Submitted transaction id: ${signedTx.tx.id}, inputs: ${signedTx.tx.body.inputs} and outputs:${signedTx.tx.body.outputs}.`
+    );
+
+    // Wait until wallet two is aware of the funds.
+    await firstValueFrom(saWallet.balance.utxo.total$.pipe(filter(({ coins }) => coins >= tAdaToSend)));
```
mkazlauskas, line 51: This seems to be redundant as you're awaiting for tx and not asserting the balance
```suggestion
```

packages/e2e/test/wallet/SingleAddressWallet/txChainHistory.test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    // Assert
+    expect(txFoundInHistory).toBeDefined();
+    expect(txFoundInHistory.id).toEqual(signedTx.tx.id);
+    expect(txFoundInHistory.body.inputs.length).toEqual(signedTx.tx.body.inputs.length);
+    expect(txFoundInHistory.body.outputs.length).toEqual(signedTx.tx.body.outputs.length);
+    expect(txFoundInHistory.body.inputs.every((txIn, i) => deepEquals(signedTx.tx.body.inputs[i], txIn))).toBe(true);
+    expect(
+      txFoundInHistory.body.outputs.every((txOut, i) => signedTx.tx.body.outputs[i].value.coins === txOut.value.coins)
+    ).toBe(true);
```
mkazlauskas, line 74: This would cover other body properties too
```suggestion
    expect(txFoundInHistory.body).toEqual(signedTx.tx.body);
```

mkazlauskas (changes requested):
Just a few suggested improvements for the e2e test

rhyslbw:
@IvayloAndonov I'm finalising the change in #463 for you to use instead, which will cover the missing snapshot data

packages/e2e/test/wallet/SingleAddressWallet/txChainHistory.test.ts
IvayloAndonov: All the e2e test feedback is applied under 1bbf6d29661f08f7030a02c3316ae7d54b3d4ff3 @mkazlauskas

packages/e2e/test/wallet/util.ts
```diff
@@ -43,3 +44,8 @@ export const walletReady = (wallet: ObservableWallet) =>
     'Took too long to be ready',
     SYNC_TIMEOUT
   );
+
+export const normalizeTxBody = (body: any) => {
```
mkazlauskas, line 48:
```suggestion
export const normalizeTxBody = (body: TxBodyAlonzo | NewTxBodyAlonzo) => {
```

mkazlauskas (dismissed):
Nice work! No need for a separate commit for the 1 requested change

mkazlauskas (approved):

rhyslbw (dismissed):
Looks good @IvayloAndonov. Let's ignore the snapshot data issues, as the e2e test covers the scenario, and this fix is now getting urgent to resolve. I'll update the integration tests when I replace the testnet snapshot. Please squash

rhyslbw (approved):

mkazlauskas (dismissed):

rhyslbw (approved):

PR #463, opened 2022-09-26
Test/local network snapshot

rhyslbw:
# Context
Our current `cardano-services` integration test DB snapshot is not ideal, given it only contains a subset of the data managed by a tedious selection process to satisfy particular test demands. For this reason, we cannot currently make assertions based on what we know from the tests that have run.

# Proposed Solution
Use the local-network state after the wallet e2e test has run as a source for running integration tests. This is done using a bash script to coordinate the network setup/teardown, test execution, and snapshot creation. While it's not possible to make concrete assertions that transactions appear in hard-coded blocks, additional setup queries should be possible to make similar assertions in an dynamic way.

- Upgrades Docker Compose to V2
- The `local-network` database is currently just being loaded alongside `testnet` for demonstration of the workflow.


mirceahasegan (approved):
LGTM 🚀

mkazlauskas (approved):
Looks good, although we probably need to update all jest snapshots too?

rhyslbw:
> Looks good, although we probably need to update all jest snapshots too?

Yeah, I'm going to see if we can switch from using the other type of snapshot completely, then will update the test setup and snapshots and assertions.

packages/cardano-services/test/ChainHistory/ChainHistoryHttpService.test.ts
```diff
@@ -464,26 +443,18 @@ describe('ChainHistoryHttpService', () => {
 
       describe('server and snapshot testing', () => {
```
rhyslbw, line 444: This describe block is a bit vague. Can you think of a better name?

packages/cardano-services/.env.test
```diff
@@ -20,3 +20,7 @@ POSTGRES_SSL_CA_FILE=./test/ssl-ca-util.pem
 
 OGMIOS_SRV_SERVICE_NAME=ogmios-test-domain
 RABBITMQ_SRV_SERVICE_NAME=rabbitmq-test-domain
+
+# UNIT TESTS - We need to support both while we complete the migration
+UNIT_TESTS_POSTGRES_CONNECTION_STRING=postgresql://postgres:doNoUseThisSecret!@127.0.0.1:5433/testnet
```
rhyslbw, line 25:
```suggestion
INTEGRATION_TESTS_POSTGRES_CONNECTION_STRING=postgresql://postgres:doNoUseThisSecret!@127.0.0.1:5433/testnet
```

packages/cardano-services/.env.test
```diff
@@ -20,3 +20,7 @@ POSTGRES_SSL_CA_FILE=./test/ssl-ca-util.pem
...
+LOCALNETWORK_UNIT_TESTS_POSTGRES_CONNECTION_STRING=postgresql://postgres:doNoUseThisSecret!@127.0.0.1:5433/localnetwork
```
rhyslbw, line 26:
```suggestion
LOCALNETWORK_INTEGRATION_TESTS_POSTGRES_CONNECTION_STRING=postgresql://postgres:doNoUseThisSecret!@127.0.0.1:5433/localnetwork
```

packages/cardano-services/.env.test
```diff
@@ -20,3 +20,7 @@ POSTGRES_SSL_CA_FILE=./test/ssl-ca-util.pem
 
 OGMIOS_SRV_SERVICE_NAME=ogmios-test-domain
 RABBITMQ_SRV_SERVICE_NAME=rabbitmq-test-domain
+
+# UNIT TESTS - We need to support both while we complete the migration
```
rhyslbw, line 24:
```suggestion
# INTEGRATION TESTS - We need to support both while we complete the migration
```

packages/e2e/local-network/scripts/make-babbage.sh
```diff
@@ -107,7 +107,6 @@ cp templates/babbage/byron-configuration.yaml "${ROOT}/configuration.yaml"
 
 $SED -i "${ROOT}/configuration.yaml" \
      -e 's/Protocol: RealPBFT/Protocol: Cardano/' \
-     -e 's/minSeverity: Info/minSeverity: Debug/' \
```
iccicci, line 110: I guess this is the reason why I wasn't able to set `Info` log level by changing the `node-config.json`!

packages/e2e/local-network/scripts/update-pools-certificate.sh
```diff
@@ -129,7 +129,7 @@ for ((i = 0; i < ${#SP_NODES[@]}; ++i)); do
   cardano-cli stake-pool registration-certificate \
       --cold-verification-key-file "$coldVKey" \
       --vrf-verification-key-file "$vrfKey" \
-      --pool-pledge 500000000 \
+      --pool-pledge 0 \
```
iccicci, line 132: Could somebody please point me a good reading to understand the meaning of this change?
Thank you

rhyslbw: https://hydra.iohk.io/build/17427421/download/1/delegation_design_spec.pdf
https://github.com/input-output-hk/cardano-ledger

rhyslbw:
Looks good @AngelCastilloB. I've just rerun the CI workflow as it was failing with the `ValueNotConserved` error

mkazlauskas (approved):
Looks good, just note that this change
> healtcheck of dbSync now waits until ithas synced up to three epochs

will make cardano-services unusable for some time when some testnet is recreated

mkazlauskas (dismissed):

mkazlauskas (approved):

mirceahasegan (approved):

PR #466, opened 2022-09-27
feat(cardano-services): get metadata should return undefined on errors

mirceahasegan:
# Context

`DbSyncAssetProvider.getAsset` returns `nftMetadata: undefined` when the asset is not an NFT.
Related to [PR-412 fix: Gracefully handle CardanoTokenRegistry errors during getAsset request ](https://github.com/input-output-hk/cardano-js-sdk/pull/412)

# Proposed Solution

It should return `nftMetadata: null` instead of undefined, similar to `tokenMetadata`.
It should catch `getTokenMetadata` errors and return `undefined`. Clients can then retry the fetch.

# Todo:

- [x] Initial implementation
- [x] Validate try/catch approach (see code REVIEW comment)
- [x] Update unit tests

### DIFF
--- a/packages/cardano-services/test/Asset/DbSyncNftMetadataService.test.ts
+++ b/packages/cardano-services/test/Asset/DbSyncNftMetadataService.test.ts
@@ -9,7 +9,7 @@ export const oneMintNft: AssetPolicyIdAndName = {
   policyId: Cardano.PolicyId('50fdcdbfa3154db86a87e4b5697ae30d272e0bbcfa8122efd3e301cb')
 };
 
-export const unexistingAsset: AssetPolicyIdAndName = {
+export const nonExistentAsset: AssetPolicyIdAndName = {
   name: Cardano.AssetName(''),
   policyId: Cardano.PolicyId('c0d07d45fe9514f80213f4020e5a61241458be626841cde717cb38a7')
 };
@@ -33,12 +33,12 @@ describe('DbSyncNftMetadataService', () => {
     await dbConnection.end();
   });
 
-  it('returns undefined for unexisting asset', async () => {
-    expect(await service.getNftMetadata(unexistingAsset)).toBeUndefined();
+  it('returns null for non-existent asset', async () => {
+    expect(await service.getNftMetadata(nonExistentAsset)).toBeNull();
   });
 
-  it('returns undefined for non-nft asset', async () => {
-    expect(await service.getNftMetadata(nonNftAsset)).toBeUndefined();
+  it('returns null for non-nft asset', async () => {
+    expect(await service.getNftMetadata(nonNftAsset)).toBeNull();
   });
 
   it('returns nft metadata when it exists', async () => {
--- a/packages/cardano-services/src/Asset/DbSyncNftMetadataService.ts
+++ b/packages/cardano-services/src/Asset/DbSyncNftMetadataService.ts
@@ -28,11 +28,11 @@ export class DbSyncNftMetadataService implements NftMetadataService {
     this.#metadataService = metadataService;
   }
 
-  async getNftMetadata(assetInfo: AssetPolicyIdAndName): Promise<Asset.NftMetadata | undefined> {
+  async getNftMetadata(assetInfo: AssetPolicyIdAndName): Promise<Asset.NftMetadata | null> {
     // Perf: could query last mint tx metadata in 1 query instead of 2
     const lastMintedTx = await this.#builder.queryLastMintTx(assetInfo.policyId, assetInfo.name);
 
-    if (!lastMintedTx) return;
+    if (!lastMintedTx) return null;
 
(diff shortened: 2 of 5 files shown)


packages/cardano-services/src/Asset/DbSyncAssetProvider.ts
```diff
@@ -64,8 +65,18 @@ export class DbSyncAssetProvider extends DbSyncProvider implements AssetProvider
...
     if (extraData?.history) await this.loadHistory(assetInfo);
-    if (extraData?.nftMetadata)
-      assetInfo.nftMetadata = await this.#dependencies.ntfMetadataService.getNftMetadata(assetInfo);
+    if (extraData?.nftMetadata) {
+      try {
+        // Use `null` instead of `undefined` when no nftMetadata is found / asset is not an NFT.
+        // Undefined means that there was a problem while querying and users can retry it.
+
+        // REVIEW: getNftMetadata does db-sync query or blockfrost query. Should we filter errros instead
+        // of catching all and transforming to undefined?
```
mirceahasegan, line 74: @IvayloAndonov, can you help with some insight here?

iccicci (approved):
Apart from possible improvements from specific errors handling; this is good

packages/cardano-services/src/Asset/DbSyncAssetProvider.ts
```diff
@@ -64,8 +65,18 @@ export class DbSyncAssetProvider extends DbSyncProvider implements AssetProvider
...
+        assetInfo.nftMetadata = (await this.#dependencies.ntfMetadataService.getNftMetadata(assetInfo)) || null;
```
mirceahasegan, line 75: I think it would be better to use the [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) operator here
- assetInfo.nftMetadata = (await this.#dependencies.ntfMetadataService.getNftMetadata(assetInfo)) || null;
+ assetInfo.nftMetadata = (await this.#dependencies.ntfMetadataService.getNftMetadata(assetInfo)) ?? null;

packages/cardano-services/src/Asset/DbSyncAssetProvider.ts
IvayloAndonov: If the motivation of this PR is the same as for the [tokenMetadata](https://github.com/input-output-hk/cardano-js-sdk/pull/412), then we should handle the errors related to services that are not under our control (same as token registry service)
Since here we have two possible providers of nft metadata: `Blokfrost` is the external one, and `db-sync` is our concern I think here we could handle the HTTP 500 errors thrown by Blockfrost, then log and mark `nftMetadata` as undefined, if the behavior should be similar as for `tokenMetadata`

Not sure about `db-sync` though, there we have service discovery and re-resolving logic, so might be worth handling connection errors, set again `nftMetadata` as `undefined` and retry on the client I guess.
@mirceahasegan

mirceahasegan: It is the same motivation as for [tokenMetadata](https://github.com/input-output-hk/cardano-js-sdk/pull/412).
Do we want to put the effort into filtering the Blockfrost errors, or can we go with the naive approach and catch all -> return undefined?

IvayloAndonov: Sorry, here in the context of `cardano-services` we perform only `db-sync` operations, blockfrost is completely separate provider, I was misled by the comment initially.
If the core types are not update then we shouldn't do anything with blockfrost
 @mirceahasegan

packages/cardano-services/test/Asset/DbSyncNftMetadataService.test.ts
```diff
@@ -33,12 +33,12 @@ describe('DbSyncNftMetadataService', () => {
     await dbConnection.end();
   });
 
-  it('returns undefined for unexisting asset', async () => {
-    expect(await service.getNftMetadata(unexistingAsset)).toBeUndefined();
+  it('returns null for unexisting asset', async () => {
```
rhyslbw, line 36:
```suggestion
  it('returns null for non-existent asset', async () => {
```

rhyslbw (changes requested):

packages/cardano-services/test/Asset/DbSyncNftMetadataService.test.ts
mirceahasegan: done in [7b87d1d](https://github.com/input-output-hk/cardano-js-sdk/commit/7b87d1ded0df482d29dd2b581a3baa9035f32d06)

rhyslbw (approved):

iccicci (approved):

IvayloAndonov (approved):
GLTM

PR #467, opened 2022-09-28
test(fix): pool certificate update will also re-register genesis dele…

AngelCastilloB:
# Context

The live stake calculation was not working on the local test network as the genesis delegators stake could not be found by our query.

# Proposed Solution

Spend these genesis UTXO once by redelegating to the pools.


mkazlauskas (approved):
Looks good! Is it a db-sync issue (and if it is, is it reported)?

rhyslbw (approved):

PR #468, opened 2022-09-28
perf(cardano-services): order by block.id in ledger tip queries

rhyslbw:
# Context
We're ordering SQL queries by an unindexed column, `block_no`, so queries on mainnet sized databases are taking ~10 seconds

# Proposed Solution
Order by the naturally indexed `id` column.


mkazlauskas (approved):

mirceahasegan (approved):
👍

PR #469, opened 2022-09-29
fix(wallet): await for updated rewards after tx confirmation

mkazlauskas:
# Context

Rewards are computed by subtracting in-flight tx withdrawals. Once tx is confirmed, it triggers a re-fetch of rewards balance from the provider. During this re-fetch, the tx is no longer 'in-flight', so there was a brief period of time when rewards balance is incorrect.

# Proposed Solution

See inline comment in the commit

## Follow-up work

ADP-2275 (see inline comment in [this wip commit](https://github.com/input-output-hk/cardano-js-sdk/commit/131a18ee33148ab55bdb456c0ba5ca7c5d6b49af#diff-a24ba91fcdc49ed6949e92d1ceeef8b221721a0c810cb4f1c54b758a9d243a84R279-R295)).

f50ce5dc84a76a0a5db7426906f13458f1df869f isn't really needed for this PR and is just some preparation work for that parked commit, but I think it's better to merge it now to have less git conflicts later.

### DIFF
--- a/packages/cardano-services/src/Rewards/DbSyncRewardProvider/queries.ts
+++ b/packages/cardano-services/src/Rewards/DbSyncRewardProvider/queries.ts
@@ -1,4 +1,13 @@
 export const findAccountBalance = `
+    WITH current_epoch AS (
+        SELECT
+            e."no" AS epoch_no,
+            optimal_pool_count
+        FROM epoch e
+        JOIN epoch_param ep ON
+            ep.epoch_no = e."no"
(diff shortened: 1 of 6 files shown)


mirceahasegan (dismissed):

mirceahasegan (dismissed):
Great work, @mkazlauskas!

mirceahasegan (approved):

iccicci (approved):
LGTM!

PR #474, opened 2022-10-03
feat(nix): Nix packaging cleanup

YorikSar:
# Context

Fix issues with Nix packaging that surfaced over time:
* size of Nix package is huge (1.2 GB)
* Nix package depends on all sources in this repo, not just the code
* Nix package in this repo cannot include cardano config, because Nix flakes don't support submodules, so we have to use this flake as input with `submodules` parameter and rebuild the package in downstream
* 

# Proposed Solution

To address these issues:
* replace Git submodule with [git-subrepo](https://github.com/git-commands/git-subrepo) to include whole cardano-configurttion repository in this repo and track updates
* filter sources in Nix package according to what's actually needed in Dockerfile
* keep only paths in resulting package that are needed, according to Dockerfile
* only keep production dependencies in Dockerfile

* Git submodule is replaced with [git-subrepo](https://github.com/git-commands/git-subrepo)


mkazlauskas (changes requested):
LGTM, just a few requests for commit messages:
- We're not using `nix` commit type. I think `build` would be good for these 2 commits.
- ee3486fc9ccae42b0262ebb8ce075c89f85e584e doesn't have a type. Was it created automatically by `git` (might be fine then :thinking:)? I suggest squashing it with the previous commit (remove submodule), as repo state is broken at that commit without adding the subrepo.

YorikSar:
@mkazlauskas Replaced `nix:` with `build(nix):` and squashed removal of submodule with addition of subrepo, also added more info to that commit.
The commit message created by `git subrepo` is important only if we do both push and pull for it, but we'll only be doing pull.

mkazlauskas (dismissed):

rhyslbw (changes requested):
It would be good to have a yarn script in the [top level `package.json`](https://github.com/input-output-hk/cardano-js-sdk/blob/master/package.json#L12):
```
"config:update": "git subrepo pull packages/cardano-services/config --message='chore: git subrepo pull packages/cardano-services/config'"
```

Otherwise, looks great! It will be a big improvements to not have the submodule.

YorikSar:
@rhyslbw Good point. Added shortcuts to both `package.json` and `flake.nix`, whichever would be more convenient.

mkazlauskas (approved):

rhyslbw (approved):

PR #475, opened 2022-10-03
feat: RewardsProvider.rewardsHistory constraints

IvayloAndonov:
# Context

This query currently returns unconstrained results, posing a DOS risk.

# Proposed Solution
Since the return type of [rewardsHistory](https://github.com/input-output-hk/cardano-js-sdk/blob/5910f552755e2a3fda75c48437f8534a913ad6c2/packages/cardano-services/src/Rewards/DbSyncRewardProvider/DbSyncRewardsProvider.ts#L18-L21) is a simple map (not a list with multiple items) : `<Map<Cardano.RewardAccount,EpochRewards[]>` and the relation between the `rewardAccounts` arg and mapped response is 1:1 then we might be better just to apply the `props.rewardAccounts` limitation of `PAGINATION_PAGE_SIZE` without pagination of the response.

- apply `rewardAccounts` arg runtime limitation of max page size limit
- extends with missing expect.assertions()


mkazlauskas (approved):
Looks good! :rocket:

iadmytro (approved):

iccicci (approved):
LGTM

PR #477, opened 2022-10-05
ADP-2273 ogmios websocket connection health check inconsistency

iccicci:
# Context

Some queries to ogmois server are performed in HTTP (as the health check) and some other are performed through WebSocket; this is possibly causing an inconsistency in calls response when the WebSocket unexpectedly close.

When the WebSocket close it happens that may requests fail with provider unhealthy while the health check continues to respond `ok`.

# Proposed Solution

I leave the stage to @IvayloAndonov , who is the actual author of this proposal.
Taking advantage from the dns resolution system, if we correctly handle the `WebSocket is closed` as a network error, we not only resolve the inconsistency issue, but we also achieve the reconnection/retry logic.
And it works! Congratulations Ivo 🤝 the actual fix is the 1 line change in the first commit of this PR!

- The fix as proposed by Ivo
- An e2e test to reproduce the error and check the fix actually does it job
- Solved some issues found while developing the e2e test

# Still missing

In my local environment the test is failing due to EACCESS on a file (owned by `root` with permission `0600`).
In order to make the test work I changed _by hands_ the file permission.
I need support from the team to make the automated test run to work; once the CI will run (and I guess fail) and the error is visible, I trust @AngelCastilloB can quickly give me some resolving hints in a while.


packages/cardano-services/src/Program/loadHttpServer.ts
```diff
@@ -11,19 +11,23 @@ import { DbSyncUtxoProvider, UtxoHttpService } from '../Utxo';
...
-import Logger, { createLogger } from 'bunyan';
 import memoize from 'lodash/memoize';
 import pg from 'pg';
 
 export interface HttpServerOptions extends CommonProgramOptions {
   serviceNames?: ServiceNames[];
   enableMetrics?: boolean;
   cardanoNodeConfigPath?: string;
+  dnsResolver?: (serviceName: string) => Promise<SrvRecord>;
+  logger?: Logger;
```
mkazlauskas, line 30: I would put these in a separate object (2nd arg to loadHttpServer). `HttpServerDependencies`?

mkazlauskas (changes requested):
Looking good! Just 1 suggestion below

packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
```diff
@@ -112,6 +112,7 @@ export class DbSyncNetworkInfoProvider extends DbSyncProvider implements Network
 
   async close(): Promise<void> {
     this.#cache.shutdown();
+    await this.#genesisDataReady;
```
IvayloAndonov, line 115: Just of a question: why execute the genesis data promise in close()?

packages/core/src/errors.ts
```diff
@@ -22,19 +22,43 @@ export const providerFailureToStatusCodeMap: { [key in ProviderFailure]: number
...
+
+/**
+ * This type check works as an "error instanceof Error" check, but it let pass also those objects
+ * which implements the Error interface without inheriting from the same base class
+ * (as the errors thrown by the fs package are)
+ *
+ * @param error the error object to evaluate
+ * @returns whether the input error is or not an ErrorLike
+ */
+const isErrorLike = (error: unknown): error is ErrorLike => {
```
IvayloAndonov, line 38: 👍🏻

packages/cardano-services/src/Program/loadHttpServer.ts
iccicci: I spent at least half an hour this morning thinking and writing about standardization of our interfaces and I lost this...
Thank you for the suggestion!

packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
iccicci: The `Promise` is created in the `constructor` and `await`ed in `genesisParameters` and `lovelaceSupply` methods.

In case `loadGenesisData` fails and `shutdown` is called without never calling neither `genesisParameters` nor `lovelaceSupply`, an unhandled rejection error is raised.

packages/e2e/test/ogmios/ogmios-interactions.test.ts
```diff
@@ -0,0 +1,162 @@
...
+    const networkInfoProvider = await networkInfoProviderFactory.create(
+      providerEnv.NETWORK_INFO_PROVIDER,
+      providerEnv.NETWORK_INFO_PROVIDER_PARAMS,
+      logger
+    );
+
+    await closeWSProxy?.();
+    await openWSProxy();
+
+    expect(await networkInfoProvider.stake()).toBeDefined();
```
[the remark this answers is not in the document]
iccicci: @IvayloAndonov , 642bcd3 should solve this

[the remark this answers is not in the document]
IvayloAndonov: Great

IvayloAndonov (approved):
Nice work!

mkazlauskas (dismissed):
:rocket:

IvayloAndonov (approved):

iadmytro (changes requested):

IvayloAndonov (approved):

iadmytro (approved):

PR #489, opened 2022-10-18
feat(web-extension)!: enhance remoteApi to allow changing observed ap…

mirceahasegan:
### Enhance remoteApi to allow changing observed api object

`ExposeApiProps` `api$` property allows changing the observed `API` objects in `exposeMessengerApi`.
This enables the following scenarios:
- `exposeMessengerApi` before having the actual `api` object and emit it once it's available. Methods returning `Promises` will await until the first `api` object is emitted.
- emit a new `api` object on `api$` observable. Consumers will start receiving values from the new object properties without having to resubscribe

# Context

Provide basis for changing wallet being observed in the lace wallet, allowing users to switch between wallets and networks.

All consumer subscriptions will complete only after the `api$` observable completes.

BREAKING CHANGE: `ExposeApiProps` `api` has changed to observable `api$`. Users can use rxjs `of` function to create an observable: `api$: of(api)` to adapt existing code to this change.


packages/web-extension/src/messaging/remoteApi.ts
```diff
@@ -242,22 +257,28 @@ export const bindObservableChannels = <API extends object>(
...
 export const exposeMessengerApi = <API extends object>(
-  { api, properties }: ExposeApiProps<API>,
+  { api$, properties }: ExposeApiProps<API>,
   dependencies: MessengerApiDependencies
 ): Shutdown => {
-  const observableChannelsSubscription = bindObservableChannels({ api, properties }, dependencies);
-  const nestedObjChannelsSubscription = bindNestedObjChannels({ api, properties }, dependencies);
+  const observableChannelsSubscription = bindObservableChannels({ api$, properties }, dependencies);
+  const nestedObjChannelsSubscription = bindNestedObjChannels({ api$, properties }, dependencies);
+  const apiTracker$ = new TrackerSubject(api$);
```
mkazlauskas, line 281: Now it can be completed in 2 ways: completing `api$`, or calling `shutdown()`. I think it would be a little better if there was only 1 way of doing that. This would rely only on `shutdown()`:
```suggestion
  const apiTracker$ = new TrackerSubject(concat(api$, NEVER));
  const observableChannelsSubscription = bindObservableChannels({ api$: apiTracker$, properties }, dependencies);
  const nestedObjChannelsSubscription = bindNestedObjChannels({ api$: apiTracker$, properties }, dependencies);
```

Also there might be edge cases where object emitted by api$ is shutdown before this. Maybe we should extend this so that `api$: Observable<API | null>` to be able to remove the local source (so that any calls and subscriptions would await for a new api obj before resolving, instead of using a destroyed obj). This is actually a legit use case with wallets: deleting a wallet and creating a new one.

mkazlauskas (changes requested):

packages/web-extension/src/messaging/remoteApi.ts
mirceahasegan: Added shutdown lifecycle in [b0b543f](https://github.com/input-output-hk/cardano-js-sdk/pull/489/commits/b0b543f4e0bc5b38be16c67535de6f0a34b5a504).
Handling `null` objects is not done yet. I am working to add better unit tests coverage for the remoteApi in general.

mirceahasegan (commented):
More unit test coverage is needed

packages/web-extension/src/messaging/remoteApi.ts
mirceahasegan: Unit tests added in [0e00207](https://github.com/input-output-hk/cardano-js-sdk/pull/489/commits/0e002077dc7bb334e6b530436a071ae66e1e9133).

packages/web-extension/test/messaging/remoteApi.test.ts
```diff
@@ -163,285 +166,455 @@ const setUp = (someNumbers$: Observable<bigint> = of(0n), nestedSomeNumbers$ = o
...
+
+        it('calls remote method from new source and resolves result', async () => {
+          expect(await sut.consumer.addOne(2n)).toBe(3n);
+          // add 11 instead of 1 to differentiate from original api object
+          otherApi.addOne = async (arg: bigint) => arg + 11n;
+          sut.api$.next(otherApi);
+          expect(await sut.consumer.addOne(2n)).toBe(13n);
+        });
+
+        // TODO: this fails in CI
```
rhyslbw, line 430: Can the tests not be `skip`ped?

packages/web-extension/src/messaging/types.ts
```diff
@@ -134,3 +134,7 @@ export interface MessengerApiDependencies {
   messenger: Messenger;
   logger: Logger;
 }
+
+export interface ApiSourceChanger<T> {
```
rhyslbw, line 138: This is unused. Is there a missing test?

rhyslbw (commented):
Looks good @mirceahasegan

packages/web-extension/test/messaging/remoteApi.test.ts
mirceahasegan: They were disabled previously. It's just the diff that is making it look as if it's a new change.
I can debug why it was failing in the first place, but I would rather do it before the PR is merged, as a separate task

packages/web-extension/src/messaging/types.ts
mirceahasegan: It's garbage from the initial implementation. I will remove it. Thanks!

mkazlauskas (dismissed):
Great work! :rocket:

packages/web-extension/src/messaging/remoteApi.ts
mirceahasegan: - null api handling to disable monitoring is implemented in: [9972fa4](https://github.com/input-output-hk/cardano-js-sdk/pull/489/commits/9972fa41de215c74487d1a50f71b1e5639d7336c).
- split unit tests into 2 files (unit & integration) ([aaa7d83](https://github.com/input-output-hk/cardano-js-sdk/pull/489/commits/aaa7d83eaa15bb88e94578e82d46e80db9214064), [7cadb7b](https://github.com/input-output-hk/cardano-js-sdk/pull/489/commits/7cadb7ba61c3e265940b7c1754fa7bc068282905))
- unit tests have been refactored to run same test suite on simple/flat object and nested object ([b14c58b](https://github.com/input-output-hk/cardano-js-sdk/pull/489/commits/b14c58bab0622dfe3d44d948bbdd19120e10f649))

packages/web-extension/src/messaging/types.ts
mirceahasegan: removed in [d82fd39](https://github.com/input-output-hk/cardano-js-sdk/pull/489/commits/d82fd39a7586375a628367d7e2481fe5a631608d)

mkazlauskas (approved):
Great work! :rocket:

rhyslbw (approved):
Nice work @mirceahasegan :sunglasses:

rhyslbw (approved):

mkazlauskas (approved):

PR #490, opened 2022-10-20
feat: Support Stake Pool Extended Metadata

IvayloAndonov:
# Context

Lace has a data-rich UI that includes information about stake pools not found in `cardano-db-sync`. Without this additional data, the Lace UX diminishes. To resolve this, we can patch this issue with a short-term solution - fetching the stake pool extended metadata from within the stake pool provider and caching the data in memory. This is not too dissimilar to how we handle asset metadata and should be modular code where possible.

It allows stake pool operators to include additional information about their stake pool at a URL defined in the extDataUrl field of a stake pool re-registration transaction, this references a JSON file that contains additional information not included. At this URL someone scraping data can find data associated with the pool's contacts, their media assets, some descriptions, and more.

As there are two competing “standards” for extended metadata:
- **CIP-6**
[CIP-6](https://cips.cardano.org/cips/cip6/) defines a standard for extended metadata of stake pools that is referenced from the pool registration data stored on-chain.
Schema: https://raw.githubusercontent.com/cardano-foundation/CIPs/master/CIP-0006/schema.json

- **AdaPools**
The more commonly used _Adaools_ format as extended metadata currently. 
For more details see here: https://a.adapools.org/extended-example

This task is to include and cache some of the data in each stake pool’s extended metadata JSON file in `cardano-services`'s stake pool provider.

# Proposed Solution

The solution is modeled based on the idea of first matching the pool to a format, then strictly validating the HTTP response against the corresponding schemas for correctness.

Having strictly JSON schema validation of the HTTP response we also provide more flexibility in error handling and more verbose logging of error details (problematic ext data URL value, exact validation error), compared with just having a single `openApi.json` schema at HTTP service root level.

Both schemas are located within `/Metadata/schemas/` folder

- define ext metadata types for both standards: **CIP-6** and **AdaPools**
- generalize `mockTokenRegistry` as generic and reusable test util
- implement `HttpMetadataService` with `getStakePoolExtendedMetadata` method
- add ext metadata schemas according to the formats: `CIP-6` or `AdaPools`
- use `jsonschema` package to validate the ext data response against the corresponding schema
- handle different errors accordingly
- fix existing typos

# Note: 
There are 2 options in order to test this code [snippet](https://github.com/input-output-hk/cardano-js-sdk/pull/490/files#diff-3a938d2c168d4b251612c6efabf201ac70e7585c3c1c2149b3c34e71981e6666R77-R90) with the error handling (the only thing left):
- we could extend the existing testnet DB snapshot data (add `extended` or `extDataUrl` JSON fields to `pool.offline_metadata`) 
- we could proceed directly with implementing e2e test with local-network data (I need support for that one to run local-network until I manage to resolve my M1 issue with dockerized ogmios)


packages/core/src/Cardano/types/StakePool/Cip6ExtendedStakePoolMetadata.ts
```diff
@@ -3,7 +3,7 @@ import { PoolIdHex } from './primitives';
 /**
  * Schema from:
  * https://raw.githubusercontent.com/cardano-foundation/CIPs/b93e77119e15d8763d7548a8a00c4cb7591714e4/CIP-0006/schema.json
- * This file was automatically generated by json-schema-to-typescript.
+ * This file was automatically generated by json-schema-to-typescript and then modified due to type names clashes.
```
rhyslbw, line 6: Can we not address the type name clashes in our definition? Modifying this auto-generated file is undesireable

packages/util-dev/src/createGenericMockServer.ts
```diff
@@ -0,0 +1,38 @@
+import { IncomingMessage, RequestListener, createServer } from 'http';
+import { getRandomPort } from 'get-port-please';
+
+export type MockHandler = (req?: IncomingMessage) => { body?: unknown; code?: number };
+export type ListenerGenerator = (handler: MockHandler) => RequestListener;
+
+export const createGenericMockServer =
```
rhyslbw, line 7: This would be good to document

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
```diff
@@ -112,14 +112,14 @@ export class CardanoTokenRegistry implements TokenMetadataService {
   }
 
   async getTokenMetadata(assetIds: Cardano.AssetId[]) {
-    this.#logger.debug(`Requested asset metatada for "${assetIds}"`);
+    this.#logger.debug(`Requested asset metadata for "${assetIds}"`);
```
rhyslbw, line 115: These spelling mistake corrections should be committed separately as a `fix` commit.

packages/cardano-services/src/Metadata/types.ts
```diff
@@ -1,11 +1,24 @@
 import { Cardano } from '@cardano-sdk/core';
 
-export interface MetadataService {
+export interface DbSyncMetadataService {
```
rhyslbw, line 3: This change is a (breaking) `refactor`, best committed separately.

packages/cardano-services/src/Metadata/util.ts
```diff
@@ -0,0 +1,35 @@
...
+ * CIP-6 schema: https://raw.githubusercontent.com/cardano-foundation/CIPs/b93e77119e15d8763d7548a8a00c4cb7591714e4/CIP-0006/schema.json
+ * AdaPools schema: https://a.adapools.org/extended-example
+ *
+ * @returns extended metadata format
+ */
+export const getSchemaFormat = (value: Cardano.StakePoolMetadata): ExtMetadataFormat =>
+  isNotNil(value.extended) ? ExtMetadataFormat.AdaPools : ExtMetadataFormat.CIP6;
+
+/**
+ * Loads json schema based on extended metadata format
```
rhyslbw, line 30:
```suggestion
 * Loads JSON schema based on extended metadata format
```

packages/cardano-services/test/Metadata/HttpMetadataService.test.ts
```diff
@@ -0,0 +1,172 @@
...
+      const extMetadata: Cardano.StakePoolMetadata = {
+        ...mainExtMetadataMock(),
+        extDataUrl: `${serverUrl}/${ExtMetadataFormat.CIP6}`
+      };
+
+      await expect(extendedMetadataService.getStakePoolExtendedMetadata(extMetadata)).rejects.toThrow(
+        new ProviderError(
+          ProviderFailure.InvalidResponse,
+          'instance requires property "pool"',
+          'Extended metadata json format validation failed against the corresponding schema for correctness'
```
rhyslbw, line 89:
```suggestion
          'Extended metadata JSON format validation failed against the corresponding schema for correctness'
```

packages/cardano-services/test/Metadata/HttpMetadataService.test.ts
```diff
@@ -0,0 +1,172 @@
...
+      const extMetadata: Cardano.StakePoolMetadata = {
+        ...mainExtMetadataMock(),
+        extended: `${serverUrl}/${ExtMetadataFormat.AdaPools}`
+      };
+
+      await expect(extendedMetadataService.getStakePoolExtendedMetadata(extMetadata)).rejects.toThrow(
+        new ProviderError(
+          ProviderFailure.InvalidResponse,
+          'instance requires property "info"',
+          'Extended metadata json format validation failed against the corresponding schema for correctness'
```
rhyslbw, line 107:
```suggestion
          'Extended metadata JSON format validation failed against the corresponding schema for correctness'
```

packages/cardano-services/test/Metadata/HttpMetadataService.test.ts
```diff
@@ -0,0 +1,172 @@
+/* eslint-disable max-len */
+import { Cardano, ProviderError, ProviderFailure } from '@cardano-sdk/core';
+import { ExtMetadataFormat, createHttpMetadataService } from '../../src/Metadata';
+import { adaPoolsExtMetadataMock, cip6ExtMetadataMock, mainExtMetadataMock } from './mocks';
+import { createGenericMockServer } from '@cardano-sdk/util-dev';
+import { dummyLogger } from 'ts-log';
```
rhyslbw, line 6: We should probably always be using the test logger now

packages/cardano-services/test/Metadata/HttpMetadataService.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    return res.end(JSON.stringify(adaPoolsExtMetadataMock));
+  } else if (reqUrl === `/${ExtMetadataFormat.CIP6}`) {
+    return res.end(JSON.stringify(cip6ExtMetadataMock));
+  }
+
+  return res.end(JSON.stringify(adaPoolsExtMetadataMock));
+});
+
+describe('HttpMetadataService', () => {
+  describe('healthy state', () => {
```
rhyslbw, line 31: Can you add a test to cover the scenario of metadata including both extended properties? In this case, CIP6 should take priority.

packages/cardano-services/test/Metadata/HttpMetadataService.test.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    beforeAll(async () => {
+      ({ closeMock, serverUrl } = await mockPoolExtMetadataServer(() => ({})));
+    });
+
+    afterAll(async () => {
+      await closeMock();
+    });
+
+    it('returns extended metadata in ada pools format when fetch from ext metadata url', async () => {
```
rhyslbw, line 44:
```suggestion
    it('returns ada pools format when extended key is present in the metadata', async () => {
```

packages/cardano-services/test/Metadata/HttpMetadataService.test.ts
```diff
@@ -0,0 +1,172 @@
...
+        ...mainExtMetadataMock(),
+        extended: `${serverUrl}/${ExtMetadataFormat.AdaPools}`
+      };
+      const result = await extendedMetadataService.getStakePoolExtendedMetadata(extMetadata);
+
+      expect(result).not.toBeNull();
+      expect(result).toEqual(adaPoolsExtMetadataMock);
+    });
+
+    it('returns extended metadata in cip-6 format when fetch from ext metadata url', async () => {
```
rhyslbw, line 55:
```suggestion
    it('returns CIP-6 format when extDataUrl is present in the metadata', async () => {
```

rhyslbw (changes requested):
Looking good so far @IvayloAndonov

packages/core/src/Cardano/types/StakePool/Cip6ExtendedStakePoolMetadata.ts
IvayloAndonov: The only reason I modified this file is that here just one of the types is auto-generated with the common name `ExtendedStakePoolMetadataFields` so by adding a prefix `Cip6` and `AP` to the exported types the difference between the formats is easily visible to the eye.

Another way would be to keep the generated ts file as it is with `ExtendedStakePoolMetadataFields` (represent CIP-6 format) and then have another type `APExtendedStakePoolMetadata` (AdaPools)?

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
IvayloAndonov: Split into a separate commit: https://github.com/input-output-hk/cardano-js-sdk/pull/490/commits/ae19731896fdcef2f95627b382ddb63b5a7925dc

packages/cardano-services/src/Metadata/types.ts
IvayloAndonov: Split into a separate breaking change commit: 28ac6c3ccea07afdbdac16f2cf0bfb5b86277064

packages/cardano-services/test/Metadata/HttpMetadataService.test.ts
IvayloAndonov: As I see we still use `dummyLogger` in a lot of places across service unit tests, ex. [here](https://github.com/input-output-hk/cardano-js-sdk/blob/159535092033a745664c399ee1273da436fd3374/packages/cardano-services/test/NetworkInfo/NetworkInfoHttpService.test.ts#L19), should I update the usage of test logger across all tests as a separate commit?

packages/cardano-services/test/Metadata/HttpMetadataService.test.ts
IvayloAndonov: All naming suggestions are applied in f99eb6f441e00d9c6dfd52d59e5a92a7a530317b @rhyslbw

packages/cardano-services/test/Metadata/HttpMetadataService.test.ts
rhyslbw: Agree

packages/core/src/Cardano/types/StakePool/Cip6ExtendedStakePoolMetadata.ts
rhyslbw: > keep the generated ts file as it is with ExtendedStakePoolMetadataFields (represent CIP-6 format) and then have another type APExtendedStakePoolMetadata (AdaPools)?

Sounds reasonable

packages/cardano-services/test/Metadata/HttpMetadataService.test.ts
IvayloAndonov: https://github.com/input-output-hk/cardano-js-sdk/pull/490/commits/eb46deb268607fe1f476d98dd6e72ce3e16caf7a

packages/cardano-services/test/Metadata/HttpMetadataService.test.ts
IvayloAndonov: d267744a4b03ced7a9f0769bc532c3a462b94575

packages/core/src/Cardano/types/StakePool/Cip6ExtendedStakePoolMetadata.ts
IvayloAndonov: The type changes are reverted back for this file as now we have types `ExtendedStakePoolMetadata`(CIP-6) and `APExtendedStakePoolMetadata` (AdaPools).

Updated with 5de0f21818716bc7d42d195ce28b17e7a349e890 (core) and 73fe4b5bb6df49342c35d925671757ca06f40f75 (cardano-services)

packages/cardano-services/src/Metadata/schemas/cip-6.json
```diff
@@ -0,0 +1,276 @@
...
+          "required": [
+            "primary"
+          ],
+          "title": "Pool contact data",
+          "type": "object",
+          "properties": {
+            "primary": {
+              "$id": "#/properties/pool/properties/contact/properties/primary",
+              "default": "email",
+              "description": "the pools prefered communication channel",
```
rhyslbw, line 130:
```suggestion
              "description": "the pools preferred communication channel",
```

rhyslbw (changes requested):
Looking good @IvayloAndonov. Just a minor spelling correction, but you can do it as part of the 
 rebasing and conflict resolution.

packages/cardano-services/src/Metadata/schemas/cip-6.json
IvayloAndonov: Resolved within rebasing: https://github.com/input-output-hk/cardano-js-sdk/pull/490/commits/8e9e0770911c9dadf87267645bed5454881161a9#diff-57ff8f1808766698a7efd474167b54754090d2269b0d0a403610f2a012a91bd7R130

mkazlauskas (changes requested):
I would classify c9295aa8244d5a63c9e3a2d6c632c6b78e4e1ffd as `feat` instead of `refactor` because it's a new feature in `util-dev` package.

Also, could you please explain why 9e7d7ced0713df8b6d9ab8d06758414e0e5ec403 was needed? Having the interface not coupled to db-sync sounds like a better name, because eventually we'll have an alternative provider for this service.

packages/core/src/Cardano/types/StakePool/APExtendedStakePoolMetadata.ts
```diff
@@ -0,0 +1,139 @@
+/**
+ * Based on:
+ * https://a.adapools.org/extended-example
```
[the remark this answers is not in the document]
IvayloAndonov: That was the initial idea of having just one common interface to expose pool extended metadata but it turned out the two formats are incompatible and not all properties are optional.

An example is `pool.contact.primary` from CIP-6 [format](https://raw.githubusercontent.com/cardano-foundation/CIPs/master/CIP-0006/schema.json) is a required one, however, it's missing from AP format so it opens an issue with the mapping to a common ext metadata interface, which is the reason we choose to support both formats.

I put that one in the core package since it is just a sibling of the other CIP-6 interface. @mkazlauskas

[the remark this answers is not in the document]
IvayloAndonov: | could you please explain why https://github.com/input-output-hk/cardano-js-sdk/commit/9e7d7ced0713df8b6d9ab8d06758414e0e5ec403 was needed
Because we introduce new [HttpMetadataService](https://github.com/input-output-hk/cardano-js-sdk/blob/c4eafd09c3d8bac4b0bdbc17849533e5c9a6fede/packages/cardano-services/src/Metadata/types.ts#L7-L11) interface, it's a HTTP one in contrast of the existing `MetadataService` that's why I saw reasonable to add the prefix to distinguish easily between both interfaces (HTTP and DbSync). I decided to not use one service with two methods since if someone needs to use only HTTP then there is no need to provide database dependency as well.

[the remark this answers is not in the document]
IvayloAndonov: If we are not obligated to return "one or another" ext data format then we can define and expose our own common `StakePoolExtendedMetadata` with all fields nullable. Probably worth to be discussed with the Lace team because they have to reconcile this variation, thoughts? @rhyslbw  @mkazlauskas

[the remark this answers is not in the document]
mkazlauskas: What is the alternative though? I really don't like having AP types in core package, and we don't want to butcher CIP6 type to fit AP either, cause then it won't be accurate to the CIP.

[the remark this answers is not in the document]
rhyslbw: Good points. The lack of adoption with CIP-6 probably makes this an obvious direction.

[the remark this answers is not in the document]
mkazlauskas: Also, if we leave it as-is, Lace will have to handle AP metadata with optional fields anyway, so it's just as difficult on their side either way.

packages/cardano-services/src/Metadata/types.ts
```diff
@@ -1,11 +1,24 @@
...
   queryTxMetadataByHashes(hashes: Cardano.TransactionId[]): Promise<Map<Cardano.TransactionId, Cardano.TxMetadata>>;
 }
 
+export interface HttpMetadataService {
```
mkazlauskas, line 7: I think a better location for new metadata types is src/StakePool

packages/core/src/Cardano/types/StakePool/APExtendedStakePoolMetadata.ts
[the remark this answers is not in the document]
teddyjfpender: I'll discuss this with the Lace team to get their input, not a problem. Making our own common type makes the most sense to me as we'll reach a greater amount of users. I can also raise the lack of CIP-6 adoption in the CIP forum and get that conversation started.

packages/cardano-services/src/Metadata/types.ts
IvayloAndonov: Agree, this one is more related to the StakePool context.

> The [interface](https://github.com/input-output-hk/cardano-js-sdk/blob/25bcbef8c529c20211c11589c40bd7f967810bb4/packages/cardano-services/src/Metadata/types.ts#L3) is not coupled to the DB
Yes, so are you saying that having one type

```
interface MetadataService{
   queryTxMetadataByHashes(...)
   getStakePoolExtendedMetadata(...)
}
```
implemented by 2 different services (DbSync and HTTP ) is a preferable way rather than 2 different services, the first located in `/Metadata/` and the second one in `/StakePool/`? And if we have one type then we have to implement both methods by both implementations.

mkazlauskas: > are you saying that having one type:

No, I'm saying to rename the other one back (or to `TxMetadataService`), and the new one to `StakePoolMetadataService` or similar.

> implemented by 2 different services (DbSync and HTTP) is a preferable way rather than having 2 different interfaces

The interface should be source-unaware. The implementation is specific to source, so it can be prefixed with `Http` or `DbSync`.

IvayloAndonov: Thx for the clarification, it's addressed with 3c52cbbefd42d95e2a198155d86450ef3ee4fa7b, where https://github.com/input-output-hk/cardano-js-sdk/commit/9e7d7ced0713df8b6d9ab8d06758414e0e5ec403 becomes redundant and will be squashed as well.

Also, I'll classify https://github.com/input-output-hk/cardano-js-sdk/commit/c9295aa8244d5a63c9e3a2d6c632c6b78e4e1ffd as `feat` instead of `refactor` when rebase and squash the commits. @mkazlauskas

packages/cardano-services/src/Metadata/HttpMetadataService.ts
```diff
@@ -0,0 +1,52 @@
...
+import { ExtMetadataResponse, StakePoolExtMetadataService } from '../StakePool/types';
+import { Logger } from 'ts-log';
+import { ValidationError, validate } from 'jsonschema';
+import { getExtMetadataUrl, getSchemaFormat, loadJsonSchema } from './util';
+import axios, { AxiosInstance } from 'axios';
+
+const HTTP_CLIENT_TIMEOUT = 1 * 1000;
+const HTTP_CLIENT_MAX_CONTENT_LENGTH = 5000;
+
+export const createHttpMetadataService = (
```
mkazlauskas, line 11:
```suggestion
export const createHttpStakePoolExtMetadataService = (
```
And maybe let's move this to src/StakePool?

mkazlauskas (changes requested):

packages/cardano-services/src/Metadata/HttpMetadataService.ts
IvayloAndonov: Yeah, since I moved the `types.ts`, I was thinking to move the HttpMetadataService.ts + schemas folder as well and leaving in `/Metadata` only Txs-related ones ? @mkazlauskas

mkazlauskas: Yep, sounds good

IvayloAndonov: Addressed in https://github.com/input-output-hk/cardano-js-sdk/pull/490/commits/edd4511f91ec1c009fb3bc17466618f44212aff8 @mkazlauskas

mkazlauskas (dismissed):
Looks good! Please squash for a final review

packages/util-dev/package.json
```diff
@@ -53,6 +53,7 @@
   "devDependencies": {
     "@types/jest": "^26.0.24",
     "eslint": "^7.32.0",
+    "get-port-please": "^2.5.0",
```
mkazlauskas, line 56: This should be in `dependencies` since it's used in the library code (`src/`). No need for a fixup for this one.

mkazlauskas (changes requested):

mkazlauskas (dismissed):
Nice work! :rocket:

rhyslbw (approved):
Great work @IvayloAndonov. I've assigned @iadmytro for review too

rhyslbw (approved):

PR #495, opened 2022-10-27
ADP-1858-upgrade-from-csl-to-cml

AngelCastilloB:
# Context

The CSL was compiled to WASM using an old version of rustwasm which had a [bug](https://github.com/rustwasm/wasm-bindgen/issues/1963) that would leak stack space when errors were thrown by the underlying rust library, This was making the CSL WASM module to crash after a several errors were thrown, this was fixed in later versions. The CML library uses the updated rustwasm and as such doesnt have this bug.

# Proposed Solution

Reemplace the CSL library with the CML library


rhyslbw (changes requested):
I had a heart-in-mouth moment :laughing::
image

Did you run `npm i`? Please ensure `yarn install` has been run again after removing the `package-lock.json`

packages/input-selection/src/selectionConstraints.ts
```diff
@@ -28,14 +28,24 @@ export const computeMinimumCost =
   ): EstimateTxFee =>
   async (selection) => {
     const tx = await buildTx(selection);
+
     return usingAutoFree((scope) => {
+      // TODO: Get this from cardano-services
```
mkazlauskas, line 33: Just flagging this to be resolved

packages/wallet/test/hardware/LedgerKeyAgent.test.ts
```diff
@@ -140,7 +140,7 @@ describe('LedgerKeyAgent', () => {
   });
 
   describe('establish, check and re-establish device connection', () => {
-    let deviceConnection: DeviceConnection;
+    let deviceConnection: any;
```
mkazlauskas, line 143: Let's try to figure out why this is needed

mkazlauskas (changes requested):

packages/e2e/package.json
```diff
@@ -65,7 +65,7 @@
     "access": "public"
   },
   "dependencies": {
-    "@cardano-foundation/ledgerjs-hw-app-cardano": "5.0.0",
+    "@cardano-foundation/ledgerjs-hw-app-cardano": "^5.0.0",
```
rhyslbw, line 68: This should be a separate `chore` commit

packages/e2e/src/factories.ts
```diff
@@ -267,14 +266,12 @@ keyManagementFactory.register('ledger', async (params: any): Promise<CreateKeyAg
...
 
-  let deviceConnection: DeviceConnection | null | undefined;
-
   return async (dependencies) => {
     const ledgerKeyAgent = await LedgerKeyAgent.createWithDevice(
       {
         accountIndex: params.accountIndex,
         communicationType: CommunicationType.Node,
-        deviceConnection,
+        deviceConnection: null,
```
rhyslbw, line 274: Separate `test` commit. Was this a bug?

packages/e2e/test/web-extension/wdio.conf.js
```diff
@@ -50,6 +50,6 @@ module.exports = {
             'chromedriver'
         ],
         specs: ['./**/*.spec.ts'],
-        waitforTimeout: 10000
+        waitforTimeout: 30000
```
rhyslbw, line 53: This doesn't belong in the same commit as the CML replacement

rhyslbw (changes requested):
The complete diff looks good, but there's a few distinct changes that have been lumped into a single commit, that need to be separated. I've commented on the ones that stick out to me, but it also appears that we missed including some CSL objects in the managed scope the first time around, so they would be best committed first as `fix`, then the lib replacement can be clean. If the new lib introduces new objects, then that's naturally part of the refactor commit.

packages/e2e/test/web-extension/wdio.conf.js
rhyslbw: The commit message is missing some context and includes redundant information already expressed in the diff. Please have another go at this commit message, with an explanation of why this is needed.

rhyslbw (changes requested):
image

Some Spanglish? :smile: 

reemplaced -> replaced

rhyslbw:
> I split the extra manage here: https://github.com/input-output-hk/cardano-js-sdk/commit/b01944c20e258097ad4e08db12e7492e99502482

Great, that's good to see there was only a small number missed in the original implementation

packages/input-selection/src/selectionConstraints.ts
mkazlauskas: Are we planning to do this TODO in a separate PR?

mkazlauskas (approved):
Nice work!

rhyslbw (approved):
Great work @AngelCastilloB :construction_worker:

mkazlauskas (approved):

PR #496, opened 2022-10-27
feat(cardano-services): Add '/meta' endpoint for deployment information

gytis-ivaskevicius:
# Context
The team keeps having issues with figuring out what is deployed, so I added a little endpoint to help us with that

# Demo
https://preprod-api.dev.lw.iog.io/meta

### DIFF
--- a/packages/cardano-services/src/Http/HttpServer.ts
+++ b/packages/cardano-services/src/Http/HttpServer.ts
@@ -98,6 +98,7 @@ export class HttpServer extends RunnableModule {
       this.app.use(`/${service.slug}`, service.router);
       this.logger.debug(`Using /${service.slug}`);
     }
+
     const servicesHealthCheckHandler = async (req: express.Request, res: express.Response) => {
       this.logger.debug('/health', { ip: req.ip });
       let body: ServicesHealthCheckResponse | Error['message'];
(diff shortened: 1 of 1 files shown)


rhyslbw (changes requested):
Thanks, @gytis-ivaskevicius, and great idea!

1. Please provide a sample of the response, so we have a record of it in the PR. The request from the link is failing so it's not clear to me what's included in the value you're setting.
2. It would be good to formalise the type too, as the frontend can then depend on explicit properties. In this type we could have an `extra` property that can be used by others if our type doesn't cover the particular use case.
3. The commit message is bound to a single package, so it should be scoped like this: `feat(cardano-services): Add '/meta' endpoint for deployment information` (there was a typo too with "endpoing"
4. Lint errors needs fixing for CI to pass.

mkazlauskas (changes requested):
Thanks @gytis-ivaskevicius,
- There are some lint errors (shown in Files Changed). Try `yarn lint --fix` to auto fix fixable problems.
- Missing tests

gytis-ivaskevicius:
Fix the lint errors. Thought I resolved them earlier
Can we skip the tests for this? I am not sure how to run this application locally and I'd probably would need to get DB snapshot and run it locally

gytis-ivaskevicius:
@rhyslbw No type definition was intentional to keep it flexible in case I need to add/remove something.

I don't have it deployed anywhere as of right now but I got this screenshot with the sample response:
image

mkazlauskas:
> I am not sure how to run this application locally and I'd probably would need to get DB snapshot and run it locally

Just `yarn test`. It should do all the setup for you if you have docker installed

rhyslbw:
> No type definition was intentional to keep it flexible in case I need to add/remove something.

This is where we need to discuss the feature and determine a solution that fits all use cases, not just nix :-). In my mental model I have `extra` being the place for ad-hoc information to be set, but concepts such as git revision is general and could be typed to present a better interface for clients.

mkazlauskas (approved):

rhyslbw:
Follow-up work to be completed in ADP-2356

rhyslbw (approved):
Thanks, @gytis-ivaskevicius!

PR #498, opened 2022-10-31
chore: git subrepo pull packages/cardano-services/config

AngelCastilloB:
Update cardano services config subrepo


rhyslbw (approved):
Thanks @AngelCastilloB

PR #499, opened 2022-11-01
chore: bump Node.js nvm version, update package engine definitions

rhyslbw:
# Context
We’re currently on the old LTS, so should bump to the latest (v18) and ensure our package.json engine definitions are consistent and cover all versions up to the next interim major (v19)

# Proposed Solution
Bump to the latest LTS version for development, shift the lower engine bound in each package and root to the latest from v14, and remove the upper bound. Update workflows to use the same v18 version.


packages/e2e/package.json
```diff
@@ -100,7 +103,6 @@
     "@types/bunyan": "^1.8.8",
     "@types/jest": "^28.1.2",
     "@types/lodash": "^4.14.182",
-    "@types/node": "14.18.12",
```
mirceahasegan, line 103: are these being picked up by the root package.json?

mirceahasegan (approved):

DominikGuzei (approved):

IvayloAndonov (dismissed):

packages/e2e/package.json
rhyslbw: Yes, but I now may need to revert this change and just duplicate it in every package, as I think there's a dependency nominating it as a peer dep. Will also look into the CI failure, which isn't failing locally

rhyslbw:
CI failure resolved in https://github.com/input-output-hk/cardano-js-sdk/pull/499/commits/f28f355e2a67e593c9dd38eefb3d6c6196217a54

It was a breaking change introduced in https://github.com/nodejs/node/pull/39987

**TLDR**; The host DNS preferences are now respected, so asserting the response is `127.0.0.1` will fail on systems that prefer IPv6.

https://stackoverflow.com/questions/72648487/node-js-16-17-changed-resolution-of-localhost

packages/e2e/package.json
rhyslbw: >  I think there's a dependency nominating it as a peer dep

I found a solution to this: https://github.com/input-output-hk/cardano-js-sdk/pull/499/commits/7e02986d4156431b4d63ee2748d10863d1482359

rhyslbw:
Added vertical Node.js testing within the nightly workflow in 25d31413e84cae362d4b5f97cd267ca326625751

mkazlauskas (dismissed):
Looks good! :rocket:

.github/workflows/nightly.yaml
```diff
@@ -11,10 +11,11 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-20.04 ]
+        nodejs: [ 14, 16, 18.12.0 ]
```
mkazlauskas, line 14: This [seems to be supported](https://github.com/actions/setup-node#supported-version-syntax)
```suggestion
        nodejs: [ 14, 16, lts/*, latest ]
```

rhyslbw: Yes that does make sense given `engines` has no upper bound. Good suggestion! I'll include it when squashing

mkazlauskas (approved):

mirceahasegan (dismissed):

mirceahasegan (approved):
Nice work @rhyslbw  🚀

PR #502, opened 2022-11-02
Feat/adp 2218 web extension wallet manager

mirceahasegan:
# Context

Create a wallet manager that keeps one wallet active at any time.
It can be used to get an ObservableWallet object in the UI scripts. The observable properties of the wallet object can be subscribed to even before any wallet is created.
Wallet manager can:
1. Activate a new wallet. The wallet object starts emitting values from this active wallet.
2. Deactivate a wallet. The wallet object observables stop emitting values but do not complete.
3. Activate a previously deactivated wallet. The previously associated store is restored.
4. Clear cached store for wallets no longer active.


packages/web-extension/src/walletManager/walletManagerWorker.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  shutdown(): void {
+    this.#hostSubscription.shutdown();
+    this.#api$.complete();
+    this.#walletStores.clear();
+  }
+
+  /** Resolves to the currently active wallet or waits until a wallet is activated. */
+  async wallet(): Promise<ObservableWallet> {
+    const wallet = firstValueFrom(this.#api$.pipe(filter((api) => !!api)));
```
mirceahasegan, line 76: use isNotNil

packages/e2e/test/web-extension/extension/background/cip30.ts
```diff
@@ -3,7 +3,7 @@ import { cip30 } from '@cardano-sdk/web-extension';
 import { logger, walletName } from '../util';
 import { runtime } from 'webextension-polyfill';
 import { cip30 as walletCip30 } from '@cardano-sdk/wallet';
-import { walletReady } from './wallet';
+import { walletReady } from './walletManager';
```
mkazlauskas, line 7: Switching the wallet also needs to switch the wallet for cip30
```ts
walletCip30.createWalletApi(walletManagerWorker.wallet$, ...)
```

mkazlauskas (changes requested):

packages/web-extension/src/walletManager/walletManager.types.ts
```diff
@@ -0,0 +1,41 @@
...
+  /** Deactivate wallet. Wallet observable properties will emit only after a new wallet is {@link activate}ed */
+  deactivate(): Promise<void>;
+  /**
+   * Remove `walletName` associated store only if is inactive.
+   * The store is recreated when wallet is activated again.
+   */
+  clearStore(walletName: string): Promise<void>;
+}
+
+export const walletManagerProperties: RemoteApiProperties<WalletManagerApi> = {
```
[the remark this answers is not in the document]
mirceahasegan: Fixed in [d33fc26](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/d33fc26dddf348c81e1053175c9d25dc42d26208)

packages/web-extension/src/walletManager/walletManagerUi.ts
```diff
@@ -0,0 +1,71 @@
...
+      { baseChannel: walletChannel, properties: observableWalletProperties },
+      dependencies
+    );
+  }
+
+  /** {@link wallet} observable properties will emit from this new wallet */
+  activate(props: WalletManagerActivateProps & { keyAgent: AsyncKeyAgent }): Promise<void> {
+    const { keyAgent, walletName, cardanoServicesUrl } = props;
+    const { logger, runtime } = this.#dependencies;
+    exposeKeyAgent(
```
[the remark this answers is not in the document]
mirceahasegan: Fixed in [0716f53](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/0716f538cf22046d1657cabf6b69b6454464760c)

packages/web-extension/src/walletManager/walletManagerWorker.ts
```diff
@@ -0,0 +1,99 @@
...
+    this.#logger = logger;
+    this.#runtime;
+    this.#hostSubscription = exposeApi(
+      { api$: this.#api$.asObservable(), baseChannel: walletChannel, properties: observableWalletProperties },
+      { logger, runtime }
+    );
+  }
+
+  async activate({ walletName, cardanoServicesUrl }: WalletManagerActivateProps): Promise<void> {
+    if (await this.#isActive(walletName)) {
```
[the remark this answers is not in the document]
mirceahasegan: Fixed in [0716f53](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/0716f538cf22046d1657cabf6b69b6454464760c)

packages/web-extension/src/walletManager/walletManagerWorker.ts
```diff
@@ -0,0 +1,99 @@
...
+    // Key agent is created by UI script on a per wallet unique channel
+    const keyAgent = consumeKeyAgent({ walletName }, { logger: this.#logger, runtime: this.#runtime });
+    const stores = this.#getStores(walletName);
+
+    // Wallet factory is responsible for creating the wallet and the providers based on cardanoServicesUrl
+    const wallet = await this.#walletFactory.create({ cardanoServicesUrl, walletName }, { keyAgent, stores });
+    this.#api$.next(wallet);
+  }
+
+  async deactivate(): Promise<void> {
```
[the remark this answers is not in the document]
mirceahasegan: Fixed in [0716f53](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/0716f538cf22046d1657cabf6b69b6454464760c)

mkazlauskas (changes requested):

packages/web-extension/src/walletManager/walletManagerWorker.ts
```diff
@@ -0,0 +1,103 @@
...
+    return stores;
+  }
+
+  async #isActive(walletName: string): Promise<boolean> {
+    return (await this.#api$.value?.getName()) === walletName;
+  }
+
+  #deactivateWallet(): void {
+    const wallet = this.#api$?.getValue();
+    this.#api$.next(null);
```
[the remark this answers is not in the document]
mirceahasegan: > * Would be better to not emit `null` before emitting new wallet if this is called in `activate`

The reason for emitting null before activating a new wallet is to avoid subscribers getting values from a wallet that is being `shutdown()`, before being switched to the new wallet. `null` is not actually seen by the clients. It's just a way to stop watching the wallet being deactivated.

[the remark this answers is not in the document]
mirceahasegan: Fixed in [6f20bd1](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/6f20bd189c069c7e3254dc989bd1cf3d055a0c4e)

packages/e2e/test/web-extension/extension/background/cip30.ts
mirceahasegan: Fixed here: [5f05e62](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/5f05e625fce3e754df58a7dc75bdac59d534364b)

packages/e2e/test/web-extension/extension/ui.ts
```diff
@@ -68,6 +69,9 @@ document
   .querySelector<HTMLButtonElement>('#clearAllowList')!
   .addEventListener('click', backgroundServices.clearAllowList);
 
+// REVIEW: There might be a problem with this tracker. The background part, using createSupplyDistributionTracker,
```
mirceahasegan, line 72: @mkazlauskas , I would like your opinion on this. Thank you!

mirceahasegan:
* [7e60722](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/7e60722174086b29b276f69de7c301b52f4156cc) - Use `observableWalletName` instead of `walletName` to make it clear we are expecting a user wallet name/identifier (e.g. `My Testnet wallet`)
* [5440c42](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/5440c42df6a0035df4f7ed07aee1ae18d8573a26) - Wallet manager channel and its internal wallet channel are now derived from the `walletName`

packages/e2e/test/web-extension/extension/ui.ts
mkazlauskas: SupplyDistributionTracker is there entirely for demo purposes, so it's not very important. Maybe just leave a comment that it won't switch when you switch the wallet.

packages/e2e/test/web-extension/.env.example
```diff
@@ -19,4 +19,5 @@ STAKE_POOL_PROVIDER_PARAMS='{"baseUrl":"http://localhost:4000/stake-pool"}'
 
 # Test Environment
 NETWORK_ID=0
-MNEMONIC_WORDS="vacant violin soft weird deliver render brief always monitor general maid smart jelly core drastic erode echo there clump dizzy card filter option defense"
+MNEMONIC_WORDS_WALLET1="vacant violin soft weird deliver render brief always monitor general maid smart jelly core drastic erode echo there clump dizzy card filter option defense"
+MNEMONIC_WORDS_WALLET2="slab gorilla reflect display cage aim silver add own arrange crew start female bitter menu inner combine exit swallow bamboo midnight wealth culture picnic"
```
rhyslbw, line 23: You can create two wallets from the same mnemonic. Any reason we can't just iterate the [account index](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/6e45d265aa8db63a787cc64a373aa31ce3a0965a#diff-721f78adca74441c90355e45c0319e02940ab5b27ca53389073a76b0b2967b6aR121)

rhyslbw (changes requested):
Good start @mirceahasegan. There's a couple of minor suggestions, and that challenges the implementation around the coupling with `cardano-services`. 

95831b3111ed6b18d96ba5a3be6016134702ecaf would read better as `chore(e2e): validate web-extension during build` or something like that. i.e say what's changed rather than assert the previous state.

packages/e2e/test/web-extension/extension/background/walletManager.ts
```diff
@@ -0,0 +1,60 @@
...
+  txSubmitProvider,
+  utxoProvider
+} from './config';
+import { logger, walletName } from '../util';
+import { of } from 'rxjs';
+import { runtime } from 'webextension-polyfill';
+
+const walletFactory: WalletFactory = {
+  create: async (
+    { observableWalletName, cardanoServicesUrl: _cardanoServicesUrl }: WalletManagerActivateProps,
```
rhyslbw, line 28: Probably best to omit `cardanoServicesUrl` until it's being used, and capture follow-up work in a ticket.

packages/web-extension/src/walletManager/walletManagerWorker.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    // Key agent is created by UI script on a per wallet unique channel
+    const keyAgent = consumeKeyAgent(
+      { walletName: observableWalletName },
+      { logger: this.#logger, runtime: this.#runtime }
+    );
+    this.#keyAgentSubscription = keyAgent;
+    const stores = this.#getStores(observableWalletName);
+
+    // Wallet factory is responsible for creating the wallet and the providers based on cardanoServicesUrl
```
rhyslbw, line 74: I agree `cardano-services` users can benefit from the convenience of just providing the API root URL, but this would be best done in the `cardano-services-client` package, with some abstract concept here to tie it together, as it's currently limiting the use of the manager to a single provider implementation. Think about the use case of local services, that may use Ogmios directly to submit transactions, and a HTTP server for the other providers.

rhyslbw (changes requested):

packages/e2e/test/web-extension/.env.example
mkazlauskas: Agree, that would be better

packages/e2e/test/web-extension/extension/background/walletManager.ts
mkazlauskas: Or keep it and add a comment. Main goal here is to demonstrate expected usage

packages/web-extension/src/walletManager/walletManagerWorker.ts
mkazlauskas: I suggest to change it to
```ts
// not destructuring
async activate(props: WalletManagerActivateProps): Promise<void> {
...
  await this.#walletFactory.create(props
```

Then WalletManagerActivateProps can become a union type to support other types of providers and this won't need to change.

packages/web-extension/src/walletManager/walletManager.types.ts
```diff
@@ -0,0 +1,34 @@
+import { AsyncKeyAgent } from '@cardano-sdk/key-management';
+import { ObservableWallet, storage } from '@cardano-sdk/wallet';
+
+export interface WalletManagerProps {
+  walletName: string;
+}
+
+export interface WalletManagerActivateProps {
+  observableWalletName: string;
+  cardanoServicesUrl?: string;
```
mkazlauskas, line 10: It's a little easier to reason about if it's typed with non-nullable props
```suggestion
  __typename: ProviderType.CardanoServices, // could omit this for now since it's the only type
  cardanoServicesUrl: string; // this should be a required prop
```

mkazlauskas (changes requested):
Looking good but I'll do 1 final review once current requested changes are addressed

packages/web-extension/src/walletManager/walletManagerWorker.ts
rhyslbw: That implies all provider types are known to the SDK though, which isn't in the spirit of making this pluggable. It would also be good to not couple the convenience setup code to the `web-extension` context, so we can set up a client by passing a URL in other contexts, including Electron or Node.js.

mkazlauskas: > That implies all provider implementations are known to the SDK though

Not necessarily, user can use his own walletFactory by design. However having that in mind, maybe we should change `WalletManagerActivateProps` to `{provider: {type: string; options: unknown;}}` instead of my suggestion [here](https://github.com/input-output-hk/cardano-js-sdk/pull/502#discussion_r1013948707)

> It would also be good to not couple the convenience setup code to the web-extension context...

Agree

packages/web-extension/src/walletManager/walletManager.types.ts
mkazlauskas: Consider suggestion [in this comment](https://github.com/input-output-hk/cardano-js-sdk/pull/502#discussion_r1013999100) instead

packages/web-extension/src/walletManager/walletManagerWorker.ts
rhyslbw: > user can use his own walletFactory 

Good point. Maybe the issue I have is just with the placement of the implementation

> maybe we should change...

Agree

packages/e2e/test/web-extension/.env.example
mirceahasegan: I need a wallet with funds. Otherwise, I can't execute the build and sign transaction step in the test

packages/web-extension/src/walletManager/walletManager.types.ts
mirceahasegan: Done here: [4d319c5](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/4d319c5241651b905ef6fb7b77000a855f60eb96)

packages/e2e/test/web-extension/extension/background/walletManager.ts
mirceahasegan: I have made the comment a bit clearer and the parameter is not destructured anymore.
Done here: [4d319c5](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/4d319c5241651b905ef6fb7b77000a855f60eb96)

mkazlauskas (changes requested):

packages/web-extension/src/walletManager/walletManager.types.ts
```diff
@@ -0,0 +1,44 @@
+import { AsyncKeyAgent } from '@cardano-sdk/key-management';
+import { ObservableWallet, storage } from '@cardano-sdk/wallet';
+
+export interface WalletManagerProps {
+  walletName: string;
+}
+
+export interface WalletManagerActivateProps<P = string, O = unknown> {
```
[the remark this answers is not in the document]
mirceahasegan: It could also be a number coming from an enum...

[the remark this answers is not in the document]
mkazlauskas: Yeah, I'm just not sure if unconstrained type will work for a type union. Maybe it will :sweat_smile: Feel free to just mark this change request as resolved, or maybe do a 'string' | 'number'

packages/web-extension/src/walletManager/walletManagerWorker.ts
```diff
@@ -0,0 +1,129 @@
...
+  #logger: Logger;
+  #runtime: MinimalRuntime;
+
+  activeWallet$: Observable<ObservableWallet>;
+
+  constructor(
+    { walletName }: WalletManagerProps,
+    { walletFactory, storesFactory, logger, runtime }: MessengerDependencies & WalletManagerDependencies
+  ) {
+    this.activeWallet$ = this.#api$.asObservable().pipe(filter(isNotNil));
```
mkazlauskas, line 45: Isn't `asObservable()` redundant as `pipe()` follows?

packages/web-extension/src/walletManager/walletManagerWorker.ts
```diff
@@ -0,0 +1,129 @@
...
+  async deactivate(): Promise<void> {
+    this.#deactivateWallet();
+  }
+
+  shutdown(): void {
+    this.#deactivateWallet();
+
+    this.#hostSubscription.shutdown();
+    this.#api$.complete();
+    this.#walletStores.clear();
```
mkazlauskas, line 89: Stores should only be cleared explicitly
```suggestion
```

packages/e2e/test/web-extension/extension/ui.html
```diff
@@ -17,7 +17,8 @@ <h1>Wallet</h1>
   </div>
   <h2>ADA price</h2>
   <div id="adaPrice">Loading...</div>
-  <button id="createKeyAgent">Create KeyAgent</button>
+  <button id="createWallet1">Create Wallet 1</button>
+  <button id="createWallet2">Create Wallet 2</button>
```
mkazlauskas, line 21:
```suggestion
  <button id="activateWallet1">Activate Wallet 1</button>
  <button id="activateWallet2">Activate Wallet 2</button>
  <button id="destroyWallets">Destroy Wallets</button>
```
Let's add an example of deactivating and clearing the stores too

packages/e2e/test/web-extension/specs/wallet.spec.ts
```diff
@@ -48,12 +57,18 @@ describe('wallet', () => {
...
+      it('can create new wallet and switch to it', async () => {
+        const prevWalletAddr = await $(spanAddress).getText();
+        await $(btnCreateWallet2).click();
+        await expect($(spanAddress)).not.toHaveTextContaining(prevWalletAddr);
+        await expect($(spanAddress)).toHaveTextContaining('addr');
+      });
+      it('can build and sign a transaction using the new wallet', async () => {
+        await buildAndSign();
+      });
+
```
mkazlauskas, line 71: Let's extend the tests to better cover functionality of the wallet manager
```suggestion
it('can switch to another wallet', async () => {
it('can switch back to the first wallet', async () => {
```

packages/e2e/test/web-extension/webpack.config.js
```diff
@@ -8,86 +8,93 @@ const distDir = path.join(__dirname, 'dist');
 require('dotenv').config({ path: path.join(__dirname, '.env') });
 
 module.exports = {
-    devtool: 'inline-source-map',
```
mkazlauskas, line 11: Are there any changes in webpack config?

mkazlauskas (changes requested):
Looking good! Just a few more requested changes :sweat_smile:

packages/e2e/test/web-extension/.env.example
rhyslbw: > I need a wallet with funds

A single mnemonic can create multiple accounts, which is all you need here to simulate multiple wallets. Maybe I'm missing some broader context?

packages/e2e/test/web-extension/extension/background/walletManager.ts
```diff
@@ -25,12 +25,19 @@ import { runtime } from 'webextension-polyfill';
 
 const walletFactory: WalletFactory = {
   create: async (
-    { observableWalletName, cardanoServicesUrl: _cardanoServicesUrl }: WalletManagerActivateProps,
+    props: WalletManagerActivateProps,
     dependencies: { keyAgent: AsyncKeyAgent; stores: storage.WalletStores }
   ) =>
-    // cardanoServicesUrl could be used to construct providers for different networks
+    /* provider could be used to pass the necessary information to construct providers for different networks.
```
rhyslbw: Is this duplication necessary? The type has the documentation, so it's not clear to me why we'd want it here too

packages/e2e/test/web-extension/.env.example
rhyslbw: OK, given you're just using what [already exists](https://github.com/input-output-hk/cardano-js-sdk/tree/master/packages/e2e#local-test-network), this is fine to leave. We probably should just be using a single mnemonic with 5 accounts derived, so we could look at refactoring this later

packages/web-extension/src/walletManager/walletManagerWorker.ts
mirceahasegan: > Stores should only be cleared explicitly

Once the walletManager is shutdown, it cannot be used anymore, so I do not see the point in keeping the stores, but I may be missing something here.

packages/web-extension/src/walletManager/walletManager.types.ts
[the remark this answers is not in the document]
mirceahasegan: Agree. Updated it in [9f664a4](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/9f664a4152ddc7853f8ce6753ba4475f3016187d)

packages/web-extension/src/walletManager/walletManagerWorker.ts
mirceahasegan: [9f664a4](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/9f664a4152ddc7853f8ce6753ba4475f3016187d)

packages/web-extension/src/walletManager/walletManagerWorker.ts
mirceahasegan: Understood. Updated here: [9f664a4](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/9f664a4152ddc7853f8ce6753ba4475f3016187d)

packages/e2e/test/web-extension/extension/background/walletManager.ts
mirceahasegan: It should not be duplicated. I have updated it in here: [9f63efb](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/9f63efb609b9314b08cce9580ec96f9108303122)

packages/e2e/test/web-extension/specs/wallet.spec.ts
mirceahasegan: Added here: [deb1c66](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/deb1c6617b48ac03ce5a59e0065db69a114cebee)

packages/e2e/test/web-extension/extension/ui.html
mirceahasegan: added here: [deb1c66](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/deb1c6617b48ac03ce5a59e0065db69a114cebee)

packages/e2e/test/web-extension/specs/wallet.spec.ts
```diff
@@ -45,15 +58,32 @@ describe('wallet', () => {
...
+      it('can switch back to the first wallet', async () => {
+        await $(btnActivateWallet1).click();
+        await expect($(spanAddress)).toHaveTextContaining(walletAddr1);
+      });
+
+      it('can deactivate the wallet and clear the stores', async () => {
+        await $(destroyWallets).click();
+        await expect($(spanAddress)).toHaveTextContaining('');
```
mkazlauskas, line 84: This assertion probably matches everything
```suggestion
        await expect($(spanAddress)).toHaveText('');
```

mkazlauskas (changes requested):

packages/e2e/test/web-extension/specs/wallet.spec.ts
mirceahasegan: 👍  Fixed in [7f4df72](https://github.com/input-output-hk/cardano-js-sdk/pull/502/commits/7f4df723ba6612846261661237effc7a1500ca80)

mkazlauskas (dismissed):
Great work! :rocket:

rhyslbw (approved):
Well done @mirceahasegan :sunflower:

mkazlauskas (approved):

rhyslbw (approved):

PR #503, opened 2022-11-03
ci: split nightly workflow into 3

mkazlauskas:
# Context

Nightly workflow uses matrix in order to run the tests on with different nodejs versions. Workflow starts by merging in `master` and we only need to do it once (not once per nodejs version).

# Proposed Solution

Split the workflow into 3: `nightly-merge` -> `nightly-test` -> `nightly-publish`

- Make a few small optimizations
- Enable mutable yarn installs on merge step c6e3bd5840c53cb42aecd500d12b0731bc1fc92c


rhyslbw (approved):
Nice work @mkazlauskas  :lotus_position:

PR #506, opened 2022-11-03
test(e2e): load test improvements

iccicci:
# Context

Looking for a good test to stress the transaction submission, we found that the tx submissions load test could be a good candidate, but it needs some fixes/improvements.

- removed the `try/catch` block around the body of the test which was hiding test failures
- coins are now self sent to the same account to avoid the too many inputs error in subsequent test runs (with high `TRANSACTIONS_NUMBER`)

### DIFF
--- a/packages/e2e/env/.env.load-testing
+++ b/packages/e2e/env/.env.load-testing
@@ -1,7 +1,9 @@
 # Logger
-LOGGER_MIN_SEVERITY=debug
+LOGGER_MIN_SEVERITY=info
 
 # Providers setup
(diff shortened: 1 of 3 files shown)


mkazlauskas (approved):
Looks good! :rocket:

rhyslbw (changes requested):
Looks good @iccicci. Please add the description you left here to the commit message, as it's important context to be included in the log, and not just here in a GitHub feature.

iadmytro:
It would be good to add example .env file for this test

rhyslbw (approved):
Nice work @iccicci :sunflower:

PR #508, opened 2022-11-04
fix(nix): fix not found /packages/cardano-services/node_modules

gytis-ivaskevicius:
# Context
Unable to build/push image without this fix

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -76,7 +76,7 @@
             cp -r packages/$p/dist $out/libexec/$sourceRoot/packages/$p/dist
             cp -r packages/$p/package.json $out/libexec/$sourceRoot/packages/$p/package.json
           done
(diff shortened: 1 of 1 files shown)


rhyslbw (changes requested):
The commit type and scope are invalid as per Conventional Commits. Fixes are reserved for software bug fixes, and we're using _scopes_ when changes are constrained to a single package. Any changes to the nix files can most likely be considered `build` type that does not require a scope.

**From**:
`fix(nix): fix not found /packages/cardano-services/node_modules`

**To**:
`build: fix not found /packages/cardano-services/node_modules in nix flake`

More about the types [here](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional#type-enum)

rhyslbw (approved):

mkazlauskas (approved):

PR #509, opened 2022-11-07
refactor: align CardanoNode and TxSubmitProvider as runnable modules

IvayloAndonov:
# Context

Refactoring which enables the switching to a `long-running` WS connection in Ogmios client.

# Note:
We still use a "One-time" WS connection atm due to missing functionality in Ogmios `TxSubmissionClient`.
Once **ADP-2370** is done and this refactoring change is merged, we could proceed with the switching to a "long-running" connection and retry strategy on the web socket closed error when not in SRV mode. (**ADP-2339**)

- rewrite `ogmiosTxSubmitProvider()` method to a class `OgmioxTxSubmitProvider` which extends `RunnableModule` the same as `OgmiosCardanoNode` (both aligned at the end)
- adjust `OgmiosTxSubmitProvider`  everywhere as a runnable module, and manage its lifecycle methods (src and tests) which enables the switching Ogmios client WS connection from “one-time” to “long-running” (calling `this.#txSubmissionClient.shutdown()` in the corresponding shutdown provider lifecycle method)
- the Ogmios mock server implementation is changed in order to support multiple distinct answers and correspond correctly to a long-running connection type (required for the tests)
- adjust and rewrite the problematic/timeout tests after runnable module refactoring
- refactor Ogmios-related errors across `CardanoNode` and `TxSubmitProvider`:
   - classify `TxSubmissionError` as a variant of `CardanoNodeError`
   - reuse Ogmios-related errors across `OgmiosCardanoNode` and `OgmiosTxSubmitProvider`
   - adjust src and tests with the new error namespace
   - generalize some error names


packages/core/src/errors.ts
```diff
@@ -110,3 +121,35 @@ export class InvalidStringError<InnerError = unknown> extends ComposableError<In
...
+  }
+}
+
+export class NotInitializedError extends CustomError {
+  constructor(methodName: string, moduleName: string) {
+    super(`${methodName} cannot be called until ${moduleName} is initialized`);
+  }
+}
+
+export const OgmiosClientErrors = {
```
mkazlauskas, line 140: I think `core` package shouldn't export ogmios-specific types. For now it's ok to re-export types from `@cardano-ogmios/client` as we're already doing that for tx submission errors. I suggest renaming this back to `CardanoNodeErrors` and generalizing `WebSocketClosed` as it's probably too specific to Ogmios. How about `ConnectionError`?

mkazlauskas (changes requested):
3c18b0c35984ab24ed01d7e7083591ef1a86bd64 should have a `test` type

packages/core/src/errors.ts
IvayloAndonov: The idea behind moving the re-exposed Ogmios-related errors to commonplace and renaming it is that some of the errors are shared across [OgmiosCardanoNode](https://github.com/input-output-hk/cardano-js-sdk/pull/509/commits/092820216057f6489e8fc7c1136520075f0b1a99#diff-f39a864229b1800d755f14614cd8496fe47244069e7b95d5ae851156616cf34bR54-R83) and [OgmiosTxSubmitProvider](https://github.com/input-output-hk/cardano-js-sdk/pull/509/commits/092820216057f6489e8fc7c1136520075f0b1a99#diff-b7119a8f737f72567611a7b9e26aba55cab3472da2b8446ce73a6623a0b2ce93R63-R66), for instance `CardanoNodeNotInitializedError` -> `NotInitializedError`

Do you think we can keep the `CardanoNodeErrors` named export for all of them and use it in both Ogmios node and provider? @mkazlauskas

mkazlauskas: Transactions are submitted to CardanoNode so yes, that makes sense. TxSubmissionError could be one of the variants of CardanoNodeError.

IvayloAndonov: Yeah, going to do it in a bit, about the TxSubmissionError I think it could be kept as it is in a separate file where we [compose and re-export](https://github.com/input-output-hk/cardano-js-sdk/blob/9d0f51fe4a3b8ae20c8e83b9209397cd99cc044b/packages/core/src/Cardano/types/TxSubmissionErrors.ts#L10-L84) bunch of TxSubmission errors under it

mkazlauskas: Would be best to have just 1 export, could do a breaking refactor to use the new one under `CardanoNodeErrors.TxSubmissionError`

IvayloAndonov:
The commit 667df25de2e1ad3f13922a033b01467527ad28f7 is not with `test` type.
I also pushed another `refactor!` breaking the commit as suggested, where the feedback from https://github.com/input-output-hk/cardano-js-sdk/pull/509#discussion_r1016332363 is addressed.

I decided to force push due to the commits type/name changes and the PR has still just one review, so now it is in a clean state.
@mkazlauskas

packages/core/src/errors.ts
IvayloAndonov: The change is applied with [93bc87f](https://github.com/input-output-hk/cardano-js-sdk/pull/509/commits/93bc87f78b402cf27c397003b10d7676da5ea543), now the `TxSubmissionError` error is classified as `CardanoNodeError` and the usage is as follows `CardanoNodeErrors.TxSubmissionError`

packages/core/src/CardanoNode/types/CardanoNode.ts
```diff
@@ -30,22 +30,6 @@ export type StakeDistribution = Map<
...
-   *
-   * @throws {CardanoNodeNotInitializedError}
-   */
-  shutdown: () => Promise<void>;
-  /**
-   * Get summaries of all Cardano eras
-   *
-   * @returns {EraSummary[]} Era summaries
-   * @throws {CardanoNodeError}
-   */
```
mkazlauskas, line 48: Please describe this breaking change in the commit message

mkazlauskas (changes requested):

packages/core/src/CardanoNode/types/CardanoNode.ts
IvayloAndonov: Described in 234305e28aefd3d9bd1736315bdf89ca31f7556f @mkazlauskas

mkazlauskas (approved):
Great work! :rocket:

rhyslbw (approved):
Great work @IvayloAndonov

PR #510, opened 2022-11-07
build/no-extraneous-dependencies

rhyslbw:
# Context
We have quite a few issues with the package dependencies, which is not unexpected, given the lack of enforcement and this being such a tedious aspect of maintaining a monorepo with Node.js' inferred dependencies design.

# Proposed Solution
This rule, with the configuration to specify test files, ensures packages used in `src` are listed in the package.json `dependencies`. It will fail if missing or if the dependency is included with `devDependencies`. Development dependencies are checked against the test directories.

There's a single disable as it's complaining about transitive dependency on `@types/pg` missing from `dependencies`. We can see if the feature added in https://github.com/import-js/eslint-plugin-import/pull/2543 handles this scenario once released.
 New similar imports may need to be overruled moving forward, but I only found one case in the whole workspace, so don't expect it to be a frequently occurring situation.


mkazlauskas (dismissed):
Nice, quite a few missing deps added! :muscle: 

I hope the new eslint rule didn't add too much to our lint time. How long did it take to commit?

mirceahasegan (approved):
LGTM 🚀

rhyslbw:
> How long did it take to commit?

No noticeable difference  :lotus_position:

mirceahasegan (dismissed):

PR #512, opened 2022-11-08
fix(cardano-services): Fix some tokens are not appearing in a UTxO & Wallet Balance

IvayloAndonov:
# Context
According to the [explorer](https://preprod.cexplorer.io/address/addr_test1qq3k65epz73c86nmf80r7hsdfff648h06feg37vgny2rz8j8kdqlntpuk5z0usn2s0evcdwydq2gahqhuwp3nfrankssq28lgp/utxo#data), address `addr_test1qq3k65epz73c86nmf80r7hsdfff648h06feg37vgny2rz8j8kdqlntpuk5z0usn2s0evcdwydq2gahqhuwp3nfrankssq28lgp` has 100 units of asset `ee1ce9d7560f48a4ba3867037dbec2d8fed776d94dd6b00a35309073`. (by metadata name: SadCoin)

However, querying [utxo-by-addresses](https://preprod-api.dev.lw.iog.io/utxo/utxo-by-addresses) returns the following for that utxo - this is missing SadCoin

Scenario: When making a request to `utxo-by-addresses` on the `preprod` network the UTxO appears in the transaction details but not all assets appear in the UTxO’s mapping and therefore does not appear in the wallet’s balance either

# Proposed Solution
- fix the UTxO mapping to accept and attach assets without names as valid case (name === '')


rhyslbw (dismissed):
> otherwise, we could keep the unit test as it is, and once the switching to the local network work is finished, the test will be replaced accordingly at that point

This option is best

Great work @IvayloAndonov

mkazlauskas (approved):
Nice work!

rhyslbw (approved):

PR #514, opened 2022-11-08
fix(cardano-services): fixed http provider error handling

iadmytro:
Extended the list of handled connection errors.

### DIFF
--- a/packages/cardano-services-client/src/HttpProvider.ts
+++ b/packages/cardano-services-client/src/HttpProvider.ts
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { Logger } from 'ts-log';
 import { ProviderError, ProviderFailure } from '@cardano-sdk/core';
-import { fromSerializableObject, isConnectionError, toSerializableObject } from '@cardano-sdk/util';
+import { fromSerializableObject, toSerializableObject } from '@cardano-sdk/util';
 import axios, { AxiosAdapter, AxiosRequestConfig } from 'axios';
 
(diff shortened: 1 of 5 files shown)


packages/cardano-services-client/src/HttpProvider.ts
```diff
@@ -110,8 +110,7 @@ export const createHttpProvider = <T extends object>({
               if (mapError) return mapError(typedError, method);
               throw new ProviderError(ProviderFailure.Unknown, typedError);
             }
-            if (mapError) return mapError(null, method);
-            if (isConnectionError(error)) {
+            if (error.request) {
               throw new ProviderError(ProviderFailure.ConnectionFailure, error, error.code);
             }
```
mkazlauskas, line 115: Can we keep `mapError`? I think it would be best for healthCheck to return `{ok: false}` for all errors returned from the server (but throw on connection errors).

We could update [mapHealthCheckError](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services-client/src/mapHealthCheckError.ts#L4) to:
- only throw on connection errors, otherwise assume that server is unhealthy
- do not wrap errors that are already ProviderError
```suggestion
            mapError ||= (err) => err;
            if (error.request) {
              throw mapError(new ProviderError(ProviderFailure.ConnectionFailure, error, error.code));
            }
            return mapError(null, method);
```

mkazlauskas (changes requested):

mkazlauskas (approved):
Nice work!

iccicci (approved):
Nice!
This would also solve ADP-2397

PR #523, opened 2022-11-15
configurable stake pools number in local network

iccicci:
# Context

We are currently in need of having more granular control over the stakepool configurations (pledge, cost, metdata etc) in the local network. This introduces a few changes that allow us to more easily set up and configure stake pools.

# Proposed Solution
Add a reusable design to stake pool set up in the local network (as proposed by @iccicci). Now in order to add a new stake pool you must simply:

Increase the number of total pools that the local network will bootstrap, this is done by modifying the nodes-configuration file:

```
AMOUNT_PER_NODE='10000000000000'
NUM_SP_NODES=11
SP_NODES_ID=$(seq 1 ${NUM_SP_NODES})
SP_NODES=$(for i in ${SP_NODES_ID} ; do echo "node-sp${i}" ; done)
```

And increasing the `NUM_SP_NODES` variable.

After that a new pool will be started by the network, you can now optionally add metadata and change its pool parameters.

To add metadata you just need to add a json file in the `file-server` directory with the name matching the new pool. For example if you increase the `NUM_SP_NODES ` to 12 then you need to add a new file named `SP12.json` in the folder with the following content:

```
{
  "name": "Awesome new Stakepool",
  "ticker": "SP12",
  "description": "Some description etc.",
  "homepage": "https://stakepool1.com"
}
```

To update the actual pool parameters, you must add a new script in the `scripts/pools` folder the script most follow the name convention `update-node-sp12.sh`, example of the script:

```
#!/usr/bin/env bash

# This script updates SP1: This pool has a positive pledge (>0), meets the pledge and has metadata.
set -euo pipefail

here="$(cd "$(dirname "$0")" >/dev/null 2>&1 && pwd)"
root="$(cd "$here/../.." && pwd)"
cd "$root"
export PATH=$PWD/bin:$PATH

# pool parameters
SP_NODE_ID=1
POOL_PLEDGE=400000000
POOL_OWNER_STAKE=500000000 # Must be greater than pledge
POOL_COST=390000000
POOL_MARGIN=0.15
METADATA_URL="http://file-server/SP${SP_NODE_ID}.json"

source ./scripts/pools/update-node-utils.sh

trap clean EXIT

updatePool ${SP_NODE_ID} ${POOL_PLEDGE} ${POOL_OWNER_STAKE} ${POOL_COST} ${POOL_MARGIN} ${METADATA_URL}

```

There you can control all the parameters of the pool (you can take a look at the other pools as an example).

- Added a more flexible schema to adding new pools
- Added 11 pools with different parameters that cover all the cases un the unit test suite
- Added a new e2e test that register two pools
- Added a new utility function to convert pubKeyHash to pool Id.


packages/e2e/local-network/scripts/nodes-configuration.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+AMOUNT_PER_NODE='10000000000000'
+NUM_SPO_NODES=11
```
rhyslbw, line 4:
```suggestion
NUM_SP_NODES=11
```
I know this is not a new variable but let's get this corrected now. SPO is an acronym for Stake Pool Operator.

packages/core/src/Cardano/types/StakePool/primitives.ts
```diff
@@ -12,6 +12,12 @@ export type PoolId = OpaqueString<'PoolId'>;
  */
 export const PoolId = (value: string): PoolId => typedBech32(value, 'pool', 45);
 
+/**
+ * @param {string} value blake2b_224 digest of an operator verification key hash
+ * @throws InvalidStringError
+ */
+export const PoolIdFromKeyHash = (value: string): PoolId => toTypedBech32('pool', value);
```
mkazlauskas, line 19: Let's move this out to where PoolId is defined and make this a property of PoolId. Also, `value` can be narrowed down.
```suggestion
export const PoolId.fromKeyHash = (value: Ed25519KeyHash): PoolId => toTypedBech32('pool', value);
```

packages/core/src/Cardano/util/primitives.ts
```diff
@@ -38,6 +38,15 @@ export const assertIsBech32WithPrefix = (
   }
 };
 
+/**
+ * Converts a hex string into a typed bech32 encoded string.
+ *
+ * @param prefix The prefix of the bech32 string.
+ * @param hexString The hex string to be encoded.
+ */
+export const toTypedBech32 = <T>(prefix: string, hexString: string): T =>
```
mkazlauskas, line 47: We have a type for hex strings, it validates the string.
```suggestion
HexBlob.toTypedBech32 = <T>(prefix: string): T =>
```
You might also want to add
```ts
HexBlob.fromEd25519KeyHash = (hash: Ed25519KeyHash) => hash as unknown as HexBlob;
```

packages/e2e/local-network/scripts/nodes-configuration.sh
mkazlauskas, line 4: Increasing # of nodes might require to also increase # of transactions in delegation rewards test, as it lowers the probability of generating rewards

packages/e2e/src/factories.ts
```diff
@@ -267,6 +267,7 @@ export type GetWalletProps = {
   name: string;
   polling?: PollingConfig;
   stores?: storage.WalletStores;
+  customKeyParams?: any;
```
mkazlauskas, line 270: This could be typed

mkazlauskas (changes requested):
- 148443c7cecd892dacf3474d50f6a091e6746c76 has incorrect scope as it has changes in `e2e` package
- I'd squash d28afa5d7e3aca47b41ebd7bfe3ddedff455784b and 24867d052ffbc6c47b162ee6dd6181357d75e63e

rhyslbw (changes requested):

packages/core/src/Cardano/util/primitives.ts
mkazlauskas: > someHexBlob.toTypedBech32('prefix');

You can't really do that without adding it to `String` prototype and I'm not sure we want to do that because it's quite invasive. I suggest to just add it as a "static" fn to be used like `HexBlob.toTypedBech32(someHexBlob)`

packages/e2e/src/factories.ts
```diff
@@ -260,13 +260,21 @@ export const keyAgentById = memoize(async (accountIndex: number, provider: strin
   return keyManagementFactory.create(provider, params, dummyLogger);
 });
 
+export type KeyAgentFactoryProps = {
+  accountIndex: number;
+  mnemonic: string;
+  networkId: number;
```
mkazlauskas, line 266:
```suggestion
  networkId: Cardano.NetworkId;
```

packages/e2e/local-network/scripts/nodes-configuration.sh
mkazlauskas: Let's keep it as-is, we'll know what to do if we see it failing with 0 rewards

rhyslbw (approved):
Great work @AngelCastilloB

mkazlauskas (approved):
Nice work! :rocket:

PR #526, opened 2022-11-20
test(e2e): wallet init test with new custom measurement and scheduler…

mirceahasegan:
Added `MeasurementUtil` helper class based on node perf hooks. Added `getLoadTestScheduler` util that schedules a given number of calls in a timeframe. Added wallet initialization load test using the two utils.

# Context

Performance/load test libraries are focused on REST/websocket primitives.
This is a POC for custom helpers for load tests.

# Proposed Solution
Solution is made up from 2 parts:
1. `MeasurementUtil` class: a wrapper over node performance hooks. Provides type safe string values for the markers and hides the complexity of matching the strings when measuring.
2. `getLoadTestScheduler` method: a helper for scheduling a number of calls in a given timeframe, using rxjs observables.

See `packages/e2e/test/load-test-custom/wallet-init.test.ts` for a usage example.

Doc comments provide more details.

# TODO:
Once the POC is validated and agreed upon:
- [x] Extend measurementUtil with more stats: min/max/avg. For now it only measures the total time.
- [x] Migrate existing stake pool search artillery.io test

### DIFF
--- a/packages/e2e/src/load-test-scheduler.ts
+++ b/packages/e2e/src/load-test-scheduler.ts
@@ -1,0 +1,35 @@
+import { Logger } from 'ts-log';
+import { Observable, map, mergeMap, take, timer } from 'rxjs';
+
+export interface LoadTestSchedulerProps<T> {
+  /** Time in seconds during which the scheduler calls the `callUnderTest` */
+  duration: number;
+  /** How many times scheduler will call `callUnderTest` in the given `duration` */
+  callsPerDuration: number;
+  /** Each call will have a unique incrementing numeric id. This is the start value. */
+  idOffset?: number;
+  /** Asynchronous call that will be called `callsPerDuration` times, evenly split in `duration` seconds */
+  callUnderTest: (id: number) => Observable<T>;
+}
+
+export interface LoadTestSchedulerDependencies {
+  logger: Logger;
+}
+
+/** Use `LoadTestSchedulerProps` to schedule calls to `callUnderTest` */
+export const getLoadTestScheduler = <T>(
+  { duration, callsPerDuration, idOffset = 0, callUnderTest }: LoadTestSchedulerProps<T>,
+  { logger }: LoadTestSchedulerDependencies
+): Observable<T> =>
+  new Observable<T>((observer) => {
+    const callIntervalMs = (duration * 1000) / callsPerDuration;
+    logger.info(`Running ${callsPerDuration} calls in ${duration}s, 1 call / ${callIntervalMs}ms`);
+
+    return timer(0, callIntervalMs)
+      .pipe(
+        map((id) => id + idOffset),
+        take(callsPerDuration),
+        mergeMap((id) => callUnderTest(id))
+      )
+      .subscribe(observer);
+  });
--- a/packages/e2e/src/measurement-util.ts
+++ b/packages/e2e/src/measurement-util.ts
@@ -1,0 +1,154 @@
+import { PerformanceObserver, performance } from 'perf_hooks';
+
+export interface MarkerDetail<T> {
+  target: T;
+  id: number;
+}
+
+export interface MeasurementData {
+  /** Stats for targets added with {@link addMeasureMarker} */
+  time?: {
+    total: number;
+    min: number;
+    max: number;
+    avg: number;
+  }; // only for measurement markers done with `addMeasureMarker`
+
+  /** The number of times {@link addStartMarker} was called for a target */
+  calls_count: number;
+}
+
+export type MeasurementResults<T extends string> = Record<T, MeasurementData>;
+
+/**
+ * Wrapper over node perf hooks.
+ * T is a user defined string union of the targets that are being
+ * measured (e.g. 'wallet-initialization | input-selection').
+ */
+export class MeasurementUtil<T extends string> {
+  #performanceObserver: PerformanceObserver;
+
+  /**
+   * Starts observing performance measurements.
+   * Must be called before adding markers.
+   */
+  start(): void {
+    this.#performanceObserver = new PerformanceObserver(() => {
+      /* just an empty callback because we don't want to do anything when markers are added */
+    });
+
+    this.#performanceObserver.observe({ buffered: true, entryTypes: ['measure', 'mark'] });
+  }
+
+  /** Stops observing performance measurements and clears previous markers */
+  stop(): void {
+    performance.clearMarks();
+    performance.clearMeasures();
+    this.#performanceObserver.disconnect();
+  }
+
+  /**
+   * Add a start time marker in the timeline.
+   *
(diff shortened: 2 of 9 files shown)


packages/e2e/src/load-test-scheduler.ts
```diff
@@ -0,0 +1,30 @@
...
+  duration: number;
+  callsPerDuration: number;
+  idOffset?: number;
+  callUnderTest: (id: number) => Observable<T>;
+}
+
+export interface LoadTestSchedulerDependencies {
+  logger: Logger;
+}
+
```
mirceahasegan, line 14: Doc comments are missing

packages/e2e/src/load-test-scheduler.ts
```diff
@@ -0,0 +1,30 @@
...
+export const getLoadTestScheduler = <T>(
+  { duration, callsPerDuration, idOffset = 0, callUnderTest }: LoadTestSchedulerProps<T>,
+  { logger }: LoadTestSchedulerDependencies
+): Observable<T> =>
+  new Observable<T>((observer) => {
+    const callIntervalMs = (duration * 1000) / callsPerDuration;
+    logger.info(`Running ${callsPerDuration} calls in ${duration}s, 1 call / ${callIntervalMs}ms`);
```
mirceahasegan, line 21: REVIEW: This is meant to be printed during the test. Is there a better way of printing it?

packages/e2e/src/measurement-util.ts
```diff
@@ -0,0 +1,118 @@
+import { PerformanceObserver, performance } from 'perf_hooks';
+
+interface MeasurementDetail<T> {
+  target: T;
+  id: number;
+}
+
+interface MeasurementData {
+  total: number;
```
mirceahasegan, line 9: Extend with `min/max/avg/count` stats

packages/e2e/test/load-test-custom/wallet-init/wallet-init.test.ts
```diff
@@ -0,0 +1,112 @@
...
+)
+  .pipe(toArray()) // Waiting until all wallets settle to call shutdown for all of them
+  .subscribe({
+    complete: () => {
+      logger.info('disconnecting');
+      showResults();
+      measurementUtil.stop();
+    },
+    // Shutdown all wallets. Remove `toArray` operator to shut them down immediately after being settled
+    next: (wallets) => wallets.map((wallet) => wallet.shutdown())
```
IvayloAndonov, line 111: Is it possible to wrap and reuse the observable shutdown logic across many similar e2e tests? (ex. wallet restoration) if we choose this approach (custom observable test scheduler)?

packages/e2e/src/load-test-scheduler.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    return timer(0, callIntervalMs)
+      .pipe(
+        map((id) => id + idOffset),
+        take(callsPerDuration),
+        mergeMap((id) => callUnderTest(id))
+      )
+      .subscribe(observer);
```
IvayloAndonov, line 29: Do we know how this behaves in comparison with artillery? Is it use a linear progression of calls under test or any other approach to spread the load for the given timeframe/duration? (sorry about my lack of knowledge around rxjs)

packages/e2e/test/load-test-custom/wallet-init/wallet-init.test.ts
mirceahasegan: The shutdown wallet is strictly related to the test needs, but you are right, we should keep an eye out for repeating shutdown sequences and extract them.
In this case it's specific to wallets and this test: wallet need shutdown() and this test intentionally shuts down all wallets at the end of the test (instead of right after settling)

packages/e2e/src/load-test-scheduler.ts
mirceahasegan: This is a simple linear progression.

IvayloAndonov: I'm not sure if it's a linear progression though, because it has `take(callsPerDuration)` which means it sends a static number of requests (VUs) per duration (they do not increase over time), which way is the same as artillery's `arrivalCount`.

To achieve a linear load progression, we must mimic the `rampTo` [rate](https://www.artillery.io/docs/guides/guides/test-script-reference#ramp-up-rate) which is the linear ramp-up phase where the number of new arrivals increases linearly over time. @mirceahasegan

packages/e2e/src/load-test-scheduler.ts
mirceahasegan: Updated in [807d2a8](https://github.com/input-output-hk/cardano-js-sdk/pull/526/commits/807d2a86f40f263f4f62b39e3a61ac5797fd3da7)

packages/e2e/src/load-test-scheduler.ts
mirceahasegan: Temporary used `logger = console` for the test

packages/e2e/src/load-test-scheduler.ts
mirceahasegan: I think you're right. I find the `rampTo` rate a bit confusing regarding to the total sessions started during the test.
In this case, the requirement is to init a number of wallets in a period of time (evenly split imho).
I would leave it like this if there isn't a specific request to have a ramping number of values started each interval.

packages/e2e/test/load-test-custom/wallet-init/wallet-init.test.ts
```diff
@@ -0,0 +1,112 @@
...
+  networkInfoProviderFactory,
+  rewardsProviderFactory,
+  stakePoolProviderFactory,
+  txSubmitProviderFactory,
+  utxoProviderFactory
+} from '../../../src';
+import { waitForWalletStateSettle } from '../../util';
+
+// Example call that creates 100 wallets in 10 seconds:
+// LOAD_TEST_DURATION=10 LOAD_TEST_COUNT_PER_DURATION=100 yarn load-test-custom:wallet-init
```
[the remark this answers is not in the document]
mirceahasegan: Updated here: [0a9afa4](https://github.com/input-output-hk/cardano-js-sdk/pull/526/commits/0a9afa462feed2f52667028456c18e037597c0c5)

packages/e2e/src/load-test-scheduler.ts
IvayloAndonov: Yep, this current load-test-scheduler implementation is fine with the wallet init load test requirements (as we are satisfied with a static total count of wallets to be initialized).

Might be it's a note that in other test cases we definitely would need to extend the scheduler to support different load-spreading options based on the configuration and env variables, same as artillery )

IvayloAndonov (approved):
Great job @mirceahasegan !

IvayloAndonov (approved):

PR #527, opened 2022-11-21
refactor(e2e): improves and centralizes env vars e2e tests configuration

iccicci:
# Context

Each e2e test is using its own custom way to access the environment variables; this is causing lot of code repetition and moreover, confusion when we need to run a test while trying to properly configure it.

# Proposed Solution
Centralized both the configuration and the reading of the configuration through environment variables.


packages/e2e/README.md
```diff
@@ -123,55 +127,6 @@ You can override this port mapping by setting the `FILE_SERVER_PORT` environment
 
 `FILE_SERVER_PORT=9874 docker compose -p local-network-e2e up`
 
-## Load Testing
-
-Cardano services end to end load tests. Please note that you must have several services up before executing the test, to start the environment(from the root):
-
-```bash
-$ cd packages/cardano-services
```
rhyslbw, line 131: This has been omitted from the replaced documentation, so looks to me like we have a gap.

packages/e2e/test/environment.ts
```diff
@@ -1,29 +1,84 @@
 import * as envalid from 'envalid';
 
-export const getEnv = (override: NodeJS.ProcessEnv = {}) =>
+/**
+ * Shared collection of all the environment configuration variables with their constraints
+ */
+const variablesConfiguration = {
```
rhyslbw, line 6:
```suggestion
const validators = {
```
Let's stick with the terminology `envalid` uses

packages/e2e/test/environment.ts
```diff
@@ -1,29 +1,84 @@
 import * as envalid from 'envalid';
 
-export const getEnv = (override: NodeJS.ProcessEnv = {}) =>
+/**
+ * Shared collection of all the environment configuration variables with their constraints
```
rhyslbw, line 4:
```suggestion
 * Shared across all tests
```
We get the intent from the rename to `validators`.

packages/e2e/test/environment.ts
```diff
@@ -1,29 +1,84 @@
...
+ */
+const variablesConfiguration = {
+  ASSET_PROVIDER: envalid.str(),
+  ASSET_PROVIDER_PARAMS: envalid.json({ default: {} }),
```
rhyslbw, line 8: You should be able to define the schema for the expected JSON for any of the `envalid.json` declarations. It would be good to cover that improvement in this PR too.

packages/e2e/test/environment.ts
```diff
@@ -1,29 +1,84 @@
...
+  TX_SUBMIT_HTTP_URL: envalid.str(),
+  TX_SUBMIT_PROVIDER: envalid.str(),
+  TX_SUBMIT_PROVIDER_PARAMS: envalid.json({ default: {} }),
+  UTXO_PROVIDER: envalid.str(),
+  UTXO_PROVIDER_PARAMS: envalid.json({ default: {} }),
+  WORKER_PARALLEL_TRANSACTION: envalid.num()
+} as const;
+
+type Entries<T> = { [K in keyof T]: [K, T[K]] }[keyof T][];
+type VariablesConfiguration = typeof variablesConfiguration;
```
rhyslbw, line 34:
```suggestion
type Validators = typeof validators;
```

packages/e2e/test/environment.ts
```diff
@@ -1,29 +1,84 @@
...
+type Variables = keyof VariablesConfiguration;
```
rhyslbw, line 35:
```suggestion
type Validator = keyof Validators;
```

rhyslbw (changes requested):
Looks good from my initial review @iccicci, I've just highlighted some documentation improvements and a suggestion to better enforce the JSON config options.

packages/e2e/README.md
iccicci: Hi @rhyslbw ,

what was called `Load Testing` is now called `RabbitMQ Load Testing`. I did this change because when we introduced this test was the only load test of the project so there was no ambiguities.
Now we are introducing new load tests, so I thought now having something called `load testing` could be prone to ambiguities.
I applied to this documentation block following changes:

1. renamed
2. moved after the `Local Network` section as that is required in order to run this test.
3. removed all the parts _cut & pasted_ from the `Local Network` section (one of the targets of this PR)

mkazlauskas (dismissed):
Nice work!

iccicci:
Hi @rhyslbw ,

all you suggestions should be addressed in 778c4ca60

packages/e2e/test/environment.ts
```diff
@@ -1,29 +1,144 @@
...
+  accountIndex: number;
+  mnemonic: string;
+  networkId: number;
+  password: string;
+}
+
+const jsonExpected = 'Expected a valid JSON object';
+
+// eslint-disable-next-line complexity
+const keyManagement = makeValidator((value) => {
```
rhyslbw, line 13: I think we're better off using a lib like this https://www.npmjs.com/package/ajv in the custom validators @iccicci, which will keep the complexity in our code low. I actually thought there was support in `envalid` to define a schema for the JSON, but in review, it just has generic types to define.

rhyslbw, line 13:
```suggestion
const keyManagemenParams = makeValidator((value) => {
```
Let's keep variables nice and expressive

packages/e2e/test/environment.ts
```diff
@@ -1,29 +1,144 @@
...
+  if (networkId === undefined) throw new Error('Missing "networkId" property');
+  if (typeof networkId !== 'number') throw new Error('"networkId" property is expected to be a number');
+
+  if (!password) throw new Error('Missing "password" property');
+  if (typeof password !== 'string') throw new Error('"password" property is expected to be a string');
+
+  return parsed;
+});
+
+const provider = makeValidator((value) => {
```
rhyslbw, line 41:
```suggestion
const providerParams = makeValidator((value) => {
```

rhyslbw (changes requested):

packages/e2e/test/environment.ts
iccicci: If I had known that link existed I would have saved a couple of hours!
Thank you very much!

rhyslbw: Good catch @IvayloAndonov, yes, let's stick with what we're already using then @iccicci. Sorry for the run-around

iccicci:
Hi @rhyslbw, @IvayloAndonov, @mkazlauskas, @AngelCastilloB, @iadmytro;

all the suggested changes should be ready to be reviewed again.
Thank you

packages/e2e/.env.example
```diff
@@ -1,11 +1,19 @@
...
-# Providers setup
+# Faucet configuration
 FAUCET_PROVIDER=cardano-wallet
 FAUCET_PROVIDER_PARAMS='{"baseUrl":"http://localhost:8090/v2","mnemonic":"fire method repair aware foot tray accuse brother popular olive find account sick rocket next"}'
+
+# Key management setup - required by getWallet
 KEY_MANAGEMENT_PROVIDER=inMemory
 KEY_MANAGEMENT_PARAMS='{"accountIndex": 0, "networkId": 0, "password":"some_password","mnemonic":"vacant violin soft weird deliver render brief always monitor general maid smart jelly core drastic erode echo there clump dizzy card filter option defense"}'
+#KEY_MANAGEMENT_PARAMS='{"accountIndex": 0, "networkId": 0, "password":"some_password","mnemonic":"slab gorilla reflect display cage aim silver add own arrange crew start female bitter menu inner combine exit swallow bamboo midnight wealth culture picnic"}'
+#KEY_MANAGEMENT_PARAMS='{"accountIndex": 0, "networkId": 0, "password":"some_password","mnemonic":"decorate survey empower stairs pledge humble social leisure baby wrap grief exact monster rug dash kiss perfect select science light frame play swallow day"}'
```
rhyslbw, line 12: Is the duplication of `KEY_MANAGEMENT_PARAMS` intentional?

rhyslbw: Oh, it's documenting the available mnemonics?

iccicci: Yes @rhyslbw , it is.
They are the the 5 mnemonics of the 5 funded accounts. The purpose is to make easier to change between a funded account and another one.

rhyslbw (approved):
Great work @iccicci :mechanical_arm:

IvayloAndonov (approved):
LGTM, nice refactoring!

PR #528, opened 2022-11-21
feat(e2e): wallet restoration load test

IvayloAndonov:
# Context

Given that we’re only testing in isolation for each of our services, we should prioritize the usage of load testing that considers the wallet because this touches on all of the services and satisfies an e2e testing environment.

Similar to **ADP-2240: Writing E2E Load Tests - Read Preview & Mainnet Network Wallet Initialisations** yet the means of generating a key agent is now different as we intend to spoof a wallet with transaction history. 

# Proposed Solution

This approach is based on `artillery` load test engine. (same as [StakePoolSearch](https://github.com/input-output-hk/cardano-js-sdk/blob/0fd40cbf154a8b900720a800dd5436ffc8540cd0/packages/e2e/test/artillery/StakePoolSearch.yml#L1-L24) load test)

## Sample of test report
The given sample creates 5 virtual users in 10 seconds phase ( appr. 1 virtual user every 2 seconds ).
In this testing context the virtual user = wallet. Executed against the local network.
The low numbers are just for better output visualization in the terminal.

Screenshot 2022-11-22 at 18 54 18
_both env variables provided inline to the script take priority over the same variables declared in .env file_


packages/e2e/env/.env.artillery
```diff
@@ -0,0 +1,29 @@
...
+UTXO_PROVIDER_PARAMS='{"baseUrl":"http://localhost:4000/utxo"}'
+STAKE_POOL_PROVIDER=http
+STAKE_POOL_PROVIDER_PARAMS='{"baseUrl":"http://localhost:4000/stake-pool"}'
+
+DB_SYNC_CONNECTION_STRING='postgresql://postgres:doNoUseThisSecret!@localhost:5432/cexplorer'
+OGMIOS_SERVER_URL='ws://localhost:1340/'
+
+# Wallet restoration load test scenario
+ARRIVAL_PHASE_DURATION_IN_SECS=60
+WALLETS_COUNT=20
```
IvayloAndonov, line 29: Or maybe `VIRTUAL_USERS_COUNT` is better as more generic?

mkazlauskas (changes requested):
The test is looking good, just need to rebase and align environment variables with changes from #527

packages/e2e/test/artillery/wallet-restoration/queries.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Query randomized distinct addresses from db associated with users who are staking
```
rhyslbw, line 2:
```suggestion
 * Query randomized distinct addresses from db with an associated stake key.
```
You'd need to join to a delegation record to make the original assertion. Also, let's be precise and avoid using the term "staking". Users register a stake key and delegate their stake :slightly_smiling_face:

packages/e2e/test/artillery/wallet-restoration/queries.ts
```diff
@@ -0,0 +1,14 @@
...
+ */
+export const findAddressesWithRewardsHistory = `
```
rhyslbw, line 4:
```suggestion
export const findAddressesWithRegisteredStakeKey = `
```
... or, update the query to join the rewards table

packages/e2e/test/artillery/wallet-restoration/queries.ts
```diff
@@ -0,0 +1,14 @@
...
+  SELECT * FROM (
+    SELECT DISTINCT txOut.address as address, sa.view as stake_address
+    FROM public.delegation d
+      LEFT JOIN public.tx_out txOut on
+    d.addr_id = txOut.stake_address_id
+      LEFT JOIN public.stake_address sa on
+    txOut.stake_address_id = sa.id
+  ) distinct_addresses
+  ORDER BY RANDOM() LIMIT $1
```
rhyslbw, line 13: The spec asks for:

> Number of transactions > 10 and less than 200

I don't think we actually need to restrict the query to addresses with greater than 10 transactions, but maybe we should just be ordering by tx count, which will make the test more consistent and ensure the significant wallets are selected.

rhyslbw (changes requested):

packages/e2e/test/artillery/wallet-restoration/queries.ts
IvayloAndonov: > You'd need to join to a delegation record to make the original assertion

Not sure I got this, should I change the query somehow? Currently the query uses delegation records and joins with another tables. @rhyslbw

packages/e2e/test/artillery/wallet-restoration/queries.ts
IvayloAndonov: I agree wee don't need to restrict the transactions count per address, might be just the ordering as you suggested @rhyslbw

packages/e2e/src/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './factories';
 export * from './FaucetProvider';
+export * from './util/StubKeyAgent';
```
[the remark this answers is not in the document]
IvayloAndonov: Addressed in [09ebf52](https://github.com/input-output-hk/cardano-js-sdk/pull/528/commits/09ebf5266b0ee99125f1d67283d611828f78d8b2) @mkazlauskas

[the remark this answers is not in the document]
IvayloAndonov: The PR is rebased with the latest master and adjusted with the new env variables changes @mkazlauskas

packages/e2e/test/artillery/wallet-restoration/queries.ts
IvayloAndonov: Updated with ddac4136263343c727ac0b5603e49a35bda614fb

packages/e2e/test/artillery/wallet-restoration/queries.ts
mkazlauskas: I think to simulate something close to real usage we need a good distribution of # of tx'es. There are a few issues with sorting by tx count:
- doesn't easily scale across multiple workers - you need to share some state, or at least know  the worker number/index to do pagination in order to not select the same wallets on every worker.
- it would probably result in addresses that all have the same # of transactions (depends on number of addresses selected of course).

IvayloAndonov: In a real-world scenario is more likely N number of users to want to restore their wallets at the same time with different number of transactions each

rhyslbw: Fair points. If we need a consistent test scenario later on we can always extend this to allow for a consistent pre-test results set.

packages/e2e/test/artillery/wallet-restoration/queries.ts
rhyslbw: > Currently, the query uses delegation records

Oh shoot, sorry, I didn't see the join. You can revert that change, as the original name was valid.

packages/e2e/test/artillery/wallet-restoration/queries.ts
rhyslbw: https://github.com/input-output-hk/cardano-js-sdk/pull/528#discussion_r1032184479

packages/e2e/test/artillery/wallet-restoration/queries.ts
IvayloAndonov: Reverted in https://github.com/input-output-hk/cardano-js-sdk/pull/528/commits/0de3a4e87486464ebb2ab565d864ac5306ff1c9f @rhyslbw

rhyslbw (approved):
Great work @IvayloAndonov :rocket:

mkazlauskas (commented):
I'm not sure about the type of 4c674444ed2be39cb3a0f1e9eceab3bb9a09499a. It has changes that could be considered `feat` in `src/`, but this change doesn't align with commit message. Thoughts @rhyslbw ?

rhyslbw:
> I'm not sure about the type of https://github.com/input-output-hk/cardano-js-sdk/commit/4c674444ed2be39cb3a0f1e9eceab3bb9a09499a. It has changes that could be considered feat in src/, but this change doesn't align with commit message?

Right, there are features being added, but without unit tests to keep the features self-contained. If we're happy the this is only needed to satisfy our own testing, then changing the commit type is sufficient

rhyslbw (approved):

mkazlauskas (approved):
Nice work! :rocket:

PR #529, opened 2022-11-22
test(e2e): removed browser usage outside of tests

mirceahasegan:
# Context
Using `chromedriver` 107.*  in local dev env, observed the following error:
_Unable to load spec files quite likely because they rely on `browser` object that is not fully initialised.
Helper files that use other `browser` commands have to be moved to `before` hook_

# Proposed Solution
Removed unused `ownOrigin` variable.
It was making web-extension tests fail with newer chromedriver versions (107+). The reason was the usage of `browser` like objects outside of tests.


mkazlauskas (approved):

rhyslbw (approved):
Nice work @mirceahasegan 🌻

rhyslbw:
Looks like a genuine CI failure since the workflow has been re-run multiple times
image

PR #530, opened 2022-11-22
db cache invalidation

iccicci:
# Context

We need to ensure DB cache is invalidated on epoch rollover.

# Proposed Solution

Added a test which registers a new stake pool and performs the same stake pool query.
The expected result is constant while in the same epoch and changes on new epochs.


IvayloAndonov (dismissed):
Nice and self-understandable test implementation!

iadmytro (dismissed):

packages/util/package.json
```diff
@@ -51,7 +51,7 @@
     "test:debug": "DEBUG=true yarn test"
   },
   "devDependencies": {
-    "@cardano-sdk/util-dev": "^0.4.1",
+    "@cardano-sdk/util-dev": "^0.5.0",
```
rhyslbw, line 54: @mkazlauskas is it expected we'll need to manually manage sibling deps? :thinking:

packages/e2e/jest.config.js
```diff
@@ -1,51 +1,18 @@
-/* eslint-disable sonarjs/no-duplicate-string */
+const project = (displayName) => ({
```
rhyslbw, line 1: :sunglasses:

rhyslbw (changes requested):
Nice work @iccicci, good to have a real-world demonstration now in an e2e test.

1. As a reminder, it's invalid to use the `fix` commit type when changing tests **only**, as that is reserved for bug fixes, however in the case of dc9b94cd4b1069afa1afad3a2e74d8792e6c8185, it would be better squash into 8830a5509677c49c0852773e187dc85cbc945f1c and make it un-scoped, since it's complementing the other, despite being a test.
2. d383c21f8cd10e900edc5bc3f66437712f137ba7 should be a `test` commit type too.

I'm happy for this to be resolved with a force-push now

packages/util/package.json
mkazlauskas: No, this was probably an error with manual merge

mkazlauskas (changes requested):
Looking good, just need to clean up the commits as @rhyslbw suggested, and also 8830a5509677c49c0852773e187dc85cbc945f1c is a breaking change (commit message has to reflect that)

packages/cardano-services/src/util/polling/DbSyncEpochPollService.ts
```diff
@@ -97,4 +97,8 @@ export class DbSyncEpochPollService implements EpochMonitor {
   async getLastKnownEpoch() {
     return await this.#currentEpoch;
   }
+
+  forceEpochRolloverCheck(): Promise<void> {
+    return this.#executePoll();
+  }
```
mkazlauskas, line 103:
```suggestion
  async #executePoll() {
    const lastEpoch = await this.#queryLastEpoch();
    const currentEpoch = await this.#currentEpoch;
    this.#onEpoch(currentEpoch);
  }
  #onEpoch(currentEpoch) {
     const shouldClearCache = !!(currentEpoch && lastEpoch > currentEpoch);
    if (shouldClearCache) {
      this.#currentEpoch = Promise.resolve(lastEpoch);
      for (const cb of this.#callbacks) {
        cb();
      }
    }
  }
  setEpoch(epoch): Promise<void> {
    return this.#onEpoch(epoch);
  }
```

iccicci: This should be solved in 1a0abce

packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
```diff
@@ -43,14 +46,33 @@ export class DbSyncNetworkInfoProvider extends DbSyncProvider(RunnableModule) im
...
     const tip = await this.#builder.queryLedgerTip();
-    return toLedgerTip(tip);
+    const result = toLedgerTip(tip);
+
+    // Perform computation only on changed tip
+    if (this.#currentHash !== result.hash) {
+      this.#currentHash = result.hash;
+
+      const slotEpochCalc = createSlotEpochCalc(await this.eraSummaries());
+      const currentEpoch = slotEpochCalc(result.slot);
```
mkazlauskas, line 64: `ledgerTip` is the most frequently used endpoint, so it would be nice to optimize this a little bit. I suggest making a change in `core` package to [memoize](https://lodash.com/docs/#memoize) both `createSlotEpochCalc` and `slotEpochCalc`

mkazlauskas (changes requested):

packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
iccicci: Thank you!

iccicci: @mkazlauskas , this should be solved in 54e09fa9c

mkazlauskas (dismissed):
Great work! :rocket:

rhyslbw (approved):
Great work @iccicci

No need to change it here, but just for future reference, "no" is the standard abbreviation for "number", rather than "nr" 40fe90748d03e72724a549dc01cec0821945a819

IvayloAndonov (approved):

PR #531, opened 2022-11-22
ADP-2420 - Fix possible APY calculation overflow on networks with fast epoch

AngelCastilloB:
# Context

There is a possible overflow error when querying the APY of a pool in the local network. This is because the calculation is sensitive to the network speed, this combined with the reduced amount of pools in the network can produce APY which overflow the 'double precision' postgres SQL type.

# Proposed Solution

The calculation was originally done with  'double precision' numeric types, which can overflow, we can cast the types to 'numeric' (which can hold a value up to 131,072 digits before the decimal point and 16,383 digits after the decimal point.), perform the calculation, clamp it to the max value of  'double precision' and then cast the result back.


mkazlauskas (approved):
Nice investigation :muscle:

rhyslbw (approved):

PR #533, opened 2022-11-23
ADP-2330 - feat(key-management): ownSignatureKeyPaths now checks for reward accounts in certificates

AngelCastilloB:
# Context

ownSignatureKeyPaths is currently not checking the type and .rewardAccount of certificates, so it will just blindly sign with stake key any transactions that have any certificates. We need to have some additional logic to determine if a transaction needs to be signed with the stake key.

# Proposed Solution

ownSignatureKeyPaths now inspect all certificates to determine if the transaction must be signed with the wallet stake key.

### DIFF
--- a/packages/key-management/src/restoreKeyAgent.ts
+++ b/packages/key-management/src/restoreKeyAgent.ts
@@ -13,6 +13,7 @@ import {
 import { InMemoryKeyAgent } from './InMemoryKeyAgent';
 import { InvalidSerializableDataError } from './errors';
 import { LedgerKeyAgent } from './LedgerKeyAgent';
+import { STAKE_KEY_DERIVATION_PATH } from './util';
 import { TrezorKeyAgent } from './TrezorKeyAgent';
 
 // TODO: use this type as 2nd parameter of restoreKeyAgent
@@ -51,6 +52,8 @@ export async function restoreKeyAgent<T extends SerializableKeyAgentData>(
   dependencies: KeyAgentDependencies,
   getPassword?: GetPassword
 ): Promise<KeyAgent> {
+  for (const address of data.knownAddresses) address.stakeKeyDerivationPath ||= STAKE_KEY_DERIVATION_PATH;
+
   switch (data.__typename) {
     case KeyAgentType.InMemory: {
       if (!data.encryptedRootPrivateKeyBytes || data.encryptedRootPrivateKeyBytes.length !== 156) {
--- a/packages/key-management/src/util/ownSignatureKeyPaths.ts
+++ b/packages/key-management/src/util/ownSignatureKeyPaths.ts
@@ -4,6 +4,52 @@ import { isNotNil } from '@cardano-sdk/util';
 import uniq from 'lodash/uniq';
 
 /**
+ * Gets whether any certificate in the provided certificate list requires a stake key signature.
+ *
+ * @param rewardAccounts The known reward accounts.
+ * @param certificates The list of certificates.
+ */
+// eslint-disable-next-line complexity
+const isStakingKeySignatureRequired = (
+  rewardAccounts: Cardano.RewardAccount[],
+  certificates: Cardano.Certificate[] | undefined
+  // eslint-disable-next-line sonarjs/cognitive-complexity
+) => {
+  if (!certificates?.length) return false;
+  if (!rewardAccounts?.length) return false;
+
+  for (const account of rewardAccounts) {
+    const stakeKeyHash = Cardano.Ed25519KeyHash.fromRewardAccount(account);
+    const poolId = Cardano.PoolId.fromKeyHash(stakeKeyHash);
+
+    for (const certificate of certificates) {
+      switch (certificate.__typename) {
+        case Cardano.CertificateType.StakeKeyRegistration:
+        case Cardano.CertificateType.StakeKeyDeregistration:
+        case Cardano.CertificateType.StakeDelegation:
+          if (certificate.stakeKeyHash === stakeKeyHash) return true;
+          break;
+        case Cardano.CertificateType.PoolRegistration:
+          // eslint-disable-next-line max-depth
+          for (const owner of certificate.poolParameters.owners) if (owner === account) return true;
+          break;
+        case Cardano.CertificateType.PoolRetirement:
+          if (certificate.poolId === poolId) return true;
+          break;
+        case Cardano.CertificateType.MIR:
+          if (certificate.rewardAccount === account) return true;
+          break;
+        case Cardano.CertificateType.GenesisKeyDelegation:
+        default:
+        // Nothing to do.
+      }
+    }
+  }
+
+  return false;
+};
+
+/**
  * Assumes that a single staking key is used for all addresses (index=0)
  *
  * @returns {AccountKeyDerivationPath[]} derivation paths for keys to sign transaction with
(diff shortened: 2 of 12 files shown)


packages/key-management/src/util/ownSignatureKeyPaths.ts
```diff
@@ -25,12 +71,11 @@ export const ownSignatureKeyPaths = async (
...
 
-  const rewardAccounts = new Set(knownAddresses.map(({ rewardAccount }) => rewardAccount));
+  const rewardAccounts = [...new Set(knownAddresses.map(({ rewardAccount }) => rewardAccount))];
 
-  const isStakingKeySignatureRequired = txBody.certificates?.length;
   if (
-    isStakingKeySignatureRequired ||
-    txBody.withdrawals?.some((withdrawal) => rewardAccounts.has(withdrawal.stakeAddress))
+    isStakingKeySignatureRequired(rewardAccounts, txBody.certificates) ||
+    txBody.withdrawals?.some((withdrawal) => rewardAccounts.includes(withdrawal.stakeAddress))
```
mkazlauskas, line 78: `isStakingKeySignatureRequired` could also check `withdrawals`, as it is also signing with stake key

packages/key-management/src/util/ownSignatureKeyPaths.ts
```diff
@@ -3,6 +3,52 @@ import { Cardano } from '@cardano-sdk/core';
...
 
+/**
+ * Gets whether any certificate in the provided certificate list requires a stake key signature.
+ *
+ * @param rewardAccounts The known reward accounts.
+ * @param certificates The list of certificates.
+ */
+// eslint-disable-next-line complexity
+const isStakingKeySignatureRequired = (
+  rewardAccounts: Cardano.RewardAccount[],
```
mkazlauskas, line 14:
```suggestion
  rewardAccounts: Cardano.RewardAccount,
```
This function should either:
- check only 1 reward account - this is perfectly fine for now since we're always signing with the same 1 stake key
- return stake key paths to sign with

mkazlauskas (changes requested):

packages/key-management/src/util/ownSignatureKeyPaths.ts
mkazlauskas: No reason, just simpler. Supporting multiple stake keys in this function would be nice, but then we need to map those keys to derivation paths to know what to sign with.

packages/key-management/src/restoreKeyAgent.ts
```diff
@@ -51,6 +52,10 @@ export async function restoreKeyAgent<T extends SerializableKeyAgentData>(
   dependencies: KeyAgentDependencies,
   getPassword?: GetPassword
 ): Promise<KeyAgent> {
+  if (data.knownAddresses && Symbol.iterator in data.knownAddresses) {
```
mkazlauskas, line 55: It's a non-nullable array, why is this check needed?
```suggestion
```

packages/key-management/src/util/ownSignatureKeyPaths.ts
```diff
@@ -1,8 +1,64 @@
...
+
+/**
+ * Gets whether any certificate in the provided certificate list requires a stake key signature.
+ *
+ * @param accounts The known reward accounts.
+ * @param txBody The transaction body.
+ */
+// eslint-disable-next-line complexity
+const getStakingKeyPaths = (
+  accounts: Account[],
```
mkazlauskas, line 19: I think it would be slightly cleaner if it was mapped inside this function. Won't need the `Account` type then.
```suggestion
  addresses: GroupedAddress[],
```

packages/key-management/test/restoreKeyAgent.test.ts
```diff
@@ -12,6 +13,11 @@ import {
 import { Cardano } from '@cardano-sdk/core';
 import { InvalidSerializableDataError } from '../src/errors';
 
+export const stakeKeyDerivationPath = {
```
mkazlauskas, line 16:
```suggestion
export const stakeKeyDerivationPath = STAKE_KEY_DERIVATION_PATH;
```

packages/key-management/test/util/ownSignaturePaths.test.ts
```diff
@@ -53,12 +150,116 @@ describe('KeyManagement.util.ownSignaturePaths', () => {
...
+        role: KeyRole.External
+      },
+      {
+        index: 0,
+        role: KeyRole.Stake
+      }
+    ]);
+  });
+
+  it('do not returns stake key derivation path when no certificate with wallet stake key hash is present', async () => {
```
mkazlauskas, line 243:
```suggestion
  it('does not return stake key derivation path when no certificate with wallet stake key hash is present', async () => {
```

packages/key-management/test/restoreKeyAgent.test.ts
```diff
@@ -12,6 +13,11 @@ import {
...
+  index: 0,
+  role: KeyRole.Stake
+};
+
 describe('KeyManagement/restoreKeyAgent', () => {
```
mkazlauskas, line 21:
```suggestion
describe('KeyManagement/restoreKeyAgent', () => {
  it('assumes default stakeKeyDerivationPath if not present in serializable data', () =>
```

mkazlauskas (changes requested):

packages/key-management/src/util/ownSignatureKeyPaths.ts
```diff
@@ -25,14 +81,10 @@ export const ownSignatureKeyPaths = async (
     ).filter(isNotNil)
   ).map(({ type, index }) => ({ index, role: Number(type) }));
 
-  const rewardAccounts = new Set(knownAddresses.map(({ rewardAccount }) => rewardAccount));
+  const rewardAccounts: Account[] = knownAddresses.map(({ rewardAccount, stakeKeyDerivationPath }) => ({
```
mkazlauskas, line 84: Could save some iterations over certificates
```suggestion
import uniqBy from 'lodash/uniqBy';
...
  const rewardAccounts: Account[] = uniqBy(knownAddresses.map(({ rewardAccount, stakeKeyDerivationPath }) => ({
  ...
  })), 'rewardAccount')
```

packages/key-management/src/restoreKeyAgent.ts
mkazlauskas: That's a problem with the test, not the implementation. Just need to add `knownAddresses: []` there

packages/key-management/test/restoreKeyAgent.test.ts
```diff
@@ -49,6 +77,14 @@ describe('KeyManagement/restoreKeyAgent', () => {
     // eslint-disable-next-line unicorn/consistent-function-scoping
     const getPassword: GetPassword = async () => Buffer.from('password');
 
+    it('assumes default stakeKeyDerivationPath if not present in serializable data', async () => {
+      const keyAgent = await restoreKeyAgent(inMemoryKeyAgentDataWithoutStakeDerivationPath, dependencies, getPassword);
+
+      for (const knownAddress of keyAgent.knownAddresses) {
```
mkazlauskas, line 83: Won't catch any issue in this case, but it's a good practice to ensure the assertions are going to be called
```suggestion
       expect.assertions(1);
       for (const knownAddress of keyAgent.knownAddresses) {
```

packages/key-management/src/util/ownSignatureKeyPaths.ts
```diff
@@ -1,7 +1,61 @@
...
+ * @param txBody The transaction body.
+ */
+// eslint-disable-next-line complexity
+const getStakingKeyPaths = (
+  groupedAddresses: GroupedAddress[],
+  txBody: Cardano.NewTxBodyAlonzo
+  // eslint-disable-next-line sonarjs/cognitive-complexity
+) => {
+  const paths: Set<AccountKeyDerivationPath> = new Set();
+  for (const account of groupedAddresses) {
```
mkazlauskas, line 20: uniqBy that I previously suggested is really an implementation detail of this function, so I'd hoist it here
```suggestion
  for (const account of uniq(groupedAddresses.map(({ rewardAccount }) => rewardAccount))) {
```

mkazlauskas (changes requested):

packages/key-management/src/restoreKeyAgent.ts
mkazlauskas: Sure it's possible, but we're trusting that users will comply with our types everywhere else. This is only happening because of `as any`.

packages/key-management/test/restoreKeyAgent.test.ts
mkazlauskas: This should be outside of `for` loop. Typically `expect.assertions` is added as 1st line of the test

mkazlauskas (dismissed):
Nice work! :rocket:

rhyslbw (approved):

PR #534, opened 2022-11-23
fix!: create a new CML scope for every call of BuildTx in selection constraints

mirceahasegan:
# Context
Calls computeMinimumCoinQuantity 4 times for every tx change for each asset. If tx is changed too fast it errors with undefined and then RangeError: offset is out of bounds.

# Proposed Solution
Fix some memory leaks when calling CML.

### DIFF
--- a/packages/input-selection/src/selectionConstraints.ts
+++ b/packages/input-selection/src/selectionConstraints.ts
@@ -4,14 +4,14 @@ import {
   ComputeSelectionLimit,
   EstimateTxFee,
   ProtocolParametersForInputSelection,
+  ProtocolParametersRequiredByInputSelection,
   SelectionConstraints,
   SelectionSkeleton,
   TokenBundleSizeExceedsLimit
 } from './types';
-import { ProtocolParametersRequiredByInputSelection } from '.';
-import { usingAutoFree } from '@cardano-sdk/util';
+import { ManagedFreeableScope, usingAutoFree } from '@cardano-sdk/util';
 
-export type BuildTx = (selection: SelectionSkeleton) => Promise<CML.Transaction>;
+export type BuildTx = (selection: SelectionSkeleton, scope: ManagedFreeableScope) => Promise<CML.Transaction>;
 
 export interface DefaultSelectionConstraintsProps {
   protocolParameters: ProtocolParametersForInputSelection;
@@ -26,10 +26,10 @@ export const computeMinimumCost =
     }: Pick<ProtocolParametersRequiredByInputSelection, 'minFeeCoefficient' | 'minFeeConstant'>,
     buildTx: BuildTx
   ): EstimateTxFee =>
-  async (selection) => {
-    const tx = await buildTx(selection);
+  (selection) =>
+    usingAutoFree(async (scope) => {
+      const tx = await buildTx(selection, scope);
 
-    return usingAutoFree((scope) => {
       // TODO: Get this from cardano-services
       const priceMem = scope.manage(
         CML.UnitInterval.new(scope.manage(CML.BigNum.from_str('577')), scope.manage(CML.BigNum.from_str('10000')))
@@ -48,7 +48,6 @@ export const computeMinimumCost =
       const minFee = scope.manage(CML.min_fee(tx, linearFee, exInitPrices));
       return BigInt(minFee.to_str());
     });
-  };
 
 export const computeMinimumCoinQuantity =
   (coinsPerUtxoByte: ProtocolParametersRequiredByInputSelection['coinsPerUtxoByte']): ComputeMinimumCoinQuantity =>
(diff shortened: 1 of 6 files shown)


packages/core/src/Cardano/util/address.ts
```diff
@@ -4,7 +4,13 @@ import { CML } from '../../CML/CML';
 /**
  * Validate input as a Cardano Address from all Cardano eras and networks
  */
-export const isAddress = (input: string): boolean => CML.Address.is_valid(input);
+export const isAddress = (input: string): boolean => {
+  try {
+    return CML.Address.is_valid(input);
```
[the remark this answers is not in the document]
mirceahasegan: The throw happens when the memory has become invalid (unexpected null pointer, etc..).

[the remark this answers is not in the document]
mirceahasegan: Yes, it became unusable due to the corrupted memory. Maybe we shouldn't encapsulate it if that is the only scenario where it throws. I'll let @mkazlauskas chip in, as there may be more scenarios 😄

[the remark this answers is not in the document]
mkazlauskas: Agree, let's drop the commit

mkazlauskas (changes requested):
If we can't reproduce in unit tests, maybe let's add a web-extension test in e2e package

packages/core/src/Cardano/util/address.ts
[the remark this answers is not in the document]
mirceahasegan: dropped the commit and force pushed

packages/input-selection/src/selectionConstraints.ts
```diff
@@ -55,10 +55,14 @@ export const computeMinimumCoinQuantity =
...
     usingAutoFree((scope) =>
       BigInt(
-        CML.min_ada_required(
-          coreToCml.txOut(scope, output),
-          scope.manage(CML.BigNum.from_str(coinsPerUtxoByte.toString()))
-        ).to_str()
+        scope
+          .manage(
+            CML.min_ada_required(
+              scope.manage(coreToCml.txOut(scope, output)),
```
rhyslbw, line 61: We've had the convention of managing any objects within the `coreTo*` translators. Can it not be done using the scope passed in?

packages/input-selection/src/selectionConstraints.ts
```diff
@@ -70,7 +74,7 @@ export const tokenBundleSizeExceedsLimit =
     }
     return usingAutoFree((scope) => {
       const value = scope.manage(CML.Value.new(cmlUtil.maxBigNum));
-      value.set_multiasset(coreToCml.tokenMap(scope, tokenBundle));
+      value.set_multiasset(scope.manage(coreToCml.tokenMap(scope, tokenBundle)));
```
rhyslbw, line 77: Here too

packages/core/src/CML/coreToCml/coreToCml.ts
```diff
@@ -90,11 +90,12 @@ export const tokenMap = (scope: ManagedFreeableScope, map: Cardano.TokenMap) =>
 
 export const value = (scope: ManagedFreeableScope, { coins, assets }: Cardano.Value): Value => {
   const result = scope.manage(Value.new(scope.manage(BigNum.from_str(coins.toString()))));
+
   if (!assets) {
     return result;
   }
   if (assets.size > 0) {
-    result.set_multiasset(tokenMap(scope, assets));
+    result.set_multiasset(scope.manage(tokenMap(scope, assets)));
```
rhyslbw, line 98: I'm a bit confused about why this changes the behavior when the returned object is already being managed here: https://github.com/input-output-hk/cardano-js-sdk/blob/a29639715f47d7d5cab7d7aa667581bb1030620f/packages/core/src/CML/coreToCml/coreToCml.ts#L63

packages/input-selection/src/selectionConstraints.ts
rhyslbw: OK, I can see it's already managed internally here: https://github.com/input-output-hk/cardano-js-sdk/blob/a29639715f47d7d5cab7d7aa667581bb1030620f/packages/core/src/CML/coreToCml/coreToCml.ts#L112

:thinking:

mkazlauskas: > We've had the convention of managing any objects within the coreTo* translators. Can it not be done using the scope passed in?

Good point and yes the `txOut` and `tokenMap` scope.manage in this PR seems redundant. Then the only functional change in this PR seems to be `scope.manage(CML.min_ada_required(`. @mirceahasegan could you please test to verify?

packages/input-selection/src/selectionConstraints.ts
```diff
@@ -87,14 +90,15 @@ const getTxSize = (tx: CML.Transaction) => tx.to_bytes().length;
...
-    const tx = await buildTx(selectionSkeleton);
-    const txSize = getTxSize(tx);
-    if (txSize <= maxTxSize) {
-      return selectionSkeleton.inputs.size;
-    }
-    return selectionSkeleton.inputs.size - 1;
-  };
+  (selectionSkeleton) =>
+    usingAutoFree(async (scope) => {
+      const tx = scope.manage(await buildTx(selectionSkeleton, scope));
```
rhyslbw, line 95:
```suggestion
      const tx = await buildTx(selectionSkeleton, scope);
```

rhyslbw (changes requested):

packages/input-selection/src/selectionConstraints.ts
mirceahasegan: [b5fcf86](https://github.com/input-output-hk/cardano-js-sdk/pull/534/commits/b5fcf86fb2f79e599e3731b79f3038023e43f679)

rhyslbw (dismissed):

rhyslbw (approved):

mkazlauskas (dismissed):
[BigNumu](https://github.com/input-output-hk/cardano-js-sdk/pull/534/commits/3b9ec98e06ea85bd2274258fae195dfeb9039406) :grin:

mkazlauskas (dismissed):

rhyslbw (approved):

mkazlauskas (approved):

PR #538, opened 2022-11-24
feat: complete ogmios tx to core mapping

IvayloAndonov:
# Context

Currently, only certificates are mapped. We need to map the rest of the transactions.

Since transaction types are supposed to be backward-compatible, this should be straightforward. Properties not available in every epoch should be `nullable?`.

Search the code for any other era-specific types and see if they can be generalized


IvayloAndonov:
In favor of https://github.com/input-output-hk/cardano-js-sdk/pull/540

PR #539, opened 2022-11-25
build(installPhase): fix node_modules copying to resulting derivation

gytis-ivaskevicius:
# Context
Resulting derivation did not had node_modules

# Proposed Solution

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -80,14 +80,13 @@
 
           mkdir -p $out/libexec/$sourceRoot $out/bin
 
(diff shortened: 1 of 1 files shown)


mkazlauskas (approved):

mkazlauskas (changes requested):
We're using package name as commit message scope. Please amend the message to `build: fix node_modules copying to resulting nix derivation`

gytis-ivaskevicius:
This change is no longer required. Also, I'm kinda surprised that it's not conflicting :thinking:

PR #541, opened 2022-11-28
fix(cardano-services): updated APY sql

iadmytro:
WIP

# Context
Formula for APY calculation -    APY = ( 1 + i_nom/N )^N - 1
Division by N inside brackets was missing in implementation

### DIFF
--- a/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/queries.ts
+++ b/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/queries.ts
@@ -346,6 +346,10 @@ export const findPoolEpochRewards = (limit?: number) => `
   ORDER BY epoch_no desc
 `;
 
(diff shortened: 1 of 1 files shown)


packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/queries.ts
```diff
@@ -363,7 +367,14 @@ export const findPoolAPY = (limit?: number) => `
           POWER(
             (
               1 + (
-                avg_daily_roi.avg_roi * (epochs.epoch_length / 86400000)
+               COALESCE( 
+                 avg_daily_roi.avg_roi * (epochs.epoch_length / 86400000) / 
+                   COALESCE(
+                     365 / NULLIF(epochs.epoch_length / 86400000, 0), 
+                     0
```
mkazlauskas, line 374: Wouldn't this be division by zero?

PR #542, opened 2022-11-28
feat: Rename era-specific types

IvayloAndonov:
# Context

Rename all era-specific types in `core` package

- `TxAlonzo` -> `HydratedTx`
- `TxBodyAlonzo` -> `HydratedTxBody`
- `TxIn` -> `HydratedTxIn`

- `NewTxAlonzo` -> `Tx`
- `NewTxBodyAlonzo` -> `TxBody`
- `NewTxIn` -> `TxIn`

### DIFF
--- a/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/DbSyncChainHistoryProvider.ts
+++ b/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/DbSyncChainHistoryProvider.ts
@@ -53,7 +53,7 @@ export class DbSyncChainHistoryProvider extends DbSyncProvider() implements Chai
     addresses,
     pagination,
     blockRange
-  }: TransactionsByAddressesArgs): Promise<Paginated<Cardano.TxAlonzo>> {
+  }: TransactionsByAddressesArgs): Promise<Paginated<Cardano.HydratedTx>> {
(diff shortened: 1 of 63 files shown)


rhyslbw (changes requested):
Looks good @IvayloAndonov, but the change is breaking, so need the correct annotation in the commit message.

mkazlauskas (commented):
Looks good! Although now I'm thinking that since we're changing this, might as well consider:
- `NewTx` -> `Tx`, as this type is aligned with ledger type
- `Tx` -> `ExtendedTx?`, not sure about the name, but this type has some extra properties that ChainHistoryProvider resolves (notably `address` on `TxIn`), but is not present in serialized types.

Thoughts @rhyslbw ?

rhyslbw:
> Tx -> ExtendedTx?, not sure about the name, but this type has some extra properties that ChainHistoryProvider resolves (notably address on TxIn), but is not present in serialized types.

`HydratedTx` ?

IvayloAndonov:
All naming suggestions are applied with the fixups, ready for review @rhyslbw @mkazlauskas

mkazlauskas (approved):
Good to squash!

rhyslbw (approved):

PR #544, opened 2022-11-29
feat: add opaque numeric types to core package

AngelCastilloB:
# Context

Opaque types add type safety and protect against using an invalid kind of value .

We should apply the same principle to ‘number’ types (OpaqueNumber) where it makes sense - where units matter (e.g. second/millisecond).

# Proposed Solution

Add the OpaqueNumber type and apply them where they make sense.

- Added new OpaqueNumber type
- Previously decelerated numeric types in the core package are now of type OpaqueNumber


packages/core/src/Cardano/types/StakePool/StakePool.ts
```diff
@@ -1,13 +1,15 @@
...
 import { PoolIdHex } from './primitives';
 import { PoolParameters } from './PoolParameters';
 import { TransactionId } from '../Transaction';
 
 /**
  * Within range [0; 1]
  */
-export type Percent = number;
+export type Percent = OpaqueNumber<'Percent'>;
+export const Percent = (value: number): Percent => value as unknown as Percent;
```
mkazlauskas, line 12: Would be nice to validate `value` range here

yarn-project.nix
```diff
@@ -1315,7 +1315,6 @@ cacheEntries = {
 "fs-minipass@npm:2.1.0" = { filename = "fs-minipass-npm-2.1.0-501ef87306-1b8d128dae.zip"; sha512 = "1b8d128dae2ac6cc94230cc5ead341ba3e0efaef82dab46a33d171c044caaa6ca001364178d42069b2809c35a1c3c35079a32107c770e9ffab3901b59af8c8b1"; };
 "fs.realpath@npm:1.0.0" = { filename = "fs.realpath-npm-1.0.0-c8f05d8126-99ddea01a7.zip"; sha512 = "99ddea01a7e75aa276c250a04eedeffe5662bce66c65c07164ad6264f9de18fb21be9433ead460e54cff20e31721c811f4fb5d70591799df5f85dce6d6746fd0"; };
 "fsevents@npm:2.3.2" = { filename = "fsevents-npm-2.3.2-a881d6ac9f-97ade64e75.zip"; sha512 = "97ade64e75091afee5265e6956cb72ba34db7819b4c3e94c431d4be2b19b8bb7a2d4116da417950c3425f17c8fe693d25e20212cac583ac1521ad066b77ae31f"; };
-"fsevents@patch:fsevents@npm%3A2.3.2#~builtin<compat/fsevents>::version=2.3.2&hash=18f3a7" = { filename = "fsevents-patch-3340e2eb10-8.zip"; sha512 = "edbd0fd80be379c14409605f77e52fdc78a119e17f875e8b90a220c3e5b29e54a1477c21d91fd30b957ea4866406dc3ff87b61432d2840ff8866b309e5866140"; };
```
mkazlauskas, line 1318: That's odd, no changes in package.json :thinking:

mkazlauskas (changes requested):

packages/core/src/Cardano/types/StakePool/StakePool.ts
mkazlauskas: Do you remember which tests? I guess APY could be >100% in local network :thinking: Anyway, please update the doc comment then

yarn-project.nix
rhyslbw: Please commit this separately @AngelCastilloB . Maybe a prior change failed to include the lockfile update

rhyslbw (approved):
Great work @AngelCastilloB :european_castle:

mkazlauskas (approved):
Nice work! :rocket:

PR #547, opened 2022-11-29
fix(web-extension): close key agent channel on walletManagerUi deacti…

mirceahasegan:
# Context

WalletManagerUI opens a keyAgent channel when activating a wallet, to give WalletManagerWorker access to the keyAgent instance.
On wallet deactivate, it does not close the channel. Opening same channel multiple times could lead to issues.

# Proposed Solution

Close the keyAgent exposed API channel every time a wallet is deactivated and when WalletManagerUi is shutdown.
Also added unit tests for WalletManagerUi since its code became more complex than just a simple forwarder of calls.


mirceahasegan:
Rebased on master

rhyslbw (approved):
Nice work @mirceahasegan, tests read well :muscle:

PR #549, opened 2022-11-30
fix(input-selection): fixed recursive use of CSL object

AngelCastilloB:
# Context

1. The CML WASM module is throwing "Error: recursive use of an object detected which would lead to unsafe aliasing in rust", This can happen when two segments of code try to borrow the same object.
The function that is throwing is using a constant which is a CML object of type BigNum, this object is then reused everytime the function is called.
2. We suspect that asm.js variant of CML is sometimes going out of memory.

# Proposed Solution

1. Create a new managed BigNum on every call and pass it to the Rust function.
2. Update test extension to use WASM variant of CML. As a positive consequence we managed to reduce bundle sizes by allowing webpack to do tree-shaking.


rhyslbw (changes requested):
Good catch @AngelCastilloB. Can you add a test to cover the bug being fixed? Also, I wonder if there are other cases of this bug. Have you done a more widespread review?

rhyslbw (changes requested):
How about converting the `cmlUtil.maxBigNum` into a function instead (which makes it a real utility)? It would need the scope passed in.

packages/wallet/test/SingleAddressWallet/load.test.ts
```diff
@@ -165,6 +165,82 @@ const assertWalletProperties2 = async (wallet: ObservableWallet) => {
...
+  return utxos;
+};
+
+/**
+ * Gets the asset list and the total amount of lovelace on a utxo set.
+ *
+ * @param utxos The utxos.
+ * @returns The total lovelace and the aggregated asset set.
+ */
+const getAssets = (utxos: Cardano.Utxo[]) => {
```
rhyslbw, line 228:
```suggestion
const getAssetsFromUtxos = (utxos: Cardano.Utxo[]) => {
```

Better if this was sourced in the `util` package

rhyslbw (changes requested):

packages/wallet/test/SingleAddressWallet/load.test.ts
rhyslbw: This is actually not possible due to the current dependency `core` has on `util`, so disregard this suggestion. We can look at hoisting when the opportunity arises.

rhyslbw (dismissed):

mkazlauskas (dismissed):
Nice work!

mkazlauskas (approved):

PR #550, opened 2022-12-01
build: Resolve circular references and add guard step in CI

IvayloAndonov:
# Context

Circular references can result in errors where exported objects are undefined. It has happened a few times already, so we should set up some tooling to guard against it in CI.

# Proposed Solution
Screenshot 2022-12-05 at 15 21 41

# Changes introduced
- Use `npx madge --circular` package to find circular dependencies across our project.
- Resolve all circular dependencies across packages in our repo
- Add `npm` script executable in our CI as a guard step


packages/cardano-services-client/package.json
```diff
@@ -39,6 +39,7 @@
     "module-fixup": "shx cp ../../build/cjs-package.json ./dist/cjs/package.json && cp ../../build/esm-package.json ./dist/esm/package.json",
     "tscNoEmit": "shx echo typescript --noEmit command not implemented yet",
     "cleanup:dist": "shx rm -rf dist",
+    "circular-deps:check": "npx madge --circular dist",
```
rhyslbw, line 42: Any reason we can't install as a `devDependency`?

rhyslbw (changes requested):
I think it's better to squash these two as `build: guard against circular dependencies`, otherwise, the changes in isolation do not have the context to show how you determined them to be circular. Adding to CI is how we enforce this change :+1:

packages/cardano-services-client/package.json
IvayloAndonov: No reason, it's now added as dev dep: bed9d46206e197323ece359fbda5ce82bf4f0f0e

IvayloAndonov:
> I think it's better to squash these two as `build: guard against circular dependencies`

The 2 commits are squashed into one `build` type as suggested. @rhyslbw

rhyslbw (approved):

IvayloAndonov:
Commits are squashed. Ready for final review

IvayloAndonov:
Rebased on master with the CI fix.

packages/core/package.json
```diff
@@ -55,6 +56,7 @@
     "@types/lodash": "^4.14.182",
     "eslint": "^7.32.0",
     "jest": "^28.1.3",
+    "madge": "^5.0.1",
```
mirceahasegan, line 59: Out of curiosity, couldn't this have been a devDependency at the root level, since it's needed by all projects?

mirceahasegan (approved):

packages/core/package.json
rhyslbw: It must be defined locally if addressing the binary in `scripts`

rhyslbw (approved):

PR #552, opened 2022-12-02
Feat/adp 1861 add tests for single address wallet shutdown

AngelCastilloB:
# Context

Not cleaning up the SingleAddressWallet properly has already caused issues twice. We need a unit test that collect all wallet Subjects (including in nested objects) and assert that they are all completed after calling wallet.shutdown()

# Proposed Solution

Write an unit tests that verifies all subscriptions to observables and services are cleaned up after shutdown.

Fixed an issue where the Rewards service provider was not being shutdown when the wallet shutdown method was invoked.

### DIFF
--- a/packages/key-management/src/util/createAsyncKeyAgent.ts
+++ b/packages/key-management/src/util/createAsyncKeyAgent.ts
@@ -1,7 +1,7 @@
 import { AsyncKeyAgent, KeyAgent } from '../';
 import { BehaviorSubject } from 'rxjs';
 
-export const createAsyncKeyAgent = (keyAgent: KeyAgent): AsyncKeyAgent => {
+export const createAsyncKeyAgent = (keyAgent: KeyAgent, onShutdown?: () => void): AsyncKeyAgent => {
   const knownAddresses$ = new BehaviorSubject(keyAgent.knownAddresses);
   return {
     async deriveAddress(derivationPath) {
@@ -16,6 +16,7 @@ export const createAsyncKeyAgent = (keyAgent: KeyAgent): AsyncKeyAgent => {
     knownAddresses$,
     shutdown() {
       knownAddresses$.complete();
+      onShutdown?.();
     },
     signBlob: keyAgent.signBlob.bind(keyAgent),
     signTransaction: keyAgent.signTransaction.bind(keyAgent)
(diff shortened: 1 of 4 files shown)


mkazlauskas (approved):
Nice work! :rocket:

packages/key-management/src/util/createAsyncKeyAgent.ts
```diff
@@ -1,7 +1,7 @@
 import { AsyncKeyAgent, KeyAgent } from '../';
 import { BehaviorSubject } from 'rxjs';
 
-export const createAsyncKeyAgent = (keyAgent: KeyAgent): AsyncKeyAgent => {
+export const createAsyncKeyAgent = (keyAgent: KeyAgent, shutdownSpy?: () => void): AsyncKeyAgent => {
```
rhyslbw, line 4:
```suggestion
export const createAsyncKeyAgent = (keyAgent: KeyAgent, onShutdown?: () => void): AsyncKeyAgent => {
```
Here's a way to ensure we're not introducing code explicitly for tests.

rhyslbw (changes requested):

packages/key-management/src/util/createAsyncKeyAgent.ts
rhyslbw: Done in https://github.com/input-output-hk/cardano-js-sdk/pull/552/commits/91b242127a566e9f052ca1fcdb9229e0c550e861

packages/key-management/src/util/createAsyncKeyAgent.ts
```diff
@@ -16,6 +16,7 @@ export const createAsyncKeyAgent = (keyAgent: KeyAgent): AsyncKeyAgent => {
     knownAddresses$,
     shutdown() {
       knownAddresses$.complete();
+      if (onShutdown) onShutdown();
```
mkazlauskas, line 19: Alternative syntax (no need to change)
```suggestion
      onShutdown?.();
```

mkazlauskas (dismissed):

packages/key-management/src/util/createAsyncKeyAgent.ts
rhyslbw: Ah yeah, I just blindly did a rename refactor 🙈  I'll update it and squash at the same time

mkazlauskas (approved):

rhyslbw (approved):

PR #553, opened 2022-12-05
Feat health check enhancement

iccicci:
# Context

Not ok health check response may not give enough information about the reason of the failed check.

# Proposed Solution

Added `projectedTip` to the `healthCheck` response of the `Provider`s extending `DbSyncProvider`.

- Since the `findLastBlockNo` query used by `DbSyncProvider.healthCheck` becomes as the `findLedgerTip` from `DbSyncNetworkInfoProvider.queryLedgerTip`; the second one was removed to reduce code repetition.
- The same changed was applied the the relative `LedgerTipModel` interface.

### DIFF
--- a/packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/NetworkInfoBuilder.ts
+++ b/packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/NetworkInfoBuilder.ts
@@ -1,12 +1,6 @@
-import {
-  ActiveStakeModel,
-  CirculatingSupplyModel,
-  EpochModel,
-  LedgerTipModel,
-  ProtocolParamsModel,
-  TotalSupplyModel
-} from './types';
+import { ActiveStakeModel, CirculatingSupplyModel, EpochModel, ProtocolParamsModel, TotalSupplyModel } from './types';
 import { Cardano } from '@cardano-sdk/core';
+import { LedgerTipModel, findLedgerTip } from '../../util/DbSyncProvider';
 import { Logger } from 'ts-log';
 import { Pool, QueryResult } from 'pg';
 import Queries from './queries';
@@ -45,7 +39,7 @@ export class NetworkInfoBuilder {
 
   public async queryLedgerTip() {
     this.#logger.debug('About to query the ledger tip');
-    const result: QueryResult<LedgerTipModel> = await this.#db.query(Queries.findLedgerTip);
+    const result: QueryResult<LedgerTipModel> = await this.#db.query(findLedgerTip);
     return result.rows[0];
   }
 
--- a/packages/cardano-services/src/util/DbSyncProvider/DbSyncProvider.ts
+++ b/packages/cardano-services/src/util/DbSyncProvider/DbSyncProvider.ts
@@ -1,5 +1,5 @@
-import { BlockNoModel, DB_BLOCKS_BEHIND_TOLERANCE, DB_MAX_SAFE_INTEGER, findLastBlockNo } from './util';
-import { CardanoNode, HealthCheckResponse, Provider, ProviderError, ProviderFailure } from '@cardano-sdk/core';
+import { Cardano, CardanoNode, HealthCheckResponse, Provider, ProviderError, ProviderFailure } from '@cardano-sdk/core';
+import { DB_BLOCKS_BEHIND_TOLERANCE, DB_MAX_SAFE_INTEGER, LedgerTipModel, findLedgerTip } from './util';
 import { Pool } from 'pg';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -25,13 +25,20 @@ export const DbSyncProvider = <
     public async healthCheck(): Promise<HealthCheckResponse> {
       try {
         const { ok, localNode } = await this.cardanoNode.healthCheck();
-        const lastBlockNo = (await this.db.query<BlockNoModel>(findLastBlockNo)).rows[0].block_no;
+        const tip = (await this.db.query<LedgerTipModel>(findLedgerTip)).rows[0];
         const isHealthy =
-          ok && lastBlockNo >= (localNode?.ledgerTip?.blockNo ?? DB_MAX_SAFE_INTEGER) - DB_BLOCKS_BEHIND_TOLERANCE;
+          ok && tip.block_no >= (localNode?.ledgerTip?.blockNo ?? DB_MAX_SAFE_INTEGER) - DB_BLOCKS_BEHIND_TOLERANCE;
+
+        const projectedTip: Cardano.Tip = {
+          blockNo: Cardano.BlockNo(tip.block_no),
+          hash: Cardano.BlockId(tip.hash.toString('hex')),
+          slot: Cardano.Slot(Number(tip.slot_no))
+        };
 
         return {
           localNode,
-          ok: isHealthy
+          ok: isHealthy,
+          projectedTip
         };
       } catch (error) {
         throw new ProviderError(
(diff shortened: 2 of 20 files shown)


packages/cardano-services/src/util/DbSyncProvider/DbSyncProvider.ts
```diff
@@ -25,13 +25,20 @@ export const DbSyncProvider = <
     public async healthCheck(): Promise<HealthCheckResponse> {
       try {
         const { ok, localNode } = await this.cardanoNode.healthCheck();
-        const lastBlockNo = (await this.db.query<BlockNoModel>(findLastBlockNo)).rows[0].block_no;
+        const tip = (await this.db.query<LedgerTipModel>(findLedgerTip)).rows[0];
         const isHealthy =
-          ok && lastBlockNo >= (localNode?.ledgerTip?.blockNo ?? DB_MAX_SAFE_INTEGER) - DB_BLOCKS_BEHIND_TOLERANCE;
+          ok && tip.block_no >= (localNode?.ledgerTip?.blockNo ?? DB_MAX_SAFE_INTEGER) - DB_BLOCKS_BEHIND_TOLERANCE;
+
```
rhyslbw, line 31: Let's add a `debug` log message here to reveal the two values used to produce the `isHealthy` value.

packages/core/test/CardanoNode/mocks.ts
```diff
@@ -48,11 +48,18 @@ export const mockStakeDistribution: StakeDistribution = new Map([
...
+  hash: Cardano.BlockId('70de6ec951af0a47b000d0f2909270896d6c3d3de3951c96b0979164941bb15f'),
+  slot: Cardano.Slot(8450)
+};
+
 export const healthCheckResponseMock = (opts?: {
   blockNo?: number;
   slot?: number;
   hash?: string;
   networkSync?: Cardano.Percent;
+  withTip?: boolean;
```
rhyslbw, line 62: Do we have scenarios where the tip is not needed in the response?

rhyslbw (commented):
Nice work with the refactor @iccicci

packages/core/src/Provider/Provider.ts
```diff
@@ -8,6 +8,7 @@ export type HealthCheckResponse = {
     ledgerTip?: Tip;
     networkSync?: Percent;
   };
+  projectedTip?: PartialBlockHeader;
```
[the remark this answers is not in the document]
iccicci: I chosen `PartialBlockHeader` (which is actually an alias for `Cardano.Tip`), to avoid ambiguity with `import { Tip } from '@cardano-ogmios/schema';`.
I'll try with an explicit `Cardano.` before `Tip`.

packages/core/test/CardanoNode/mocks.ts
iccicci: Yes, I had to add it in order to distinguish tests which was expecting a result with tip from those whose don't.

packages/cardano-services/test/Program/services/ogmios.test.ts
```diff
@@ -282,7 +282,7 @@ describe('Service dependency abstractions', () => {
             headers: { 'Content-Type': APPLICATION_JSON }
           });
           expect(res.status).toBe(200);
-          expect(res.data).toEqual(healthCheckResponseMock());
+          expect(res.data).toEqual(healthCheckResponseMock({ withTip: true }));
```
IvayloAndonov, line 285: I think `withTip: true` could be a default behavior of `healthCheckResponseMock()` rather than this repetition, since now `DbSyncProvider's` [healthCheck](https://github.com/input-output-hk/cardano-js-sdk/pull/553/files#diff-4737374e2ac27a76e30bc8c8858b7890aee4d442cffc3fe03ce9e9ebe77b0247R25-R42) contains the `projectedTip` and it's reused across all providers (they extend it) @iccicci

packages/cardano-services/test/cli.test.ts
```diff
@@ -202,13 +207,13 @@ describe('CLI', () => {
               { env: {}, stdio: 'pipe' }
             );
 
-            await assertServiceHealthy(apiUrl, ServiceNames.Asset);
-            await assertServiceHealthy(apiUrl, ServiceNames.ChainHistory);
-            await assertServiceHealthy(apiUrl, ServiceNames.NetworkInfo);
-            await assertServiceHealthy(apiUrl, ServiceNames.StakePool);
+            await assertServiceHealthy(apiUrl, ServiceNames.Asset, false, true);
```
IvayloAndonov, line 210: probably same here

packages/core/src/Provider/Provider.ts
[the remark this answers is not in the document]
rhyslbw: Yes, that would be better semantically

packages/core/test/CardanoNode/mocks.ts
rhyslbw: :thinking: Which tests are those? I must be missing when the scenario of not having projected tip information is valid.

iccicci: @rhyslbw , my guess is: the tests which use some test `Provider` implementation rather than real `Provider`s...
but I need to check this in order to provide a real answer.

rhyslbw: Ah, yes!

packages/cardano-services/src/util/DbSyncProvider/DbSyncProvider.ts
iccicci: Now I remember: I thought something similar, but I didn't because `DbSyncProvider` has no `logger` (we already spoken about this and the outcome was that we should refactor/standardize `Provider`s dependencies types in order to add the `logger` here).

Once said that... are we sure that's actually required as the core of this PR is to forward exactly the same data in the output JSON?

rhyslbw: Please create a ticket to capture the refactor + addition of debug logging

packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/NetworkInfoBuilder.ts
```diff
@@ -45,7 +39,7 @@ export class NetworkInfoBuilder {
 
   public async queryLedgerTip() {
     this.#logger.debug('About to query the ledger tip');
-    const result: QueryResult<LedgerTipModel> = await this.#db.query(Queries.findLedgerTip);
+    const result: QueryResult<LedgerTipModel> = await this.#db.query(findLedgerTip);
```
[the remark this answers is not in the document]
iccicci: No actions taken here as both the comment suggest to postpone this

packages/core/src/Provider/Provider.ts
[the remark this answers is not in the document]
iccicci: ea59ed748 should solve this

packages/cardano-services/src/util/DbSyncProvider/DbSyncProvider.ts
iccicci: ADP-2492 created

packages/cardano-services/test/Program/services/ogmios.test.ts
iccicci: ea59ed748 and 0576735d8 should solve this

IvayloAndonov (approved):
Nice work!

rhyslbw (approved):

IvayloAndonov:
There is a type in the commit [title](https://github.com/input-output-hk/cardano-js-sdk/pull/553/commits/58d9bbba58085e30fca86e8e0439ab5b476b3709): "test(ogmios): upgrades tests to be **cmpliant** with health check response"

I think this could be squashed into the previous 0576735d817ea462bb6c955692f31fdd2e613210 commit with no scope.

rhyslbw:
Good catch @IvayloAndonov. This can be squashed into a single commit IMO

rhyslbw (approved):

IvayloAndonov (approved):

PR #558, opened 2022-12-10
Feat/adp 2441 spending transaction to byron address

AngelCastilloB:
# Context

There was a small bug in the SDK that was causing transactions with Byron era addresses as targets for outputs to fail when being de-serialized. Constructing, signing and submitting a transaction to the blockchain with the SDK involves serializing and de-serializing the transaction at several spots, which means, constructing transaction with Byron addresses was impossible.

# Proposed Solution

Fix the de-serialization of transaction to account for Byron addresses.

- Transaction can now include Byron addresses as output targets.
- Added a new unit test to validate deserialization of outputs with Byron addresses.
- Added a new e2e test to validate that outputs can be spend to Byron addresses.
- Fixed a small memory leak when parsing Byron addresses.

### DIFF
--- a/packages/e2e/test/wallet/SingleAddressWallet/byron.test.ts
+++ b/packages/e2e/test/wallet/SingleAddressWallet/byron.test.ts
@@ -1,0 +1,57 @@
+/* eslint-disable sonarjs/no-duplicate-string */
+import { Cardano } from '@cardano-sdk/core';
+import { SingleAddressWallet, buildTx } from '@cardano-sdk/wallet';
+import { assertTxIsValid } from '../../../../wallet/test/util';
+import { filter, firstValueFrom, map, take } from 'rxjs';
+import { getEnv, walletVariables } from '../../environment';
+import { getLogger, getWallet } from '../../../src';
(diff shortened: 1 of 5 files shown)


packages/e2e/test/wallet/SingleAddressWallet/byron.test.ts
```diff
@@ -0,0 +1,56 @@
...
+        {
+          address: Cardano.Address('5oP9ib6ym3Xc2XrPGC6S7AaJeHYBCmLjt98bnjKR58xXDhSDgLHr8tht3apMDXf2Mg'),
+          value: {
+            coins: 3_000_000n
+          }
+        }
+      ])
+    };
+
+    const unsignedTx = await wallet.initializeTx(txProps);
```
mirceahasegan, line 37: I would propose using the txBuilder because new users searching for examples on how to build transactions will use this as a starting point.

mirceahasegan (dismissed):
Added only one small testing suggestion which is not related to the actual scope.
Great work @AngelCastilloB! 👍

mirceahasegan (approved):
LGTM 🚀

PR #560, opened 2022-12-11
Feat/key agent chain

mirceahasegan:
# Context

Currently there is an observableWalletName prop in the wallet manager. It is used as a messaging channel name and it’s quite confusing to use, as there is also a ‘walletName’ in cip30, which has a different purpose. We can get rid of this prop and make it easier to use by having a chainId property for the KeyAgent (see https://cips.cardano.org/cips/cip34/) - then we can compute channel name by combining:
- root public key hash
- chainId

# Proposed Solution
- Replace KeyAgent.networkId with chainId and use it to derive messaging channel
- WalletManager now creates an internal walletId for each activated wallet, based on networkId, networkMagic and hash of root public key.
- observableWalletName does not have to be unique anymore, it is just the user configured name for the activated wallet.
- Fixed a bug where stores were not destroyed on WalletManager.destroy().

BREAKING_CHANGES
- removed WalletManagerApi.clearStore
- added WalletManager.destroy() which deactivates the active wallet and
  deletes its store


packages/e2e/test/web-extension/extension/ui.ts
```diff
@@ -126,7 +129,7 @@ const createWallet = async (accountIndex: number) => {
     logger
   });
 
-  // TODO: remove observableWalletName here
```
mirceahasegan: @mkazlauskas , I left the observableWalletName because it will be the user configured name. However, it is not used as unique ID anymore.

packages/web-extension/src/walletManager/util.ts
```diff
@@ -7,6 +10,19 @@ export const walletChannel = (walletName: WalletManagerProps['walletName']) =>
...
+ * Compute a unique walletId from the keyAgent chainId and the root public key hash.
+ * `networkId-networkMagic-blake2bHashOfExtendedAccountPublicKey`
+ */
+export const getWalletId = async (keyAgent: AsyncKeyAgent): Promise<string> => {
+  const { networkId, networkMagic } = await keyAgent.getChainId();
+  const extendedAccountPublicKey = await keyAgent.extendedAccountPublicKey();
+  const pubKey = Buffer.from(extendedAccountPublicKey, 'hex');
+  const pubKeyHash = blake2b(64).update(pubKey).digest('hex');
+
+  return `${networkId}-${networkMagic}-${pubKeyHash}`;
```
mirceahasegan, line 27: We can discuss whether this is the best way to build the ID. It is kind of lengthy, and the `extendedAccountPublicKey` and its hash have the same length

packages/web-extension/src/walletManager/walletManagerWorker.ts
```diff
@@ -58,19 +59,20 @@ export class WalletManagerWorker implements WalletManagerApi {
...
+    const { walletId } = props;
+    if (this.#isActiveWallet(walletId)) {
       return;
     }
     this.#deactivateWallet();
+    this.#activeWalletId = walletId;
 
-    // Key agent is created by UI script on a per wallet unique channel
+    // Key agent is created by UI script on a per wallet unique channel derived from the keyAgent
     const keyAgent = consumeKeyAgent(
```
mirceahasegan, line 70: We could open a single keyAgent channel for all wallets and wait here for the keyAgent with the rightID. This way we could stop having to manage the keyAgent channel open/close cycle.

mirceahasegan:
I have to rebase and force push because a test from master if failing with these changes

packages/web-extension/src/walletManager/util.ts
rhyslbw: It's semantically nice. If length is an issue we could truncate the public key hash to 7 characters or so, but probably best to keep it in original form until there's a reason to think about shortening.

packages/web-extension/src/walletManager/walletManagerWorker.ts
rhyslbw: Seems like a good improvement. How complicated would that be to implement?

mirceahasegan: It is an easy change.

packages/key-management/src/types.ts
```diff
@@ -171,14 +171,16 @@ export interface KeyAgent {
 
 export type AsyncKeyAgent = Pick<KeyAgent, 'deriveAddress' | 'derivePublicKey' | 'signBlob' | 'signTransaction'> & {
   knownAddresses$: Observable<GroupedAddress[]>;
+  getChainId(): Promise<Cardano.ChainId>;
+  extendedAccountPublicKey(): Promise<Cardano.Bip32PublicKey>;
```
mkazlauskas, line 175: We should align the naming (there's also `chainId` -> `getChainId` etc.). I think it's better to go with a new name for async method instead of using the name of the synchronous variant.
```suggestion
  getExtendedAccountPublicKey(): Promise<Cardano.Bip32PublicKey>;
```

packages/web-extension/src/keyAgent/util.ts
```diff
@@ -6,6 +6,7 @@ export const keyAgentChannel = (walletName: string) => `${walletName}$-keyAgent`
 export const keyAgentProperties: RemoteApiProperties<AsyncKeyAgent> = {
   deriveAddress: RemoteApiPropertyType.MethodReturningPromise,
   derivePublicKey: RemoteApiPropertyType.MethodReturningPromise,
+  extendedAccountPublicKey: RemoteApiPropertyType.MethodReturningPromise,
```
mkazlauskas, line 9: I think this belongs in f905f2a72d8c0707bd18cd53d49d41503cb12ff7

packages/web-extension/test/walletManager/util.test.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  // Avoid using the same hashing logic as production.
+  // Used `b2sum` to get blake2b hash from pubKey.
+  const keyHashFromB2Sum =
+    // eslint-disable-next-line max-len
+    '1e73bcd61983f86626359b11a437bfc5e1a85ca3db90c79ea02f93dab150ad3f2ddf2e2f1c123b5d353827576a2a758487a6df59fcc47e17a28bfc62fd6098eb';
+
+  it('creates unique walletId from chainId and root public key hash', async () => {
+    const expectedWalletId = `${chainId.networkId}-${chainId.networkMagic}-${keyHashFromB2Sum}`;
+    expect(await getWalletId(mockKeyAgent)).toEqual(expectedWalletId);
```
mkazlauskas, line 27: I don't think this test has much value as it's very tightly coupled with implementation (basically duplicates the code, just tests the hashing algorithm used). How about calling the function with different combinations of networkId/networkMagic/pubKey and expecting the returned strings to be unique?

packages/e2e/test/web-extension/extension/ui.ts
mkazlauskas: Maybe let's rename it to better reflect the purpose. How about `friendlyName`?

mkazlauskas (changes requested):

packages/web-extension/src/keyAgent/util.ts
mirceahasegan: I will extract this before squashing.

packages/key-management/src/types.ts
mirceahasegan: Agree. Updated name here: [99ac5b5](https://github.com/input-output-hk/cardano-js-sdk/pull/560/commits/99ac5b5ca12d3e19dc63ca767c00e19e54371b59)

packages/web-extension/test/walletManager/util.test.ts
mirceahasegan: Good point. Updated tests to verify that walletId is unique when any of the networkId/networkMagic/publicKey changes: [7fb0dd5](https://github.com/input-output-hk/cardano-js-sdk/pull/560/commits/7fb0dd528e1c2d3b9d0eb4556d3f7e35ffe08f48)

packages/web-extension/src/walletManager/util.ts
mirceahasegan: Done: [7fb0dd5](https://github.com/input-output-hk/cardano-js-sdk/pull/560/commits/7fb0dd528e1c2d3b9d0eb4556d3f7e35ffe08f48)

packages/key-management/test/InMemoryKeyAgent.test.ts
```diff
@@ -232,11 +259,11 @@ describe('InMemoryKeyAgent', () => {
                 .as_bytes()
             ).toString('hex')
           ),
-          getPassword: jest.fn().mockResolvedValue(Buffer.from('nMmys*X002')), // daedelus enforces min length of 10
-          knownAddresses: [],
-          networkId: Cardano.NetworkId.testnet
+          getPassword: jest.fn().mockResolvedValue(Buffer.from('nMmys*X002')),
+          // daedelus enforces min length of 10
```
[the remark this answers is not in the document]
mirceahasegan: [9d24965](https://github.com/input-output-hk/cardano-js-sdk/pull/560/commits/9d24965a790c6a73895eac74a10487686074f044)

packages/e2e/test/web-extension/extension/ui.ts
mirceahasegan: For the moment, let's keep observableWalletName as a way of identifying the wallets in the logs. The walletId is calculated internally and may be difficult to follow logs if are coming from wallets on the same network.

packages/core/src/util/slotCalc.ts
```diff
@@ -38,12 +38,13 @@ export const testnetEraSummaries: EraSummary[] = [
...
-export type EraSummariesMap = { [key in CardanoNetworkMagic]: EraSummary[] };
+export type EraSummariesMap = { [key in NetworkMagics]: EraSummary[] };
 
 export const eraSummariesConfig: EraSummariesMap = {
-  [CardanoNetworkMagic.Mainnet]: mainnetEraSummaries,
-  [CardanoNetworkMagic.Testnet]: testnetEraSummaries
-};
+  [NetworkMagics.Mainnet]: mainnetEraSummaries,
+  [NetworkMagics.Testnet]: testnetEraSummaries
+  // TODO: add missing era summaries
```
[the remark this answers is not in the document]
mirceahasegan: [1852610](https://github.com/input-output-hk/cardano-js-sdk/pull/560/commits/185261090771094ed866e1fbf19bd19ab15d2bef): Moved  `testnetEraSummaries` to util-dev package. 
Removed mainnetEraSummaries

packages/core/src/Cardano/ChainId.ts
```diff
@@ -16,5 +19,34 @@ export enum CardanoNetworkMagic {
...
+  networkMagic: NetworkMagic;
+}
+
+/**
+ * Common Cardano ChainIds
+ */
+export const ChainIds = {
+  LegacyTestnet: {
+    networkId: NetworkId.testnet,
+    networkMagic: NetworkMagics.Testnet
```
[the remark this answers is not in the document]
mirceahasegan: Updated all enums to use titlecase to match GenesisData.networkId type: [2565cd4](https://github.com/input-output-hk/cardano-js-sdk/pull/560/commits/2565cd47f7a95c6d2b579d5d8ba841f08a0c349a)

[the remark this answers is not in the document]
mkazlauskas: I would categorize this commit as `refactor!`

packages/web-extension/src/keyAgent/util.ts
mirceahasegan: done

mkazlauskas (approved):
Nice work! :rocket:

PR #561, opened 2022-12-12
feat(wallet): adds a retry strategy to single address wallet

iccicci:
# Context

Currently the wallet is retrying operations even the raised error can't be tried.

# Proposed Solution

Added a _filter_ to `coldObservableProvider` retry strategy.

Added a `SingleAddressWallet.error$` observable which emits those errors that can't be retried.

### DIFF
--- a/packages/wallet/src/SingleAddressWallet/SingleAddressWallet.ts
+++ b/packages/wallet/src/SingleAddressWallet/SingleAddressWallet.ts
@@ -162,6 +162,7 @@ export class SingleAddressWallet implements ObservableWallet {
   readonly protocolParameters$: TrackerSubject<Cardano.ProtocolParameters>;
   readonly genesisParameters$: TrackerSubject<Cardano.CompactGenesis>;
   readonly assets$: TrackerSubject<Assets>;
+  readonly fatalError$: Subject<unknown>;
   readonly syncStatus: SyncStatus;
   readonly name: string;
   readonly util: WalletUtil;
@@ -244,6 +245,10 @@ export class SingleAddressWallet implements ObservableWallet {
       )
     );
 
+    this.fatalError$ = new Subject();
+
+    const onFatalError = this.fatalError$.next.bind(this.fatalError$);
+
     this.name = name;
     const cancel$ = connectionStatusTracker$.pipe(
       tap((status) => (status === ConnectionStatus.up ? 'Connection UP' : 'Connection DOWN')),
@@ -256,6 +261,7 @@ export class SingleAddressWallet implements ObservableWallet {
       minPollInterval: pollInterval,
       provider$: coldObservableProvider({
         cancel$,
+        onFatalError,
         provider: this.networkInfoProvider.ledgerTip,
         retryBackoffConfig
       }),
@@ -279,6 +285,7 @@ export class SingleAddressWallet implements ObservableWallet {
       coldObservableProvider({
         cancel$,
         equals: deepEquals,
+        onFatalError,
         provider: this.networkInfoProvider.eraSummaries,
         retryBackoffConfig,
         trigger$: eraSummariesTrigger.pipe(tap(() => 'Trigger request era summaries'))
@@ -301,6 +308,7 @@ export class SingleAddressWallet implements ObservableWallet {
       coldObservableProvider({
         cancel$,
         equals: isEqual,
+        onFatalError,
         provider: this.networkInfoProvider.protocolParameters,
         retryBackoffConfig,
         trigger$: epoch$
@@ -311,6 +319,7 @@ export class SingleAddressWallet implements ObservableWallet {
       coldObservableProvider({
         cancel$,
         equals: isEqual,
+        onFatalError,
         provider: this.networkInfoProvider.genesisParameters,
         retryBackoffConfig,
         trigger$: epoch$
@@ -327,6 +336,7 @@ export class SingleAddressWallet implements ObservableWallet {
       inFlightTransactionsStore: stores.inFlightTransactions,
       logger: contextLogger(this.#logger, 'transactions'),
       newTransactions: this.#newTransactions,
+      onFatalError,
       retryBackoffConfig,
       tip$: this.tip$,
       transactionsHistoryStore: stores.transactions
@@ -352,6 +362,7 @@ export class SingleAddressWallet implements ObservableWallet {
     this.utxo = createUtxoTracker({
       addresses$,
       logger: contextLogger(this.#logger, 'utxo'),
+      onFatalError,
       retryBackoffConfig,
       stores,
       tipBlockHeight$,
@@ -363,6 +374,7 @@ export class SingleAddressWallet implements ObservableWallet {
       epoch$,
       eraSummaries$,
       logger: contextLogger(this.#logger, 'delegation'),
+      onFatalError,
       retryBackoffConfig,
       rewardAccountAddresses$: this.addresses$.pipe(
         map((addresses) => addresses.map((groupedAddress) => groupedAddress.rewardAccount))
@@ -379,6 +391,7 @@ export class SingleAddressWallet implements ObservableWallet {
         assetProvider: this.assetProvider,
         balanceTracker: this.balance,
         logger: contextLogger(this.#logger, 'assets$'),
+        onFatalError,
         retryBackoffConfig
       }),
       stores.assets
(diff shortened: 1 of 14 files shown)


packages/wallet/src/services/util/coldObservableProvider.ts
```diff
@@ -49,7 +52,22 @@ export const coldObservableProvider = <T>({
                     )
               )
             )
-          ).pipe(retryBackoff(retryBackoffConfig))
+          ).pipe(
+            retryBackoff({
+              ...retryBackoffConfig,
+              shouldRetry: (error) => {
+                if (retryBackoffConfig.shouldRetry && !retryBackoffConfig.shouldRetry(error)) return false;
```
mkazlauskas, line 59: Alternative syntax:
```suggestion
                if (retryBackoffConfig.shouldRetry?.(error)) return false;
```

packages/wallet/test/SingleAddressWallet/load.test.ts
```diff
@@ -390,3 +391,31 @@ describe('SingleAddressWallet creates big UTXO', () => {
...
+        ...mocks.mockNetworkInfoProvider(),
+        ledgerTip: jest.fn().mockRejectedValue(new InvalidStringError('Test invalid string error'))
+      },
+      rewardsProvider: mocks.mockRewardsProvider(),
+      utxoProvider: mocks.mockUtxoProvider(utxoSet)
+    });
+
+    try {
+      await firstValueFrom(wallet.tip$);
+      expect(false).toBeTruthy();
```
mkazlauskas, line 412: I think either `expect.assertions(n)` or this
```suggestion
      throw new Error('Expected tip$' to throw)
```
would be a little clearer.

Also, do we want to throw on both `error$` and other observables? When I suggested `error$`, I thought it would be a 'catch-all' and other observables wouldn't throw. If we throw on other observables then maybe we don't need `error$`? Thoughts @rhyslbw @mirceahasegan ?

mkazlauskas (changes requested):

iccicci:
Both the suggestions have a considerable impact on the implementation and on the tests as well, so I had to refactor the unit tests.
Now `SingleAddressWallet.error$` is a `Subject` (rather than a `TrackedSubject`) so it must be observed before all other `Observable`s and it has a _catch all_ behavior: i.e. it emits all the fatal errors which are no longer thrown by other `Observable`s.

packages/wallet/test/SingleAddressWallet/load.test.ts
```diff
@@ -390,3 +391,31 @@ describe('SingleAddressWallet creates big UTXO', () => {
...
+    // wallet.error$ must be observed till the beginning of time
+    const errorPromise = expect(firstValueFrom(wallet.error$)).resolves.toBeInstanceOf(InvalidStringError);
+
+    wallet.tip$.subscribe(() => {
+      throw new Error('Expected tip$ to not emit');
+    });
+
+    await errorPromise;
+
+    wallet.shutdown();
```
mkazlauskas, line 419: A more expressive way to assert it doesn't emit:
```suggestion
    let tipHandler = jest.fn();
    wallet.tip$.subscribe(tipHandler);
    await errorPromise;
    wallet.shutdown();
    expect(tipHandler).not.toBeCalled();
```

mkazlauskas (changes requested):

packages/wallet/test/services/util/coldObservableProvider.test.ts
```diff
@@ -47,6 +47,28 @@ describe('coldObservableProvider', () => {
...
+    const testValue = { test: 'value' };
+    const testError = new InvalidStringError('Test invalid string error');
+    const underlyingProvider = jest
+      .fn()
+      .mockRejectedValueOnce(new Error('Test error'))
+      .mockResolvedValueOnce(testValue)
+      .mockRejectedValueOnce(testError)
+      .mockResolvedValueOnce(testValue);
+    const onFatalError = jest.fn();
+    const retryBackoffConfig: RetryBackoffConfig = { initialInterval: 1, shouldRetry: () => true };
```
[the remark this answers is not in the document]
iccicci: Nope, it is provided to prove the retry strategy works also providing a `shouldRetry` function.
Do you think is it better to write two distinct tests?

packages/wallet/test/SingleAddressWallet/load.test.ts
```diff
@@ -407,14 +407,17 @@ describe('SingleAddressWallet.error$', () => {
       utxoProvider: mocks.mockUtxoProvider(utxoSet)
     });
 
+    // wallet.error$ must be observed till the beginning of time
```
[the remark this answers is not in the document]
iccicci: It should, probably it's better to add a dedicated test to prove this.

packages/wallet/test/services/util/coldObservableProvider.test.ts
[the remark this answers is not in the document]
mkazlauskas: No it's ok, just wanted to make sure we didn't change the behavior

packages/wallet/test/SingleAddressWallet/load.test.ts
iccicci: efd3bd734 should solve this

packages/wallet/src/SingleAddressWallet/SingleAddressWallet.ts
```diff
@@ -162,6 +164,7 @@ export class SingleAddressWallet implements ObservableWallet {
   readonly protocolParameters$: TrackerSubject<Cardano.ProtocolParameters>;
   readonly genesisParameters$: TrackerSubject<Cardano.CompactGenesis>;
   readonly assets$: TrackerSubject<Assets>;
+  readonly error$: TrackerSubject<unknown>;
```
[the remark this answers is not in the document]
iccicci: ffbc1017a should solve this

packages/wallet/test/SingleAddressWallet/load.test.ts
iccicci: ffbc1017a should solve this

packages/wallet/test/SingleAddressWallet/load.test.ts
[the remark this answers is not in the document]
iccicci: 6f8cb2673 adds the test
also renames `error$` in `fatalError$`

packages/wallet/src/types.ts
```diff
@@ -82,6 +82,7 @@ export interface ObservableWallet {
   readonly protocolParameters$: Observable<Cardano.ProtocolParameters>;
   readonly addresses$: Observable<GroupedAddress[]>;
   readonly assets$: Observable<Assets>;
+  readonly fatalError$: Observable<unknown>;
```
mkazlauskas, line 85: A doc comment on this would be nice (explaining it's a catch-all and that wallet is probably 'bricked' when this emits)

mkazlauskas (dismissed):
Nice work! Please remove scope from the commit when squashing as it also has a change in web-extension package

packages/wallet/src/services/util/coldObservableProvider.ts
```diff
@@ -49,10 +56,24 @@ export const coldObservableProvider = <T>({
...
+              }
+            })
+          )
         ),
         distinctUntilChanged(equals),
-        takeUntil(cancel$)
+        takeUntil(internalCancel$)
       )
       .subscribe(subscriber);
     return () => sub.unsubscribe();
```
mirceahasegan, line 79: I think it would be good to also complete the `cancelOnFatalError$` subject in the unsubscribe callback:
```diff
-     return () => sub.unsubscribe();
+     return () => { sub.unsubscribe(); cancelOnFatalError$.complete(); };

Comment (mkazlauskas) [reply packages/wallet/src/services/util/coldObservableProvider.ts:79]: Nice catch!

File: packages/wallet/test/services/util/coldObservableProvider.test.ts [hunk 9]
```diff
@@ -16,7 +17,6 @@ describe('coldObservableProvider', () => {
     expect(await firstValueFrom(provider$)).toBe(true);
     expect(underlyingProvider).toBeCalledTimes(2);
     expect(retryBackoff).toBeCalledTimes(2);
-    expect(retryBackoff).toBeCalledWith(backoffConfig);
```
mirceahasegan, line 19: is this expectation not valid anymore?

mirceahasegan (approved):
Nice work @iccicci!

packages/wallet/src/services/util/coldObservableProvider.ts
iccicci: dcda82612 should solve this

mkazlauskas (approved):
:rocket:

mkazlauskas (approved):

mirceahasegan (approved):
Great work, @iccicci 🚀

PR #563, opened 2022-12-13
feat: added new babbage era types in Transactions and Outputs

AngelCastilloB:
# Context

To support the new Babbage Era features, we must:

- Add the new fields (e.g. collateral return, inline datums, etc.) to the Core package’s types.
- Ensure serialization of those new fields is possible & well tested
- Including all existing Alonzo fields that have not yet been added.

# Proposed Solution

Add the new Babbage types to our core types.

New fields of transaction body include:

1. collateral return;
2. total collateral
3. reference inputs;

New fields of transaction outputs include:

1. inline datum   
2. script referen


packages/core/src/Cardano/types/Utxo.ts
```diff
@@ -15,7 +16,30 @@ export interface HydratedTxIn extends TxIn {
 export interface TxOut {
   address: Address;
   value: Value;
-  datum?: Hash32ByteBase16;
```
mkazlauskas, line 18: This makes it a breaking change: please update commit msg

packages/core/src/Cardano/types/Utxo.ts
```diff
@@ -15,7 +16,30 @@ export interface HydratedTxIn extends TxIn {
...
+  /**
+   * Datum hash, this allows to specify a Datum without publicly revealing its value. To spend an output which specifies
+   * this type of datum, the actual Datum value must be provided and will be added to the witness set of
+   * the transaction.
+   */
+  datumHash?: Hash32ByteBase16;
+
+  /**
+   * Inline datum reveals the value on the blockchain at the time of output creation. This way of attaching datums lets
+   * users consume this output without specifying the datum value.
```
mkazlauskas, line 29: They are still specifying the datum value, but they can find it on-chain.
```suggestion
   * users consume this output without specifying the datum value separately.
```

packages/core/src/Cardano/types/Utxo.ts
```diff
@@ -15,7 +16,30 @@ export interface HydratedTxIn extends TxIn {
...
+   */
+  inlineDatum?: HexBlob;
+
+  /**
+   * Reference scripts can be used to satisfy script requirements during validation, rather than requiring the spending
+   * transaction to do so. This will allow transactions using common scripts to be much smaller.
```
mkazlauskas, line 35:
```suggestion
   * transaction to do so. This allows transactions using common scripts to be much smaller.
```

packages/core/src/Cardano/types/Utxo.ts
```diff
@@ -15,7 +16,30 @@ export interface HydratedTxIn extends TxIn {
...
+   *
+   * The key idea is to use reference inputs and outputs which carry actual scripts ("reference scripts"), and allow
+   * such reference scripts to satisfy the script witnessing requirement for a transaction. This means that the
+   * transaction which uses the script will not need to provide it at all, so long as it referenced an output
+   * which contained the script.
+   */
+  scriptRef?: Script;
```
mkazlauskas, line 42: This should probably be a different type
```suggestion
  scriptRef?: ScriptReference;
```

packages/ogmios/src/ogmiosToCore/tx.ts
```diff
@@ -249,15 +249,22 @@ const mapTxIn = (txIn: Schema.TxIn): Cardano.TxIn => ({
   txId: Cardano.TransactionId(txIn.txId)
 });
 
-const mapDatum = (datum: Schema.TxOut['datum']) => {
+const mapInlineDatum = (datum: Schema.TxOut['datum']) => {
+  if (!datum) return;
+  if (typeof datum === 'string') return Cardano.util.HexBlob(datum);
+  if (typeof datum === 'object') return Cardano.util.HexBlob(JSON.stringify(datum));
```
mkazlauskas, line 255: Is this correct? :thinking: Need a test with ogmios inline datum

mkazlauskas (changes requested):

packages/core/src/Cardano/types/Utxo.ts
mkazlauskas: Right, you only specify `{txId,index}`

packages/core/src/Cardano/types/Utxo.ts
mkazlauskas: Oh so it's actually the 'Script' type. Maybe let's go back to using the existing `Script` type then, but also rename the field? Like this:
```ts
script: Script
```
Then, how is it actually referenced (used as witness)?

packages/ogmios/src/ogmiosToCore/tx.ts
mkazlauskas: Do we have test data when ogmios datum is an `object`?

packages/core/src/Cardano/types/Utxo.ts
mkazlauskas: > but if we want to change it to just script I am also okay with that

I can't think of a reason to have a separate type if it refers to the same thing, so yeah, let's do that unless there's a reason to have separate types.

> ...as a reference input and the node will search for the eUTXO and fetch the script automatically. What I have done here allows us to create these reference outputs (or inputs), but we are still missing the part of using them, that will come once we have smart-contract support

Thanks for an explanation. The part of using them is what's unclear to me, are all scripts in reference inputs implicitly used as witness? :thinking:

packages/ogmios/src/ogmiosToCore/tx.ts
mkazlauskas: Maybe it's only relevant for `datum_hash`: https://github.com/CardanoSolutions/ogmios/blob/master/clients/TypeScript/packages/schema/src/index.ts#L1120

My guess is that key = scripthash and value=datumhash there

mkazlauskas: I think our current implementation is incorrect. Key and value there needs to be interpreted separately:
- key=hash32 (script hash? not sure)
- value=plutus_data (the hexblob)

https://github.com/input-output-hk/cardano-ledger/blob/master/eras/babbage/test-suite/cddl-files/babbage.cddl#L421

Let's try to figure out what is the difference between 'Datum' and key-value object: https://github.com/CardanoSolutions/ogmios/blob/master/clients/TypeScript/packages/schema/src/index.ts#L1120-L1123

mkazlauskas: In that case we should drop the key and only use the value?

Or can a single output have multiple datums? :thinking: In this case we'd need to update our core type

packages/ogmios/src/ogmiosToCore/tx.ts
```diff
@@ -249,15 +249,21 @@ const mapTxIn = (txIn: Schema.TxIn): Cardano.TxIn => ({
   txId: Cardano.TransactionId(txIn.txId)
 });
 
-const mapDatum = (datum: Schema.TxOut['datum']) => {
+const mapInlineDatum = (datum: Schema.TxOut['datum']) => {
+  if (!datum || typeof datum !== 'string') return;
```
mkazlauskas, line 253: Can there be an empty datum string? Either way, I think `typeof` should be sufficient here.
```suggestion
  if (typeof datum !== 'string') return;
```

mkazlauskas (changes requested):

mkazlauskas (approved):
Nice work! Great to see extensive doc comments :muscle:

IvayloAndonov (dismissed):
Great work!🔝

mkazlauskas (approved):

IvayloAndonov (approved):

PR #564, opened 2022-12-13
fix: cip30 wallet has to accept hex encoded address

mirceahasegan:
# Context

Using the signData method of cardano-js-sdk with a test dapp with a hex encoded address throws a DataSignError code 2 (Invalid address). This is because cardano-js-sdk only expects it to be in bech32 format.

According to the CIP-30 spec, the address argument should accept either bech32 or hex-encoded bytes. However, it seems that the cardano-js-sdk only accepts bech32. Support for hex-encoded addresses should be added.

# Proposed Solution
Accept hex encoded address. 

### DIFF
--- a/packages/core/src/Cardano/types/Address.ts
+++ b/packages/core/src/Cardano/types/Address.ts
@@ -1,5 +1,8 @@
+import { usingAutoFree } from '@cardano-sdk/util';
+
+import { CML } from '../../CML';
 import { InvalidStringError } from '../../errors';
-import { OpaqueString, assertIsBech32WithPrefix } from '../util/primitives';
+import { OpaqueString, assertIsBech32WithPrefix, assertIsHexString } from '../util/primitives';
 import { isAddress } from '../util/address';
 
 /**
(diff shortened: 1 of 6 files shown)


rhyslbw (changes requested):
Looks good @mirceahasegan.

The address util is best committed first as a feature, given it's standalone with test coverage. The only time to bundle this type of change into the same commit is if the fix expresses the behaviour completely, i.e you only have integration test coverage.

packages/core/src/Cardano/types/Address.ts
```diff
@@ -27,3 +30,23 @@ export const Address = (value: string): Address => {
...
 };
+
+/**
+ * Transform hex-encoded bytes address into `Cardano.Address`,
+ * base58 in case of Byron era or bech32 in case of Shelley era or newer.
+ *
+ * @param value hex-encoded bytes address.
+ * @throws {InvalidStringError} if value is invalid
+ */
+Address.fromHex = (value: string): Address =>
```
mkazlauskas, line 41: Maybe we could update the existing `Address()` function to use this. It should currently throw given hex-encoded address, so we could call this in the `catch` block

mkazlauskas (changes requested):

packages/core/src/Cardano/types/Address.ts
rhyslbw: Nice, that would be quite elegant. Taking this approach would mean a single commit @mirceahasegan, as it is just a fix for the current function

mirceahasegan: Updated in [2bf0f29](https://github.com/input-output-hk/cardano-js-sdk/pull/564/commits/2bf0f29a886dcf145317864d6dda84d7e9e33c0a)

packages/core/src/Cardano/types/Address.ts
```diff
@@ -24,29 +24,28 @@ const isRewardAccount = (address: string) => {
...
+ * Transform {@link value} into `Cardano.Address`,
+ * Resulting Address will be base58 in case of Byron era or bech32 in case of Shelley era or newer.
+ *
+ * @param value bech32 string, base58 string or hex-encoded bytes address.
+ * @throws {InvalidStringError} if value is invalid
+ */
+
 export const Address = (value: string): Address => {
   if (isAddress(value) && !isRewardAccount(value)) {
     return value as unknown as Address;
```
mkazlauskas, line 37: Could early exit to not try to parse hex bytes from reward account, as well as give a better error message here
```ts
  if (isAddress(value)) {
    if (isRewardAccount(value)) {
      throw new InvalidStringError(`Address type can only be used for payment addresses`);
    }
    return value as unknown as Address;
  }
```

mkazlauskas (changes requested):

packages/core/src/Cardano/types/Address.ts
mirceahasegan: [78c38c6](https://github.com/input-output-hk/cardano-js-sdk/pull/564/commits/78c38c6e4c75be7137e3718273e27d560c5e9dce). Added specific errors for:
1. non payment address
2. format not accepted
3. invalid address

mkazlauskas (dismissed):
Nice work! :rocket:

mkazlauskas (approved):

PR #565, opened 2022-12-16
fix(wallet):  Incorrect Rewards Amount

IvayloAndonov:
# Context
Some Lace Open Beta testing feedback has highlighted that the “total rewards” displayed (in the delegation centre) are incorrect when compared to Eternl or [Cardano Explorer | Stake pools | Cexplorer.io](http://preprod.cexplorer.io/). For the user who reported this, their rewards balance shown was 37.31 tADA (incorrect), according to Eternl and [Cexplorer.io](http://cexplorer.io/) their actual “total rewards” were 32.73 tADA (correct). This is causing confusion for users. Other users have not reported this yet, and this may be an issue when restoring a wallet that already has staking rewards available. 

The value the user is familiar with & reporting is actually "Rewards Withdrawn" which is arguably different to “Total Rewards”.

# Proposed Solution
Investigation details and solution in the [thread](https://input-output-rnd.slack.com/archives/G01MHNME0UV/p1671113387674349)

- fix the formula for determining the first delegation epoch, included in the rewards calculation afterward

### DIFF
--- a/packages/wallet/src/services/DelegationTracker/RewardsHistory.ts
+++ b/packages/wallet/src/services/DelegationTracker/RewardsHistory.ts
@@ -12,6 +12,11 @@ import first from 'lodash/first';
 import flatten from 'lodash/flatten';
 import groupBy from 'lodash/groupBy';
 
+const DELEGATION_EPOCHS_AHEAD_COUNT = 2;
+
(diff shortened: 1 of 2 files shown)


mkazlauskas (approved):
Great work! :detective:

PR #566, opened 2022-12-20
feat(wallet): Enable pouchdb auto-compaction

IvayloAndonov:
# Context

By default, PouchDB stores all document revisions forever. SingleAddressWallet doesn’t take advantage of it, so we should keep it small.

# Proposed Solution
Enable [auto-compaction](https://pouchdb.com/guides/compact-and-destroy.html#auto-compaction).

### DIFF
--- a/packages/wallet/test/persistence/pouchDbStores.test.ts
+++ b/packages/wallet/test/persistence/pouchDbStores.test.ts
@@ -13,7 +13,7 @@ describe('pouchDbStores', () => {
 
   afterAll(async () => {
     // delete files from the filesystem
-    await new PouchDB(dbName).destroy();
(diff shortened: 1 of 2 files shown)


packages/wallet/test/persistence/pouchDbStores.test.ts
```diff
@@ -13,7 +13,7 @@ describe('pouchDbStores', () => {
 
   afterAll(async () => {
     // delete files from the filesystem
-    await new PouchDB(dbName).destroy();
+    await new PouchDB(dbName, { auto_compaction: true }).destroy();
```
mkazlauskas, line 16: Don't think this change does anything, as it's being destroyed here

mkazlauskas (approved):
:rocket:

PR #567, opened 2022-12-21
refactor!: Provider dependencies standartization

IvayloAndonov:
# Context

As per [this conversation](https://github.com/input-output-hk/cardano-js-sdk/pull/553#discussion_r1041822915), we need a refactoring action to standardize the Provider dependencies in order to enhance the DbSyncProvider parameters to provide it a logger and finally adding the same log line described in the same conversation.
# Proposed Solution

- restructure `Asset` directory to be aligned with all other providers
- standardize `DbSyncProvider` dependencies, create interface `DbSynProviderDependencies`
- add debug log on `DbSyncProvider` health check
- align all provider dependencies structure
- add doc blocks for all props and deps interfaces 

**BREAKING CHANGE:**
- all provider constructors are updated to use a standardized form of deps


mkazlauskas (approved):
Nice work!

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/DbSyncChainHistoryProvider.ts
```diff
@@ -42,7 +46,7 @@ export class DbSyncChainHistoryProvider extends DbSyncProvider() implements Chai
     { paginationPageSizeLimit }: ChainHistoryProviderProps,
     { db, cardanoNode, metadataService, logger }: ChainHistoryProviderDependencies
   ) {
-    super(db, cardanoNode);
+    super({ cardanoNode, db, logger });
     this.#logger = logger;
```
iccicci, line 50: The `#logger` property can be removed from from `DbSyncChainHistoryProvider` as it now inherits the `logger` property from `DbSyncProvider`.
The same, can be applied to all other `Provider`s extending `DbSyncProvider`.

iccicci (commented):
Excellent job! Just a single comment.

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/DbSyncChainHistoryProvider.ts
IvayloAndonov: Good one, applied in 95e990040301ec5bf26de7346cc2b152dae8556e @iccicci

iccicci (dismissed):

IvayloAndonov:
Rebased with master and commits are squashed.

mkazlauskas (approved):

rhyslbw (changes requested):
Looks good @IvayloAndonov, and good job with the attention to adding docblocks and refactoring for standardisation. There's just a few types that need to be reconsidered though, as they'll be used for other implementations:

https://github.com/input-output-hk/cardano-js-sdk/blob/4b9fd5690fbe7cff0eeae8313536459ec0a0ce7a/packages/cardano-services/src/Asset/DbSyncAssetProvider/types.ts#L4-L31

rhyslbw:
> What is meant here by "to be reconsidered"?

They should not be nested under a `db-sync` scope as they can be used in other contexts. For now you could just leave these interfaces where they were, and only move the `db-sync` models into the new directory.

IvayloAndonov:
> > What is meant here by "to be reconsidered"?
> 
> They should not be nested under a `db-sync` scope as they can be used in other contexts. For now you could just leave these interfaces where they were, and only move the `db-sync` models into the new directory.

Aham, both interfaces are not coupled with `db-sync` at all, only the actual implementations could be, like `DbSyncNftMetadataService`.
So I'll move both types out of `Asset/DbSyncAssetProvider`, good catch!

rhyslbw (changes requested):
https://github.com/input-output-hk/cardano-js-sdk/blob/4b9fd5690fbe7cff0eeae8313536459ec0a0ce7a/packages/cardano-services/src/Asset/DbSyncAssetProvider/types.ts#L33-L50

These are specific to `db-sync`, so best to nest them as you had it originally

rhyslbw (approved):

packages/cardano-services/src/util/DbSyncProvider/DbSyncProvider.ts
```diff
@@ -1,9 +1,34 @@
...
+  ProviderFailure
+} from '@cardano-sdk/core';
 import { DB_BLOCKS_BEHIND_TOLERANCE, DB_MAX_SAFE_INTEGER, LedgerTipModel, findLedgerTip } from './util';
+import { Logger } from 'ts-log';
 import { Pool } from 'pg';
 
+/**
+ * Properties that are need to create DbSyncProvider
+ */
+export interface DbSynProviderDependencies extends ProviderDependencies {
```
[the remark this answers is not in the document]
IvayloAndonov: The typo is fixed with 48438d21bc630df092ac23cf68d9fed3ad7a49e7 @iadmytro

iadmytro (approved):

IvayloAndonov:
Commits are squashed and ready for merge once the CI pass.

rhyslbw (approved):

iccicci (approved):

PR #572, opened 2023-01-10
Fix/adp 2519 tx id log

iccicci:
# Context

Some messages logged by `DbSyncChainHistoryProvider` contains transaction ids in `Buffer` format; this makes the same messages hard to be read.

# Proposed Solution

Changed the logged messages including the transaction ids in the hex string format.
Added test checks as well in order to ensure the logged format.

Added a `TestLogger.reset()` method the reset the recorded messages logged by the `TestLogger`.

### DIFF
--- a/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
+++ b/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
@@ -46,8 +46,8 @@ export class ChainHistoryBuilder {
   }
 
   public async queryTransactionInputsByHashes(hashes: Cardano.TransactionId[], collateral = false): Promise<TxInput[]> {
+    this.#logger.debug(`About to find inputs (collateral: ${collateral}) for transactions:`, hashes);
     const byteHashes = hashes.map((hash) => hexStringToBuffer(hash.toString()));
-    this.#logger.debug(`About to find inputs (collateral: ${collateral}) for transactions:`, byteHashes);
     const result: QueryResult<TxInputModel> = await this.#db.query(
       collateral ? Queries.findTxCollateralsByHashes : Queries.findTxInputsByHashes,
       [byteHashes]
@@ -62,8 +62,8 @@ export class ChainHistoryBuilder {
   }
 
   public async queryTransactionOutputsByHashes(hashes: Cardano.TransactionId[]): Promise<TxOutput[]> {
+    this.#logger.debug('About to find outputs for transactions:', hashes);
     const byteHashes = hashes.map((hash) => hexStringToBuffer(hash.toString()));
-    this.#logger.debug('About to find outputs for transactions:', byteHashes);
     const result: QueryResult<TxOutputModel> = await this.#db.query(Queries.findTxOutputsByHashes, [byteHashes]);
     if (result.rows.length === 0) return [];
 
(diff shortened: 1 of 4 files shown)


packages/util-dev/src/TestLogger.ts
```diff
@@ -129,8 +129,9 @@ export const createLogger = (options: TestLoggerOptions = {}) => {
   };
 
   const logger = Object.fromEntries((<LoggerEntry[]>Object.entries(logLevels)).map((_) => [_[0], getLogFunction(_)]));
+  const reset = () => <void>(<unknown>(messages.length = 0));
```
rhyslbw, line 132: Adding a new method requires unit test coverage, but wouldn't it be better to just create a new instance for each test?

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
```diff
@@ -47,7 +47,7 @@ export class ChainHistoryBuilder {
 
   public async queryTransactionInputsByHashes(hashes: Cardano.TransactionId[], collateral = false): Promise<TxInput[]> {
     const byteHashes = hashes.map((hash) => hexStringToBuffer(hash.toString()));
-    this.#logger.debug(`About to find inputs (collateral: ${collateral}) for transactions:`, byteHashes);
+    this.#logger.debug(`About to find inputs (collateral: ${collateral}) for transactions:`, hashes);
```
rhyslbw, line 50: Since the log line no longer requires the variable set on the above line, it would be better to move this above it to line 49. The same feedback applies to all log lines where this change was made.

packages/cardano-services/test/ChainHistory/DbSyncChainHistoryProvider/ChainHistoryBuilder.test.ts
```diff
@@ -3,7 +3,30 @@ import { ChainHistoryBuilder } from '../../../src';
...
+import { createLogger } from '@cardano-sdk/util-dev';
+
+const logger = createLogger({ record: true });
+
+/**
+ * Checks tx_ids are logged as hex strings
+ *
+ * @param text the text of the loggd message containing tx_ids
+ */
+const checkLoggedTxids = (text: string) => {
```
rhyslbw, line 15:
```suggestion
const assertLoggedTxIds = (text: string): void => {
```

rhyslbw (changes requested):

packages/cardano-services/test/ChainHistory/DbSyncChainHistoryProvider/ChainHistoryBuilder.test.ts
iccicci: This was inferred by the absence of a `return` statement, but better to make it explicit.
ed6416dc7 should solve this

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
iccicci: ed6416dc7 should solve this

packages/util-dev/src/TestLogger.ts
iccicci: Test added with 0557f8148

> wouldn't it be better to just create a new instance for each test?

As general rule, yes, I agree.
Considering our tests, we usually need to pass the logger to several dependencies of the `class`/`function` to test: doing that would bean to destroy several object at the end of each test and re-creating them at each new test; this would make our tests much more complex.

rhyslbw (approved):
Nice work @iccicci

mkazlauskas (dismissed):
Nice work!

iccicci:
Squashed. This should be ready for final review

rhyslbw (approved):

PR #573, opened 2023-01-10
feat(cardano-services)!: upgrade /meta endpoint

IvayloAndonov:
# Context
There was a [contributed a solution ](https://github.com/input-output-hk/cardano-js-sdk/pull/496)to expose deployment metadata, which has made it possible to understand critical details that are important for dev and test engineers, and is being consumed in this wallet-world document https://github.com/input-output-hk/wallet-world/blob/master/ENVIRONMENTS.md. Feedback has been provided in the PR about how to turn it into a feature we can support, but due to time constraints it was decided to merge as-is, then dedicate some time for improvements in a future sprint.

# Proposed Solution
With provided a valid BUILD_INFO JSON format
Screenshot 2023-01-10 at 11 20 37

With an invalid BUILD_INFO JSON format on server startup
Screenshot 2023-01-11 at 12 23 01

- introduce `ServerMetadata` type
- add `--build-info` CLI arg and `BUILD_INFO` env
- add metadata JSON validation on startup
- add the endpoint to `OpenApi` def
- add CLI and unit tests

**BREAKING CHANGE:**

Nested `narHash`, `path` and `sourceInfo` under top level `extra` property of BUILD_INFO

### DIFF
--- a/packages/cardano-services/src/Http/HttpServer.ts
+++ b/packages/cardano-services/src/Http/HttpServer.ts
@@ -1,3 +1,4 @@
+import * as OpenApiValidator from 'express-openapi-validator';
 import { Gauge, Registry } from 'prom-client';
 import { HttpServerConfig, ServiceHealth, ServicesHealthCheckResponse } from './types';
 import { HttpService } from './HttpService';
@@ -9,6 +10,8 @@ import bodyParser from 'body-parser';
 import express from 'express';
 import expressPromBundle from 'express-prom-bundle';
 import http from 'http';
+import path from 'path';
+
 export const CONTENT_TYPE = 'Content-Type';
 export const APPLICATION_JSON = 'application/json';
 
@@ -91,6 +94,7 @@ export class HttpServer extends RunnableModule {
     }
     const requestLogger = contextLogger(this.logger, 'request');
     this.app.use((req, _res, next) => {
+      // eslint-disable-next-line @typescript-eslint/no-shadow
       const { body, method, path, query } = req;
       requestLogger.debug({ body, method, path, query });
       next();
@@ -122,11 +126,22 @@ export class HttpServer extends RunnableModule {
       HttpServer.sendJSON(res, new ProviderError(ProviderFailure.Unhealthy, err), err.status || 500);
     });
 
-    const buildInfo = JSON.parse(process.env.BUILD_INFO ?? '{}');
-    const deploymentInfo = { ...buildInfo, startupDate: new Date() };
-    const metaHandler = (_: express.Request, res: express.Response) => res.json(deploymentInfo);
-    this.app.get('/meta', metaHandler);
-    this.app.post('/meta', metaHandler);
+    const apiSpec = path.join(__dirname, 'openApi.json');
+    this.app.use(
+      OpenApiValidator.middleware({
+        apiSpec,
+        ignoreUndocumented: true,
+        validateRequests: true,
+        validateResponses: true
+      })
+    );
+
+    const serverMetadataHandler = async (req: express.Request, res: express.Response) => {
+      this.logger.debug('/meta', { ip: req.ip });
+      return HttpServer.sendJSON(res, this.#config.meta);
+    };
+
+    this.app.use('/meta', serverMetadataHandler);
   }
 
   static sendJSON<ResponseBody>(
--- a/packages/cardano-services/src/Http/openApi.json
+++ b/packages/cardano-services/src/Http/openApi.json
@@ -1,0 +1,77 @@
+{
+  "openapi": "3.0.0",
+  "info": {
+    "title": "Server Metadata",
+    "license": {
+      "name": "Apache 2.0",
+      "url": "http://www.apache.org/licenses/LICENSE-2.0.html"
+    },
+    "version": "1.0.0"
+  },
+  "paths": {
+    "/meta": {
+      "post": {
+        "summary": "HTTP Server metadata POST",
+        "operationId": "server-metadata-post",
+        "responses": {
+          "200": {
+            "description": "HTTP Server metadata",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/ServerMetadata"
+                }
+              }
+            }
+          }
+        }
+      },
+      "get": {
+        "summary": "HTTP server metadata GET",
+        "operationId": "server-metadata-get",
+        "responses": {
+          "200": {
+            "description": "HTTP Server metadata",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/ServerMetadata"
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  },
+  "components": {
+    "schemas": {
+      "ServerMetadata": {
+        "type": "object",
+        "required": [
+          "startupTime"
+        ],
+        "properties": {
+          "startupTime": {
+            "type": "number"
+          },
+          "lastModified": {
+            "type": "number"
+          },
+          "lastModifiedDate": {
+            "type": "string"
+          },
+          "rev": {
+            "type": "string"
+          },
+          "shortRev": {
+            "type": "string"
+          },
+          "extra": {
+            "type": "object"
+          }
+        }
+      }
+    }
+  }
+}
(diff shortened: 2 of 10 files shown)


packages/cardano-services/src/Http/HttpServer.ts
```diff
@@ -122,11 +126,34 @@ export class HttpServer extends RunnableModule {
...
-    this.app.get('/meta', metaHandler);
-    this.app.post('/meta', metaHandler);
+    const serverMetadataHandler = async (req: express.Request, res: express.Response) => {
+      this.logger.debug('/meta', { ip: req.ip });
+      let metadata: ServerMetadata;
+      try {
+        const buildInfo: BuildInfo = JSON.parse(this.#config.buildInfo ?? '{}');
+        metadata = { ...buildInfo, startupTime: Date.now() };
+      } catch (error) {
+        this.logger.error(`Invalid JSON format of process.env.BUILD_INFO: ${error}`);
```
rhyslbw, line 136: Better to validate this at startup

packages/cardano-services/src/Program/loadHttpServer.ts
```diff
@@ -34,6 +34,7 @@ import pg from 'pg';
 export interface HttpServerOptions extends CommonProgramOptions {
   serviceNames?: ServiceNames[];
   enableMetrics?: boolean;
+  buildInfo?: string;
```
rhyslbw, line 37:
```suggestion
  meta?: ServerMetadata;
```

Considering we're just mapping this value to the endpoint, terminology alignment makes sense. Also, it's not strictly build information, given there's `startupTime`.

I suggest renaming all uses of `buildInfo` to `meta`

rhyslbw (changes requested):

packages/cardano-services/src/Program/loadHttpServer.ts
rhyslbw: > Then we [pass it down](https://github.com/input-output-hk/cardano-js-sdk/pull/573/files#diff-cfc78f77666af1469e2e670c258fecabed5a923995563490c52370f7a6b4888cR211) to the config.buildInfo and then [compose](https://github.com/input-output-hk/cardano-js-sdk/pull/573/files#diff-db63771378e2c7af74fab64645861ab1f0556c3320ce46fabf60eedf796ab08eR133-R134) the matadata : ServerMetadata obj as the final endpoint response (buildInfo + startupTime).

Ah yes, fine as is then :+1:

packages/cardano-services/src/Http/HttpServer.ts
IvayloAndonov: Applied with 88a41a2a0027e9769d42f40a23bf0f851b81b5da together with your suggestion. @rhyslbw

iccicci (dismissed):
Good job

packages/cardano-services/src/util/validators.ts
```diff
@@ -1,5 +1,41 @@
...
+import { validate } from 'jsonschema';
+import fs from 'fs';
+
+const buildInfoSchema = {
+  additionalProperties: false,
+  properties: {
+    extra: { type: 'object' },
+    lastModified: { type: 'number' },
+    lastModifiedDate: { type: 'string' },
+    narHash: { type: 'string' },
```
rhyslbw, line 13: Better to nest `narHash`, `path`, and `sourceInfo` under extra properties, since these are specific to nix, or rather, not generic enough to be typed.

packages/cardano-services/test/cli.test.ts
```diff
@@ -22,7 +22,7 @@ import axios, { AxiosError } from 'axios';
 import http from 'http';
 import path from 'path';
 
-jest.setTimeout(60_000);
+jest.setTimeout(90_000);
```
rhyslbw, line 25: Was it being flaky with a 60 second timeout?

rhyslbw (changes requested):

packages/cardano-services/test/cli.test.ts
IvayloAndonov: No actually, but I noticed the average `cli.test.ts` execution time is around 60 secs, so I increased it to 1.5 min as a precaution

packages/cardano-services/src/util/validators.ts
IvayloAndonov: You mean to nest them in [ServerMetadata](https://github.com/input-output-hk/cardano-js-sdk/blob/88a41a2a0027e9769d42f40a23bf0f851b81b5da/packages/cardano-services/src/Http/types.ts#L15-L25) response type, correct? As the `BUILD_INFO` input all fields are [flat structured](https://preprod-api.dev.lw.iog.io/meta). @rhyslbw

rhyslbw: In effect, you can remove these properties from our type, and instead expect them to be included as `extra` arguments.

packages/cardano-services/test/cli.test.ts
rhyslbw: It's OK here for now, but in the future it would be better to commit this separately with a description to explain the change.

packages/cardano-services/src/util/validators.ts
IvayloAndonov: Okay, that will force the SRE to pass `BUILD_INFO` value with the new expected structure with `extra` field.
Going to change it now, just not sure wether it should be marked as a breaking change or not? @rhyslbw

rhyslbw: Good observation. Yes, it's a breaking change

IvayloAndonov: 4a727e918cae826a61de8d53e82aedc70e2ec4d5 @rhyslbw

packages/cardano-services/src/Http/openApi.json
```diff
@@ -0,0 +1,106 @@
...
+            "description": "HTTP Server metadata",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/ServerMetadata"
+                }
+              }
+            }
+          },
+          "400": {
```
mkazlauskas, line 27: Does it ever respond with 400? There are no arguments to this request :thinking:

IvayloAndonov: RIght, there is no such case since we validate the build info on startup, will remove it from open API schema @mkazlauskas

packages/cardano-services/src/Http/types.ts
```diff
@@ -12,6 +12,21 @@ export type ServicesHealthCheckResponse = {
...
+  path?: string;
+  sourceInfo?: string;
+};
+
+export type ServerMetadata = {
+  lastModified?: number;
+  lastModifiedDate?: string;
+  rev?: string;
+  shortRev?: string;
+  extra?: ExtraMeta;
```
rhyslbw, line 26:
```suggestion
  extra?: JSON;
```

rhyslbw (changes requested):

packages/cardano-services/src/util/validators.ts
rhyslbw: Extra should remain generic though, as I've suggested here: https://github.com/input-output-hk/cardano-js-sdk/pull/573#discussion_r1069382716

So the only place we see `narHash`, `path`, and `sourceInfo` is in tests, although you could also just use dummy values there.

IvayloAndonov: Yeah, for some reason I have paid attention only to the first part (`"to nest narHash, path, and sourceInfo under extra properties"`), and inadvertently missed the second (`"not generic enough to be typed"`). 
It is applied with 3f3931ccdb4fd7b4befe7609b74565678a8f4aee commit @rhyslbw

packages/cardano-services/src/Http/types.ts
IvayloAndonov: 3f3931ccdb4fd7b4befe7609b74565678a8f4aee

rhyslbw (approved):

IvayloAndonov:
Commits are squashed, the breaking change commit type is annotated and rebased with master.

rhyslbw (approved):

iccicci (approved):
LGTM

rhyslbw:
@gytis-ivaskevicius I've reviewed your implementation and can see this probably won't be breaking for you, so we'll merge ahead of review.

PR #574, opened 2023-01-10
feat(crypto)!: libsodium crypto implementation

AngelCastilloB:
# Context

Currently, the SDK uses the Cardano Multiplatform Lib to perform all Ed25519-related operations. We want to replace this dependency by implementing our own package that uses [libsodium](https://github.com/jedisct1/libsodium.js/) for cryptographic operations.

# Proposed Solution

Introduce a new set of ed25519 classes to the crypto package based on [libsodium](https://github.com/jedisct1/libsodium.js/). There is also a new BIP32 set of classes implemented natively in javascript for key public and private key derivation.

# Critical Segments

This PR introduces some code that must be audited and verified, as we consider it to be critical:
  - BIP-32 key derivation: The core logic that does the derivation is contained in this ts file [Bip32KeyDerivation.ts ](https://github.com/input-output-hk/cardano-js-sdk/blob/feat/implement-ed25519-crypto-provider-interface/packages/crypto/src/Bip32/Bip32KeyDerivation.ts), and the following classes are implemented in terms of the functions defined in such file: [Bip32PrivateKey.ts ](https://github.com/input-output-hk/cardano-js-sdk/blob/feat/implement-ed25519-crypto-provider-interface/packages/crypto/src/Bip32/Bip32PrivateKey.ts
) and [Bip32PublicKey.ts](https://github.com/input-output-hk/cardano-js-sdk/blob/feat/implement-ed25519-crypto-provider-interface/packages/crypto/src/Bip32/Bip32PublicKey.ts)
  - Ed25519 Extended: Libsodium doesn't support out-of-the-box extended keys in its high-level API  (**crypto_sign** and **crypto_sign_detached**), so we had to implement our own sign method using the Ed25519 primitives provided by the [libsodium](https://github.com/jedisct1/libsodium.js/) library, said function can be found here: [signExtendedDetached](https://github.com/input-output-hk/cardano-js-sdk/blob/feat/implement-ed25519-crypto-provider-interface/packages/crypto/src/Ed25519e/Ed25519PrivateKey.ts#L72-L87), the following classes will be also a point of interest for reviewers: [Ed25519PrivateKey.ts](https://github.com/input-output-hk/cardano-js-sdk/blob/feat/implement-ed25519-crypto-provider-interface/packages/crypto/src/Ed25519e/Ed25519PrivateKey.ts )  and [Ed25519PublicKey.ts](https://github.com/input-output-hk/cardano-js-sdk/blob/feat/implement-ed25519-crypto-provider-interface/packages/crypto/src/Ed25519e/Ed25519PublicKey.ts) 

 - A new set of classes were introduced to the  'crypto' package, which contains several classes and utilities to perform Ed25519 cryptographic operations using [libsodium](https://github.com/jedisct1/libsodium.js/).
 - Added a new implementation of the Bip32Ed25519 interface.
 - Unit tests and e2e test now uses the new SodiumBip32Ed25519 concrete class by default.


packages/crypto/src/index.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export * from './Ed25519e';
+export * from './Bip32';
+export * from './errors';
+
+/**
+ * Resolves when the crypto module is ready.
+ */
+export const initialize = async () => {
+  await Sodium.ready;
```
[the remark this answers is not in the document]
mkazlauskas: It's fine but might require some refactoring

packages/crypto/src/Bip32/Bip32PrivateKey.ts
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * Bip32PrivateKey private key. This type of key have the ability to derive additional keys from them
+ * following the BIP32 derivation scheme variant called BIP32-Ed25519.
+ *
+ * @see <a href="https://ieeexplore.ieee.org/document/7966967">
+ *        BIP32-Ed25519: Hierarchical Deterministic Keys over a Non-linear Keyspace
+ *      </a>
+ */
+export class Bip32PrivateKey {
```
mkazlauskas, line 51: I think we should keep the original types as OpaqueString (just move them to `crypto` package), because they are being used as string and we still need a type for that.

Classes that provides operations using the key could have a different name and a constructor that takes in the opaque hex string.

mkazlauskas (changes requested):

packages/crypto/src/Bip32/Bip32PrivateKey.ts
mkazlauskas: Not really, OpaqueString is actually just String at runtime, which is very handy for serialization/deserialization

mkazlauskas: For example, data returned from http providers is serialized in cardano-services and deserialized in cardano-services-client. Having those types as classes would make this serialization and deserialization more complicated. That's not relevant for private key types, but relevant for public keys, hashes and signatures.

Also just having a string type that is not interchangeable with other strings is very handy, because you can be sure that you have the correct type of key/hash etc., and can't use another hex string in place of it. An example of that would be that tsc would complain if you tried using private key in place of a public key.

mkazlauskas: Also have in mind that keys, hashes and signatures are almost always passed around as hex strings

mkazlauskas: I think we should keep `core` types as plain javascript objects. It'll keep it more interoperable and you won't need to make any changes to serialization functions that are currently used. What do you think @rhyslbw ?

packages/governance/src/cip36.ts
```diff
@@ -20,8 +19,8 @@ export const SIG_MAP_KEY = 1n;
  * CIP-36 voting key derivation path constants
  */
 export enum VotingKeyDerivationPath {
-  PURPOSE = keyManagementUtil.harden(1694),
-  COIN_TYPE = keyManagementUtil.harden(1815)
+  PURPOSE = 1694,
+  COIN_TYPE = 1815
```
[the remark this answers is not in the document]
mkazlauskas: Yeah seems like a bug. The fix will be breaking so let's make sure it's in a separate commit so that it's highlighted in the changelog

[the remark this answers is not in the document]
mkazlauskas: Can we safely assume that no use cases would use indexes >2147483648? :thinking:

packages/crypto/src/Bip32/Bip32KeyDerivation.ts
```diff
@@ -0,0 +1,175 @@
...
+  const left = truc28Mul8(kl, zl);
+  const right = add(kr, zr);
+
+  return Buffer.concat([left, right, chainCode]);
+};
+
+/**
+ * Derive the given public key with the given index.
+ *
+ * Public key derivation os only possible with un hardened indices.
```
rhyslbw, line 145:
```suggestion
 * Public key derivation is only possible with non-hardened indices.
```

rhyslbw (changes requested):
Fantastic work here @AngelCastilloB

packages/crypto/src/Bip32/Bip32KeyDerivation.ts
```diff
@@ -0,0 +1,175 @@
...
+ *
+ * @param index The derivation index.
+ * @param scalar Ed25519 curve scalar.
+ * @param chainCode The chain code.
+ */
+const deriveSoft = (index: number, scalar: Buffer, chainCode: Buffer): { zMac: Uint8Array; ccMac: Uint8Array } => {
+  const data = Buffer.allocUnsafe(1 + 32 + 4);
+  data.writeUInt32LE(index, 1 + 32);
+
+  const vk = Buffer.from(crypto_scalarmult_ed25519_base_noclamp(scalar));
```
iquerejeta, line 54: Just a thought here, but it might be more efficient to store the VK together with the secret key, rather than compute it each time that is needed.

packages/crypto/src/Bip32/Bip32KeyDerivation.ts
```diff
@@ -0,0 +1,175 @@
...
+  new BN(lhs, 16, 'le').add(new BN(rhs.slice(0, 28), 16, 'le').mul(new BN(8))).toArrayLike(Buffer, 'le', 32);
+
+/**
+ * Computes `(8 * sk[:28])*G` where `sk` is a little-endian encoded int and `G` is the curve's base point.
+ *
+ * @param sk The secret key.
+ */
+const pointOfTrunc28Mul8 = (sk: Uint8Array) => {
+  const left = new BN(sk.slice(0, 28), 16, 'le').mul(new BN(8));
+  return ed25519.curve.g.mul(left);
```
iquerejeta, line 82: Out of curiosity, why are we using `ed25519.curve.g.mul` here instead of `crypto_scalarmult_ed25519_base_noclamp` as used in the ed25519 implementation?

packages/crypto/src/Bip32/Bip32KeyDerivation.ts
```diff
@@ -0,0 +1,175 @@
...
+ *
+ * @param lhs Left hand side Little-Endian big number.
+ * @param rhs Right hand side Little-Endian big number.
+ */
+const add = (lhs: Uint8Array, rhs: Uint8Array): Buffer => {
+  let r = new BN(lhs, 16, 'le').add(new BN(rhs, 16, 'le')).toArrayLike(Buffer, 'le').subarray(0, 32);
+
+  if (r.length !== 32) {
+    r = Buffer.from(r.toString('hex').padEnd(32, '0'), 'hex');
+  }
```
iquerejeta, line 96: What happens if r.length >= 32? This could happen in practice, right? `Z_R` and `k_{R}^{P}` are both 32 bytes, and they could overflow.

packages/crypto/src/Bip32/Bip32KeyDerivation.ts
```diff
@@ -0,0 +1,175 @@
...
+  data.writeUInt32LE(index, 1 + 32);
+
+  if (isHardenedDerivation(index))
+    throw new InvalidArgumentError('index', 'Public key can not be derived from a hardened index.');
+
+  pk.copy(data, 1);
+  data[0] = 0x02;
+  const z = crypto_auth_hmacsha512(data, cc);
+  data[0] = 0x03;
+  const i = crypto_auth_hmacsha512(data, cc);
```
iquerejeta, line 165: why is this variable called `i` and not `c`?

packages/crypto/src/Bip32/Bip32PrivateKey.ts
```diff
@@ -0,0 +1,182 @@
...
+ *
+ *  1. clearing the 3 lower bits.
+ *  2. clearing the three highest bits.
+ *  3. setting the second-highest bit.
+ *
+ * @param scalar The clamped scalar.
+ */
+const clampScalar = (scalar: Buffer): Buffer => {
+  scalar[0] &= 0b1111_1000;
+  scalar[31] &= 0b0001_1111;
```
iquerejeta, line 29: why are we clearing the highest three bits? The paper says `the highest bit of the last byte is cleared`, not the "three highest bits".

iquerejeta (commented):
First round of comments.

mkazlauskas (changes requested):
Some additional refactoring is needed:
- typedHex() should probably live in `util` package just like opaqueTypes (seems to be duplicated in core & crypto packages right now)
- Base hash types (e.g. `Hash32ByteBase16`) should probably live in `crypto` package

packages/crypto/src/Bip32/Bip32KeyDerivation.ts
iquerejeta: The savings are not massive (though it is one of the most expensive cryptographic operations). The reason why I said this was because I saw the use of `crypto_scalarmult_ed25519_base_noclamp` in the signature function as well, and thought that it could save some computation to simply store it as part of the extended private key for instance. However, performance does not seem to be paramount to the point of wanting to reduce 50us, so it should be ok.

packages/crypto/src/Bip32/Bip32KeyDerivation.ts
iquerejeta: I'm not sure I fully understood the problem. You could also use `crypto_core_ed25519_add` from libsodium for the arithmetic. And what format are you trying to decode the point to?

packages/crypto/src/Bip32/Bip32KeyDerivation.ts
iquerejeta: yes, this is correct. If the function trunks in case its larger, it is good. I'm no expert of the language, and my understanding was that the only thing we were doing was pad with zeros. 

But now that I look into it again, I see that we are taking the `subarray(0,32)`.
All good 👍

packages/crypto/src/Bip32/Bip32PrivateKey.ts
iquerejeta: But that is due to another condition of the paper that says `If the third highest bit of the last byte of kL is not zero, discard ek.` So basically, if that third bit is not zero, we discard the key. Some libraries seem to _force_ that condition, but it is not what the paper says. I'll check what we use in the node. 

Would it be too much of a pain to _discard_ a key?

iquerejeta: Actually, I can't really understand why they say in the paper that the key should be discarded. Let me think about it.

rhyslbw (changes requested):
Let's add a notification in the `crypto` package README, warning users the implementation has not yet been audited, and use at this stage is at own risk. Can also add this in the `package.json` description.

packages/crypto/src/Bip32/Bip32PrivateKey.ts
rhyslbw: @AngelCastilloB can you capture this and any other issues in a new MD file, so we can close out this PR without missing any context. Once the audit is complete we can replace this file with a report.

packages/crypto/README.md
```diff
@@ -2,3 +2,7 @@
 
 This package provides a set of high level primitives to perform hashing, signature generation/verification, import private keys from BIP39
 mnemonics, and derive BIP32-Ed25519 extended signing keys.
+
+| :warning: WARNING                                                                                              |
```
rhyslbw, line 6: Try https://github.com/orgs/community/discussions/16925

rhyslbw (changes requested):

rhyslbw (approved):

PR #579, opened 2023-01-17
fix(cardano-services): stake pool query by metrics results no longer …

iccicci:
# Context

The `DbSyncStakePoolProvider` in case of query sorted by metrics or apy, in case the requested pagination parameters exceed the total stake pools count, fills the page anyway.
This behavior was introduced as workaround when some pools couldn't be found by the query.
Another problem of the same workaround is that, in such cases, the result does not respect the requested sort.

# Proposed Solution

Being the problem no longer live, the workaround was removed.


iccicci:
Moving this to a new PR

PR #580, opened 2023-01-18
initial work on ada handle

greatertomi:
# Context

# Proposed Solution

### DIFF
--- a/packages/projection/src/operators/certificates/index.ts
+++ b/packages/projection/src/operators/certificates/index.ts
@@ -1,3 +1,4 @@
 export * from './withCertificates';
 export * from './withStakeKeys';
 export * from './withStakePools';
+export * from './withAdaHandle';
--- a/packages/projection/src/operators/certificates/withAdaHandle.ts
+++ b/packages/projection/src/operators/certificates/withAdaHandle.ts
@@ -1,0 +1,15 @@
+import { WithCertificates } from './withCertificates';
+import { WithEpochNo } from '../withEpochNo';
+import { unifiedProjectorOperator } from '../utils';
+
+export interface WithAdaHandle {
+  handle: {
+    address: string;
+    quantity: number;
+  };
+}
+
+export const withAdaHandle = unifiedProjectorOperator<WithCertificates & WithEpochNo, WithAdaHandle>((evt) =>
+  // console.log('evt', evt);
+  ({ ...evt, handle: { address: '', quantity: 0 } })
+);
--- a/packages/projection/src/projections/adaHandle.ts
+++ b/packages/projection/src/projections/adaHandle.ts
@@ -1,0 +1,6 @@
+import { createProjection } from './utils';
+import { withAdaHandle, withCertificates, withEpochNo } from '../operators';
+
+export const adaHandles = createProjection((evt$) => evt$.pipe(withEpochNo(), withCertificates(), withAdaHandle()));
+
+export type AdaHandleProjection = typeof adaHandles;


packages/projection/src/operators/certificates/withAdaHandle.ts
```diff
@@ -0,0 +1,15 @@
+import { WithCertificates } from './withCertificates';
+import { WithEpochNo } from '../withEpochNo';
+import { unifiedProjectorOperator } from '../utils';
+
+export interface WithAdaHandle {
+  handle: {
+    address: string;
+    quantity: number;
```
rhyslbw, line 8:
```suggestion
    handle: string;
```

packages/projection/src/operators/certificates/withAdaHandle.ts
```diff
@@ -0,0 +1,15 @@
...
+  };
+}
+
+export const withAdaHandle = unifiedProjectorOperator<WithCertificates & WithEpochNo, WithAdaHandle>((evt) =>
```
rhyslbw, line 12:
```suggestion
export const withAdaHandle = unifiedProjectorOperator<WithAdaHandle>((evt) =>
```

packages/projection/src/projections/adaHandle.ts
```diff
@@ -0,0 +1,6 @@
+import { createProjection } from './utils';
+import { withAdaHandle, withCertificates, withEpochNo } from '../operators';
+
+export const adaHandles = createProjection((evt$) => evt$.pipe(withEpochNo(), withCertificates(), withAdaHandle()));
```
rhyslbw, line 4:
```suggestion
export const adaHandles = createProjection((evt$) => evt$.pipe(withAdaHandle()));
```

packages/projection/src/operators/certificates/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './withCertificates';
 export * from './withStakeKeys';
 export * from './withStakePools';
+export * from './withAdaHandle';
```
mkazlauskas, line 4: Please create a new `operators/misc` directory for this
```suggestion
export * from './misc';
```

.gitignore
```diff
@@ -10,6 +10,7 @@ lerna-debug.log
 !.yarn/releases
 !.yarn/sdks
 !.yarn/versions
+.idea
```
mkazlauskas, line 13: Please extract this change to a separate `chore:` typed commit

packages/projection/src/operators/certificates/withAdaHandle.ts
```diff
@@ -0,0 +1,12 @@
+import { WithCertificates } from './withCertificates';
+import { unifiedProjectorOperator } from '../utils';
+
+export interface WithAdaHandle {
+  address: string;
+  handle: string;
+}
```
mkazlauskas, line 7: Let's generalize this and convert it to array, because block can have multiple handles:
```suggestion
export interface AddressAlias {
  address: string;
  alias: string;
  protocol: string;
}
export interface WithAddressAliases {
  addressAliases: AddressAlias[];
}
```

then we can have
```ts
export interface AdaHandle extends AddressAlias {
  address: string;
  alias: string;
  protocol: 'AdaHandle';
}
```

packages/projection/src/operators/certificates/withAdaHandle.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const withAdaHandle = unifiedProjectorOperator<WithCertificates, WithAdaHandle>((evt) =>
```
mkazlauskas, line 9: The 1st generic argument is for dependencies. This operator will probably not need any?
```suggestion
export const withAdaHandle = unifiedProjectorOperator<{}, WithAdaHandle>((evt) =>
```

packages/projection/src/sinks/inMemory/types.ts
```diff
@@ -10,6 +10,7 @@ export type InMemoryStore = {
       retirements: PoolRetirement[];
     }
   >;
+  adaHandles: Map<Cardano.Address, string>;
```
mkazlauskas, line 13: In order to implement RollBackward in `sink()` we need to store history of handle owners (e.g. an array): rolling back a transaction that transfers ownership of a handle should result in re-assigning that handle to the previous owner. Also, shouldn't it be a mapping from handle to address (not the other way around)?
```suggestion
  // type AddressAlias = string;
  // type AddressAliasProtocol = string;
  addressAliases: Map<AddressAlias, Map<AddressAliasProtocol, Cardano.Address[]>>;
```

mkazlauskas (changes requested):
Thanks @greatertomi , good start. We're using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) as our commit message format, please take a look at SDK's commit history for some examples.

packages/projection/src/operators/certificates/withAdaHandle.ts
rhyslbw: The protocol should actually be `HandleStandard` 
image
and I think we should be able to avoid using the brand name in our codebase. The handle standard can be expressed as: Resolve the address of the asset's owner, by providing the `assetName` of a known policy. The `policyId` should be a service configuration that is included in a store of assets.

PR #581, opened 2023-01-20
fix(cardano-services): Fix cache TTL validator

IvayloAndonov:
# Context
Currently, both [DbCacheTtl](https://github.com/input-output-hk/cardano-js-sdk/blob/e9c3a5b74a2cc1f69c53f562af8a5b4e693bd20a/packages/cardano-services/src/Program/Options.ts#L4) CLI arg and [InMemoryCache](https://github.com/input-output-hk/cardano-js-sdk/blob/670d0c12589a4385b72c445e5f332e195f3b8000/packages/cardano-services/src/InMemoryCache/InMemoryCache.ts#L10-L16) are aligned and accept the TTL value in seconds since the latest PR [update](https://github.com/input-output-hk/cardano-js-sdk/pull/530/commits/f0eb80f73e61ea48f10809fb3c329fb5c4022e6b).

Noticed we forgot to update the [cacheTtlValidator](https://github.com/input-output-hk/cardano-js-sdk/blob/e9c3a5b74a2cc1f69c53f562af8a5b4e693bd20a/packages/cardano-services/src/util/validators.ts#L37-L46) itself - there is no need to transform minutes to seconds anymore: `cacheTtl * 60` is obsolete

We need to remove the conversion from seconds to minutes and cover it with tests.

- removed the obsolete seconds to minutes conversion
- added tests

### DIFF
--- a/packages/cardano-services/test/util/validators.test.ts
+++ b/packages/cardano-services/test/util/validators.test.ts
@@ -1,0 +1,37 @@
+import {
+  CACHE_TTL_LOWER_LIMIT,
+  CACHE_TTL_UPPER_LIMIT,
+  MissingProgramOption,
+  ProgramOptionDescriptions,
+  ServiceNames
+} from '../../src';
+import { cacheTtlValidator } from '../../src/util/validators';
+
+describe('utils/validators', () => {
+  describe('cacheTtlValidator', () => {
+    it('returns TTL in seconds with a valid value within the range', async () => {
+      const ttl = '240';
+      expect(cacheTtlValidator(ttl)).toEqual(Number.parseInt(ttl, 10));
+    });
+
+    it('throws a validation error if TTL is not a valid number', async () => {
+      expect(() => cacheTtlValidator('not a number')).toThrowError(
+        new MissingProgramOption(ServiceNames.NetworkInfo, ProgramOptionDescriptions.DbCacheTtl)
+      );
+    });
+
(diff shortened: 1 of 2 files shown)


iccicci (dismissed):
Concise and precise

rhyslbw (approved):
Great work @IvayloAndonov :european_castle:

packages/cardano-services/test/util/validators.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  CACHE_TTL_UPPER_LIMIT,
+  MissingProgramOption,
+  ProgramOptionDescriptions,
+  ServiceNames
+} from '../../src';
+import { cacheTtlValidator } from '../../src/util/validators';
+
+describe('utils/validators', () => {
+  describe('cacheTtlValidator', () => {
+    it('returns ttl in seconds with a valid value within the range', async () => {
```
rhyslbw, line 12:
```suggestion
    it('returns TTL in seconds with a valid value within the range', async () => {
```
and the other descriptions

rhyslbw (changes requested):
Sorry, missed this before submitting approval. Can squash and force push

packages/cardano-services/test/util/validators.test.ts
IvayloAndonov: It's done, force-pushed.

rhyslbw (approved):

PR #584, opened 2023-01-23
Fix/wallet manager worker unassigned runtime

mirceahasegan:
# Context

WalletManagerWorker.runtime is never assigned the value provided in the constructor, which 

# Proposed Solution
Added unit test to catch this problem.
Added  no-unused-expressions typescript eslint rule to catch this kind of errors.

### DIFF
--- a/packages/util/test/freeable.test.ts
+++ b/packages/util/test/freeable.test.ts
@@ -81,7 +81,6 @@ describe('freeable', () => {
         resolve(returnValue);
         await expect(result).resolves.toBe(returnValue);
         expect(freeableSpy).toBeCalledTimes(1);
-        reject;
       });
 
       it('returns promise and frees scope after promise rejects', async () => {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,6 +16,7 @@ module.exports = {
     '@typescript-eslint/explicit-module-boundary-types': 0,
     '@typescript-eslint/no-non-null-assertion': 0,
     '@typescript-eslint/no-shadow': ['error'],
+    '@typescript-eslint/no-unused-expressions': ['error', { allowShortCircuit: true, allowTernary: true }],
     '@typescript-eslint/no-unused-vars': ['warn', { argsIgnorePattern: '^_', varsIgnorePattern: '^_' }],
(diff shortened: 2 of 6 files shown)


packages/util/test/freeable.test.ts
```diff
@@ -81,6 +81,7 @@ describe('freeable', () => {
         resolve(returnValue);
         await expect(result).resolves.toBe(returnValue);
         expect(freeableSpy).toBeCalledTimes(1);
+        // eslint-disable-next-line @typescript-eslint/no-unused-expressions
         reject;
```
mkazlauskas, line 85: This doesn't seem to be doing anything, maybe just delete it?

mkazlauskas (changes requested):
I'm wondering how it was even working, is `runtime` available as global in the extension? :thinking:

packages/util/test/freeable.test.ts
mirceahasegan: I was unsure wether it does something or not 😄

lucas-barros:
> I'm wondering how it was even working, is `runtime` available as global in the extension? 🤔

@mkazlauskas in the extension it is working because the `consumeRemoteApi` resolves to `_backgroundConsumeRemoteApi` which calls `getBackgroundMessenger` and returns the `backgroundMessenger` if it was already created. For desktop the `consumeRemoteApi` will resolve to `_nonBackgroundConsumeRemoteApi` which doesnt use `getBackgroundMessenger` and always creates the `backgroundMessenger` and has the runtime as dependency.

packages/util/test/freeable.test.ts
mirceahasegan: Updated in [a56da0f](https://github.com/input-output-hk/cardano-js-sdk/pull/584/commits/a56da0ffc0122fb0e48d6478f3f811d0d8b7e89d)

rhyslbw (approved):

mkazlauskas (dismissed):
Nice work!

mirceahasegan:
squashed and rebased on top of master

mkazlauskas (approved):

rhyslbw (approved):

PR #591, opened 2023-02-02
feat(cardano-services): include healthCheck response as detail in Provider.Unhealthy errors

rhyslbw:
# Context
We're currently lacking detail to help understand the reason for throwing `Provider.Unhealthy` errors.

image

# Proposed Solution
Include the `healthCheck` response as stringified JSON in the error detail. I've elected to leave [this](https://github.com/input-output-hk/cardano-js-sdk/blob/1058e5491d19b34789289826250055f6728ca148/packages/cardano-services/src/TxSubmit/TxSubmitHttpService.ts#L41-L42) instance alone due to the error handling logic being different and the intent to refactor it anyway.

``` console
{"name":"http-server","hostname":"741e2e41edef","pid":1,"level":30,"msg":"[OgmiosCardanoNode] Initialized","time":"2023-02-02T12:26:08.726Z","v":0}
{"name":"http-server","hostname":"741e2e41edef","pid":1,"level":30,"msg":"[asset] Initializing...","time":"2023-02-02T12:26:08.726Z","v":0}
ProviderError: UNHEALTHY ({
  "localNode": {
    "ledgerTip": {
      "slot": 12747005,
      "hash": "cb4b27f6096e6eb2b83befddb73613418409a4c84b386b7ad0ecc1e7fe2d8c41",
      "blockNo": 274973
    },
    "networkSync": 0.67554
  },
  "ok": false,
  "projectedTip": {
    "blockNo": 267674,
    "hash": "7137509e0209cdf3cdefaa04db2807df91238a71e0df25c8ba902daef7b44b41",
    "slot": 12594484
  }
})
```

### DIFF
--- a/packages/cardano-services/src/Http/HttpService.ts
+++ b/packages/cardano-services/src/Http/HttpService.ts
@@ -41,8 +41,9 @@ export abstract class HttpService extends RunnableModule {
   async initializeImpl(): Promise<void> {
     if (this.provider instanceof RunnableModule) await this.provider.initialize();
 
-    if (!(await this.healthCheck()).ok) {
-      throw new ProviderError(ProviderFailure.Unhealthy);
+    const health = await this.healthCheck();
+    if (!health.ok) {
+      throw new ProviderError(ProviderFailure.Unhealthy, null, JSON.stringify(health, null, 2));
     }
   }
 
--- a/packages/cardano-services/src/TxSubmit/rabbitmq/TxSubmitWorker.ts
+++ b/packages/cardano-services/src/TxSubmit/rabbitmq/TxSubmitWorker.ts
@@ -156,9 +156,9 @@ export class TxSubmitWorker extends EventEmitter {
     }
     this.#dependencies.logger.info(`${moduleName} init: checking tx submission provider health status`);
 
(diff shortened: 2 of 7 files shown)


mkazlauskas (dismissed):
Mixed feelings having this as part of error message. Separate property isn't ideal too though, because it wouldn't make sense for other ProviderErrors

rhyslbw:
> Mixed feelings having this as part of error message. Separate property isn't ideal too though, because it wouldn't make sense for other ProviderErrors

What about if we make all error messages structured JSON @mkazlauskas?

mkazlauskas:
> What about if we make all error messages structured JSON @mkazlauskas?

Don't think it's a good idea, best to be compatible with [Error.prototype.message](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/message). Let's just keep the approach you implemented here, it's fine for the use case we have (logging)

mkazlauskas (dismissed):

iccicci (approved):
I agree we could do something more, but the risk to waste time overengineering is very high.

This solves the problem we currently have, so I'm fine.

PR #594, opened 2023-02-03
Add support for CIP-0025 version 2

iccicci:
# Context

We are currently supporting CIP-0025 v1, but not v2

# Proposed Solution

Changed `metadatumToCip25` to support CIP-0025 version 2 as well.

-  added the development version of local-network;
- NFT metadata are no longer read from DB in json format as it can be prone to ambiguity; now are read in raw bytes format;
- added an e2e test to assert CIP-0025 v2 is supported.

### DIFF
--- a/packages/cardano-services/src/Metadata/mappers.ts
+++ b/packages/cardano-services/src/Metadata/mappers.ts
@@ -1,10 +1,21 @@
-import { Cardano, ProviderUtil } from '@cardano-sdk/core';
+import { CML, Cardano, cmlToCore } from '@cardano-sdk/core';
 import { TxMetadataModel } from './types';
+import { usingAutoFree } from '@cardano-sdk/util';
 
-export const mapTxMetadata = (metadataModel: Pick<TxMetadataModel, 'json_value' | 'key'>[]): Cardano.TxMetadata =>
+export const mapTxMetadata = (metadataModel: Pick<TxMetadataModel, 'bytes' | 'key'>[]): Cardano.TxMetadata =>
   metadataModel.reduce((map, metadatum) => {
-    const { key, json_value } = metadatum;
-    if (!json_value || !key) return map;
-    map.set(BigInt(key), ProviderUtil.jsonToMetadatum(json_value));
+    const { bytes, key } = metadatum;
+
+    if (bytes && key) {
+      const biKey = BigInt(key);
+      const metadata = usingAutoFree((_) => cmlToCore.txMetadata(CML.GeneralTransactionMetadata.from_bytes(bytes)));
+
+      if (metadata) {
+        const datum = metadata.get(biKey);
+
+        if (datum) map.set(biKey, datum);
+      }
+    }
+
     return map;
   }, new Map<bigint, Cardano.Metadatum>());
--- a/packages/core/src/Asset/util/metadatumToCip25.ts
+++ b/packages/core/src/Asset/util/metadatumToCip25.ts
@@ -2,7 +2,6 @@ import { AssetInfo, ImageMediaType, MediaType, NftMetadata, NftMetadataFile, Uri
 import { Cardano } from '../..';
 import { CustomError } from 'ts-custom-error';
 import { Logger } from 'ts-log';
-import { Metadatum, MetadatumMap } from '../../Cardano/types/AuxiliaryData';
 import { asMetadatumArray, asMetadatumMap } from '../../util/metadatum';
 import { assetIdFromPolicyAndName } from './assetId';
 import { isNotNil } from '@cardano-sdk/util';
@@ -18,7 +17,7 @@ const asString = (obj: unknown): string | undefined => {
   }
 };
 
-const asStringArray = (metadatum: Metadatum | undefined): string[] | undefined => {
+const asStringArray = (metadatum: Cardano.Metadatum | undefined): string[] | undefined => {
   if (Array.isArray(metadatum)) {
     const result = metadatum.map(asString);
     if (result.some((str) => typeof str === 'undefined')) {
@@ -32,13 +31,13 @@ const asStringArray = (metadatum: Metadatum | undefined): string[] | undefined =
   }
 };
 
-const mapOtherProperties = (metadata: MetadatumMap, primaryProperties: string[]) => {
+const mapOtherProperties = (metadata: Cardano.MetadatumMap, primaryProperties: string[]) => {
   const extraProperties = difference([...metadata.keys()].filter(isString), primaryProperties);
   if (extraProperties.length === 0) return;
   return extraProperties.reduce((result, key) => {
     result.set(key, metadata.get(key)!);
     return result;
-  }, new Map<string, Metadatum>());
+  }, new Map<string, Cardano.Metadatum>());
 };
 
 const toArray = <T>(value: T | T[]): T[] => (Array.isArray(value) ? value : [value]);
@@ -46,7 +45,7 @@ const toArray = <T>(value: T | T[]): T[] => (Array.isArray(value) ? value : [val
 const missingFileFieldLogMessage = (fieldType: string, assetId: Cardano.AssetId) =>
   `Omitting cip25 metadata file: missing "${fieldType}". AssetId: ${assetId}`;
 
-const mapFile = (metadatum: Metadatum, assetId: Cardano.AssetId, logger: Logger): NftMetadataFile | null => {
+const mapFile = (metadatum: Cardano.Metadatum, assetId: Cardano.AssetId, logger: Logger): NftMetadataFile | null => {
   const file = asMetadatumMap(metadatum);
   if (!file) throw new InvalidFileError();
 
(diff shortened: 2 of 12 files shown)


packages/cardano-services/src/Metadata/mappers.ts
```diff
@@ -1,10 +1,21 @@
...
+export const mapTxMetadata = (metadataModel: Pick<TxMetadataModel, 'bytes' | 'key'>[]): Cardano.TxMetadata =>
   metadataModel.reduce((map, metadatum) => {
-    const { key, json_value } = metadatum;
-    if (!json_value || !key) return map;
-    map.set(BigInt(key), ProviderUtil.jsonToMetadatum(json_value));
+    const { bytes, key } = metadatum;
+
+    if (bytes && key) {
+      const biKey = BigInt(key);
+      const datumMap = usingAutoFree((_) => cmlToCore.txMetadatum(CML.TransactionMetadatum.from_bytes(bytes)));
```
mkazlauskas, line 11: - `datum` refers to an output field in Cardano, so best not to use it here
- Object that's being parsed here appears to have label, so we should be mapping the correct core object here (`Metadata` vs `Metadatum`)

```suggestion
      const metadata = usingAutoFree((_) => cmlToCore.txMetadata(CML.TransactionMetadatum.from_bytes(bytes)));
```

packages/core/src/Asset/util/metadatumToCip25.ts
```diff
@@ -85,28 +84,67 @@ const mapFile = (metadatum: Metadatum, assetId: Cardano.AssetId, logger: Logger)
...
-const getAssetMetadata = (policy: MetadatumMap, asset: Pick<AssetInfo, 'name'>) =>
-  asMetadatumMap(policy.get(asset.name.toString()) || policy.get(Buffer.from(asset.name, 'hex').toString('utf8')));
+const getPolicyMetadata = (policy: Cardano.MetadatumMap, policyId: Cardano.PolicyId) => {
+  const policyIdString = policyId.toString();
+
+  return asMetadatumMap(
+    policy.get(policyIdString) ||
+      (() => {
+        for (const [key, value] of policy.entries()) {
+          if (key instanceof Uint8Array && Buffer.from(key).toString('hex') === policyIdString) return value;
```
mkazlauskas, line 103: This is a more reliable check. Same change request for `getAssetMetadata`.
```suggestion
          if (ArrayBuffer.isView(key) && Buffer.from(key).toString('hex') === policyIdString) return value;
```

mkazlauskas (changes requested):
`packages/e2e/test/wallet/SingleAddressWallet/cip0025-v2.test.ts` is also testing v1 - should probably rename. Maybe consolidate with `nft.test.ts`?

packages/core/src/Asset/util/metadatumToCip25.ts
iccicci: 9909f4149 should solve this

packages/cardano-services/src/Metadata/mappers.ts
iccicci: 7c51e61d6 should solve this

mkazlauskas (approved):
Nice work!

iccicci:
> `packages/e2e/test/wallet/SingleAddressWallet/cip0025-v2.test.ts` is also testing v1 - should probably rename. Maybe consolidate with `nft.test.ts`?

@mkazlauskas , ee5d93170 and 23743b100 should solve this

packages/core/src/Asset/util/metadatumToCip25.ts
```diff
@@ -85,28 +84,67 @@ const mapFile = (metadatum: Metadatum, assetId: Cardano.AssetId, logger: Logger)
...
 
 /**
- * Also considers asset name encoded in utf8 within metadata valid
+ * Gets the `Map<AssetName, NFTMetadata>` relative to the given policyId from the given
+ * `Map<PolicyId, Map<AssetName, NFTMetadata>>`.
+ *
+ * The policyId in the `policy` Map can be encoded as per CIP-0025 v1 or v2 specifications.
+ *
+ * @param policy The `MetadatumMap` containing the NFT metadata for all the NFT assets
+ * @param policyId The policyId of the assent looking for
```
rhyslbw, line 93:
```suggestion
 * @param policyId
```
This is self-descriptive. Is it legal based on our eslint rules to do this?

packages/core/src/Asset/util/metadatumToCip25.ts
```diff
@@ -85,28 +84,67 @@ const mapFile = (metadatum: Metadatum, assetId: Cardano.AssetId, logger: Logger)
...
+  );
+};
+
+/**
+ * Gets the `NFTMetadata` relative to the given assetName from the given `Map<AssetName, NFTMetadata>`.
+ *
+ * The assetName in the `policy` Map can be encoded as per CIP-0025 v1 (hex or utf8) or v2 specifications.
+ *
+ * @param policy The `MetadatumMap` containing the NFT metadata for all the NFT assets with a specific policyId.
+ * @param assetName The assetName of the assent looking for.
```
rhyslbw, line 115:
```suggestion
 * @param assetName
```

packages/e2e/README.md
```diff
@@ -113,6 +113,39 @@ Address:    addr_test1qr0c3frkem9cqn5f73dnvqpena27k2fgqew6wct9eaka03agfwkvzr0zyq
 
 You can configure any of these five wallets in your test and use any amount of tADA you need.
 
+### Local Test Network - development version
+
+In addition to **local-network** there is its development version, which mounts the `packages` directory in the
```
rhyslbw, line 118:
```suggestion
In addition to the **local-network** npm script there is a script suitable for development, which mounts the `packages` directory in
```

packages/e2e/README.md
```diff
@@ -113,6 +113,39 @@ Address:    addr_test1qr0c3frkem9cqn5f73dnvqpena27k2fgqew6wct9eaka03agfwkvzr0zyq
...
+http-server container and in the worker container. It helps to apply changes to the two components on a running
```
rhyslbw, line 119:
```suggestion
the service containers. It helps to apply changes to the components without shutting down the **local-network** each time.
```

packages/e2e/README.md
```diff
@@ -113,6 +113,39 @@ Address:    addr_test1qr0c3frkem9cqn5f73dnvqpena27k2fgqew6wct9eaka03agfwkvzr0zyq
...
+**local-network** whithout the need to shutdown and restart it.
+
+It can be started with:
+
+```bash
+$ yarn workspace @cardano-sdk/e2e local-network:dev
+```
+
+and a good idea is to run following command before starting it.
```
rhyslbw, line 128: I would make this a `prelocal-network:dev` script, then remove this manual step.

packages/e2e/README.md
```diff
@@ -113,6 +113,39 @@ Address:    addr_test1qr0c3frkem9cqn5f73dnvqpena27k2fgqew6wct9eaka03agfwkvzr0zyq
...
+
+```bash
+$ yarn cleanup && yarn && yarn build
+```
+
+Once it is running, to restart the http-server it is enough to issue the following command.
```
rhyslbw, line 134:
```suggestion
Restart the http-server:
```

rhyslbw (changes requested):
Looks good @iccicci. Few minor suggestions

packages/core/src/Asset/util/metadatumToCip25.ts
iccicci: Nope @rhyslbw , if I remove the `@param` description, eslint is not happy.
It seems we must provide at least a simple description.

packages/e2e/README.md
iccicci: af738bc2c should solve this

packages/e2e/README.md
iccicci: af738bc2c should solve this

packages/e2e/README.md
iccicci: af738bc2c should solve this

packages/core/src/Asset/util/metadatumToCip25.ts
mkazlauskas: I think you can remove the entire @param line. Docs should be able to figure out that there is a param from typescript

iccicci: I've checked and @mkazlauskas 's suggestion seems to be a good solution.
Being the function not exported it is not included in the `typedoc` output.
I've tried running it exporting the function and the produced doc was fine even if without the parameter description

f9eceac70 should solve this

packages/core/src/Asset/util/metadatumToCip25.ts
iccicci: f9eceac70 should solve this

rhyslbw (dismissed):

mkazlauskas (approved):

rhyslbw (approved):
Great work @iccicci :muscle:

PR #596, opened 2023-02-07
fix: ledger mapper validity interval prop name

mkazlauskas:
# Context

Ledger was signing a different transaction hash due to incorrect property name when mapping transaction props

# Proposed Solution

Fix it to the correct property name and also add a check to throw an error if Ledger signs a different transaction hash

A few chores that fix minor issues with SDK setup:
- 389f8d4eff9c29543ff4a14c282ec87107635bd7
- 5a96d0b2f0829159b0046848cb8c7473cbdccfdd

### DIFF
--- a/packages/key-management/src/LedgerKeyAgent.ts
+++ b/packages/key-management/src/LedgerKeyAgent.ts
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import * as Crypto from '@cardano-sdk/crypto';
-import { AuthenticationError, TransportError } from './errors';
+import { AuthenticationError, HwMappingError, TransportError } from './errors';
 import { Cardano, NotImplementedError, coreToCml } from '@cardano-sdk/core';
 import {
   CardanoKeyConst,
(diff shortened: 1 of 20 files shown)


rhyslbw (dismissed):
Nice work @mkazlauskas. Great improvement to verifying the hardware result to fail early

mirceahasegan (dismissed):
Nice work!

packages/key-management/src/LedgerKeyAgent.ts
```diff
@@ -239,6 +239,9 @@ export class LedgerKeyAgent extends KeyAgentBase {
         this.deviceConnection
       );
       const result = await deviceConnection.signTransaction(ledgerTxData);
+      if (result.txHashHex !== hash.toString()) {
```
iadmytro, line 242: It would be good to add test for this case.

rhyslbw: It's a tricky setup in a unit test, for not that much value, as this exception will be covered by integration tests that demonstrate we are able to construct transactions for each property.

mkazlauskas: Added a test in 2ba7be9
Also fixed existing test, it was failing due to using different (mocked) output addresses and producing a different transaction hash

rhyslbw (approved):

iadmytro (dismissed):

PR #599, opened 2023-02-09
fix/cip30 addresses as cbor instead of bech32

mirceahasegan:
# Context

Most DAPPs expect and use CBOR addresses.

# Proposed Solution

Return cip30 addresses as Cbor (hex string) instead of Cardano.Address (bech32)


packages/wallet/src/cip30.ts
```diff
@@ -74,6 +74,15 @@ const compareUtxos = (utxo: Cardano.Utxo, comparedTo: Cardano.Utxo) => {
   return 0;
 };
 
+const cardanoAddressToCbor = (address: Cardano.Address | Cardano.RewardAccount): Cbor =>
+  usingAutoFree((scope) => {
+    const cmlAddr = parseCmlAddress(scope, address.toString());
```
mirceahasegan, line 79: I am not 100% sure this is the right way to do it for reward accounts too. @mkazlauskas, do you know if this is correct?

packages/wallet/src/cip30.ts
```diff
@@ -172,7 +181,7 @@ export const createWalletApi = (
       if (!rewardAccount) {
         throw new ApiError(APIErrorCode.InternalError, 'could not get reward address');
       } else {
-        return [rewardAccount.toString()];
+        return [cardanoAddressToCbor(rewardAccount)];
```
mirceahasegan, line 184: Is this a breaking change? Considering the return value was already specified as `Promise<Cbor[]>`?

packages/wallet/src/cip30.ts
mkazlauskas: Yes this seems to be the correct way to do it for reward accounts too. Would be best to test it e2e with a dapp

packages/wallet/src/cip30.ts
mkazlauskas: Yes, it should be pointed out in the commit message. It was a bug, wouldn't have happened if Cbor was an OpaqueString :sweat_smile:

packages/wallet/src/cip30.ts
```diff
@@ -74,6 +74,15 @@ const compareUtxos = (utxo: Cardano.Utxo, comparedTo: Cardano.Utxo) => {
   return 0;
 };
 
+const cardanoAddressToCbor = (address: Cardano.Address | Cardano.RewardAccount): Cbor =>
+  usingAutoFree((scope) => {
+    const cmlAddr = parseCmlAddress(scope, address.toString());
+    if (!cmlAddr) {
+      throw new ApiError(APIErrorCode.InternalError, `could transform address ${address} to CBOR`);
```
rhyslbw, line 81:
```suggestion
      throw new ApiError(APIErrorCode.InternalError, `could not transform address ${address} to CBOR`);
```

packages/wallet/src/cip30.ts
rhyslbw: Should we add an OpaqueString now then?

rhyslbw (changes requested):

packages/wallet/src/cip30.ts
mirceahasegan: added e2e test in [cda9ceb](https://github.com/input-output-hk/cardano-js-sdk/pull/599/commits/cda9ceb71622602d026fd3ba637637d419d14d31)

packages/wallet/src/cip30.ts
mirceahasegan: I will mark it resolved after final squash with commit message updated

mkazlauskas (approved):
Nice work!

rhyslbw (dismissed):

packages/wallet/src/cip30.ts
rhyslbw: Should we make a task to convert the alias to an `OpaqueString`?

rhyslbw (approved):
Nice work @mirceahasegan :rocket:

packages/wallet/src/cip30.ts
mirceahasegan: > I will mark it resolved after final squash with commit message updated
Updated the commit message, thus marking this as resolved.

mkazlauskas (approved):

PR #600, opened 2023-02-11
fix: cip30 wallet interface now allows submitting raw CBOR transactions

AngelCastilloB:
# Context

The CIP-30 `submitTx` mapping is deserializing the input, to conform with the `ObservableWallet` interface, which results in a round trip that can invalidate the tx, since it can produce a different CBOR representation because there is no reliable way to encode certain structs; It's implementation-specific. For example, an indefinite length array after deserialization/serialization roundtrip can be a definite length array with the same elements. Also, `TransactionOutput` can be serialized in two ways.

We need to rework the wallet to handling hex-encoded blobs, rather than just modelling everything on expecting the SDK `Tx` object to be the source, then pass the raw value submitted via the CIP-30 mapping instead.

# Proposed Solution

## Accept `TxCBOR` in `ObservableWallet.submitTx`
- This method now can handle `Cardano.Tx` and `TxCBOR`
- If the latter is submitted, only use the converted object for supplementary purposes, with the input sent to the provider without round-tripping through a serialization step.
## CIP30 wallet mapping
- Change the Cip30 wallet mapping to use the `TxCBOR` in the wallet submission invocation, using the deserialized object for supplementary purposes.

- `ObservableWallet.submitTx` now returns the transaction ID

### DIFF
--- a/packages/core/src/CBOR/TxCBOR.ts
+++ b/packages/core/src/CBOR/TxCBOR.ts
@@ -1,0 +1,25 @@
+import { HexBlob, OpaqueString, bufferToHexString, usingAutoFree } from '@cardano-sdk/util';
+import { cmlUtil, coreToCml } from '../CML';
+import type { Cardano } from '..';
+
+/**
+ * Transaction serialized as CBOR, encoded as hex string
+ */
+export type TxCBOR = OpaqueString<'TxCbor'>;
+
+/**
+ * @param tx Serialized as CBOR, encoded as hex string
+ * @throws InvalidStringError
+ */
+export const TxCBOR = (tx: string): TxCBOR => HexBlob(tx) as unknown as TxCBOR;
+
+/**
+ * Serialize transaction to hex-encoded CBOR
+ */
+TxCBOR.serialize = (tx: Cardano.Tx): TxCBOR =>
+  bufferToHexString(Buffer.from(usingAutoFree((scope) => coreToCml.tx(scope, tx).to_bytes()))) as unknown as TxCBOR;
+
+/**
+ * Deserialize transaction from hex-encoded CBOR
+ */
+TxCBOR.deserialize = (tx: TxCBOR): Cardano.Tx => cmlUtil.deserializeTx(tx.toString());
--- a/packages/core/src/CBOR/index.ts
+++ b/packages/core/src/CBOR/index.ts
@@ -1,0 +1,2 @@
+export * from './TxCBOR';
+export * from './TxBodyCBOR';
--- a/packages/wallet/src/SingleAddressWallet/SingleAddressWallet.ts
+++ b/packages/wallet/src/SingleAddressWallet/SingleAddressWallet.ts
@@ -491,7 +491,7 @@ export class SingleAddressWallet implements ObservableWallet {
   async submitTx(
     input: Cardano.Tx | TxCBOR | OutgoingTx,
     { mightBeAlreadySubmitted }: SubmitTxOptions = {}
-  ): Promise<void> {
+  ): Promise<Cardano.TransactionId> {
     const outgoingTx = processOutgoingTx(input);
     this.#logger.debug(`Submitting transaction ${outgoingTx.id}`);
     this.#newTransactions.submitting$.next(outgoingTx);
@@ -502,6 +502,7 @@ export class SingleAddressWallet implements ObservableWallet {
       const { slot: submittedAt } = await firstValueFrom(this.tip$);
       this.#logger.debug(`Submitted transaction ${outgoingTx.id} at slot ${submittedAt}`);
       this.#newTransactions.pending$.next(outgoingTx);
+      return outgoingTx.id;
     } catch (error) {
       if (
         mightBeAlreadySubmitted &&
@@ -514,7 +515,7 @@ export class SingleAddressWallet implements ObservableWallet {
       ) {
         this.#logger.debug(`Transaction ${outgoingTx.id} appears to be already submitted...`);
         this.#newTransactions.pending$.next(outgoingTx);
-        return;
+        return outgoingTx.id;
       }
       this.#newTransactions.failedToSubmit$.next({
         error: error as CardanoNodeErrors.TxSubmissionError,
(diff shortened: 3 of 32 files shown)


packages/wallet/src/SingleAddressWallet/SingleAddressWallet.ts
```diff
@@ -131,15 +133,23 @@ export const DEFAULT_POLLING_CONFIG = {
   pollInterval: 5000
 };
 
+const isCardanoTx = (input: Cardano.Tx | Cardano.SerializedTx): input is Cardano.Tx => {
+  const { id, body, witness } = input as Cardano.Tx;
+  return id !== undefined && body !== undefined && witness !== undefined;
+};
+
+const isSerializedTx = (input: Cardano.Tx | Cardano.SerializedTx): input is Cardano.SerializedTx =>
+  typeof (input as Cardano.SerializedTx) === 'string';
```
mirceahasegan, line 142: This is a bit confusing. We already have `isCardanoTx` narrowing predicate, which, if negated, says that the input is `Cardano.SerializedTx`. However, the `isSerializedTx` narrowing predicate takes a different verification approach.
Ideally, it should be like this:
```diff
-  typeof (input as Cardano.SerializedTx) === 'string';
+  !isCardanoTx(input);

Comment (mkazlauskas) [reply packages/wallet/src/SingleAddressWallet/SingleAddressWallet.ts]: It now has 3 variants and 2 narrowing functions: https://github.com/input-output-hk/cardano-js-sdk/pull/600/commits/b31b69edec251b41788a866bcc3e4dd06530fa48#diff-aeb17003218a7a98485f6c851ac5be47b6876308204497ddd74a552f69fa9393R130-R156

Please review again

File: packages/core/src/CBOR/index.ts [hunk 2]
```diff
@@ -0,0 +1,2 @@
+export * from './TxCBOR';
+export * from './TxBodyCBOR';
```
mkazlauskas, line 2: Could extract all `core` features to a separate commit

packages/wallet/src/cip30.ts
```diff
@@ -312,7 +312,7 @@ export const createWalletApi = (
     if (shouldProceed) {
       try {
         const wallet = await firstValueFrom(wallet$);
-        await wallet.submitTx(txData);
+        await wallet.submitTx(TxCBOR(tx));
```
mkazlauskas, line 315: Need to map InvalidStringError into cip 30 error type

mkazlauskas (changes requested):

packages/core/src/CBOR/index.ts
rhyslbw: Agree. It needs some unit test coverage too

packages/core/src/CBOR/TxCBOR.ts
```diff
@@ -0,0 +1,27 @@
+// To be hoisted to a new package.
```
rhyslbw, line 1: Maybe just make a ticket for this instead

packages/wallet/src/persistence/pouchDbStores/pouchDbWalletStores.ts
```diff
@@ -59,7 +59,7 @@ export const createPouchDbWalletStores = (
     destroyed: false,
     eraSummaries: new PouchDbEraSummariesStore(docsDbName, 'EraSummaries', logger),
     genesisParameters: new PouchDbGenesisParametersStore(docsDbName, 'genesisParameters', logger),
-    inFlightTransactions: new PouchDbInFlightTransactionsStore(docsDbName, 'transactionsInFlight', logger),
+    inFlightTransactions: new PouchDbInFlightTransactionsStore(docsDbName, 'transactionsInFlightV2', logger),
```
rhyslbw, line 62:
```suggestion
    inFlightTransactions: new PouchDbInFlightTransactionsStore(docsDbName, 'transactionsInFlight.v2', logger),
```
Is this a legal name? Might be a little nicer to segment it this way

mkazlauskas: How about using _? I think `.` is has too much potential to cause issues as it's used with different underlying stores. If you feel strongly about '.' I can test it with IndexedDB and LevelDB

rhyslbw: Happy with the underscore too. Just wanting to set a standard now since it's our first version bump

packages/core/src/CBOR/TxCBOR.ts
mkazlauskas: ADP-2687
Will remove inline comment in a fixup commit

packages/core/src/CBOR/index.ts
mkazlauskas: 2fcc3f6806e0638ede2095d4d180e262d9160f43 and did a small refactor for an inconsistency that I noticed in the existing test 21daef2f50cd6b05bc74f1bdc21a36ef4a361657

packages/core/src/CBOR/TxCBOR.ts
mkazlauskas: 2fcc3f6806e0638ede2095d4d180e262d9160f43

packages/wallet/src/cip30.ts
mkazlauskas: 37188c22c30cdce8584fa5899890a8b68ac4f80f

packages/wallet/src/persistence/pouchDbStores/pouchDbWalletStores.ts
mkazlauskas: 43953b4be4f5fd744d3648fb71d6bf54eb384a9a

rhyslbw (approved):

rhyslbw (approved):
Excellent work @mkazlauskas. I can't see any issues, other than the test that I've fixed up.

mkazlauskas (dismissed):

mirceahasegan (approved):

PR #601, opened 2023-02-13
Fix pledge met filter

iccicci:
# Context

Stake pool queries with `pledgeMet` filter exclude some pools from the result.

Found a minor bug on stake pool query sort by name: the `->` operator doesn't convert the field, but only extracts a property as json type. This is originating a minor problem: the double quotes of json representation of string was included in the sort algorithm.

# Proposed Solution

Added a `LEFT` clause to the failing `JOIN`.

Changed `->` to `->>` in the `SORT` clause of stake pools query sort by name.

- added dev version of _mainnet_, _preprod_ and _preview_ docker compose;
- fixed a bug in stake pools query which was causing a division by zero;
- fixed sort by name stake pools query e2e tests to be case-insensitive;


packages/e2e/docker-compose-dev.yml
```diff
@@ -1,9 +1,7 @@
-version: "3.9"
-
 services:
   http-server:
     volumes:
-      - ..:/app/packages
+      - ../..:/app
```
rhyslbw, line 4: These changes are in the wrong commit

rhyslbw (changes requested):
Looks good so far @iccicci

packages/e2e/docker-compose-dev.yml
iccicci: 18b0bc171 plus 8643ce86a should solve this

rhyslbw (approved):
LGTM!

IvayloAndonov (approved):
LGTM as well, good new knowledge for the `->>` operator :)

IvayloAndonov (approved):

PR #602, opened 2023-02-14
Fix/get collateral cip30 return value

mirceahasegan:
# Context

[getCollateral](https://github.com/input-output-hk/cardano-js-sdk/blob/99a57171e9bead364e8751474213a58522d9e271/packages/wallet/src/cip30.ts#L123) currently[ throws when there are no UTxOs](https://github.com/input-output-hk/cardano-js-sdk/blob/99a57171e9bead364e8751474213a58522d9e271/packages/wallet/test/integration/cip30mapping.test.ts#L267-L275), which, based on the return type of Promise<    Cbor[] | null should be returning null.

# Proposed Solution

getCollateral should then return null, not throw an error in this case.

There should no longer be ApiError(APIErrorCode.InternalError, ‘Nope'); errors in the module. The message could be error?.message || 'Unknown error' or better

### DIFF
--- a/packages/wallet/src/cip30.ts
+++ b/packages/wallet/src/cip30.ts
@@ -15,11 +15,9 @@ import {
 } from '@cardano-sdk/dapp-connector';
 import { CML, Cardano, cmlToCore, coreToCml, parseCmlAddress } from '@cardano-sdk/core';
 import { HexBlob, ManagedFreeableScope, usingAutoFree } from '@cardano-sdk/util';
-import { InputSelectionError } from '@cardano-sdk/input-selection';
 import { Logger } from 'ts-log';
 import { Observable, firstValueFrom } from 'rxjs';
 import { ObservableWallet } from './types';
-import { errors } from '@cardano-sdk/key-management';
 
 export type Cip30WalletDependencies = {
   logger: Logger;
@@ -83,6 +81,8 @@ const cardanoAddressToCbor = (address: Cardano.Address | Cardano.RewardAccount):
     return Buffer.from(cmlAddr.to_bytes()).toString('hex');
   });
 
+const formatUnknownError = (error: unknown): string => (error as Error)?.message || 'Unknown error';
+
 export const createWalletApi = (
   wallet$: Observable<ObservableWallet>,
   confirmationCallback: CallbackConfirmation,
@@ -116,7 +116,7 @@ export const createWalletApi = (
       if (error instanceof ApiError) {
         throw error;
       }
-      throw new ApiError(APIErrorCode.InternalError, 'Nope');
+      throw new ApiError(APIErrorCode.InternalError, formatUnknownError(error));
     }
   },
   // eslint-disable-next-line max-statements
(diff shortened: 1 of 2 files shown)


packages/wallet/src/cip30.ts
```diff
@@ -116,7 +118,7 @@ export const createWalletApi = (
       if (error instanceof ApiError) {
         throw error;
       }
-      throw new ApiError(APIErrorCode.InternalError, 'Nope');
+      throw new ApiError(APIErrorCode.InternalError, (error as Error)?.message || UNKNOWN_ERROR);
```
mkazlauskas, line 121: Seems to be duplicated quite a few times
```suggestion
      throw new ApiError(APIErrorCode.InternalError, formatUnknownErrorMessage(error));
```

packages/wallet/test/integration/cip30mapping.test.ts
```diff
@@ -298,7 +298,7 @@ describe('cip30', () => {
         expect(utxos).toHaveLength(1);
       });
       test('throws when there is no given amount and wallet has no UTxOs', async () => {
```
mkazlauskas, line 300:
```suggestion
      test('returns null when there is no given amount and wallet has no UTxOs', async () => {
```

packages/wallet/test/integration/cip30mapping.test.ts
```diff
@@ -270,7 +270,7 @@ describe('cip30', () => {
       });
       test('throws an error when there are no UTxOs in the wallet', async () => {
```
mkazlauskas, line 271:
```suggestion
      test('returns null when there are no "unspendable" UTxOs in the wallet', async () => {
```

mkazlauskas (changes requested):

packages/wallet/test/integration/cip30mapping.test.ts
mirceahasegan: Fixed in [ca144bd](https://github.com/input-output-hk/cardano-js-sdk/pull/602/commits/ca144bd8b8d13c7a628d493ed3dcf073898c4129)

packages/wallet/test/integration/cip30mapping.test.ts
mirceahasegan: [ca144bd](https://github.com/input-output-hk/cardano-js-sdk/pull/602/commits/ca144bd8b8d13c7a628d493ed3dcf073898c4129)

packages/wallet/src/cip30.ts
mirceahasegan: Fixed in [65a3784](https://github.com/input-output-hk/cardano-js-sdk/pull/602/commits/65a3784f00c6574efed79cbdcc530954c50b77d1)

rhyslbw (approved):
Great work @mirceahasegan

mkazlauskas (approved):
Nice work!

mirceahasegan:
Added one more test for unknown errors in [75e3659](https://github.com/input-output-hk/cardano-js-sdk/pull/602/commits/75e36591c855b49b331c97d7a81dd14157764bb1)

mkazlauskas (approved):

rhyslbw (approved):

PR #604, opened 2023-02-15
React native with detox

wklos-iohk:
# Context

https://input-output.atlassian.net/browse/LW-5012

# Proposed Solution


rhyslbw (changes requested):
Thanks for the contribution @wklos-iohk! 

I need to review this in more depth, but from my quick scan, this code should be added to the `e2e` package, as it's here only to support testing IIUC. The only time we'd need a `mobile` packages is if there's code that applications will use. Is that the case?

wklos-iohk:
> Thanks for the contribution @wklos-iohk!
> 
> I need to review this in more depth, but from my quick scan, this code should be added to the `e2e` package, as it's here only to support testing IIUC. The only time we'd need a `mobile` packages is if there's code that applications will use. Is that the case?

@rhyslbw 
It's related to https://input-output.atlassian.net/browse/LW-5012
Should I put the `mobile` folder directly under `/cardano-js-sdk/tree/master/packages/e2e`?

rhyslbw:
@wklos-iohk yes, that's where we we're expecting this :+1:

mkazlauskas:
I think the setup should be the same as with our web-extension test: `packages/e2e/test/mobile`

rhyslbw:
CI is failing @wklos-iohk

PR #605, opened 2023-02-16
Feat/adp 2677 handle collateral consumption and rollback

AngelCastilloB:
# Context

We have no logic for detecting collateral consumption, which will result in transactions being constructed with a UTxO that’s no longer owned by the wallet if a prior transaction produces a phase 2 validation failure.

Chain rollbacks are not being considered either, so a newly created output or consumption could be rolled back without unspendable$ being updated.

 setUnspendable is also not part of the ObservableWallet interface, only the SingleAddressWallet implementation. It needs to be defined in the interface for applications to access via the web extension messaging transport.

# Proposed Solution

1. Unset the unspendables array within the UTxOs tracker by filtering the emitted values by existence in total$ utxo
2. Make setUnspendable asynchronous.
3. Change the type of the utxo property in the ObservableWallet from TransactionalObservables<Cardano.Utxo[]> to UtxoTracker so setUnspendable is available in the 
ObservableWallet interface.

### DIFF
--- a/packages/wallet/src/services/UtxoTracker.ts
+++ b/packages/wallet/src/services/UtxoTracker.ts
@@ -1,6 +1,6 @@
 import { Cardano, UtxoProvider } from '@cardano-sdk/core';
 import { Logger } from 'ts-log';
-import { NEVER, Observable, combineLatest, concat, map, of, switchMap } from 'rxjs';
+import { NEVER, Observable, combineLatest, concat, distinctUntilChanged, map, of, switchMap } from 'rxjs';
 import { PersistentCollectionTrackerSubject, coldObservableProvider, txInEquals, utxoEquals } from './util';
 import { RetryBackoffConfig } from 'backoff-rxjs';
 import { TxInFlight, UtxoTracker } from './types';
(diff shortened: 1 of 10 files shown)


rhyslbw (changes requested):
See if you can separate 7e090914c0dc890bde4d2bc3f3c9ce53f33aebe5 into these commits. There's also a typo in the commit message that can be corrected "funciton".

1. refactor(wallet)!: convert `SingleAddressWallet.setUnspendable` to async
2. feat: expose `setUnspendable` on `ObservableWallet` interface
3. feat(wallet): managed unspendables after collateral consumption or chain rollbacks

mirceahasegan (approved):
Nice work @AngelCastilloB 🧙

mirceahasegan (dismissed):
🚀

packages/wallet/src/services/UtxoTracker.ts
```diff
@@ -119,13 +119,20 @@ export const createUtxoTracker = (
...
     shutdown: () => {
       utxoSource$.complete();
       unspendableUtxoSource$.complete();
       logger.debug('Shutdown');
     },
     total$,
-    unspendable$: unspendableUtxoSource$
+    unspendable$: combineLatest([unspendableUtxoSource$, total$]).pipe(
+      map(([unspendableUtxo, utxo]) =>
+        utxo.filter(([utxoTxIn]) => unspendableUtxo.some(([unspendable]) => txInEquals(utxoTxIn, unspendable)))
```
mkazlauskas, line 133: There will be less `unspendableUtxo` than `total`, so this would be slightly more efficient
```suggestion
        unspendableUtxo.filter(([unspendable]) => utxo.some(([utxoTxIn]) => txInEquals(utxoTxIn, unspendable)))
```

mkazlauskas (approved):
Nice work! :rocket:

packages/wallet/src/services/UtxoTracker.ts
rhyslbw: Done in https://github.com/input-output-hk/cardano-js-sdk/pull/605/commits/e6007465e617698954a61774b3c2a461678c322a (squashed due to time constraints)

mkazlauskas (approved):

rhyslbw (approved):
Fantastic work @AngelCastilloB :muscle:

PR #583, opened 2023-01-23
stake pool apy queries

iccicci:
# Context

The APY output of our APIs seems to be not so precise. We need to investigate/fix it.

# Proposed Solution

Applied some refactoring and fix
1. to make our APIs more ledger compliant
2. to fix some issues identified during the investigation
3. to make the computed APY more accurate

# Important Changes Introduced

- fixed a bug: stake pool query pagination when sorting by `saturation` or `apy` was providing more results than the ones expected;
- fixed the indentation of stake pool queries to improve readability;
- fixed a bug: stake pool cached result was not honoring `rewardsHistoryLimit`;
- reworked `StakePoolEpochRewards` fields to make them more meaningful and ledger compliant;
- fixed a bug: stake pool reward history now contains up to `rewardsHistoryLimit` elements no longer only one;
- refactored the way the genesis file is read to make it available as a dependency for all the providers which need it: preparatory to provide genesis data to `DbSyncStakePoolProvider`;
- fixed a bug: epoch length in stake pool rewards history is now the length from genesis data for all epochs but current one;

PR #608, opened 2023-02-17
Fix apy compute

iccicci:
# Context

While working on #601 we realized that in #583 I gave to `pool_update.pledge` the wrong meaning: we need to use the actual amount delegated to the pool by the owners to compute the APY instead of the value of that field.

The SQL queries we perform around stake pools have some code repetition, one more would be required to address the previous problem.

# Proposed Solution

Introduced some `TEMPORARY VIEW`s to try reduce code repetition.


iccicci:
This is no longer required

PR #613, opened 2023-02-23
Fix/failed tx undetected when undefined invalid hereafter

mirceahasegan:
# Context

In case of undefined invalidHereafter in transaction, failed transactions rxjs pipe
used an EMPTY observable, which completed immediately, causing the whole
observable to close before detecting the failed transaction.

# Proposed Solution
Use NEVER observable since it waits for other conditions (completed, rollback, failed).

None

### DIFF
--- a/packages/wallet/test/services/TransactionsTracker.test.ts
+++ b/packages/wallet/test/services/TransactionsTracker.test.ts
@@ -338,7 +338,9 @@ describe('TransactionsTracker', () => {
         const failedToSubmit$ = hot<FailedTx>('---a|', {
           a: { reason: TransactionFailure.FailedToSubmit, ...outgoingTx }
         });
-        const failedFromReemitter$ = cold<FailedTx>('-a|', { a: { reason: TransactionFailure.Timeout, ...outgoingTxReemit } });
+        const failedFromReemitter$ = cold<FailedTx>('-a|', {
+          a: { reason: TransactionFailure.Timeout, ...outgoingTxReemit }
+        });
         const transactionsTracker = createTransactionsTracker(
           {
             addresses$,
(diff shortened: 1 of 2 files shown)


packages/wallet/test/services/TransactionsTracker.test.ts
```diff
@@ -338,7 +338,9 @@ describe('TransactionsTracker', () => {
         const failedToSubmit$ = hot<FailedTx>('---a|', {
           a: { reason: TransactionFailure.FailedToSubmit, ...outgoingTx }
         });
-        const failedFromReemitter$ = cold<FailedTx>('-a|', { a: { reason: TransactionFailure.Timeout, ...outgoingTxReemit } });
+        const failedFromReemitter$ = cold<FailedTx>('-a|', {
```
rhyslbw, line 341: It's surprising the lint rules allow either of these options (the before and now) :thinking: Seems like we're not enforcing some rules in tests?

rhyslbw (approved):
Nice work @mirceahasegan :european_castle:

packages/wallet/test/services/TransactionsTracker.test.ts
mirceahasegan: The rules are disabled for the whole file. Probably to allow the lengthy test names.

mkazlauskas: If I remember right it was disabled for more readable test setup:
```
cold('------a----', { a: 'some long object definition' });
cold('-------a---');
```
vs
```
cold('------a----', {
  a: 'some long object definition'
});
cold('-------a---');
```
You can understand the setup easier when there are many of those aligned one after another. Even when you use a variable of this long object definition, the names are sometimes too long or there are simply too many of emitted items.

mkazlauskas (approved):
Nice catch! :muscle:

packages/wallet/test/services/TransactionsTracker.test.ts
rhyslbw: OK, it makes sense, I'm just expecting the lint rules to be enforcing style. I wonder if there's one to favour this automatically, but it's not important now.

mirceahasegan: Unfortunately, it was lost at some point, probably due to editors like mine with prettier formatter enabled, which ignores the eslint file rules.

mkazlauskas: Agree @rhyslbw, we should probably just have different rules for 'test/' to not have to eslint-disable as much. We do that a lot in tests. `any` is typically acceptable for example.

PR #614, opened 2023-02-23
refactor: better opaque primitive types

mkazlauskas:
# Context

OpaqueString and OpaqueNumber is not assignable to `number`, which causes some inconveniences when working with SDK.

# Proposed Solution

- Update the types to be assignable, with a side-effect of types having an extra property that doesn't actually exist at runtime (`__opaqueString: T` etc.)
- Remove all redundant `.toString()` and `.valueOf()` calls in SDK


mirceahasegan (approved):
much cleaner this way 🚀

rhyslbw (approved):
Nice refactor @mkazlauskas

PR #616, opened 2023-02-24
feat: pg sinks

mkazlauskas:
# Context

Projection service needs to have persistent sinks

# Proposed Solution

Implement PostgreSQL sinks using typeorm. This PR includes:
- New package setup
- Tests setup (package-level and e2e projection)
- Persistent StabilityWindowBuffer implementation
- `stakeKeys` projection sink as an example of a modular sink

Generated schema:
image

Also includes some refactors and new utilities - see commit log


packages/projection-typeorm/src/entity/Block.entity.ts
```diff
@@ -0,0 +1,25 @@
+import { Cardano } from '@cardano-sdk/core';
+import { Column, Entity, Index, PrimaryColumn, RelationOptions } from 'typeorm';
+
+@Entity({ name: 'block' })
+export class BlockEntity {
+  @PrimaryColumn({ length: 64, type: 'char' })
```
iccicci, line 6: Using a 64 character field as primary key doesn't sound to be a good idea: all other entries referring to this table will need to spend a lot of space to refer to this table.
A 32 bytes integer field (up to 2 billions blocks) as primary key for this table would be much more efficient.

packages/projection-typeorm/src/entity/StakeKey.entity.ts
```diff
@@ -0,0 +1,8 @@
+import { Ed25519KeyHashHex } from '@cardano-sdk/crypto';
+import { Entity, PrimaryColumn } from 'typeorm';
+
+@Entity('stake_key')
+export class StakeKeyEntity {
```
iccicci, line 5: I would use `BlockCascadeRelationOptions` for all the entries; this will reduce the code to handle rollback to just a single `DELETE` statement.

packages/projection-typeorm/src/entity/BlockData.entity.ts
```diff
@@ -0,0 +1,17 @@
...
+@Entity('block_data')
+export class BlockDataEntity {
+  @PrimaryGeneratedColumn()
+  id?: number;
+
+  @OneToOne(() => BlockEntity, BlockCascadeRelationOptions)
+  @JoinColumn()
+  block?: BlockEntity;
+
+  @Column({ transformer: [serializableObj, json, stringBytea], type: 'bytea' })
```
iccicci, line 15: The advantages of a byte array Vs a string type are:

1. the ability to write non printable characters
2. avoiding some escape/un-escape of single quote character

the main disadvantage is: while developing/debugging, querying a byte array from DB results in an opaque string.

Using a byte array field to store a JSON encoded string has two great disadvantages:

1. it uses a lot of space
2. it produces an opaque string

If we want to settle on an opaque string (less handy while developing/debugging) I think it would be better to optimize the data space: CBOR serialization produces a binary tiny string which perfectly fit into a bytes array field.

On the other hand, we can renounce to heavy space optimization to gain the benefit of a human readable string. In this case a string field rather than a bytes array string is required.

iccicci (commented):
I have to admit my low knowledge of `rx-js` makes my a bit hard to read this PR.

Some inline _typedoc_ documentation could be helpful (regardless from my ignorance).

I appreciate so much the refactoring of docker related stuff in `util-dev`! 👍

packages/projection-typeorm/src/entity/Block.entity.ts
```diff
@@ -0,0 +1,25 @@
+import { Cardano } from '@cardano-sdk/core';
+import { Column, Entity, Index, PrimaryColumn, RelationOptions } from 'typeorm';
+
+@Entity({ name: 'block' })
+export class BlockEntity {
```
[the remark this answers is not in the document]
mkazlauskas: This model is the bare minimum that the projector needs to function, we can add `epoch` later if we need it, or we can add it to other entities in a de-normalized fashion

packages/projection-typeorm/src/entity/StakeKey.entity.ts
mkazlauskas: That doesn't work well for all use cases. For example, this table represents "an active set of stake keys", valid for latest 'block'. We're not projecting the certificates directly, so there is no association of specific stake key with a specific block. Sure we could change the model to fit `BlockCascadeRelationOptions`, but the "state at latest block" type of projections are really nice when it's possible to do.

packages/projection-typeorm/src/entity/BlockData.entity.ts
mkazlauskas: Hm, this is not supposed to be queryable and needs to be optimized for disk space.

> it uses a lot of space

Are you sure? I chose `bytea` because it should take less space :sweat_smile: Need to do some additional research...

packages/projection-typeorm/src/entity/StakeKey.entity.ts
iccicci: Initially I didn't got the meaning of the table. Thank you for the explanation.

packages/projection-typeorm/src/entity/BlockData.entity.ts
iccicci: I'm not so prepared on the headers of the two field types, but I thing it's negligible if compared to the size of a block.
I would focus on the content of the field: if you think we will no need to query for its content while developing debugging I would say this is enough to opt for a CBOR serialization.

mirceahasegan (changes requested):
[removed comment posted from vscode and added it in the context of the code]

packages/projection/src/sinks/inMemory/InMemoryStabilityWindowBuffer.ts
```diff
@@ -1,39 +1,45 @@
...
+        tap(({ eventType, block, genesisParameters: { securityParameter } }) => {
+          if (eventType === ChainSyncEventType.RollForward) {
+            // clear blocks that are past stability window
+            while (this.#blocks.length > securityParameter) this.#blocks.shift();
+            // add current block to cache and return the event unchanged
+            this.#blocks.push(block);
+            this.tip$.next(block);
+            this.#setTail(this.#blocks[0]);
+          } else if (eventType === ChainSyncEventType.RollBackward) {
+            for (let i = this.#blocks.length - 1; i >= 0; i--) {
```
mirceahasegan, line 25: I will start with the assumption that a rollback always happens on the tip$. If that is correct, then iterating the blocks backwards, trying to find the rollback block does not make sense: It's either the tip$ or nothing.
On the other hand, if the rollback event could be on an older block, all blocks newer than it should be deleted.
The for loop looks for the rollback block and removes it, leaving newer blocks untouched.

mirceahasegan (changes requested):

packages/projection-typeorm/src/entity/Block.entity.ts
mkazlauskas: Note that it can sometimes make queries more complicated, if you already have block hash like [here](https://github.com/input-output-hk/cardano-js-sdk/pull/616/commits/1eee6b82e4fe3c5e3ff16ea545f2e41c2b7f4313#diff-fb271f632f8478f3ec6b07bbc5218c7bdbd118313fe542f37be2e91acff55208R183)

packages/projection-typeorm/src/entity/BlockData.entity.ts
mkazlauskas: Note it's not CBOR that we're saving, it's JSON. In the future it's likely we'd change it to CBOR.

We could create a SPIKE for this, we're free to change it until the service is deployed.

packages/projection/src/sinks/inMemory/InMemoryStabilityWindowBuffer.ts
mkazlauskas: In context of stability window buffer it will always pop the tip - a rollback can happen to an older block, but this gets called for each block.

So yeah, this implementation isn't great, but in-memory buffer is out of scope of this PR, this was just a refactor. We'll probably only be using this buffer in tests, but feel free to open a ticket to fix this.

packages/util-dev/src/patchObject.ts
```diff
@@ -0,0 +1,11 @@
+export type ObjectPatches<T> = {
```
mirceahasegan, line 1: This looks like the builtin typescript `Partial`. I don't think we need it

packages/projection/src/sinks/inMemory/InMemoryStabilityWindowBuffer.ts
mkazlauskas: 90595eac41c9eececcd58fe2be21f91b95bf8f80

packages/projection-typeorm/src/types.ts
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
mirceahasegan, line 1: there doesn't seem to be any `any` in the code 
```diff
- /* eslint-disable @typescript-eslint/no-explicit-any */

Comment (mkazlauskas) [reply packages/projection-typeorm/src/types.ts]: 96ec056b6ae41868b0729e4e1b319a45c4f9f028

File: packages/projection-typeorm/src/createSinks.ts [hunk 8]
```diff
@@ -0,0 +1,169 @@
...
+        await queryRunner.connect();
+        // buffer has to emit tip$ and tail$ for projectIntoSink
+        // to find intersection before emitting any events
+        await buffer.initialize(queryRunner);
+        return { queryRunner };
+      })()
+    ),
+    NEVER
+  ).pipe(
+    finalizeWithLatest(async (evt) => {
```
mirceahasegan, line 67: I don't think this will ever finalize due to `NEVER` being the last observable in the concat pipe 🤔

mkazlauskas: It does, when unsubscribed by the subscriber

packages/projection-typeorm/src/TypeormStabilityWindowBuffer.ts
```diff
@@ -0,0 +1,204 @@
...
+    } = block;
+    const repository = queryRunner.manager.getRepository(BlockDataEntity);
+    if (eventType === ChainSyncEventType.RollForward) {
+      this.#logger.debug('Add block data at height', blockNo);
+      const blockData = repository.create({
+        block: blockEntity,
+        data: block
+      });
+      await Promise.all([repository.insert(blockData), this.#deleteOldBlockData(evt)]);
+      transactionCommit$.subscribe(() => {
```
mirceahasegan, line 112: I think `transactionCommitted$` would be a better name, as it triggers every time the transaction is committed.
`transactionCommit$` sounds like a commit was triggered, and we should take action, not that it was done.
Just a suggestion in case you see it this way too.

packages/projection-typeorm/src/TypeormStabilityWindowBuffer.ts
```diff
@@ -0,0 +1,204 @@
...
+}
+
+export class TypeormStabilityWindowBuffer
+  implements StabilityWindowBuffer<Operators.WithNetworkInfo & WithTypeormContext>
+{
+  #tail: Cardano.Block | 'origin';
+  readonly #logger: Logger;
+  readonly #compactEvery: number;
+  readonly #tip$ = new ReplaySubject<Cardano.Block | 'origin'>(1);
+  readonly #tail$ = new ReplaySubject<Cardano.Block | 'origin'>(1);
```
mirceahasegan, line 44: This could be a `BehaviorSubject` and get rid of #tail property.
```diff
-  readonly #tail$ = new ReplaySubject<Cardano.Block | 'origin'>(1);
+  readonly #tail$ = new BehaviorSubject<Cardano.Block | 'origin'>('origin');
+  // use this.#tail$.value instead of this.#tail

Comment (mkazlauskas) [reply packages/projection-typeorm/src/TypeormStabilityWindowBuffer.ts:44]: Hm not really, BehaviorSubject requires initial value and we don't want to emit `origin`. Having it with `| null` could work but then we'd need to expose it as `pipe(filter(isNotNil))` which I'm not sure is any better than existing implementation

File: packages/projection-typeorm/src/TypeormStabilityWindowBuffer.ts (hunk 9 above)
Comment (rhyslbw) [reply packages/projection-typeorm/src/TypeormStabilityWindowBuffer.ts:112]: Agree with @mirceahasegan. `Committed` is less vague

File: packages/projection-typeorm/src/entity/BlockData.entity.ts [hunk 11]
```diff
@@ -0,0 +1,17 @@
+import { BlockCascadeRelationOptions, BlockEntity } from './Block.entity';
```
mirceahasegan, line 1: Why are you using two entities, `Block.entity & BlockData.entity` instead of a single one that has all the columns?

mkazlauskas: Great question! I was thinking about the buffer as a separate subsystem and haven't really considered a nullable `data` column on the `block` table. Let's consider it, here are my thoughts:

Pros:
- Would save 1 `INSERT` on RollForward, which is great as the projection would be a little bit faster.
- Faster query to select block data for `tip$`.

Cons:
- Slower query to select block data for `tail$`, as you would need a `WHERE` not null. This query is made on initialization and once every 100 (by default) RollForward events.
- If we wanted to support using another implementation of the buffer (not PostgreSQL), but keep the projection sinks, then it'd end up with a trash column that's always null. Not very likely to happen, but possible.

Any more Pros you have in mind @mirceahasegan ?

packages/projection-typeorm/src/TypeormStabilityWindowBuffer.ts
mkazlauskas: 2020e6ba9

packages/projection-typeorm/src/entity/BlockData.entity.ts
mirceahasegan: Nope. Thanks for the clarification.

mirceahasegan (approved):
Excelent implementation, @mkazlauskas  🚀

packages/projection-typeorm/src/TypeormStabilityWindowBuffer.ts
```diff
@@ -0,0 +1,204 @@
...
+          data: true as any,
+          id: true
+        },
+        where: {
+          block: {
+            height: nextTailBlockHeight
+          }
+        }
+      }),
+      // Review: this is fragile, may be better to make 2 type-safe(r) queries instead:
```
rhyslbw, line 186: Let's stick with the most performant option. Either drop the comment or move it to a Jira ticket

packages/projection-typeorm/package.json
```diff
@@ -0,0 +1,81 @@
+{
+  "name": "@cardano-sdk/projection-typeorm",
+  "version": "0.1.0",
+  "description": "Project Chain Sync events into PostgreSQL via typeorm",
```
rhyslbw, line 4:
```suggestion
  "description": "Project Chain Sync events into PostgreSQL via TypeORM",
```

packages/projection-typeorm/src/util.ts
```diff
@@ -0,0 +1,41 @@
...
+  NoVersionOrUpdateDateColumnError,
+  PersistedEntityNotFoundError,
+  PessimisticLockTransactionRequiredError,
+  QueryRunnerAlreadyReleasedError,
+  QueryRunnerProviderAlreadyReleasedError,
+  TransactionAlreadyStartedError,
+  TransactionNotStartedError
+} from 'typeorm';
+
+// Might have to adjust this list - classes were picked based on names and their doc comments
```
rhyslbw, line 20: Anything need to be done here, or is it a clue for a future investigation :smile: ?

packages/projection-typeorm/test/sinks/types.ts
```diff
@@ -0,0 +1,5 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+// Consider moving this to projection package src types
```
rhyslbw, line 3: Seems reasonable to do this

rhyslbw (changes requested):

packages/projection-typeorm/src/util.ts
mkazlauskas: Just a clue. We'll need to extend this as we use new features (create new kind of db objects). Default typeorm naming convention isn't very readable, it uses hashes to name some things.

packages/projection-typeorm/src/TypeormStabilityWindowBuffer.ts
mkazlauskas: Dropped the comment e9ad0e196f5bc35ad6a85b22520f0b0a10e8cb44

packages/projection-typeorm/package.json
mkazlauskas: 1b766590cf7194edd7cde66b8752ee350a8d82c5

packages/projection-typeorm/test/sinks/types.ts
mkazlauskas: 760c8990fdbbd5564b9c360a3fd6caed17728a70 and 3e5853d654f3947a68a52f78d7981ecb94de814d

rhyslbw (approved):
Ready for squashing

packages/projection-typeorm/src/entity/BlockData.entity.ts
mkazlauskas: Created ADP-2818

rhyslbw (approved):
Awesome work @mkazlauskas :rocket:

mirceahasegan (approved):
🥇  Looks awesome

PR #617, opened 2023-02-24
Fix/hardware mint/burn mapping + bump Ledger deps

rhyslbw:
# Context
There's a logical error in the hardware wallet mappings, causing any transaction containing a mint operation to throw. There's also a need to bump the Ledger package dependencies to reduce risk of firmware incompatibility. This version is still not compatible with Ledger transactions of Plutus type, but we need to do a rework of the solution foundations first to gain better confidence in adding new features.

# Proposed Solution
- Hoist the logic into a separate function, add unit tests, and fix bug.
- Bump Ledger package dependencies

Trezor unit tests are failing, but it's happening in master too. Not in scope to resolve.
image


mkazlauskas (dismissed):
Looks good! :rocket:

packages/key-management/src/util/mapHardwareSigningData.ts
```diff
@@ -87,6 +87,12 @@ const matchGroupedAddress = (knownAddresses: GroupedAddress[], outputAddress: Bu
   return knownAddresses.find(({ address }) => address === outputAddressBech32);
 };
 
+export const mintValueAsString = (scope: ManagedFreeableScope, amount: CML.Int): string => {
```
mirceahasegan, line 90: I think we could use the `to_str()` directly here. [The docs](https://github.com/dcSpark/cardano-multiplatform-lib/blob/685354495562ff879101fd023eb89555db9bbb1e/rust/pkg/cardano_multiplatform_lib.js.flow#L3401) suggest it will add the `-` in anyway.
```diff
-  const positiveAmount = scope.manage(amount?.as_positive())?.to_str();
-  const negativeAmount = scope.manage(amount?.as_negative())?.to_str();
-  return positiveAmount ? positiveAmount : `-${negativeAmount}`;
+  return amount?.to_str();

File: packages/key-management/src/util/mapHardwareSigningData.ts [hunk 2]
```diff
@@ -87,6 +87,12 @@ const matchGroupedAddress = (knownAddresses: GroupedAddress[], outputAddress: Bu
...
+  const positiveAmount = scope.manage(amount?.as_positive())?.to_str();
+  const negativeAmount = scope.manage(amount?.as_negative())?.to_str();
+  return positiveAmount ? positiveAmount : `-${negativeAmount}`;
```
mirceahasegan, line 93: In the case of `undefined | null` amount, the function will return `-undefined`

packages/key-management/src/util/mapHardwareSigningData.ts
```diff
@@ -395,15 +401,12 @@ const prepareTrezorMintBundle = (
         for (let k = 0; k < assetsKeys.len(); k++) {
           const assetName = scope.manage(assetsKeys.get(k));
           const amount = scope.manage(assets.get(assetName));
-          const positiveAmount = scope.manage(amount?.as_positive())?.to_str();
-          const negativeAmount = scope.manage(amount?.as_negative())?.to_str();
-          if (!amount || !positiveAmount || !negativeAmount) {
-            throw new HwMappingError('Missing token amount.');
+          if (amount) {
+            tokens.push({
```
mirceahasegan, line 405: I'm not sure if it's correct, but the logic has changed a bit.
Previously, in case of `undefined` amount it would throw.
Now, in case of `undefined` amount it will just not add the token.

packages/key-management/test/util/mapHardwareSigningData.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    it('returns "1" when passed a CML.Int representing 1', () => {
+      expect(mintValueAsString(scope, scope.manage(CML.Int.from_str('1')))).toBe('1');
+    });
+    it('returns "-1" when passed a CML.Int representing -1', () => {
+      expect(mintValueAsString(scope, scope.manage(CML.Int.from_str('-1')))).toBe('-1');
+    });
+    it('returns "0" when passed a CML.Int representing 0', () => {
+      expect(mintValueAsString(scope, scope.manage(CML.Int.from_str('0')))).toBe('0');
+    });
+  });
```
mirceahasegan, line 24: I think a test capturing [the limits cases](https://github.com/dcSpark/cardano-multiplatform-lib/blob/685354495562ff879101fd023eb89555db9bbb1e/rust/pkg/cardano_multiplatform_lib.js.flow#L3369) would be helpful.
Although we would just be testing the CML implementation, so maybe we shouldn't add it 😄

packages/key-management/src/util/mapHardwareSigningData.ts
mkazlauskas: Nice catch, although minting an `undefined` or `null` amount is not a thing, so shouldn't really be a problem

mirceahasegan: True. In this case, we shouldn't even take it into account:
```diff
- amount?.as_positive()
+ amount.as_positive()

File: packages/key-management/src/util/mapHardwareSigningData.ts (hunk 1 above)
Comment (rhyslbw) [reply packages/key-management/src/util/mapHardwareSigningData.ts:90]: Great catch. Will update

File: packages/key-management/src/util/mapHardwareSigningData.ts (hunk 3 above)
Comment (rhyslbw) [reply packages/key-management/src/util/mapHardwareSigningData.ts:399]: No longer exists in 67e3f4aa61259851b735f3652c1fd95e64000848

File: packages/key-management/src/util/mapHardwareSigningData.ts (hunk 1 above)
Comment (rhyslbw) [reply packages/key-management/src/util/mapHardwareSigningData.ts]: 67e3f4aa61259851b735f3652c1fd95e64000848

File: packages/key-management/src/util/mapHardwareSigningData.ts (hunk 2 above)
Comment (rhyslbw) [reply packages/key-management/src/util/mapHardwareSigningData.ts]: > True. In this case, we shouldn't even take it into account:

Yes, the optional chaining operator is wrong. Gone in 67e3f4aa61259851b735f3652c1fd95e64000848

File: packages/key-management/test/util/mapHardwareSigningData.test.ts (hunk 4 above)
Comment (rhyslbw) [reply packages/key-management/test/util/mapHardwareSigningData.test.ts]: Removed in https://github.com/input-output-hk/cardano-js-sdk/commit/67e3f4aa61259851b735f3652c1fd95e64000848

File: packages/key-management/src/util/mapHardwareSigningData.ts (hunk 3 above)
Comment (mirceahasegan) [reply packages/key-management/src/util/mapHardwareSigningData.ts:399]: The original code threw a `HwMappingError('Missing token amount.')` in case of falsy `amount`.
The current code doesn't throw anymore. It simply skips adding it to the token map.
If that's the desired flow, I'm ok with it, and we can mark this resolved 🤝

Comment (rhyslbw) [reply packages/key-management/src/util/mapHardwareSigningData.ts:397]: That logic was only there due to the design of the CML needing multiple steps to get the value, so this is more of a type issue. I've improved this in c2cec975920ae58f76deb6ee5c084722bd54a2d6

File: packages/key-management/src/util/mapHardwareSigningData.ts [hunk 5]
```diff
@@ -466,19 +464,17 @@ const prepareLedgerOutputs = (outputs: CML.TransactionOutputs, knownAddresses: G
...
-            for (let k = 0; k < assetsKeys.len(); k++) {
-              const assetName = scope.manage(assetsKeys.get(k));
-              const amount = scope.manage(assets.get(assetName));
-              if (assetName && amount) {
-                tokens.push({
-                  amount: amount.to_str(),
-                  assetNameHex: Buffer.from(assetName.name()).toString('hex')
-                });
-              }
+          const assetsKeys = scope.manage(assets.keys());
```
mirceahasegan, line 469: The code is nearly identical to prepareTrezorOutputs, but I think that should be a bigger task of refactoring this code to be more DRY

mirceahasegan (dismissed):
Great work, @rhyslbw 🚀

packages/key-management/src/util/mapHardwareSigningData.ts
rhyslbw: Agree. It's arguable to even be committing this change now, given there's other reasons why this transaction will still fail, but it will at least remove a confusing error, so that's the value the partial fix is adding.

mkazlauskas (approved):

PR #618, opened 2023-02-27
Feat/adp 2474 parsing and extracting information from addresses

AngelCastilloB:
# Context

We want to implement a simple set of typescript classes that encodes all the rules of [CIP-19](https://cips.cardano.org/cips/cip19/), both Base58 and Bech32 encoding algorithms are required.

# Proposed Solution

Implement CIP-19 natively in typescript.

- Added new address types to parse and create Shelly and Byron addresses.
- Reeplaced the CML functions related to CIP-19 with our new custom classes.


packages/core/src/Cardano/types/Address.ts
```diff
@@ -1,23 +1,1259 @@
...
+   * @throws if is a Shelley address. In principle, it is possible for Byron address to be encoded in Bech32. However,
+   * implementations are discouraged to encode addresses against the convention, as this helps with the goal that lay
+   * users only encounter a single, canonical version of every address. (See CIP-19.)
+   */
+  toBech32(): AddressString {
+    const words = BaseEncoding.bech32.toWords(Buffer.from(this.toBytes(), 'hex'));
+
+    if (this.#props.type === AddressType.Byron) throw new Error('Only Shelley addresses will be encoded in bech32');
+    const prefix = Address.getBech32Prefix(this.#props.type, this.#props.networkId!);
+    return BaseEncoding.bech32.encode(prefix, words, MAX_BECH32_LENGTH_LIMIT) as AddressString;
```
mkazlauskas, line 370: We should probably remove the existing `bech32` package dependency since we have this now

packages/core/src/Cardano/types/Address.ts
```diff
@@ -1,23 +1,1259 @@
...
+   * Byron addresses content (kept for backward compatibility with Byron addresses).
+   */
+  byronAddressContent?: ByronAddressContent;
+};
 
 /**
  * mainnet or testnet address (Shelley as bech32 string, Byron as base58-encoded string)
  */
-export type Address = OpaqueString<'Address'>;
+export type AddressString = OpaqueString<'AddressString'>;
```
mkazlauskas, line 235: How about thiis naming? To most people an address can't be anything other but a string :grin: 
```suggestion
export type PaymentAddress = OpaqueString<'PaymentAddress'>;
```

packages/core/src/Cardano/ChainId.ts
```diff
@@ -1,11 +1,21 @@
...
- * Network type (mainnet or **some** testnet)
+ * The "network ID" identifies a network in the Cardano blockchain, it is included in every address and also
+ * optionally present in the transaction body.
+ *
+ * This can only store 16 possibilities (4 bits).
  */
 export enum NetworkId {
-  Mainnet = 1,
-  Testnet = 0
+  Mainnet = 0b0001,
```
mkazlauskas, line 8: Isn't this the same as it was? :thinking:

packages/core/test/Cardano/util/address.test.ts
```diff
@@ -1,11 +1,6 @@
-import * as parseCmlAddress from '../../../src/CML/parseCmlAddress';
```
mkazlauskas, line 1: This test suite seems to be testing `Address.ts`, so would be best to consolidate the tests in 1 file

mkazlauskas (changes requested):

packages/core/src/Cardano/types/Address.ts
rhyslbw: Do we even need this as a separate type now? Maybe performance related? With multiple accepted encodings it seems to me that we're better off moving to just passing a class instance around, using `fromBech32` or `fromHex` static methods, and can then `toBech32` and `toHex` methods to convert the other way.

mkazlauskas: Yes, we still need a separate opaque string for addresses. Using class objects everywhere would complicate serialization

packages/core/src/Cardano/types/Address.ts
```diff
@@ -1,23 +1,1259 @@
...
+  }
+
+  /**
+   * Encodes this address to bech32.
+   *
+   * @throws if is a Shelley address. In principle, it is possible for Byron address to be encoded in Bech32. However,
+   * implementations are discouraged to encode addresses against the convention, as this helps with the goal that lay
+   * users only encounter a single, canonical version of every address. (See CIP-19.)
+   */
+  toBech32(): AddressString {
```
mkazlauskas, line 365: The original `Address` type was used only for payment addresses. Stake addresses are called 'RewardAccount'
```suggestion
  toBech32(): AddressString | RewardAccount {
```

mkazlauskas (changes requested):

packages/core/src/Cardano/types/Address.ts
rhyslbw: I'm not convinced without seeing how complicated it is, but if we do need a separate type, it should be specific: `Bech32PaymentAddress`

mkazlauskas: This can also be Byron address. We use this type for external inputs as well, e.g. `initializeTx`

mkazlauskas: > but to me it feels a bit redundant (and cost inefficient) to be converting from string/to string everything we interact with these types
We don't need to convert every time, once you create `Cardano.Address` you can pass it around without re-validating, so I think the boundary here is correct - it's as soon as it gets into the system, after that we trust that it has the expected format.

mkazlauskas: If you know that it's correct type then there's no need to re-validate. `str as KeyType` should do the job?

mkazlauskas: I think we're mixing up 2 topics here:
1. OpaqueString types - I think those are very useful and we should definitely keep them.
2. Constructors of OpaqueString types that do validation - this is debatable. We could make all validations explicit, e.g.
```ts
   type SomeKey = OpaqueString<'SomeKey'>;
   const SomeKey = (str: string): SomeKey => str as SomeKey // this doesn't necessarily have to do validation, we could do
   // 1. const SomeKey = (str: string, validate = false) => ...
   // 2. const SomeKey.validate = (str): boolean =>
   // 3. const SomeKey.parse = (str): SomeKey => // same as original SomeKey()
```

rhyslbw: > internal implementation could benefit from passing around objects though.

OK, I think we're in agreement here. Having the `Address` would be more convenient when doing internal transformations, which is something I'm working on now, so I can present a proposal to push the opaque string to the outer edges.

packages/core/src/Cardano/ChainId.ts
rhyslbw: The change conflicts with the domain concept, which represents these as integers, so reverting is preferred.

mkazlauskas (dismissed):
Great work! :rocket:

packages/core/src/Cardano/types/Address.ts
```diff
@@ -1,23 +1,1264 @@
...
+  }
+
+  /**
+   * Gets the address network id.
+   */
+  getNetworkId(): NetworkId {
+    if (this.#props.type === AddressType.Byron) {
+      // Byron addresses do not encode the network identifier, however, they include an optional attribute
+      // 'network magic', the network magic is only set to specify which testnet this address is for (if any). We can
+      // then assume that if the network magic is undefined the network id is mainnet; otehrwise; testnet.
```
rhyslbw, line 501:
```suggestion
      // then assume that if the network magic is undefined the network id is mainnet; otherwise; testnet.
```

packages/core/src/Cardano/types/Address.ts
```diff
@@ -1,23 +1,1264 @@
...
+}
+
+/**
+ * Enterprise addresses carry no stake rights, so using these addresses means that you are opting out of participation
+ * in the proof-of-stake protocol.
+ *
+ * Note that using addresses with no stake rights effectively decreases the total amount of stake, which plays
+ * into the hands of a potential adversary.
+ */
+export class EnterpriseAddress {
```
rhyslbw, line 884: This file is pretty large. I would move this out to it's own file

rhyslbw (changes requested):
Looks great @AngelCastilloB! The docs are particularly nice, top job!

Let's just make sure the position in the codebase is correct. Currently the address classes are defined in `Cardano/types`, so I'd suggest we move them out to `Address/Address.ts` + `Address/RewardAddress.ts`, `Address/util/` etc, and make sure their tests also reflect this new structure.

packages/core/src/Cardano/Address/ByronAddress.ts
```diff
@@ -0,0 +1,81 @@
+/* eslint-disable no-bitwise */
+import { Address, AddressProps, AddressType, ByronAddressContent, ByronAddressType, ByronAttributes } from './Address';
```
rhyslbw, line 2: Any associated types, such as `ByronAddressContent` or `ByronAddressType` should be in the relevant file. So those ones can be here, and imported elsewhere if needed.

rhyslbw (changes requested):

packages/core/src/Cardano/Address/Address.ts
```diff
@@ -0,0 +1,739 @@
...
+   *
+   * - 0000 PaymentKeyHash StakeKeyHash
+   * - 0001 ScriptHash     StakeKeyHash
+   * - 0010 PaymentKeyHash ScriptHash
+   * - 0011 ScriptHash     ScriptHash
+   *
+   * @param type The address type.
+   * @param data The serialized address data.
+   */
+  private static unpackBaseAddressParts(type: number, data: Uint8Array): Address {
```
rhyslbw, line 470: Better to move these to their respective classes, then you can generalise to `packParts` and `unpackParts`. It would reduce the complexity of this file too (still over 700 lines)

rhyslbw (changes requested):

rhyslbw (dismissed):

rhyslbw (changes requested):
https://github.com/input-output-hk/cardano-js-sdk/pull/618/commits/0cf75ca58b7e87234a585682a2c22b49f2344e2a is also best to mark as breaking change.

mkazlauskas (approved):

rhyslbw (approved):
:tada:

PR #619, opened 2023-03-01
Hybrid blockfrost approach

iccicci:
# Context

Stake pool queries against _mainnet_ turned out to be too much heavy. We need a dedicated schema to do more preforming queries.

# Proposed Solution

The solution was actually proposed by @rhyslbw .

The new schema and populate it depends on the projector service. As an intermediate solution this change fetches stake pools metrics from Blockfrost and saves the result in the DB in order to change the queries in new ones more performing.


packages/cardano-services/src/Blockfrost/BlockfrostCache.ts
```diff
@@ -0,0 +1,86 @@
...
+    this.#cacheTtl = cacheTtl;
+    this.#db = db;
+  }
+
+  protected initializeImpl() {
+    return Promise.resolve();
+  }
+
+  protected startImpl() {
+    return Promise.resolve();
```
rhyslbw, line 45: Seems this doesn't need to be a `RunnableModule`?

packages/cardano-services/src/Program/programs/blockfrostWorker.ts
```diff
@@ -0,0 +1,71 @@
...
+export type BlockfrostWorkerArgs = CommonProgramOptions & PosgresProgramOptions & BlockfrostWorkerConfig;
+
+export interface LoadBlockfrostWorkerDependencies {
+  dnsResolver?: (serviceName: string) => Promise<SrvRecord>;
+  logger?: Logger;
+}
+
+const blockfrostWorker = 'Blockfrost worker';
+
+export const loadAndStartBlockfrostWorker = async (
```
rhyslbw, line 38:
```suggestion
export const loadBlockfrostWorker = async (
```
Leave the lifecycle management up to the HTTP server

packages/cardano-services/src/Program/programs/blockfrostWorker.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  if (!db)
+    throw new MissingProgramOption(blockfrostWorker, [
+      PostgresOptionDescriptions.ConnectionString,
+      PostgresOptionDescriptions.ServiceDiscoveryArgs
+    ]);
+
+  const blockfrost = new BlockfrostWorker(args, { db, logger });
+
+  await blockfrost.initialize();
```
rhyslbw, line 67:
```suggestion
```

packages/cardano-services/src/Program/programs/blockfrostWorker.ts
```diff
@@ -0,0 +1,71 @@
...
+  await blockfrost.start();
```
rhyslbw, line 68:
```suggestion
```

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
```diff
@@ -26,10 +26,20 @@ import { toStakePoolResults } from './mappers';
...
+   */
+  enableBlockfrost?: boolean;
+
   /**
    * Pagination page size limit used for provider methods constraint.
    */
   paginationPageSizeLimit: number;
+
+  /**
+   * Enables Blockfrost hibryd cache.
```
rhyslbw, line 40:
```suggestion
   * Enables Blockfrost hybrid cache.
```

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
```diff
@@ -26,10 +26,20 @@ import { toStakePoolResults } from './mappers';
  * Properties that are need to create DbSyncStakePoolProvider
  */
 export interface StakePoolProviderProps {
+  /**
+   * Enables useBlockfrost search parameters.
```
rhyslbw, line 30: Can you elaborate on this? It's still not clear why we need two params

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
```diff
@@ -63,11 +73,15 @@ export class DbSyncStakePoolProvider extends DbSyncProvider(RunnableModule) impl
   #epochLength: number;
   #epochMonitor: EpochMonitor;
   #epochRolloverDisposer: Disposer;
-  #paginationPageSizeLimit: number;
+  #maxLovelaceSupply: bigint;
```
rhyslbw, line 76:
```suggestion
  #maxLovelaceSupply: Lovelace;
```

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
```diff
@@ -174,6 +192,21 @@ export class DbSyncStakePoolProvider extends DbSyncProvider(RunnableModule) impl
...
+        undefined,
+        'The parameter "useBlockfrost" is not allowed in this mode'
+      );
+    }
+
+    this.logger.error(useBlockfrost);
+    this.logger.error(useBlockfrost);
+    this.logger.error(useBlockfrost);
+    this.logger.error(useBlockfrost);
+    this.logger.error(useBlockfrost);
```
rhyslbw, line 209: :smile: ?

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/types.ts
```diff
@@ -124,8 +124,8 @@ export interface SubQuery {
   query: string;
 }
 
-export interface TotalAdaModel {
-  total_ada: string;
+export interface ReservesModel {
```
rhyslbw, line 127: _Reserves_ is a specific pot of funds in the network. This is the total UTxOs value right?

packages/cardano-services/src/cli.ts
```diff
@@ -181,6 +214,71 @@ withCommonOptions(
...
+  )
+  .addOption(
+    new Option('--scan-interval <scanInterval>', BlockfrostWorkerOptionDescriptions.ScanInterval)
+      .env('SCAN_INTERVAL')
+      .default(SCAN_INTERVAL_DEFAULT)
+      .argParser((interval) => Number.parseInt(interval, 10))
+  )
+  .action(async (args: { apiUrl: URL } & BlockfrostWorkerArgs) => {
+    process.stdout.write('Blockfrost worker\n');
+    const blockfrostWorker = await loadAndStartBlockfrostWorker({
```
rhyslbw, line 272: Here you can start the HTTP server, just like `providerServer`

packages/cardano-services/src/Program/programs/blockfrostWorker.ts
```diff
@@ -0,0 +1,71 @@
...
+  dnsResolver?: (serviceName: string) => Promise<SrvRecord>;
+  logger?: Logger;
+}
+
+const blockfrostWorker = 'Blockfrost worker';
+
+export const loadAndStartBlockfrostWorker = async (
+  args: BlockfrostWorkerArgs,
+  deps: LoadBlockfrostWorkerDependencies = {}
+) => {
```
rhyslbw, line 41:
```suggestion
): Promise<HttpServer> => {
```

rhyslbw (changes requested):
Looking good so far @iccicci!
A few things to do for compatibility with the refactor, in addition to some inline feedback:
- Add a `healthCheck` method To the `BlockfrostWorker`
- Add `BlockfrostWorkerHttpService`, supplying it with a `BlockfrostWorker` instance as the `provider` (cast it `as Provider`)

packages/cardano-services/docker-compose.yml
```diff
@@ -153,6 +155,34 @@ services:
...
+  blockfrost-worker:
+    build:
+      context: ../../
+      target: blockfrost-worker
+    depends_on:
+      postgres:
+        condition: service_healthy
+    environment:
+      - LOGGER_MIN_SEVERITY=${LOGGER_MIN_SEVERITY:-info}
+      - BLOCKFROST_API_KEY=${BLOCKFROST_API_KEY}
```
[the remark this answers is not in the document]
iccicci: Right! I mainly thought to let it optional, I worked in following direction:
by default the docker compose should start normally; if enabled through environment, it can start with Blockfrost enabled

packages/cardano-services/src/Blockfrost/BlockfrostCache.ts
iccicci: Not yet, the intention is to check at the beginning of iteration in the main loop if a shutdown was requested to immediately exit rather than completing to refresh all the pools

packages/cardano-services/src/Program/programs/blockfrostWorker.ts
iccicci: As far as I remember calling `start` on a `RunnableModule` without calling `initialize` throws.

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
iccicci: You discovered one of my secret development techniques! 🤣

packages/cardano-services/src/cli.ts
iccicci: Wow!

packages/cardano-services/docker-compose.yml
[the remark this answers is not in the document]
rhyslbw: By "secret" I mean, a [Docker Secret](https://docs.docker.com/engine/reference/commandline/secret/): here: https://github.com/input-output-hk/cardano-js-sdk/pull/619/files/131595f8409b0473bab790932475b3eaaa704b1e#diff-a5f22c45d19dce606e9bae0a8cd3a11ce586442845b16c7e1a056ac7e028ea79R186docker

packages/cardano-services/src/Program/programs/blockfrostWorker.ts
rhyslbw: Right, but the lifecycle will be handled by the `HttpService`

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/types.ts
rhyslbw: Ah, the diff is confusing me :-)

packages/core/src/Provider/StakePoolProvider/types/StakePoolProvider.ts
```diff
@@ -50,6 +50,11 @@ export interface QueryStakePoolsArgs {
    * Will return all stake pools matching the query if not specified
    */
   pagination: PaginationArgs;
+
+  /**
+   * Blockfrost hibryd cache. When enabled, default: true
+   */
+  useBlockfrost?: boolean;
```
mkazlauskas, line 57: Can we keep this out of provider interfaces? Shouldn't be a client's concern where it comes from

mkazlauskas (changes requested):

packages/core/src/Provider/StakePoolProvider/types/StakePoolProvider.ts
rhyslbw: Good catch, I didn't notice this being on the provider interface. It's specific to the `DbSyncStakePoolProvider` implementation, so would be best constrained there.

packages/cardano-services/src/Program/services/blockfrost.ts
```diff
@@ -0,0 +1,130 @@
...
+  #timeOut?: NodeJS.Timeout;
+
+  constructor(cfg: BlockfrostWorkerConfig, deps: BlockfrostWorkerDependencies) {
+    const { apiUrl, createSchema, dropSchema, dryRun, scanInterval } = cfg;
+    const { db, logger } = deps;
+    const blockfrostCache = new BlockfrostCache(cfg, deps);
+
+    super(
+      { listen: getListen(apiUrl), name: 'blockfrost-worker' },
+      { logger, runnableDependencies: [blockfrostCache], services: [] }
```
rhyslbw, line 49:
```suggestion
      { logger, runnableDependencies: [blockfrostCache], services: [blockfrostCache] }
```

rhyslbw (changes requested):

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
rhyslbw: OK, I understand now. Having this exposed in the search params to facilitate testing is a code smell. Can we not instead just load two `ProviderServer`s, one with Blockfrost cache enabled, the other without, then compare?

packages/cardano-services/blockfrost-keys/README.md
```diff
@@ -0,0 +1,7 @@
+### Blockfrost API keys
+
+Store here the files containing the Blockfrost API key for each network.
```
mkazlauskas, line 3: This breaks the 'self-containment' of the application. Previously all files required to run it were under `dist/`. A cleaner approach would be using env variables, but if we don't want to do that then we should at least copy the keys to 'dist/cjs/' during `yarn build`

mkazlauskas (changes requested):

packages/cardano-services/blockfrost-keys/README.md
rhyslbw: The existing strategy for sensitive information is to offer both secrets support and ENV

packages/e2e/.env.example
```diff
@@ -40,3 +40,6 @@ TX_SUBMIT_HTTP_URL=http://localhost:3456/tx-submit
 TRANSACTIONS_NUMBER=10
 START_LOCAL_HTTP_SERVER=true
 WORKER_PARALLEL_TRANSACTION=3
+
+# Required by test:providers, test:blockfrost
+STAKE_POOL_PROVIDER_URL="http://localhost:4000/stake-pool"
```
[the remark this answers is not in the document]
mkazlauskas: Why can't it take the URL from the other variable though? Does the test need to use both variables at the same time?

packages/cardano-services/blockfrost-keys/README.md
mkazlauskas: Thanks for an explanation, no change required

mkazlauskas (approved):
Great work! :rocket:

rhyslbw:
Ready for squashing @iccicci

packages/e2e/.env.example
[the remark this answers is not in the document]
iccicci: With last refactoring this is no longer used by blockfrost hybrid solution, this is used only by `yarn test:provider`.

`yarn test:provider` can be performed against a docker compose instance; in these case all the default values makes the test to work.

`yarn test:provider` can be performed against a deployed instance, in that case docker compose instance doesn't need to be started and `STAKE_POOL_PROVIDER_URL` needs to be set to point to the deployed instance API URL.

In the end `STAKE_POOL_PROVIDER_PARAMS` is used by `yarn test:wallet` and `STAKE_POOL_PROVIDER_URL` is used by `yarn test:provides`; they have two distinct meaning, the reason why they have the same value is to simplify running the tests against the docker compose instances (_local-network_, _preprod_ ect...)

To deduplicate this value we could do something similar to:

```
COMMON_STAKE_POOL_URL="http://localhost:4000/stake-pool"
STAKE_POOL_PROVIDER_PARAMS='{"baseUrl":"'$COMMON_STAKE_POOL_URL'"}'
STAKE_POOL_PROVIDER_URL=$COMMON_STAKE_POOL_URL
```

but in my opinion the lost readability doesn't worth the duplication removal.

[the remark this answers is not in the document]
mkazlauskas: Sorry, but I still don't understand how they have a different meaning :sweat_smile: They both provide a stake pool provider URL :thinking: I understand that different test suites use different variables right now, but don't understand why can't the `yarn test:provider` just get `STAKE_POOL_PROVIDER_PARAMS` instead of `STAKE_POOL_PROVIDER_URL`

PR #620, opened 2023-03-01
Feat/adp 2627 cip30 sign tx should throw

AngelCastilloB:
# Context

cip30 signTx should throw (reject) when it doesn't find any keys to sign with.

if partialSign=true and returned signatures length is 0 then throw (not explicit in cip, so not 100% sure about this)

if partialSign=false and not ALL inputs of the tx can be signed by the wallet then throw

From the cip:

```
If partialSign is true, the wallet only tries to sign what it can.
If partialSign is false and the wallet could not sign the entire transaction,
```
DataSignErrorCode shall be returned with the ProofGeneration code.

# Proposed Solution

- add a new wallet util 'requiresForeignSignatures' to check if the wallet can account for all inputs/certificates in the transaciton.
- check in the signTx function if all inputs/certs are accounted for, and throw depending on the value of the partialSign param.

### DIFF
--- a/packages/key-management/src/util/ownSignatureKeyPaths.ts
+++ b/packages/key-management/src/util/ownSignatureKeyPaths.ts
@@ -32,7 +32,6 @@ const getStakingKeyPaths = (
 
     for (const certificate of txBody.certificates) {
       switch (certificate.__typename) {
-        case Cardano.CertificateType.StakeKeyRegistration:
         case Cardano.CertificateType.StakeKeyDeregistration:
         case Cardano.CertificateType.StakeDelegation:
           if (certificate.stakeKeyHash === stakeKeyHash) paths.add(account.stakeKeyDerivationPath);
@@ -49,6 +48,7 @@ const getStakingKeyPaths = (
         case Cardano.CertificateType.MIR:
           if (certificate.rewardAccount === account.rewardAccount) paths.add(account.stakeKeyDerivationPath);
           break;
+        case Cardano.CertificateType.StakeKeyRegistration:
         case Cardano.CertificateType.GenesisKeyDelegation:
         default:
         // Nothing to do.
--- a/packages/key-management/test/util/ownSignaturePaths.test.ts
+++ b/packages/key-management/test/util/ownSignaturePaths.test.ts
@@ -1,3 +1,4 @@
+/* eslint-disable sonarjs/no-duplicate-string */
 import { AccountKeyDerivationPath, AddressType, GroupedAddress, KeyRole, util } from '../../src';
 import { Cardano } from '@cardano-sdk/core';
 
@@ -62,8 +63,7 @@ describe('KeyManagement.util.ownSignaturePaths', () => {
   });
 
   it(
-    'doesnt returns stake key derivation path when a StakeKeyRegistration' +
-      // eslint-disable-next-line sonarjs/no-duplicate-string
+    'does not return stake key derivation path when a StakeKeyRegistration' +
       ' certificate with the wallet stake key hash is present',
     async () => {
       const txBody = {
--- a/packages/key-management/test/util/ownSignaturePaths.test.ts
+++ b/packages/key-management/test/util/ownSignaturePaths.test.ts
@@ -62,7 +62,7 @@ describe('KeyManagement.util.ownSignaturePaths', () => {
   });
 
   it(
-    'returns stake key derivation path when a StakeKeyRegistration' +
+    'doesnt returns stake key derivation path when a StakeKeyRegistration' +
       // eslint-disable-next-line sonarjs/no-duplicate-string
       ' certificate with the wallet stake key hash is present',
     async () => {
@@ -75,10 +75,6 @@ describe('KeyManagement.util.ownSignaturePaths', () => {
         {
           index: 0,
           role: KeyRole.External
-        },
-        {
-          index: 0,
-          role: KeyRole.Stake
         }
       ]);
     }
--- a/packages/wallet/src/cip30.ts
+++ b/packages/wallet/src/cip30.ts
@@ -18,6 +18,7 @@ import { HexBlob, ManagedFreeableScope, usingAutoFree } from '@cardano-sdk/util'
 import { Logger } from 'ts-log';
 import { Observable, firstValueFrom } from 'rxjs';
 import { ObservableWallet } from './types';
+import { requiresForeignSignatures } from './services';
 
 export type Cip30WalletDependencies = {
   logger: Logger;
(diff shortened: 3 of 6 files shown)


packages/key-management/src/util/ownSignatureKeyPaths.ts
```diff
@@ -59,6 +59,32 @@ const getStakingKeyPaths = (
...
+ * Predicate that evaluates whether the given input set contains inputs that do not belong to the wallet.
+ *
+ * @param inputs The set of inputs to inspect.
+ * @param knownAddresses The wallet known address.
+ * @param inputResolver the input resolver.
+ */
+export const hasForeignInputs = async (
+  inputs: Cardano.TxIn[],
+  knownAddresses: GroupedAddress[],
+  inputResolver: Cardano.util.InputResolver
```
mkazlauskas, line 72: Let's be explicit here (and in `ownSignatureKeyPaths` signature), because we'll be adding an input resolver that uses cardano-services provider. In the context of determining key paths, we're only interested in our own inputs, which can be resolved using data that the wallet already has (not making any extra requests).
```suggestion
  ownInputResolver: Cardano.util.InputResolver
```

packages/key-management/src/util/ownSignatureKeyPaths.ts
```diff
@@ -59,6 +59,32 @@ const getStakingKeyPaths = (
...
+) =>
+  inputs.some(async (input) => {
+    const ownAddress = await inputResolver.resolveInputAddress(input);
+    if (!ownAddress) return null;
```
mkazlauskas, line 76: The logic seems to be inverted here: `hasForeignInputs` should be `true` if we find some input that is **not own**
```suggestion
    if (!ownAddress) return true;
```

packages/key-management/src/util/ownSignatureKeyPaths.ts
```diff
@@ -68,18 +94,25 @@ export const ownSignatureKeyPaths = async (
...
+          if (foundAddress === undefined) partialSign = true;
+
+          return foundAddress;
         })
       )
     ).filter(isNotNil)
   ).map(({ type, index }) => ({ index, role: Number(type) }));
 
-  return [...paymentKeyPaths, ...getStakingKeyPaths(knownAddresses, txBody)];
+  return { keyPaths: [...paymentKeyPaths, ...getStakingKeyPaths(knownAddresses, txBody)], partialSign };
```
mkazlauskas, line 117: `partialSign` should also be `true` if there are certificates that require signature for a foreign stake key

packages/key-management/src/util/ownSignatureKeyPaths.ts
```diff
@@ -68,18 +94,25 @@ export const ownSignatureKeyPaths = async (
...
-): Promise<AccountKeyDerivationPath[]> => {
+): Promise<SignatureKeyPathsResult> => {
+  let partialSign = false;
+
   const paymentKeyPaths = uniq(
     (
       await Promise.all(
         txBody.inputs.map(async (input) => {
           const ownAddress = await inputResolver.resolveInputAddress(input);
           if (!ownAddress) return null;
```
mkazlauskas, line 105: See comment above about renaming argument to `ownInputResolver`. Let's add a doc comment about the input resolver argument to document it's expected behavior.
```suggestion
          if (!ownAddress) {
            partialSign = true;
            return null;
          }
```

packages/wallet/src/cip30.ts
```diff
@@ -290,21 +290,44 @@ export const createWalletApi = (
...
         const {
-          witness: { signatures }
+          tx: {
+            witness: { signatures }
+          },
+          partialSign
         } = await wallet.finalizeTx({ tx: { ...coreTx, hash } });
 
+        // If partialSign is false and the wallet could not sign the entire transaction
+        if (!_partialSign && partialSign)
```
mkazlauskas, line 305: Let's reserve the `partialSign` for the name of the argument (as it is in the cip), and rename our variable here.
```suggestion
        if (!partialSign && signedPartially)
```

mkazlauskas (changes requested):

packages/wallet/src/cip30.ts
```diff
@@ -293,18 +294,38 @@ export const createWalletApi = (
     if (shouldProceed) {
       const wallet = await firstValueFrom(wallet$);
       try {
+        const needsForeignSignature = await requiresForeignSignatures(coreTx, wallet);
+
+        // If partialSign is false and the wallet could not sign the entire transaction
+        if (!_partialSign && needsForeignSignature)
```
mkazlauskas, line 300: Now that the argument is used, let's rename it
```suggestion
        if (!partialSign && needsForeignSignature)
```

mkazlauskas (changes requested):
Feel free to squash, no need for fixup

mkazlauskas (approved):
Nice work! :rocket:

packages/wallet/src/services/WalletUtil.ts
```diff
@@ -157,3 +159,82 @@ export const createLazyWalletUtil = (): WalletUtil & { initialize: SetWalletUtil
     }
   };
 };
+
+/**
+ * Gets whether the given TX requires signatures that can not be provided by the given wallet.
+ *
+ * @param tx The transaction to inspect.
+ * @param wallet The wallet that will provide the signatures.
+ * @returns true if the wallet can not sing all inputs/certificates; otherwise; false.
```
rhyslbw, line 168:
```suggestion
 * @returns true if the wallet can not sign all inputs/certificates; otherwise; false.
```

packages/wallet/test/integration/cip30mapping.test.ts
```diff
@@ -380,5 +394,281 @@ describe('cip30', () => {
...
+          mockWallet.finalizeTx = () => Promise.resolve(tx as unknown as Cardano.Tx);
+
+          expect(tx.body.inputs.length).toBeGreaterThanOrEqual(1);
+          expect(tx.body.certificates!.length).toBe(2);
+          await expect(mockApi.signTx(cbor, true)).resolves.not.toThrow();
+        }
+      );
+
+      it(
+        'throw DataSignError with ProofGeneration as error code if the wallet cant sing a single input/cert and ' +
```
rhyslbw, line 593:
```suggestion
        'throw DataSignError with ProofGeneration as error code if the wallet cant sign a single input/cert and ' +
```

packages/key-management/test/util/ownSignaturePaths.test.ts
```diff
@@ -62,7 +62,7 @@ describe('KeyManagement.util.ownSignaturePaths', () => {
   });
 
   it(
-    'returns stake key derivation path when a StakeKeyRegistration' +
+    'doesnt returns stake key derivation path when a StakeKeyRegistration' +
```
rhyslbw, line 65:
```suggestion
    'does not return stake key derivation path when a StakeKeyRegistration is present' +
```

rhyslbw (changes requested):
Great work @AngelCastilloB :tada:

rhyslbw (approved):
:mechanical_arm:

PR #625, opened 2023-03-06
Fix stake pools queries fixes

iccicci:
# Context

Due to recent investigations we discovered we have two bugs in stake pools search related queries.

1. In order compute the _saturation_, the total staked amount is required: we are currently using the total UTXOs amount;
2. in case of phase 2 validation fail, **db-sync** stores the collateral inputs and outputs in `tx_in` and `tx_out` as if they were non collateral: we are excluding these UTXOs

# Proposed Solution

1. Replaced the query to get total UTXOs with the call to `OgmiosCardanoNode` to get the required value.
2. Removed the `tx.valid_contract = TRUE` condition from the `WHERE` clauses in stake pools search related queries.

### DIFF
--- a/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
+++ b/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
@@ -63,8 +63,8 @@ export class DbSyncStakePoolProvider extends DbSyncProvider(RunnableModule) impl
   #epochLength: number;
   #epochMonitor: EpochMonitor;
   #epochRolloverDisposer: Disposer;
-  #paginationPageSizeLimit: number;
   #metadataService: StakePoolExtMetadataService;
+  #paginationPageSizeLimit: number;
 
   constructor(
     { paginationPageSizeLimit }: StakePoolProviderProps,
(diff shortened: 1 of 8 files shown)


packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
```diff
@@ -199,10 +199,18 @@ export class DbSyncStakePoolProvider extends DbSyncProvider(RunnableModule) impl
...
     );
-    // Get total amount of ada cached
-    const totalAdaAmount = await this.#cache.get(queryCacheKey(StakePoolsSubQuery.TOTAL_ADA_AMOUNT), () =>
-      this.#builder.getTotalAmountOfAda()
-    );
+    // Get cached total staked amount
+    const totalStake = await this.#cache.get(queryCacheKey(StakePoolsSubQuery.TOTAL_STAKE), async () => {
+      const distribution = await this.cardanoNode.stakeDistribution();
+
+      for (const [_, value] of distribution) return value.stake.supply.toString();
```
rhyslbw, line 206:
```suggestion
      const first = [...distribution.values()][0];
      if (!first) throw ...
      return first.stake.supply.toString();
```
Slightly improved semantics?

iccicci: Yes, I focused on the optimization:

- mine version access the `Map` through its built in `Iterator`, only once
- proposed version requires the creation of an `Array` from a `Map`

rhyslbw (approved):
Great work @iccicci :muscle:

IvayloAndonov (dismissed):
Nice optimization!
Are there any minor discrepancies between the values (previous `totalAdaAmount` from db-sync vs. `totalStake` from the node)?

rhyslbw:
> Are there any minor discrepancies between the values (previous totalAdaAmount from db-sync vs. totalStake from the node)?

Yes, as expected, since the prior value is invalid to use in the saturation calculation. It was using total of the global UTxO set, which is not the total network stake.

rhyslbw (approved):

PR #627, opened 2023-03-07
feature: projector service

rhyslbw:
# Context

To run a projector we need a new service, with configuration for our current node and database infrastructure choices, Ogmios and PostgreSQL.
 
# Proposed Solution

Add a new cli program `start-projector` that has the same Ogmios and PostgreSQL configuration arguments/variables as the Provider Server. It accepts a set of  `--projection-names`, which are used to determine which entities, mappers and stores to use in the projection.

Added but not used: 
- `ObservableCardanoNode.healthCheck$`


packages/ogmios/src/CardanoNode/OgmiosObservableCardanoNode/OgmiosObservableCardanoNode.ts
```diff
@@ -52,6 +62,15 @@ export class OgmiosObservableCardanoNode implements ObservableCardanoNode {
       distinctUntilChanged(isEqual),
       shareReplay({ bufferSize: 1, refCount: true })
     );
+    this.healthCheck$ = this.#connectionConfig$.pipe(
+      switchMap((connectionConfig) =>
+        from(
+          getServerHealth({
+            connection: createConnectionObject(connectionConfig)
+          })
+        ).pipe(map(ogmiosServerHealthToHealthCheckResponse))
```
mkazlauskas, line 71:
```suggestion
        ).pipe(
          map(ogmiosServerHealthToHealthCheckResponse),
          timeoutWith(...), // we should probably not keep waiting forever for healthchecks
          catchError(error => {
            // I don't remember how it behaves when you subscribe to #interactionContext$, the errors might be swallowed - would need to test it
            if (isConnectionError(err)) {
              return { ok: false };
            }
            throw error;
          })
        )
```

packages/ogmios/src/CardanoNode/OgmiosObservableCardanoNode/OgmiosObservableCardanoNode.ts
```diff
@@ -52,6 +62,15 @@ export class OgmiosObservableCardanoNode implements ObservableCardanoNode {
       distinctUntilChanged(isEqual),
       shareReplay({ bufferSize: 1, refCount: true })
     );
+    this.healthCheck$ = this.#connectionConfig$.pipe(
```
mkazlauskas, line 65: Can't we use a potentially already existing connection?
```suggestion
    this.healthCheck$ = this.#interactionContext$.pipe(
```

mkazlauskas (changes requested):

packages/ogmios/src/CardanoNode/OgmiosObservableCardanoNode/OgmiosObservableCardanoNode.ts
rhyslbw: I found that it was conflicting in my test, since the WS connection would disconnect and then cause problems with the whole context, so thought it would be best to keep the two concerns separate. The health check just makes a HTTP GET request after all.

packages/cardano-services/src/Program/services/ogmios.ts
```diff
@@ -163,6 +169,25 @@ export const ogmiosCardanoNodeWithDiscovery = async (
...
+ * @returns ObservableOgmiosCardanoNode instance
+ */
+export const ogmiosObservableCardanoNodeWithDiscovery = async (
+  dnsResolver: DnsResolver,
+  logger: Logger,
+  serviceName: string
+): Promise<OgmiosObservableCardanoNode> =>
+  new OgmiosObservableCardanoNode(
+    {
+      connectionConfig$: from(dnsResolver(serviceName).then(({ name, port }) => ({ host: name, port })))
```
mkazlauskas, line 186: Each subscription must re-resolve
```suggestion
      connectionConfig$: defer(() => from(dnsResolver(serviceName).then(({ name, port }) => ({ host: name, port }))))
```

mkazlauskas (dismissed):

packages/cardano-services/test/util.ts
```diff
@@ -85,3 +86,30 @@ export const clearDbPools = async ({ main, healthCheck }: DbPools) => {
...
+export const projectorConnectionConfig = {
+  database: process.env.POSTGRES_DB_PROJECTION!,
+  host: process.env.POSTGRES_HOST!,
+  password: process.env.POSTGRES_PASSWORD!,
+  port: Number.parseInt(process.env.POSTGRES_PORT!),
+  username: process.env.POSTGRES_USER!
+};
+export const projectorConnectionConfig$ = of(projectorConnectionConfig);
+
+// export const initializeDataSource = async (props: Pick<CreateDataSourceProps, 'entities' | 'extensions'>) => {
```
mirceahasegan, line 99: Do we want to keep this commented code?

mkazlauskas: No it's trash, e70b23b8f34004aa600ecc9427c854bf8401c4a0

mirceahasegan (dismissed):
Great work! :rocket:

packages/cardano-services/README.md
```diff
@@ -129,6 +129,16 @@ RABBITMQ_SRV_SERVICE_NAME=some-domain-for-rabbitmq \
...
 ```
 
+**`start-projector` using CLI options with Ogmios and PostgreSQL running on localhost:**
+
+```bash
+./dist/cjs/cli.js \
+  start-projector \
+    --ogmios-url 'ws://localhost:1339' \
+    --postgres-connection-string 'postgresql://postgres:doNoUseThisSecret!@localhost/projection' \
+    stake-pool,stake-pool-metadatajob
```
rhyslbw, line 139:
```suggestion
    stake-pool,stake-pool-metadata-job
```

rhyslbw (commented):
Also need to add the Docker build stage and Docker Compose service

iccicci (approved):

mkazlauskas (approved):

mirceahasegan (approved):
Great work! :rocket:

PR #628, opened 2023-03-09
feat!: upgrade resolveInputAddress to resolveInput

rhyslbw:
# Context
Applications wishing to display Values relating to the resolved input need to independently fetch the entire transaction to resolve every input. This is sub-optimal in both needing to remotely fetch transactions already present in the wallet, and that most of the data is not needed, so the we’re placing significant unnecessary load on the provider backend. As a result, the UX is impacted as a result of network latency.

# Proposed Solution
Rename the function and return `{address, value}` instead of just the address.

### DIFF
--- a/packages/core/src/Cardano/Address/PaymentAddress.ts
+++ b/packages/core/src/Cardano/Address/PaymentAddress.ts
@@ -6,7 +6,7 @@ import {
   assertIsBech32WithPrefix,
   assertIsHexString
 } from '@cardano-sdk/util';
-import { HydratedTx, HydratedTxIn, TxIn } from '../types';
+import { HydratedTx, HydratedTxIn, TxIn, TxOut } from '../types';
 import { NetworkId } from '../ChainId';
 import { RewardAccount } from './RewardAccount';
 
@@ -72,13 +72,13 @@ export const inputsWithAddresses = (tx: HydratedTx, ownAddresses: PaymentAddress
   tx.body.inputs.filter(isAddressWithin(ownAddresses));
 
 /**
- * @param txIn transaction input to resolve address from
- * @returns input owner address
+ * @param txIn transaction input to resolve associated txOut from
+ * @returns txOut
  */
-export type ResolveInputAddress = (txIn: TxIn) => Promise<PaymentAddress | null>;
+export type ResolveInput = (txIn: TxIn) => Promise<TxOut | null>;
 
 export interface InputResolver {
-  resolveInputAddress: ResolveInputAddress;
+  resolveInput: ResolveInput;
 }
 
 /**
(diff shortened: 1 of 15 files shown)


iccicci (dismissed):
LGTM

IvayloAndonov (approved):
LGTM as well

packages/core/src/Cardano/Address/PaymentAddress.ts
```diff
@@ -75,10 +75,10 @@ export const inputsWithAddresses = (tx: HydratedTx, ownAddresses: PaymentAddress
  * @param txIn transaction input to resolve address from
  * @returns input owner address
  */
-export type ResolveInputAddress = (txIn: TxIn) => Promise<PaymentAddress | null>;
+export type ResolveInput = (txIn: TxIn) => Promise<{ address: PaymentAddress; value: Value } | null>;
```
mkazlauskas, line 78: Any reason not to resolve the entire output?
```suggestion
export type ResolveInput = (txIn: TxIn) => Promise<TxOut | null>;
```
Either way, would be great to export a type that this resolves.

mkazlauskas (changes requested):

packages/core/src/Cardano/Address/PaymentAddress.ts
rhyslbw: > Any reason not to resolve the entire output?

Good point. I recall having this discussion when implementing the feature originally, but we opted for just the address to meet the immediate demands. Now that we're resolving an object, it makes sense to return the whole `txOut`

rhyslbw: https://github.com/input-output-hk/cardano-js-sdk/pull/628/commits/1159f9fd69adafc1a45fa237af1b53e223aaac43

packages/key-management/package.json
```diff
@@ -47,6 +47,7 @@
     "prepack": "yarn build"
   },
   "devDependencies": {
+    "@cardano-sdk/tx-construction": "^0.2.1",
```
mkazlauskas, line 50: I think when you do a relative import then you need to update test project tsconfig
```
"include": [
    "../../../tx-construction/test/testData"
  ],
```
This dependency doesn't feel right, I'd remove it from package.json - can still use it with "includes"

mkazlauskas (changes requested):

packages/key-management/package.json
rhyslbw: Done and squashed fcfa035a3498f675945dafcc82b8f05c08318dd8

mkazlauskas (approved):
Nice work! :rocket:

iccicci (approved):

PR #629, opened 2023-03-09
refactor(cardano-services): tx.valid_contract filter is not needed to…

mirceahasegan:
… calculate withdrawals

Withdrawals are ignored for in phase 2 validation failed transactions. No need to join and filter withdrawal table by transaction valid_contract.

# Context

see commits


iccicci (approved):
that's all 🚀

rhyslbw (approved):
Nice work @mirceahasegan. Small change, but a good knowledge exercise.
:lotus_position:

PR #632, opened 2023-03-10
fix(wallet): map tx submission errors to TxSendError in cip30 submitTx

mkazlauskas:
# Context

cip30 submitTx can throw an error that is uncompliant with the spec


rhyslbw (approved):
Good job @mkazlauskas

mirceahasegan (approved):
🥇

PR #635, opened 2023-03-11
build: correct invalid production dependencies

rhyslbw:
# Context
A couple of packages have `@types/*` packages listed as `dependencies`, rather than `devDependencies`. This causes complexities when consuming the package in some environments.

# Proposed Solution
Move them to `devDependencies`


mirceahasegan (approved):
👍

PR #638, opened 2023-03-12
Test/prepare tests for mainnet

mirceahasegan:
# Context
Review and configure tests to be ran on Mainnet

# Proposed Solution
1. `yarn test:mainnet-real-ada` picks only the tests suitable for a network with real ada
2. Added a `KEY_MANAGEMENT_PARAMS_PEER` env to be used in tests with two wallets.
3. Added `MINT_POLICY_SIGNER_MNEMONICS1` and `...2` to be used for minting tests to avoid losing the keys previously generated randomly in tests.
4. Added burn tokens step to cleanup after minting tests.
 
### DIFF
--- a/packages/e2e/jest.config.js
+++ b/packages/e2e/jest.config.js
@@ -1,9 +1,13 @@
-const project = (displayName) => ({
-  displayName,
+const commonProjectProps = {
   preset: 'ts-jest',
   setupFiles: ['dotenv/config'],
-  testMatch: [`<rootDir>/test/${displayName}/**/*.test.ts`],
   transform: { '^.+\\.test.ts?$': 'ts-jest' }
+};
+
+const project = (displayName) => ({
+  displayName,
+  testMatch: [`<rootDir>/test/${displayName}/**/*.test.ts`],
+  ...commonProjectProps
 });
 
 // jest tests suitable to run in an environment with real ADA (not tADA)
@@ -25,17 +29,19 @@ module.exports = {
     project('load-testing'),
     project('local-network'),
     project('long-running'),
-    project('measurement-util'),
     project('ogmios'),
     project('projection'),
     project('providers'),
     project('wallet'),
     {
+      ...commonProjectProps,
       displayName: 'wallet-real-ada',
-      preset: 'ts-jest',
-      setupFiles: ['dotenv/config'],
-      testMatch: [`<rootDir>/test/wallet/SingleAddressWallet/(${realAdaTestFileNames.join('|')}).test.ts`],
-      transform: { '^.+\\.test.ts?$': 'ts-jest' }
+      testMatch: [`<rootDir>/test/wallet/SingleAddressWallet/(${realAdaTestFileNames.join('|')}).test.ts`]
+    },
+    {
+      ...commonProjectProps,
+      displayName: 'utils',
+      testMatch: ['<rootDir>/test/measurement-util/*.test.ts', '<rootDir>/test/util.test.ts']
     }
   ],
   testTimeout: 1000 * 60 * 25
--- a/packages/e2e/jest.config.js
+++ b/packages/e2e/jest.config.js
@@ -6,6 +6,19 @@ const project = (displayName) => ({
   transform: { '^.+\\.test.ts?$': 'ts-jest' }
 });
 
+// jest tests suitable to run in an environment with real ADA (not tADA)
+const realAdaTestFileNames = [
+  'delegation',
+  'metadata',
+  'mint',
+  'multisignature',
+  'nft',
+  'pouchDbWalletStores',
+  'txChainHistory',
+  'txChaining',
+  'unspendableUtxos'
+];
+
 module.exports = {
   projects: [
     { ...project('blockfrost'), globalSetup: './test/blockfrost/setup.ts' },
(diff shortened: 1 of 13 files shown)


iadmytro (approved):

packages/e2e/test/wallet/SingleAddressWallet/mint.test.ts
```diff
@@ -25,7 +25,7 @@ describe('SingleAddressWallet/mint', () => {
     const genesis = await firstValueFrom(wallet.genesisParameters$);
 
     const aliceKeyAgent = await createStandaloneKeyAgent(
-      util.generateMnemonicWords(),
+      env.MINT_POLICY_SIGNER_MNEMONICS1.split(' '),
```
rhyslbw, line 28: Is there a reason we can't use accounts from the credentials provided in `KEY_MANAGEMENT_PARAMS`, for both Alice and Bob?

rhyslbw (changes requested):
Well done categorising the tests here @mirceahasegan, this is a great start.

In general though we should aim to only require one set of credentials, to ensure we're not adding unnecessary complexity. All of the operations in our tests can be performed to the requirements with just a single key, including the minting operations, so here's some pointers based on the ask:

> Make sure that receiving wallet address is configurable and specified in .env so funds would not be sent to a random address

The important part here is "so funds would not be sent to a random address". The two options I see here are:

1. Send from the account 0 address to account 1 address, then after completion, return the maximum to the original address. This is maybe the nicest end result on the face of it, but it has a higher cost in both lovelaces and time to run the test.
2. Dynamically select the account with the highest balance to be the sender, so the roles alternate each test run. I prefer this option since it also places an emphasis on not getting too attached with a specific address in assertions, plus there's no cleanup to fail, and it's more sustainable. It's also addressing the other ask:

> ... it would be nice to spend as little amount of funds as possible...

So it would be best to revert the account index convention change in the tests and just ensure the funds remain in control of the single key we already have configured, and have the account 0 perform the minting and burning. If funds remain in control of the key being supplied, manually configuring a return address is a risk we should not expose administrators of the test to.

rhyslbw (approved):
Great work @mirceahasegan! Ready for squashing

rhyslbw (changes requested):
Lint failure

rhyslbw (approved):
Good catch on the extra cleanup logic :muscle:

rhyslbw (approved):

packages/e2e/jest.config.js
```diff
@@ -1,22 +1,48 @@
...
 });
 
+// jest tests suitable to run in an environment with real ADA (not tADA)
+const realAdaTestFileNames = [
+  'delegation',
+  'metadata',
+  'mint',
+  'multisignature',
+  'nft',
+  'puchDbWalletStores',
```
mkazlauskas, line 20:
```suggestion
  'pouchDbWalletStores',
```

packages/e2e/test/wallet/SingleAddressWallet/mint.test.ts
```diff
@@ -159,97 +105,7 @@ describe('SingleAddressWallet/mint', () => {
...
-      )
-    ]);
-
-    // Transaction should be part of the history
-    expect(txFoundInHistory).toBeTruthy();
-    // But it should also be failed since it failed the phase2 validation
-    expect(failedTx.id).toEqual(signedTx.id);
-    expect(failedTx.reason).toBe(TransactionFailure.Phase2Validation);
-    expect(txFoundInHistory.body.fee).toEqual(collateralCoinValue);
+    await burnTokens({ policySigners: [alicePolicySigner], scripts: [policyScript], wallet });
```
mkazlauskas, line 109: Is this really necessary? If we're always minting the same AssetId, then it doesn't really 'trash' the chain that much more. This will cost a fee that could be avoided.

mkazlauskas (changes requested):
Since there are no changes in `src/`, all commits should have `test` type.

Maybe we could use jest setup/teardown scripts to check the total balance of all used wallets/accounts and console.log how much it costs to run the tests. It would provide some confidence that we're not burning coins.

packages/e2e/jest.config.js
```diff
@@ -1,22 +1,48 @@
...
+
+const project = (displayName) => ({
+  displayName,
+  testMatch: [`<rootDir>/test/${displayName}/**/*.test.ts`],
+  ...commonProjectProps
 });
 
+// jest tests suitable to run in an environment with real ADA (not tADA)
+const realAdaTestFileNames = [
+  'delegation',
```
mkazlauskas, line 15: This would take at least 10 days on mainnet
```suggestion
```

mkazlauskas (changes requested):

packages/e2e/jest.config.js
mirceahasegan: Great catch 🕵️ !

mirceahasegan: If this is the only required change, I would like to force-push it to avoid the extra CI run and round of reviews needed when squashing.

rhyslbw: Yes, force-push is fine

rhyslbw (approved):

mkazlauskas (approved):
:rocket:

PR #639, opened 2023-03-12
perf: Optimize transactions by addresses SQL queries

iccicci:
# Context

The `/chain-history/txs/by-addresses` endpoint (called by wallet restoration procedure) may take a long time depending on the number of the transactions affecting the addresses.

# Proposed Solution

The `transactionsByAddresses` exposed function takes advantage from `transactionsByHashes` exposed function.

- Added a new `transactionsByIds` internal function.
- Changed the two previously said function to call the new one.

This let to change the main query performed by `transactionsByAddresses` to return the `tx.id` rather than `tx.hash` optimizing it.

Changed all the `ChainHistoryBuilder` methods called by `transactionsByHashes` to accept `tx.id` rather than `tx.hash`.

### DIFF
--- a/packages/cardano-services/src/Metadata/util.ts
+++ b/packages/cardano-services/src/Metadata/util.ts
@@ -1,0 +1,16 @@
+import { Cardano } from '@cardano-sdk/core';
+import { TxMetadataByHashes } from './DbSyncMetadataService';
+import { TxMetadataModel } from './types';
+import { mapTxMetadata } from './mappers';
+
+export const mapTxMetadataByHashes = (listOfMetadata: TxMetadataModel[]): TxMetadataByHashes => {
+  const metadataMap: Map<Cardano.TransactionId, TxMetadataModel[]> = new Map();
+
+  for (const metadata of listOfMetadata) {
+    const txId = metadata.tx_id.toString('hex') as Cardano.TransactionId;
+    const currentMetadata: TxMetadataModel[] = metadataMap.get(txId) ?? [];
+    metadataMap.set(txId, [...currentMetadata, metadata]);
+  }
+
+  return new Map([...metadataMap].map(([id, metadata]) => [id, mapTxMetadata(metadata)]));
+};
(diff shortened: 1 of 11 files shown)


packages/cardano-services/test/ChainHistory/DbSyncChainHistoryProvider/ChainHistoryBuilder.test.ts
```diff
@@ -47,61 +50,64 @@ describe('ChainHistoryBuilder', () => {
 
   beforeEach(() => logger.reset());
 
-  describe('queryTransactionInputsByHashes', () => {
+  const hashesToId = async (hashes: Cardano.TransactionId[]) => {
```
mkazlauskas, line 53: This can return IDs in different order. I suggest doing this and then if you need multiple you can do `Promise.all(arr.map(getTxId))`
```suggestion
  const getTxId = async (hash: Cardano.TransactionId) => {
```
Alternatively could return a Set or just add a docblock to make the behavior clear.

packages/cardano-services/src/Metadata/util.ts
```diff
@@ -0,0 +1,16 @@
+import { Cardano } from '@cardano-sdk/core';
+import { TxMetadataByHashes } from './DbSyncMetadataService';
+import { TxMetadataModel } from './types';
+import { mapTxMetadata } from './mappers';
+
+export const mapTxMetadataByHashes = (listOfMetadata: TxMetadataModel[]): TxMetadataByHashes => {
+  const metadataMap: Map<Cardano.TransactionId, TxMetadataModel[]> = new Map();
+
+  for (const metadata of listOfMetadata) {
+    const txId = metadata.tx_id.toString('hex') as unknown as Cardano.TransactionId;
```
mkazlauskas, line 10: This should be enough since 5ff77c8a8a4cce5c521c39afd9f23d546a5bfa1f
```suggestion
    const txId = metadata.tx_id.toString('hex') as Cardano.TransactionId;
```

mkazlauskas (changes requested):

packages/cardano-services/test/ChainHistory/DbSyncChainHistoryProvider/ChainHistoryBuilder.test.ts
iccicci: Nice catch!
4cb75f987 should solve this

packages/cardano-services/src/Metadata/util.ts
iccicci: 4cb75f987 should solve this

mkazlauskas (approved):
Great work! :rocket:

rhyslbw (approved):
Awesome work @iccicci and @IvayloAndonov :fire:

PR #640, opened 2023-03-13
Feat/stake pool metadata fetcher

AngelCastilloB:
# Context

We must be able to fetch the off-chain pool metadata, verify the base hash, and extended stake pool metadata signature using the extSigUrl (A URL with the extended metadata signature) and extVkey (the public key for verification) as specified in [CIP-6](https://cips.cardano.org/cips/cip6/).

We also want to add support for AdaPools extended metadata. (make use of extended property if presented within JSON response)

# Proposed Solution

Implement a new StakePoolMetadataService that can fetch the metadata and perform the proper validations

### DIFF
--- a/packages/cardano-services/src/StakePool/HttpStakePoolMetadata/HttpStakePoolMetadataService.ts
+++ b/packages/cardano-services/src/StakePool/HttpStakePoolMetadata/HttpStakePoolMetadataService.ts
@@ -1,6 +1,6 @@
 import { Cardano, ProviderError, ProviderFailure } from '@cardano-sdk/core';
 import { Logger } from 'ts-log';
-import { StakePoolExtMetadataResponse, StakePoolExtMetadataService } from '../types';
+import { StakePoolExtMetadataResponse, StakePoolMetadataService } from '../types';
 import { ValidationError, validate } from 'jsonschema';
 import { getExtMetadataUrl, getSchemaFormat, loadJsonSchema } from './util';
 import { mapToExtendedMetadata } from './mappers';
@@ -9,13 +9,13 @@ import axios, { AxiosInstance } from 'axios';
 const HTTP_CLIENT_TIMEOUT = 1 * 1000;
 const HTTP_CLIENT_MAX_CONTENT_LENGTH = 5000;
 
-export const createHttpStakePoolExtMetadataService = (
+export const createHttpStakePoolMetadataService = (
   logger: Logger,
   axiosClient: AxiosInstance = axios.create({
     maxContentLength: HTTP_CLIENT_MAX_CONTENT_LENGTH,
     timeout: HTTP_CLIENT_TIMEOUT
   })
-): StakePoolExtMetadataService => ({
+): StakePoolMetadataService => ({
   async getStakePoolExtendedMetadata(metadata: Cardano.StakePoolMetadata): Promise<Cardano.ExtendedStakePoolMetadata> {
     const url = getExtMetadataUrl(metadata);
     try {
@@ -30,14 +30,14 @@ export const createHttpStakePoolExtMetadataService = (
           throw new ProviderError(
             ProviderFailure.NotFound,
             error,
-            `StakePoolExtMetadataService failed to fetch extended metadata from ${url} due to resource not found`
+            `StakePoolMetadataService failed to fetch extended metadata from ${url} due to resource not found`
           );
         }
 
         throw new ProviderError(
           ProviderFailure.ConnectionFailure,
           error,
-          `StakePoolExtMetadataService failed to fetch extended metadata from ${url} due to connection error`
+          `StakePoolMetadataService failed to fetch extended metadata from ${url} due to connection error`
         );
       }
       if (error instanceof ValidationError) {
--- a/packages/cardano-services/src/StakePool/HttpStakePoolMetadata/HttpStakePoolMetadataService.ts
+++ b/packages/cardano-services/src/StakePool/HttpStakePoolMetadata/HttpStakePoolMetadataService.ts
@@ -1,13 +1,23 @@
-import { Cardano, ProviderError, ProviderFailure } from '@cardano-sdk/core';
+/* eslint-disable max-len */
+/* eslint-disable complexity */
+/* eslint-disable sonarjs/cognitive-complexity */
+/* eslint-disable max-depth */
+
+import * as Crypto from '@cardano-sdk/crypto';
+import { CML, Cardano } from '@cardano-sdk/core';
+import { Hash32ByteBase16 } from '@cardano-sdk/crypto';
+import { HexBlob } from '@cardano-sdk/util';
 import { Logger } from 'ts-log';
 import { StakePoolExtMetadataResponse, StakePoolMetadataService } from '../types';
+import { StakePoolMetadataResponse, StakePoolMetadataServiceError, StakePoolMetadataServiceFailure } from '../..';
 import { ValidationError, validate } from 'jsonschema';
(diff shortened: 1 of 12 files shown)


packages/cardano-services/src/StakePool/HttpStakePoolMetadata/HttpStakePoolMetadataService.ts
```diff
@@ -93,6 +93,7 @@ export const createHttpStakePoolMetadataService = (
           return { errors: [new CustomError('Invalid metadata: Missing ext signature or public key')], metadata };
         }
 
+        // TODO: We probably need a delay here
```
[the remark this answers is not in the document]
IvayloAndonov: I haven't seen this behavior when testing it manually with a few pool metadata URLs.
So the case is when the first metadata [fetch](https://github.com/input-output-hk/cardano-js-sdk/pull/640/files#diff-a125636e823a5a5ec5840f90b0ce619515a214b5f0523bdd41db32eaa011c1f4R77) has finished, the [second](https://github.com/input-output-hk/cardano-js-sdk/pull/640/files#diff-a125636e823a5a5ec5840f90b0ce619515a214b5f0523bdd41db32eaa011c1f4R97) one sometimes fails with what error? Is it reproducible with the same URLs always?
If it depends on the external pool metadata servers and the delay helps, not sure what other approach we can take atm either. @AngelCastilloB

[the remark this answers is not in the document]
rhyslbw: We need to be resilient to network outages too, so ideally we would make an immediate attempt, then if it fails, retry a few times before giving up for this job anyway.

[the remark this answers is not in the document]
IvayloAndonov: Yes, for the first we can use the same `pRetry` used by [createDnsResolver](https://github.com/input-output-hk/cardano-js-sdk/blob/23bdb6ebb664da90ae5bf877d7a631e13eedfacf/packages/cardano-services/src/Program/utils.ts#L35-L40).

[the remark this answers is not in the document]
mkazlauskas: We should push the retry logic to the worker, because that will be needed either way. The fetcher just needs to return more granular error types (either with some enum like with ProvideFailure, or just distinct error types), so that it can decide what to do.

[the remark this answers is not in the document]
IvayloAndonov: The error granularity looks good to me👍🏼

IvayloAndonov (dismissed):
Great finish of the work and comprehensive test coverage. Nice job!

packages/cardano-services/package.json
```diff
@@ -101,6 +101,7 @@
     "@cardano-sdk/crypto": "^0.1.3",
     "@cardano-sdk/ogmios": "^0.10.0",
     "@cardano-sdk/util": "^0.8.2",
+    "@dcspark/cardano-multiplatform-lib-nodejs": "^3.1.1",
```
rhyslbw, line 104:
```suggestion
```

packages/cardano-services/src/StakePool/HttpStakePoolMetadata/HttpStakePoolMetadataService.ts
```diff
@@ -0,0 +1,171 @@
+/* eslint-disable max-len */
+/* eslint-disable complexity */
+/* eslint-disable sonarjs/cognitive-complexity */
+/* eslint-disable max-depth */
+
+import * as CML from '@dcspark/cardano-multiplatform-lib-nodejs';
```
rhyslbw, line 6:
```suggestion
```

packages/cardano-services/src/StakePool/HttpStakePoolMetadata/HttpStakePoolMetadataService.ts
```diff
@@ -0,0 +1,171 @@
...
+import * as Crypto from '@cardano-sdk/crypto';
+import { Cardano } from '@cardano-sdk/core';
```
rhyslbw, line 8:
```suggestion
import { Cardano, CML } from '@cardano-sdk/core';
```

rhyslbw (changes requested):

packages/cardano-services/src/StakePool/HttpStakePoolMetadata/HttpStakePoolMetadataService.ts
```diff
@@ -0,0 +1,170 @@
...
+              );
+            }
+
+            // If signature url failed -> omit extended metadata from response and add specific error
+          } catch {
+            extMetadata = undefined;
+            errors.push(
+              new StakePoolMetadataServiceError(
+                StakePoolMetadataServiceFailure.FailedToFetchExtendedSignature,
+                null,
```
mkazlauskas, line 147:
```suggestion
          } catch (error) {
            extMetadata = undefined;
            errors.push(
              new StakePoolMetadataServiceError(
                StakePoolMetadataServiceFailure.FailedToFetchExtendedSignature,
                error,
```

packages/cardano-services/src/StakePool/HttpStakePoolMetadata/HttpStakePoolMetadataService.ts
```diff
@@ -0,0 +1,170 @@
...
+      if (axios.isAxiosError(error)) {
+        errors.push(
+          new StakePoolMetadataServiceError(
+            StakePoolMetadataServiceFailure.FailedToFetchMetadata,
+            error.toJSON(),
+            `${SERVICE_NAME} failed to fetch metadata JSON from ${url} due to ${error.message}`
+          )
+        );
+      } else if (error instanceof StakePoolMetadataServiceError) {
+        errors.push(new StakePoolMetadataServiceError(error.reason, error.innerError, error.detail));
```
mkazlauskas, line 164: Is there a reason to clone here?
```suggestion
        errors.push(error);
```

yarn.lock
```diff
@@ -2370,6 +2370,7 @@ __metadata:
     "@cardano-sdk/ogmios": ^0.10.0
     "@cardano-sdk/util": ^0.8.2
     "@cardano-sdk/util-dev": ^0.8.0
+    "@dcspark/cardano-multiplatform-lib-nodejs": ^3.1.1
```
mkazlauskas, line 2373: There are no changes in package.json, please check if running `yarn` still adds this
```suggestion
```

mkazlauskas (changes requested):

packages/cardano-services/src/StakePool/HttpStakePoolMetadata/HttpStakePoolMetadataService.ts
IvayloAndonov: I think we just can push the thrown error

rhyslbw:
Looks like a genuine test failure @AngelCastilloB

mkazlauskas (approved):
Nice work! :rocket:

IvayloAndonov (approved):
Nice work!

mkazlauskas (approved):
:rocket:

rhyslbw (approved):
Great work @IvayloAndonov and @AngelCastilloB  :muscle:

PR #609, opened 2023-02-18
Feat/optionally disable Stake Pool APY metric

rhyslbw:
# Context
On _mainnet_ the current solution of calculating stake pool APY, to cache for later queries, is too resource intensive. As a short-term solution it's acceptable to omit this field from the response.

# Proposed Solution
- Add support for disabling select fields from the Stake Pool provider search response
- Add an option to the CLI to disable the APY metric

PR #641, opened 2023-03-13
feat(cardano-services): support disabling stake pool APY metric

rhyslbw:
# Context
On _mainnet_ the current solution of calculating stake pool APY, to cache for later queries, is too resource intensive. As a short-term solution it's acceptable to omit this field from the response.

# Proposed Solution
- Add support for disabling select fields from the Stake Pool provider search response
- Add an option to the CLI to disable the APY metric

New PR for the same branch reviewed in https://github.com/input-output-hk/cardano-js-sdk/pull/609

### DIFF
--- a/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
+++ b/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
@@ -21,6 +21,7 @@ import {
 import { RunnableModule, isNotNil } from '@cardano-sdk/util';
 import { StakePoolBuilder } from './StakePoolBuilder';
 import { toStakePoolResults } from './mappers';
+import merge from 'lodash/merge';
 
 /**
  * Properties that are need to create DbSyncStakePoolProvider
@@ -30,6 +31,17 @@ export interface StakePoolProviderProps {
    * Pagination page size limit used for provider methods constraint.
    */
   paginationPageSizeLimit: number;
+
+  /**
+   * Configure the response optional fields
+   */
+  responseConfig?: {
+    search?: {
+      metrics?: {
+        apy?: boolean;
+      };
+    };
+  };
 
   /**
    * Enables Blockfrost hybrid cache.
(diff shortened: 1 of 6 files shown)


packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
```diff
@@ -107,6 +126,9 @@ export class DbSyncStakePoolProvider extends DbSyncProvider(RunnableModule) impl
         return (options?: QueryStakePoolsArgs) =>
           this.#builder.queryPoolMetrics(hashesIds, totalStake, useBlockfrost, options);
       case 'apy':
+        if (this.#responseConfig?.search?.metrics?.apy === false) {
```
mkazlauskas, line 129:
```suggestion
        if (this.#responseConfig?.search?.metrics?.apy) {
```
If there's a reason for explicit check then there's a bug [here](https://github.com/input-output-hk/cardano-js-sdk/pull/641/files#diff-3a938d2c168d4b251612c6efabf201ac70e7585c3c1c2149b3c34e71981e6666R273)

mkazlauskas (dismissed):

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
rhyslbw: 6a9b545095363a8468e81f7f5e5a606c148cc86b

rhyslbw:
Also had a bug due to the command rename. Fixed in https://github.com/input-output-hk/cardano-js-sdk/pull/641/commits/402e5bd61ec035c0ef2d5ee3e76cee5df2e3a20b

mkazlauskas (approved):

IvayloAndonov (dismissed):
LGTM! Already approved it btw

mkazlauskas (dismissed):

iccicci (approved):
Great!

PR #642, opened 2023-03-14
feat(cardano-services): configurable pgpool size

rhyslbw:
# Context
There's no exposed interface to control the pgpool size, which impacts ability to tune the application at runtime.

# Proposed Solution
- Plumb in the `PoolConfig` `max` property, exposed on the CLI. The original type covered the previous options, but it was technically incorrect.


iccicci (approved):
Excellent!

mkazlauskas (approved):

IvayloAndonov (approved):

PR #645, opened 2023-03-15
refactor(core)!: move parseAssetId and createAssetId from util.assetId

AngelCastilloB:
# Context

The `parseAssetId` and `createAssetId` util are out of place. Their purpose is to convert to/from CML objects from/to the Core type `Cardano.AssetId`

# Proposed Solution

Move both utils to the CML modules `coreToCml` and `cmlToCore`.


mirceahasegan (approved):

rhyslbw (approved):
Nice work @AngelCastilloB :rocket:

PR #647, opened 2023-03-15
feat: metric for sync status

iadmytro:
# Context

Added new metrics "node_sync_percentage" and "projection_sync_percentage".

### DIFF
--- a/packages/cardano-services/src/Http/HttpServer.ts
+++ b/packages/cardano-services/src/Http/HttpServer.ts
@@ -1,11 +1,11 @@
 import * as OpenApiValidator from 'express-openapi-validator';
+import { DB_BLOCKS_BEHIND_TOLERANCE, listenPromise, serverClosePromise } from '../util';
 import { Gauge, Registry } from 'prom-client';
 import { HttpServerConfig, ServiceHealth, ServicesHealthCheckResponse } from './types';
 import { HttpService } from './HttpService';
 import { Logger } from 'ts-log';
 import { ProviderError, ProviderFailure } from '@cardano-sdk/core';
 import { RunnableModule, contextLogger, fromSerializableObject, toSerializableObject } from '@cardano-sdk/util';
-import { listenPromise, serverClosePromise } from '../util';
 import bodyParser from 'body-parser';
 import express from 'express';
 import expressPromBundle from 'express-prom-bundle';
@@ -69,31 +69,7 @@ export class HttpServer extends RunnableModule {
       })
     );
     if (this.#config?.metrics?.enabled) {
-      const promRegistry = new Registry();
-      const getServicesHealth = async () => await this.#getServicesHealth();
-      this.#healthGauge = new Gauge<string>({
-        async collect() {
-          const currentValue = Number((await getServicesHealth()).ok);
-          // Set a health check gauge value with type number.Health check values: 0/1 (unhealthy/healthy)
-          this.set(currentValue);
-        },
-        help: 'healthcheck_help',
-        labelNames: ['method', 'statusCode'],
-        name: 'healthcheck',
-        registers: [promRegistry]
-      });
-      this.app.use(
-        expressPromBundle({
-          includeMethod: true,
-          promRegistry,
-          ...this.#config.metrics.options
-        })
-      );
-      this.logger.info(
-        `Prometheus metrics: ${(await promRegistry.getMetricsAsArray()).map(({ name }) => name)} configured at ${
-          this.#config.metrics.options?.metricsPath || '/metrics'
-        }`
-      );
+      await this.initMetrics();
     }
     const requestLogger = contextLogger(this.logger, 'request');
     this.app.use((req, _res, next) => {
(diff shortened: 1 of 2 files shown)


IvayloAndonov:
Could you please attach some visuals of how the metrics are printed at the end? @iadmytro

mkazlauskas (changes requested):

packages/cardano-services/src/Http/HttpServer.ts
```diff
@@ -180,4 +156,49 @@ export class HttpServer extends RunnableModule {
 
     return serverClosePromise(this.server);
   }
+
+  private async initMetrics() {
+    const promRegistry = new Registry();
+    for (const service of this.#dependencies.services) {
```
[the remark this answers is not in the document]
rhyslbw: You're right @mkazlauskas, based on our program sharing a node between all registered services, the scaling of a backend to use multiple nodes is managed by running multiple instances of the provider server. The service-level metric is more suited to reporting the projected sync progress, which is what I had in mind when discussing the implementation with @iadmytro, which is the other metric we need here to visualise the health of the service.

Let's switch this to be reporting the projected sync status, then add another top level metric that reports the node sync status, which can be mapped to the first item in the services array.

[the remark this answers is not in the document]
iadmytro: > Could you please attach some visuals of what is the metrics output and how it looks with the new sync status? 

image

rhyslbw (changes requested):

packages/cardano-services/src/Http/HttpServer.ts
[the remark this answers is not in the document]
rhyslbw: > ...then add another top level metric that reports the node sync status, which can be mapped to the first item in the services array.

It looks like the implementation has the **node sync percentage** being reported at the service level @iadmytro, rather than what I suggested above.

packages/ogmios/src/CardanoNode/OgmiosCardanoNode.ts
```diff
@@ -99,7 +99,7 @@ export class OgmiosCardanoNode extends RunnableModule implements CardanoNode {
       });
       response.localNode = {
         ledgerTip: lastKnownTip,
-        networkSync: Cardano.Percent(networkSynchronization)
+        networkSync: Cardano.Percent(networkSynchronization * 100)
```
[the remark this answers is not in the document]
iadmytro: moved to separate commit [c110961](https://github.com/input-output-hk/cardano-js-sdk/pull/647/commits/c11096155c54c1cb344e5e79870e9cb039c5e75e)

packages/cardano-services/src/Http/HttpServer.ts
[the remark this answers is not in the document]
iadmytro: 'node sync percentage' moved to top level in [b806319](https://github.com/input-output-hk/cardano-js-sdk/pull/647/commits/b8063199a9caa511ed58b5aa65f578054be19be4)

packages/core/test/CardanoNode/mocks.ts
```diff
@@ -69,7 +69,7 @@ export const healthCheckResponseMock = (opts?: {
       hash: opts?.hash ?? '9ef43ab6e234fcf90d103413096c7da752da2f45b15e1259f43d476afd12932c',
       slot: opts?.slot ?? 52_819_355
     },
-    networkSync: opts?.networkSync ?? Cardano.Percent(0.999)
+    networkSync: opts?.networkSync ?? Cardano.Percent(99.9)
```
mkazlauskas, line 72: `Cardano.Percent` type should only be used with original format. Please create a new opaque number type, or just use `number` for `networkSync`.
```suggestion
    networkSync: opts?.networkSync ?? 99.9
```

packages/cardano-services/src/Http/HttpServer.ts
```diff
@@ -180,4 +156,69 @@ export class HttpServer extends RunnableModule {
...
+      new Gauge<string>({
+        async collect() {
+          const data = await service.healthCheck();
+          let syncStatus;
+          if (data.projectedTip && data.localNode?.ledgerTip) {
+            syncStatus =
+              (data.projectedTip.blockNo * 100) / (data.localNode?.ledgerTip?.blockNo - DB_BLOCKS_BEHIND_TOLERANCE);
+          }
+          if (syncStatus) {
+            if (syncStatus > 100) this.set(100);
```
mkazlauskas, line 187: This probably means that a rollback happened and projector hasn't caught up yet. Would be best to not show it as 100%
```suggestion
            if (syncStatus > 100) this.set(syncStatus - 100);
```

iadmytro: The intention is to show 100% sync status if it is within DB_BLOCKS_BEHIND_TOLERANCE range.
When `projectedTip.blockNo == ledgerTip.blockNo` the formula above will show result slightly more than 100, so `if (syncStatus > 100) this.set(100);` just ensures that reslut will 100% when projectedTip is somewhere in `(ledgerTip - DB_BLOCKS_BEHIND_TOLERANCE <=  projectedTip <= ledgerTip)`

mkazlauskas: Oh I see, ok. It seems like an odd behavior to me. Why not just report the truth (the 99.9x%)?

mkazlauskas (dismissed):
Nice work! Please squash the commit and scope it to `feat(cardano-services):`

mkazlauskas (approved):

packages/cardano-services/src/Http/HttpServer.ts
```diff
@@ -180,4 +156,69 @@ export class HttpServer extends RunnableModule {
 
     return serverClosePromise(this.server);
   }
+
+  // eslint-disable-next-line sonarjs/cognitive-complexity
+  private async initMetrics() {
+    const promRegistry = new Registry();
+    const anyService = this.#dependencies.services[0];
```
IvayloAndonov, line 163: Question: doesn't matter which service we pick here as first, correct? @iadmytro

iadmytro: yes, since services reporting same node sync status

iadmytro:
> Shouldn't we reset the gauge value in service shutdown() similarly to the `healthGauge` [here](https://github.com/input-output-hk/cardano-js-sdk/blob/51f30a1ee827586da720ad6bc25262b3ef41e5cd/packages/cardano-services/src/Http/HttpServer.ts#L176). For that purpose we need to assign the gauge to a variable and to set the default value in `shutdownImpl()` ? 

I don't think that we need to do something special for shutdown, whenever server is unavailable it just stops sending metrics to Prometheus and will be showed as interrupted line.

mkazlauskas (approved):

IvayloAndonov (approved):
LGTM!

PR #651, opened 2023-03-16
refactor!: the TxSubmit endpoint no longer adds the stack trace when …

AngelCastilloB:
# Context

We need to strip the stack trace from domain error at the HTTP API when running in a production env.

# Proposed Solution
 - Strip the stack trace from domain error at the HTTP API

### DIFF
--- a/packages/util/src/environment.ts
+++ b/packages/util/src/environment.ts
@@ -1,0 +1,22 @@
+/**
+ * Node.js environment configurations.
+ */
+export enum Environment {
+  /**
+   * Production environment.
+   *
+   * Node.js assumes it's always running in a development environment. You can signal Node.js that you are running
+   * in production by setting the NODE_ENV=production environment variable.
+   */
+  Production = 'production',
+
+  /**
+   * Development environment.
+   */
+  Development = 'development'
+}
+
+export const isProductionEnvironment = (): boolean => process.env.NODE_ENV === Environment.Production;
+
+export const getEnvironmentConfiguration = (): Environment =>
+  isProductionEnvironment() ? Environment.Production : Environment.Development;
--- a/packages/util/src/errors.ts
+++ b/packages/util/src/errors.ts
@@ -5,6 +5,18 @@ interface ErrorLike {
   stack: string;
 }
 
+interface WithInnerError {
+  innerError: string | Error;
+}
+
+/**
+ * Gets whether the given error has an innerError.
+ *
+ * @param error The error to be checked for.
+ */
+const isWithInnerError = (error: unknown): error is WithInnerError =>
+  error !== null && typeof error === 'object' && 'innerError' in (error as never);
+
 /**
  * This type check works as an "error instanceof Error" check, but it let pass also those objects
  * which implements the Error interface without inheriting from the same base class
@@ -20,6 +32,23 @@ const isErrorLike = (error: unknown): error is ErrorLike => {
   const { message, stack } = error as ErrorLike;
 
   return typeof message === 'string' && typeof stack === 'string';
+};
+
+/**
+ * Strips the stack trace of all errors and their inner errors recursively.
+ *
+ * @param error The error to be stripped of its stack trace.
+ */
+export const stripStackTrace = (error: unknown) => {
+  if (!error) return;
+
+  if (isErrorLike(error)) {
+    delete (error as Error).stack;
+  }
+
+  if (isWithInnerError(error)) {
+    stripStackTrace(error.innerError);
+  }
 };
 
 export class ComposableError<InnerError = unknown> extends CustomError {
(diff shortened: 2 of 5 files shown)


packages/ogmios/src/CardanoNode/OgmiosCardanoNode.ts
```diff
@@ -65,7 +65,12 @@ export class OgmiosCardanoNode extends RunnableModule implements CardanoNode {
       this.#logger.info('Getting system start');
       return await this.#stateQueryClient.systemStart();
     } catch (error) {
-      throw CardanoNodeUtil.asCardanoNodeError(error) || new CardanoNodeErrors.UnknownCardanoNodeError(error);
+      const domainError =
+        CardanoNodeUtil.asCardanoNodeError(error) || new CardanoNodeErrors.UnknownCardanoNodeError(error);
```
mkazlauskas, line 69: Maybe let's keep the stack trace for UnknownCardanoNodeError - it could be useful

mkazlauskas (changes requested):
If we're stripping stack trace from `systemStart` and `stakeDistribution`, then we should also do it for `eraSummaries`. However I would limit the scope of this task to just stripping the stack trace of `submitTx`.

rhyslbw:
How about stripping it here: https://github.com/input-output-hk/cardano-js-sdk/blob/87f4160424c0f21a050365af75b0b22056bdb3c2/packages/cardano-services/src/Http/HttpService.ts#L69-L73

It moves the stripping of this server-relevant information in **all** responses from a single point in, arguably, the correct layer; From the HTTP interactions, which is a boundary we're looking to protect. This will result in smaller payloads for all exceptions, and more assurance we're not missing particular cases.

packages/util/src/errors.ts
```diff
@@ -5,6 +5,18 @@ interface ErrorLike {
   stack: string;
 }
 
+interface HasInnerErrorLike {
```
mkazlauskas, line 8: I think that's a little nicer, but feel free to keep the original
```suggestion
interface WithInnerError {
```

packages/util/src/errors.ts
```diff
@@ -22,6 +34,23 @@ const isErrorLike = (error: unknown): error is ErrorLike => {
   return typeof message === 'string' && typeof stack === 'string';
 };
 
+/**
+ * Strips the stack trace of all errors and their inner errors recursively.
+ *
+ * @param error The error to be stripped of its stack trace.
+ */
+export const stripStackTrace = (error: unknown) => {
```
mkazlauskas, line 42: Need unit tests since new utils are exported from index

packages/util/src/environment.ts
```diff
@@ -0,0 +1,35 @@
...
+  Development = 'development',
+
+  /**
+   * Unknown environment.
+   */
+  Unknown = 'Unknown'
+}
+
+export const isProductionEnvironment = (): boolean => process.env.NODE_ENV === Environment.Production;
+export const isDevelopmentEnvironment = (): boolean => process.env.NODE_ENV === Environment.Production;
```
mkazlauskas, line 27: No NODE_ENV is typically interpreted as development. I'd remove the Unknown variant too.
```suggestion
export const isDevelopmentEnvironment = (): boolean => process.env.NODE_ENV !== Environment.Production;
```

mkazlauskas (changes requested):

packages/util/src/environment.ts
```diff
@@ -0,0 +1,27 @@
...
+  /**
+   * Development environment.
+   *
+   * You can signal Node.js that you are running by setting the NODE_ENV=development environment variable.
+   */
+  Development = 'development'
+}
+
+export const isProductionEnvironment = (): boolean =>
+  process.env.NODE_ENV ? process.env.NODE_ENV === Environment.Production : false;
```
mkazlauskas, line 20:
```suggestion
  process.env.NODE_ENV === Environment.Production;
```

mkazlauskas (approved):
Nice work, no need for fixup for last change request

mirceahasegan (dismissed):
Great work, @AngelCastilloB !

rhyslbw (dismissed):

mkazlauskas (dismissed):

packages/util/src/errors.ts
```diff
@@ -5,6 +5,18 @@ interface ErrorLike {
...
+  innerError: string | Error;
+}
+
+/**
+ * Gets whether the given error has an innerError.
+ *
+ * @param error The error to be checked for.
+ */
+const isWithInnerError = (error: unknown): error is WithInnerError =>
+  error !== undefined && typeof error === 'object' && 'innerError' in (error as never);
```
iccicci, line 18:
```suggestion
  error !== null && typeof error === 'object' && 'innerError' in (error as never);
```
`typeof null === "object"`
while
`typeof undefined === "undefined"`

PR #655, opened 2023-03-17
test(cardano-services): improve CLI test - assertion and logging

rhyslbw:
# Context
The CLI test is producing hard to debug failures in CI. There's a fragile assertion that swallows the underlying failure, and we're blind to the child process failures since their output is not logged.

# Proposed Solution
- Improve assertion to better reveal failure
- Add a configurable logger to the CLI tests

Not in scope is a way to dynamically set the logger level, so for now, if needing to debug, we can just update the min level in the code. Ideally we should be able to choose this at runtime, or use some other indicator such as commit message content to set it in the workflow.


iccicci (approved):
GJ!

mkazlauskas (approved):
Nice work! :rocket:

iadmytro (approved):

PR #656, opened 2023-03-17
yarn test flakiness

iccicci:
# Context

`yarn workspace @cardano-sdk/cardano-services test` suffers some flakiness.

# Proposed solution

Handling the output chunks should resolve one source of flakiness.

- The change to `jest-setup.ts` does not actually addresses the other source of flakiness, but at least makes the not yet identified error to immediately throw rather than making the test hanging silently. Hopefully this can give us some more info to address the root cause.
- Fixed 4 occurrences of the typo `--service_names` instead of `--service-names`
- `dataMatchOnError` is now required to be sure the got error is the expected one.


mkazlauskas (approved):
Great work!

rhyslbw (approved):
Great work @iccicci @and @IvayloAndonov :muscle:

PR #657, opened 2023-03-17
Chore/yarn test flakiness insights

IvayloAndonov:
# Context

Add extra jest flags to get more insights in case of hanging/errors in CI.

# Proposed Solution

- fix wong usage of CLI arg in tests
- removed `--forceExit` from CI/CD commands
- `--expand`: Use this flag to show full diffs and errors instead of a patch
- `--detectOpenHandles` logs out errors preventing your tests from exiting clearly and it implies `--runInBand` which ensures your tests run in the same threads so no overlapping. 
-  `--showConfig` print your Jest config and then exits. The reason I call it to print the jest config of the remote CI/CD machine and get more insights such as `maxConcurreny` and `maxWorkers`

### DIFF
--- a/.github/workflows/continuous-integration-unit-tests.yaml
+++ b/.github/workflows/continuous-integration-unit-tests.yaml
@@ -36,4 +36,4 @@ jobs:
       run: |
         yarn test:build:verify
         yarn test --showConfig
-        yarn test --forceExit --expand --detectOpenHandles
+        TL_LEVEL=debug yarn test --forceExit --expand --detectOpenHandles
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration-unit-tests.yaml
```diff
@@ -35,4 +35,5 @@ jobs:
     - name: 🔬 Test
       run: |
         yarn test:build:verify
-        yarn test --forceExit
+        yarn test --showConfig
+        TL_LEVEL=info yarn test --expand --detectOpenHandles
```
mkazlauskas, line 39: `--detectOpenHandles` [makes the tests run serially](https://jestjs.io/docs/cli#--detectopenhandles) - it will slow down most of the test suites that can be run in parallelized mode. However it might make no performance difference when run in CI, so this might be ok.
```suggestion
        TL_LEVEL=info yarn test --expand
```

PR #661, opened 2023-03-20
fix(cardano-services): explicitly process.exit(1) on unhandledRejections

mkazlauskas:
# Context

Nightly builds have been failing on Node v14

# Proposed Solution

Process is not existing on unhandledRejections by default in Node 14. Explicitly exit when it happens.


iccicci (dismissed):
LGTM

rhyslbw (changes requested):

rhyslbw (approved):

IvayloAndonov (approved):
LGTM

PR #662, opened 2023-03-20
Test/stake pool search load test review

iccicci:
# Context

The stake pool search load test as was born is not configurable, making it configurable will let us to perform different load test scenarios.

While working on it I identified some actions, from minor to major, and I tried to address them as well.

# Proposed Solution

- Changed the stake pool search load test to accept configuration variables
- Changed the `DbSyncStakePoolProvider` to skip the query for stake distribution when configured to use Blockfrost hybrid solution: the total stake amount is required to compute the `saturation` which is a value cached from Blockfrost.
- Changed the `DbSyncStakePoolProvider` to not repeat the same on going queries.

- Removed a few lines of unused code
- Applied some minor improvements to the `docker-compose.yml` file.

### DIFF
--- a/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
+++ b/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
@@ -269,7 +269,7 @@ export class DbSyncStakePoolProvider extends DbSyncProvider(RunnableModule) impl
     const totalCount = poolUpdates.length;
     // Get last epoch data
     const lastEpoch = await this.#builder.getLastEpochWithData();
-    const { optimalPoolCount, no: lastEpochNo } = lastEpoch;
+    const { no: lastEpochNo } = lastEpoch;
     // Get stake pools data cached
     const { orderedResultHashIds, orderedResultUpdateIds, orderedResult, poolDatas, hashesIds, sortType } =
       await this.#cache.get(queryCacheKey(StakePoolsSubQuery.POOLS_DATA_ORDERED, options), () =>
@@ -317,10 +317,6 @@ export class DbSyncStakePoolProvider extends DbSyncProvider(RunnableModule) impl
     );
     const { results, poolsToCache } = toStakePoolResults(orderedResultHashIds, fromCache, useBlockfrost, {
       lastEpochNo: Cardano.EpochNo(lastEpochNo),
-      nodeMetricsDependencies: {
-        optimalPoolCount,
-        totalAdaAmount: BigInt(totalStake)
-      },
       poolAPYs,
       poolDatas,
       poolMetrics,
--- a/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
+++ b/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
@@ -114,7 +114,7 @@ export class DbSyncStakePoolProvider extends DbSyncProvider(RunnableModule) impl
 
   private getQueryBySortType(
     sortType: PoolSortType,
-    queryArgs: { hashesIds: number[]; updatesIds: number[]; totalStake: string },
+    queryArgs: { hashesIds: number[]; updatesIds: number[]; totalStake: string | null },
     useBlockfrost: boolean
   ) {
     const { hashesIds, updatesIds, totalStake } = queryArgs;
@@ -161,7 +161,7 @@ export class DbSyncStakePoolProvider extends DbSyncProvider(RunnableModule) impl
 
   private async getPoolsDataOrdered(
     poolUpdates: PoolUpdate[],
-    totalStake: string,
+    totalStake: string | null,
     useBlockfrost: boolean,
     options?: QueryStakePoolsArgs
   ) {
@@ -200,7 +200,7 @@ export class DbSyncStakePoolProvider extends DbSyncProvider(RunnableModule) impl
   private async queryExtraPoolsData(
     idsToFetch: PoolUpdate[],
     sortType: PoolSortType,
-    totalStake: string,
+    totalStake: string | null,
     orderedResult: OrderedResult,
     useBlockfrost: boolean
   ) {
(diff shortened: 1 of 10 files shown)


packages/e2e/test/artillery/StakePoolSearch.yml
```diff
@@ -6,12 +6,11 @@ config:
     timeout: 180
   phases:
     - name: 'Stake Pool Search'
-      duration: 10
-      arrivalRate: 1
+      duration: 1
+      arrivalCount: 1
```
IvayloAndonov, line 10: Is this 1 VU enough for the phase?

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
```diff
@@ -254,22 +250,24 @@ export class DbSyncStakePoolProvider extends DbSyncProvider(RunnableModule) impl
       this.#builder.queryPoolHashes(query, params)
     );
     // Get cached total staked amount
-    const totalStake = await this.#cache.get(queryCacheKey(StakePoolsSubQuery.TOTAL_STAKE), async () => {
-      const distribution = await this.cardanoNode.stakeDistribution();
+    const totalStake = this.#useBlockfrost
+      ? '1'
```
IvayloAndonov, line 254: Why do we use '1' here? Some defaulting if Blockfrost is used?

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
```diff
@@ -297,42 +295,61 @@ export class DbSyncStakePoolProvider extends DbSyncProvider(RunnableModule) impl
...
+        poolOwners,
+        poolRegistrations,
+        poolRelays,
+        poolRetirements,
+        poolRewards: poolRewards.filter(isNotNil),
+        totalCount
+      });
+    };
+
+    const resultPromise = queriesAndMap();
```
IvayloAndonov, line 331: The name of the internal func`queriesAndMap` sounds a bit confusing to me. It does many things + maps at the end.

packages/e2e/test/artillery/StakePoolSearch.yml
iccicci: Yes, that's the phase for the first user, the first user needs to perform a query without filters to fetch all the pools

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
```diff
@@ -297,42 +295,61 @@ export class DbSyncStakePoolProvider extends DbSyncProvider(RunnableModule) impl
...
+        resultPromise.then(({ poolsToCache }) => {
+          const pool = poolsToCache[hashId as unknown as number];
+
+          this.#cache.set(cacheKey, pool, UNLIMITED_CACHE_TTL);
+
+          return pool;
+        }),
+        UNLIMITED_CACHE_TTL
+      );
+    }
```
IvayloAndonov, line 349: Could it be extracted into a method?

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
iccicci: That value is unused when blockfrost is used, any value is fine.

IvayloAndonov: Maybe const `STAKE_NOT_USED = '0' | 'NotUsed'` ?

iccicci: 9d34b4639 should solve this

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
iccicci: 9d34b4639 should solve this

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
iccicci: 9d34b4639 should solve this

rhyslbw (changes requested):
Looks good @iccicci.

Minor improvement to the commit message of https://github.com/input-output-hk/cardano-js-sdk/pull/662/commits/524d170e2c7d0397403d410d762b9e92e9a22965

> util to resolve all promises contained in an object

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
mkazlauskas: `null` typically represents absence of value. Why not just use it? `queryExtraPoolsData` argument type can also be updated to `string | null` - then it's more explicit.

packages/util/src/util.ts
```diff
@@ -0,0 +1,12 @@
+export type PromiseOrValue<T> = Promise<T> | T;
+
+export const resolveObjectOfPromises = async <T>(obj: {
+  [k: string]: PromiseOrValue<T>;
+}): Promise<{ [k: string]: T }> =>
+  Object.fromEntries(
+    await Promise.all(
+      Object.entries(obj).map(([key, promise]) =>
+        promise instanceof Promise ? promise.then((value) => [key, value]) : [key, promise]
```
mkazlauskas, line 9: `instanceof Promise` can be brittle, there are many libraries that use their own Promise-like types, polyfills etc.
```suggestion
        Promise.resolve(promise).then((value) => [key, value])
```

Nice util btw :1st_place_medal:

packages/util/src/util.ts
```diff
@@ -0,0 +1,12 @@
+export type PromiseOrValue<T> = Promise<T> | T;
+
+export const resolveObjectOfPromises = async <T>(obj: {
```
mkazlauskas, line 3: How about this name?
```suggestion
export const resolveObjectValues = async <T>(obj: {
```

mkazlauskas (changes requested):

packages/util/src/util.ts
iccicci: Nice! Thank you

packages/util/src/util.ts
iccicci: 60b094607 should solve this

packages/util/src/util.ts
iccicci: 60b094607 should solve this

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/DbSyncStakePoolProvider.ts
iccicci: @IvayloAndonov , @mkazlauskas 's suggestion is even better.
abb9b0178 should solve this

mkazlauskas (dismissed):
Nice work! :rocket:

IvayloAndonov (dismissed):
Great job @iccicci ! Just one conflict needs to be resolved.

rhyslbw (approved):

PR #663, opened 2023-03-21
feat: Make request timeout to external services configurable

IvayloAndonov:
# Context

We need to expose an interface for request timeouts to external sources such as the TokenMetadata registry, so it’s configurable while performing fetching through HTTP.

# Proposed Solution

Logged variation of internal network errors related to the Token Metadata Server
Screenshot 2023-03-21 at 17 17 31

- expose `TOKEN_METADATA_REQUEST_TIMEOUT` env var
- expose `--token-metadata-request-timeout` CLI arg
- use default request timeout of `3` seconds for the token metadata registry
- log corresponding extended error message when the client doesn't get the actual error
- refactor `mockTokenRegistry()` with async handler

### DIFF
--- a/packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
+++ b/packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
@@ -5,6 +5,7 @@ import { TokenMetadataService } from './types';
 import axios, { AxiosInstance } from 'axios';
 
 export const DEFAULT_TOKEN_METADATA_CACHE_TTL = 600;
+export const DEFAULT_TOKEN_METADATA_REQUEST_TIMEOUT = 3 * 1000;
 export const DEFAULT_TOKEN_METADATA_SERVER_URL = 'https://tokens.cardano.org';
 
 interface NumberValue {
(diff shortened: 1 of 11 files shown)


iccicci (approved):
LGTM

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
```diff
@@ -152,10 +162,9 @@ export class CardanoTokenRegistry implements TokenMetadataService {
         throw new ProviderError(
           ProviderFailure.ConnectionFailure,
```
mkazlauskas, line 163: Server responding with ConnectionFailure is leaking internals and could be confusing from HTTP client's perspective. Maybe Unhealthy would be better? Not a change request since this was already here.

mkazlauskas (approved):
Nice work!

packages/cardano-services/src/Asset/CardanoTokenRegistry.ts
rhyslbw: Connection failure error is fine since it must already be interpreted by the `AssetProvider` to set the HTTP response as `undefined`. We could improve this by not using `ProviderError` though, given `TokenMetadataService` isn't a provider.

mkazlauskas: Good points :+1:

rhyslbw (approved):
Great work @IvayloAndonov

mkazlauskas (approved):

PR #666, opened 2023-03-22
feat: Optimised health checks with dedicated pool

rhyslbw:
# Context
If the connection pool is saturated, health check queries to the DB will timeout, which can signal a false positive for an unhealthy service. We're also making identical requests to the DB when calling the top-level `/health` endpoint, due to the modular server design, resulting in the over-consumption of clients.
In addition, we're not caching the result to serve independent API requests that come close together.

# Proposed Solution
- Add a separate short-lived cache for the health check results, covering both the DB and local state query results.
- Provision a dedicated DB Pool for health checks within the `DbSyncProvider`, so now we have two distinct pools, one for the main DB operations and the second for health checks only


packages/cardano-services/src/Program/services/postgres.ts
```diff
@@ -51,6 +51,41 @@ export const getPoolWithServiceDiscovery = async (
...
+ * - use passed srv service name in order to resolve the port
+ * - make dealing with failover (re-resolving the port) opaque
+ * - use exponential backoff retry internally with default timeout and factor
+ * - intercept 'query' operation and handle connection errors runtime
+ * - all other operations are bind to pool object without modifications
+ *
+ * @returns pg.Pool instance
+ */
+export const getPoolClient = async (pgConnectionCreator: PgConnectionCreator, logger: Logger): Promise<PoolClient> => {
+  let poolClient = await pgConnectionCreator('fixme');
```
iccicci, line 65: `'fixme'`?

rhyslbw: Yes, pushing early to run in CI. I'm getting a error in the test global setup that I want to ensure is reproducible

IvayloAndonov:
@rhyslbw you are also in the list of reviewers here, but since the draft PR was raised by you, I can not add you right now 😀

packages/cardano-services/src/util/DbSyncProvider/DbSyncProvider.ts
```diff
@@ -1,16 +1,38 @@
...
+
 /**
  * Properties that are need to create DbSyncProvider
  */
 export interface DbSyncProviderDependencies extends ProviderDependencies {
   /**
-   * DB connection
+   * Cache engines. Default: InMemoryCache with HealthCheck.cacheTTL as default TTL
+   */
+  cache?: {
```
iccicci, line 29: Both `cache` and `healthCheck` should be required.
All the `DbSyncProvider`s should share the same cache instance; if each `DbSyncProvider` instantiate its own cache, we are going to repeat the same query for each of them.

packages/cardano-services/src/util/DbSyncProvider/DbSyncProvider.ts
```diff
@@ -26,27 +48,37 @@ export const DbSyncProvider = <
...
-      const { db, cardanoNode, logger } = dependencies;
-
+      const { cache, dbPools, cardanoNode, logger } = dependencies;
       super(...baseArgs);
 
-      this.db = db;
+      this.dbPools = dbPools;
       this.cardanoNode = cardanoNode;
       this.logger = logger;
+      this.#cache = cache?.healthCheck
```
iccicci, line 66: This can be simplified once made required the cache engine

iccicci (changes requested):
Nice, let's just do it a bit more strict changing some optional attributes in required options

packages/cardano-services/src/util/DbSyncProvider/DbSyncProvider.ts
```diff
@@ -1,16 +1,38 @@
...
 import { Pool } from 'pg';
 
+/**
+ * Dedicated DB pools
+ */
+export interface DbPools {
+  /**
+   * Main operational db pool
+   */
+  main?: Pool;
```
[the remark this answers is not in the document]
IvayloAndonov: Good call, the usage of `Required<>` is more times than `Partial<>`. The only case where `db` could be `undefined` is if we have started on `TxSubmit` service which doesn't depend on `DbSync`. 
That's why [getPool()](https://github.com/input-output-hk/cardano-js-sdk/blob/897b9e1f2fc4fdd12350a5f4d80daa9ba36e078a/packages/cardano-services/src/Program/services/postgres.ts#L56-L76) still can return `undefined`

packages/cardano-services/src/util/DbSyncProvider/DbSyncProvider.ts
IvayloAndonov: I think the intention for this to be optional was to facilitate our tests preps, so the `cache` creation can be skipped there. As in the [providerServer](https://github.com/input-output-hk/cardano-js-sdk/blob/1f1a2c83a1c8208ddb7dd2d6decc88c74a1f719b/packages/cardano-services/src/Program/programs/providerServer.ts#L82-L83) impl we always create the shared `healthCheckCache`.
Going to make it required @iccicci

packages/cardano-services/src/util/DbSyncProvider/DbSyncProvider.ts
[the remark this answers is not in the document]
IvayloAndonov: 426169643da6936df4ff1d85b19c29fc897c9135 should address it

packages/cardano-services/src/util/DbSyncProvider/DbSyncProvider.ts
IvayloAndonov: 426169643da6936df4ff1d85b19c29fc897c9135

packages/cardano-services/src/util/DbSyncProvider/DbSyncProvider.ts
IvayloAndonov: 426169643da6936df4ff1d85b19c29fc897c9135

iccicci (dismissed):
Much better now

iccicci (approved):

mkazlauskas (approved):
Nice work! :rocket:

PR #667, opened 2023-03-23
test(cardano-services): fixes a not deterministic error in tests

iccicci:
# Context

Our unit tests are fetching `http://some_url/unfetchable` to check error handling.
The problem is that the error may depend on external system (the dns server) causing the error to be not deterministic: this can make the tests to fail.

# Proposed Solution

Changed the URL to call into a mocked URL responding with error.

### DIFF
--- a/packages/cardano-services/test/StakePool/HttpStakePoolMetadataService/HttpMetadataService.test.ts
+++ b/packages/cardano-services/test/StakePool/HttpStakePoolMetadataService/HttpMetadataService.test.ts
@@ -5,6 +5,7 @@ import { Cardano } from '@cardano-sdk/core';
 import { DataMocks } from '../../data-mocks';
 import { ExtMetadataFormat } from '../../../src/StakePool/types';
 import { Hash32ByteBase16 } from '@cardano-sdk/crypto';
+import { IncomingMessage } from 'http';
 import {
   StakePoolMetadataServiceError,
   StakePoolMetadataServiceFailure,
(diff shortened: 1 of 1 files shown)


IvayloAndonov (approved):
LGTM!

packages/cardano-services/test/StakePool/HttpStakePoolMetadataService/HttpMetadataService.test.ts
```diff
@@ -202,23 +203,19 @@ describe('StakePoolMetadataService', () => {
...
-        };
+        return { body: 'not found', code: 404 };
       };
 
       ({ closeMock, serverUrl } = await mockPoolExtMetadataServer(handler));
 
       metadata = {
         ...mainExtMetadataMock,
         extDataUrl: `${serverUrl}/${ExtMetadataFormat.CIP6}`,
-        extSigUrl: UNFETCHABLE,
```
mkazlauskas, line 221: Is this variable still needed? If yes, maybe update other tests that use it as well?

mkazlauskas (approved):

packages/cardano-services/test/StakePool/HttpStakePoolMetadataService/HttpMetadataService.test.ts
iccicci: It is used in other tests which do not perform the fetch.
I tried to do that but that would increase the test complexity as now a pre computed hash can be used (the URL is always the same), changing it requires computing the hash on some more tests... I think this doesn't worth.

mkazlauskas: Alright, let's keep it

PR #668, opened 2023-03-23
fix: circular deps check in CI

IvayloAndonov:
# Context
Currently `circular-deps:check` script does not exit appropriately if there are circular refs in the compiled source code. This is cause by `--summary` flag we use for more structured and beautiful output but it masks the exiting and always exits with `0` which is not documented.

- remove `--summary` flag so exit codes are correct
- resolve all circular refs within packages

# Note 
Packages `projection-typeorm` and `core` have circular dependencies logically which is fine. 
For example, in `typeorm` 2 distinct entities are referencing each other from two files.
In `core` the case is with `Address` and `BaseAddress` ts files references.


packages/cardano-services/src/Asset/AssetHttpService.ts
```diff
@@ -2,7 +2,7 @@ import * as OpenApiValidator from 'express-openapi-validator';
 import { AssetProvider } from '@cardano-sdk/core';
 import { HttpService } from '../Http';
 import { Logger } from 'ts-log';
-import { ServiceNames } from '../Program';
+import { ServiceNames } from '../Program/programs/types';
```
mkazlauskas, line 5: This appears to be in a wrong file as it's not a type. Not required to resolve to merge this PR

mkazlauskas (approved):
Nice work!

iccicci (approved):
Very nice!

iccicci (changes requested):
Better to change `madge --circular dist` to `madge --circular dist/cjs 2> /dev/null`.
`dist` to `dist/cjs` to make it faster;
`2> /dev/null` to produce a more clean output.
This in all `package.json` files

iccicci (approved):

IvayloAndonov:
Squashed and ready for merge @mkazlauskas  @iccicci

mkazlauskas (approved):

PR #670, opened 2023-03-27
feat(cardano-services): ledgerTip query result is now cached with a 1 second TTL

AngelCastilloB:
# Context

The ledger tip API result remains the same for each tick of the chain, so [querying the db ](https://github.com/input-output-hk/cardano-js-sdk/blob/96fabf383b31eafc90f27ada4afaa83d65535af8/packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts#L63) for every request places unnecessary load on the DB, and consumes clients from the limited pool, when we could just be caching the value in memory.

# Proposed Solution

add a short-lived cache with a non-configurable TTL of 3 seconds

### DIFF
--- a/packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
+++ b/packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
@@ -4,6 +4,7 @@ import {
   CardanoNodeUtil,
   EraSummary,
   NetworkInfoProvider,
+  Seconds,
   SlotEpochCalc,
   StakeSummary,
   SupplySummary,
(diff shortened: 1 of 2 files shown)


packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
```diff
@@ -17,6 +17,8 @@ import { RunnableModule } from '@cardano-sdk/util';
 import { toGenesisParams, toLedgerTip, toProtocolParams, toSupply } from './mappers';
 import memoize from 'lodash/memoize';
 
+export const LEDGER_TIP_TTL = 3; // In seconds
```
mkazlauskas, line 20: We can use the type we have instead of comment. Also since the tip query is now quite efficient let's go with 1s.
```suggestion
export const LEDGER_TIP_TTL = Seconds(1);
```

mkazlauskas (changes requested):

packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/DbSyncNetworkInfoProvider.ts
mirceahasegan: We could consider setting the cache value based on the slot length

mkazlauskas (dismissed):
Nice work! Please squash

mkazlauskas (approved):

PR #671, opened 2023-03-27
Draft: Feat/adp 2802 implement serialization of tx main body

AngelCastilloB:
# Context

We want to remove the dependency of the SDK with the CML library.

# Proposed Solution

Create a Transaction class which can serialize and serialize to CBOR format. This class must also be able to convert to and from Core.Tx type


packages/core/test/Serialization/Transaction.test.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    tx.setAuxiliaryData(scope.manage(tx2.auxiliaryData()));
+
+    // Perform a round trip serialization.
+    const tx3 = await Transaction.fromCbor(tx.toCbor());
+
+    expect(tx3.auxiliaryData()).toEqual(undefined);
+  });
+
+  it('can performs a deep clone of the object', async () => {
```
mkazlauskas, line 87:
```suggestion
  it('can perform a deep clone of the object', async () => {
```

mkazlauskas (changes requested):
Looking great @AngelCastilloB :tada: 
Could we replace usages of `coreToCml.tx` and `cmlToCore.newTx` with this class? Existing tests would add extra confidence.

mkazlauskas (approved):
Great work! :rocket: please squash for a final review

mkazlauskas (dismissed):

PR #672, opened 2023-03-27
fix(wallet): adjust time to wait before resubmit

mirceahasegan:
# Context

Lace preprod network logs show some failed ValueNotConservedError transactions with consumed coins: 0.

# Proposed Solution
Blocks produced more than 250 slots apart are rare (1 per epoch or less), given the current Preprod/Mainnet params of activeSlotCoeff: 0.05 and epochLength: 432000.
Updated the default, currently derived from the poll.maxInterval of 100 seconds, to be multiplied with 2.5 when used as reemit timeout.


mkazlauskas (approved):
Nice investigative work :detective: 
We should probably add a separate SingleAddressWallet prop for this instead of using `maxInterval` (or use genesis params to compute it), but this will do the job for now

PR #674, opened 2023-03-29
feat(ogmios): ogmios TxSubmit client now uses a long-running ws connection

AngelCastilloB:
# Context

We discovered an ogmios client bug affecting the submission of multiple transactions through a single long lasting WebSocket connection.

Basically the transaction submission client needs to implement the mirror/reflection feature as the [query client](https://github.com/CardanoSolutions/ogmios/blob/31a2d6e8ca17e1e6e7eb4bf72a041bece49ec80e/clients/TypeScript/packages/client/src/StateQuery/Query.ts#L28) does.

# Proposed Solution

 Implement the mirror/reflection feature from the StateQuery client in the TxSubmit client.

### DIFF
--- a/packages/cardano-services/test/Program/services/ogmios.test.ts
+++ b/packages/cardano-services/test/Program/services/ogmios.test.ts
@@ -328,7 +328,7 @@ describe('Service dependency abstractions', () => {
       }
     });
 
-    it('should resolve initialize without reconnection logic with one time ws connection type', async () => {
+    it('should resolve DNS twice during initialization without reconnection logic with long ws connection type', async () => {
       const srvRecord = { name: 'localhost', port: ogmiosPortDefault, priority: 1, weight: 1 };
       const failingOgmiosMockPort = await getRandomPort();
       let resolverAlreadyCalled = false;
@@ -347,10 +347,7 @@ describe('Service dependency abstractions', () => {
       });
 
       await expect(provider.initialize()).resolves.toBeUndefined();
-      // This test should fail once we switch to a long-running ws connection
-      // dnsResolverMock should be called twice and try to dns resolve
-      // while init the txSubmitClient within `provider.initialize()`
-      expect(dnsResolverMock).toBeCalledTimes(1);
+      expect(dnsResolverMock).toBeCalledTimes(2);
       await provider.start();
       await provider.shutdown();
     });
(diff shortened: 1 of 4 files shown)


packages/ogmios/package.json
```diff
@@ -52,7 +52,7 @@
     "prepack": "yarn build"
   },
   "devDependencies": {
-    "@cardano-ogmios/schema": "5.6.0",
+    "@cardano-ogmios/schema": "^5.6.0",
```
mkazlauskas, line 55: We shouldn't depend on ^ range because we're using internal functions directly
```suggestion
    "@cardano-ogmios/schema": "5.6.0",
```

packages/ogmios/src/Ogmios/TxSubmissionClient.ts
```diff
@@ -0,0 +1,123 @@
...
+        );
+
+        const response = TxSubmission.handleEvaluateTxResponse((await evaluateTxResponse.next()).value);
+
+        if (TxSubmission.isEvaluationResult(response)) {
+          return response;
+        }
+        throw response;
+      }, context);
+    },
```
mkazlauskas, line 99: Let's just remove it until we need it since we're not implementing mirror/reflection for this right now and not using this function anywhere
```suggestion
```

mkazlauskas (changes requested):

packages/cardano-services/test/Program/services/ogmios.test.ts
```diff
@@ -328,33 +328,6 @@ describe('Service dependency abstractions', () => {
...
-      });
-
-      await expect(provider.initialize()).resolves.toBeUndefined();
-      // This test should fail once we switch to a long-running ws connection
-      // dnsResolverMock should be called twice and try to dns resolve
-      // while init the txSubmitClient within `provider.initialize()`
-      expect(dnsResolverMock).toBeCalledTimes(1);
-      await provider.start();
-      await provider.shutdown();
-    });
```
IvayloAndonov, line 356: Shouldn't we update it rather than delete it?

rhyslbw (approved):
Great work @AngelCastilloB

mkazlauskas (approved):
:rocket:

IvayloAndonov (approved):
Nice work!

PR #675, opened 2023-03-29
feat: Parametrize pagination queries and add support for prepared statements in txs by addresses

IvayloAndonov:
# Context

PostgreSQL has the concept of a [prepared statement](https://www.postgresql.org/docs/9.3/static/sql-prepare.html). `node-postgres` supports this by supplying a name parameter to the query config object. If you supply a name parameter the query execution plan will be cached on the PostgreSQL server on a per-connection basis.

We need to optimize the paginated queries and take the advantage of prepared statements.

# Proposed Solution
Check the commit messages

- `withPagination()` is refactored to use parametrized queries with `$` signs
- added prepared statements support for `txs-by-addresses` queries

### DIFF
--- a/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/queries.ts
+++ b/packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/queries.ts
@@ -712,9 +712,14 @@ export const getStatusWhereClause = (
   return `(${whereClause.join(' OR ')})`;
 };
 
-export const withPagination = (query: string, pagination?: QueryStakePoolsArgs['pagination']) => {
-  if (pagination) return `${query} LIMIT ${pagination.limit} OFFSET ${pagination.startAt} `;
-  return query;
+export const withPagination = (query: string, args: unknown[], pagination?: QueryStakePoolsArgs['pagination']) => {
+  if (pagination) {
+    return {
+      args: [...args, pagination.limit, pagination.startAt],
+      query: `${query} LIMIT $${++args.length} OFFSET $${++args.length}`
+    };
+  }
+  return { args, query };
 };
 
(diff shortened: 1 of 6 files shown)


packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/queries.ts
```diff
@@ -712,9 +712,14 @@ export const getStatusWhereClause = (
...
 };
 
-export const withPagination = (query: string, pagination?: QueryStakePoolsArgs['pagination']) => {
-  if (pagination) return `${query} LIMIT ${pagination.limit} OFFSET ${pagination.startAt} `;
-  return query;
+export const withPagination = (query: string, args: unknown[], pagination?: QueryStakePoolsArgs['pagination']) => {
+  if (pagination) {
+    return {
+      args: [...args, pagination.limit, pagination.startAt],
+      query: `${query} LIMIT $${++args.length} OFFSET $${++args.length}`
```
iccicci, line 719: `++Array.length` doesn't sounds as a good idea.
Better to use `Array.length + 1` and `Array.length + 2`

iccicci (changes requested):
Very good till here, just a minor change

packages/cardano-services/src/StakePool/DbSyncStakePoolProvider/queries.ts
IvayloAndonov: 0b6d9ee92aaf683c2cfb3d45bd4ff434f354575d

iccicci (dismissed):
Awesome!

mkazlauskas (dismissed):
Nice work! Please squash, hopefully CI actions will get unstuck

mkazlauskas (approved):

PR #677, opened 2023-03-29
fix(cardano-services): unlimited_cache_ttl

iadmytro:
# Context

Fixed value for unlimited cache ttl


mkazlauskas (changes requested):
It might be actually using the default TTL instead of `undefined` now: https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services/src/InMemoryCache/InMemoryCache.ts#L28

We could either remove ` = this.#ttlDefault`, or set this to sth like `-1` and then in `InMemoryCache.get`/`InMemoryCache.set` do sth like `ttl === UNLIMITED_CACHE_TTL ? undefined : tll`

PR #679, opened 2023-03-30
refactor(cardano-services): aggregates the first two queries of txs/b…

iccicci:
# Context

At its beginning, the `txs/by-addresses` endpoint performs two similar queries to get the total number of transactions and the `tx.id`s of the requested page.

# Proposed Solution

Merged the two queries in a single one to get the entire list of `tx.id`s and performed the _count_ and the _pagination_ at TypeScript level.
This should reduce the DB server workload, but simultaneously increases the amount of data transferred from the DB server to the SDK.
We need to measure the impact of this change, moreover on the restoration of wallets with high number of transactions, before taking a decision on merging this or not.

The increased workload on SDK side shouldn't be a big problem as SDK servers can be simply scaled adding more sibling servers while scaling the DB server would require much more expensive actions: the reduction of the DB server workload (to be measured) should be another aspect to take care while taking the final decision about this branch.

# QA request

Hi @iadmytro , could you please perform load tests on wallet restoration procedure to verify this change yields better performance?

It would be useful to perform load tests on a set of wallets with high number of transactions and on a set of wallets with low number of transactions; following the expected result

- restoration of wallets with low number of transaction is expected to have benefit: please verify this
- restoration of wallets with high number of transaction unknown: the main reason of this QA request.

### DIFF
--- a/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
+++ b/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
@@ -1,8 +1,7 @@
 import * as Queries from './queries';
-import { Cardano, PaginationArgs } from '@cardano-sdk/core';
+import { Cardano } from '@cardano-sdk/core';
 import {
   CertificateModel,
-  CountModel,
   DelegationCertModel,
   MirCertModel,
   MultiAssetModel,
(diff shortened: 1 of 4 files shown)


rhyslbw (approved):
Nice work @iccicci

iadmytro (approved):

mkazlauskas (approved):

PR #682, opened 2023-03-31
feat(cardano-services): add path metrics on provider server

IvayloAndonov:
# Context
The current metrics exposed by[npm: express-prom-bundle](https://www.npmjs.com/package/express-prom-bundle) has no granular metrics per route.
We need to enable it and check if a prom client needs an additional path normalization function.

# Proposed Solution
Screenshot 2023-03-31 at 12 54 48


rhyslbw (approved):
Simple! Nice work @IvayloAndonov

PR #685, opened 2023-04-06
fix: Nft metadata fields in base64 format

IvayloAndonov:
# Context
Asset HTTP service doesn’t return a specific asset with NFT metadata when the image is in base64 encoding format.

# Proposed Solution
Extend the metadata image and files validations within CIP25 mappings

Check commit messages


packages/core/test/Asset/util/metadatumToCip25.test.ts
```diff
@@ -197,6 +191,57 @@ describe('NftMetadata/metadatumToCip25', () => {
...
+      'ggyy4qu3LsyJe5rwB/ZIIDsgQOyBI4IEvggCyBA7IEDsgSOCBL4IAsgQOyBA7IEj',
+      'ggyy7qgr3buDL3FYDfTHBAlsABWQIHZAkckCVwQJbAAVkCB2QJHJAlcECWwAFZAg',
+      'dkLd0u6rLtinq3dLG2758fev7pu2+Hnr9sTHBAlsABWQIHZAkckCVwQJbAAVkCB2',
+      'QJHJAlcECWwAFZAgdk/Xe7qHZFWWZ2XRfLBAdkCRyQJXBAlsABWQIHZAkckCVwQJ',
+      'bAAVkCB2QJHJAlcEDWL3M8PyEcpB1DAAAAAElFTkSuQmCC'
+    ];
+
+    const metadatum = createMetadatumWithFiles([], base64DecodedImage);
+    const result = metadatumToCip25(validAsset, metadatum, logger);
+    expect(result?.image).toEqual([base64DecodedImage.join('')]);
```
mkazlauskas, line 242: I suggest refactoring our type to change from `Uri[]` to `Uri`, because if I understand the spec correctly it's always exactly 1 image. An array is probably used due to metadatum string length limitations and there is no such limitation in javascript - we can provide it in a more convenient format.
```suggestion
    expect(result?.image).toEqual(base64DecodedImage.join(''));
```

packages/core/test/Asset/util/metadatumToCip25.test.ts
```diff
@@ -294,8 +415,9 @@ describe('NftMetadata/metadatumToCip25', () => {
           mediaType: 'image/png',
           name: 'file2',
           otherProperties: new Map([['extraProp', 'extra']]),
-          src: ['https://file2.location']
-        }
+          src: ['ipfs://file2.location']
+        },
+        { mediaType: 'image/jpg', name: 'file3', src: [base64FileSrc.join('')] }
```
mkazlauskas, line 420:
```suggestion
        { mediaType: 'image/jpg', name: 'file3', src: base64FileSrc.join('') }
```

packages/core/src/Asset/types/NftMetadata.ts
```diff
@@ -7,6 +7,9 @@ export const Uri = (uri: string) => {
   if (/^[a-z]+:\/\/.+/.test(uri)) {
     return uri as unknown as Uri;
   }
+  if (uri.startsWith('data:')) {
```
mkazlauskas, line 10: Need to update the `InvalidStringError` message to also allow `data:`

mkazlauskas (changes requested):

packages/core/test/Asset/util/metadatumToCip25.test.ts
mkazlauskas: I take this change request back due to it being a major breaking change. Captured in ADP-2939

packages/core/src/Asset/types/NftMetadata.ts
IvayloAndonov: 79a11d500edcedf286d2384ef9ed858047b97707

rhyslbw (approved):

PR #687, opened 2023-04-07
refactor(cardano-services): changed circulating supply query

iadmytro:
# Context

Getting live circulating supply value from db-sync is very expensive operation.

# Proposed Solution

Get circulating supply from Ada pots to improve performance of the query.

# Notes

Value from Ada pots updated once per epoch, while live circulating supply updating each block, so this change is trading off value accuracy for performance.

### DIFF
--- a/packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/queries.ts
+++ b/packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/queries.ts
@@ -1,23 +1,5 @@
 export const findCirculatingSupply = `
-    WITH total_rewards AS (
-        SELECT COALESCE(SUM(amount), 0) as rewards_amount
-        FROM reward
-    ),
-    total_withdrawals AS (
-        SELECT COALESCE(SUM(amount), 0) as withdrawals_amount
-        FROM withdrawal
-    ),
-    total_utxo AS (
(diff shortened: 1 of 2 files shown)


packages/cardano-services/src/NetworkInfo/DbSyncNetworkInfoProvider/queries.ts
```diff
@@ -1,23 +1,5 @@
...
-            SUM(tx_out.value) AS utxo_amount
-        FROM tx_out
-        LEFT JOIN tx_in 
-            ON tx_out.tx_id = tx_in.tx_out_id
-            AND tx_out.index = tx_in.tx_out_index
-        WHERE tx_in.id IS NULL    
-    )
-    SELECT CAST(utxo_amount + (rewards_amount - withdrawals_amount) AS BIGINT) as circulating_supply
-    FROM total_rewards, total_withdrawals, total_utxo
+  SELECT ( utxo + rewards ) as circulating_supply FROM ada_pots ORDER BY id DESC LIMIT 1;
```
IvayloAndonov, line 2: How the result looks comparable with the previous query version? Are they almost the same or the difference is big?

mkazlauskas (approved):
Nice work!

rhyslbw (approved):

PR #692, opened 2023-04-12
feat(ogmios): log successful transaction submission with info level

rhyslbw:
# Context
We're currently not logging successful tx submissions, so can only rely on requests for gathering metrics on system metrics.

# Proposed Solution
Log the tx hash returned by Ogmios.

The default logging level will now include a log of all submitted transactions. This may appear to be a bit noisy but can be reworked with a more sophisticated metrics solution in future. For now it's not too impactful, and does fit within the "info" category IMO.


mkazlauskas (approved):

mirceahasegan (approved):
🚀

PR #693, opened 2023-04-12
Feat/adp 2802 implement serialization of tx main body

AngelCastilloB:
# Context

We want to remove the dependency of the SDK with the CML library. For this we need to develop our own serialization logic around the Cardano domain types.

# Proposed Solution

- Implement a set of CBOR classes that follow the Reader/Writer pattern. These classes will allow us to be very precise and control how the types are serialized. This will also allow us to solve the round trip problem we currently have due difference on transactions generated by the CLI or the CSL/CML library.

- Create a Transaction class which can serialize and serialize to CBOR format. This class must also be able to convert to and from Core.Tx type

- Removed the dependency on borc library.
- CborReader and CborWriter introduced.


packages/cardano-services/src/TxSubmit/rabbitmq/TxSubmitWorker.ts
```diff
@@ -294,7 +294,7 @@ export class TxSubmitWorker extends EventEmitter {
       const txBody = new Uint8Array(content);
 
       // Register the handling of current transaction
-      txId = cmlUtil.deserializeTx(txBody).id;
+      txId = (await cmlUtil.deserializeTx(txBody)).id;
```
mkazlauskas, line 297: Just noticed that we have a more efficient way to do this:
```suggestion
      txId = TransactionId.fromTxBodyCbor(TxBodyCBOR.fromTxCBOR(txBody))
```
Maybe we can also update those utils to (at least partially) use new serialization?

mkazlauskas (changes requested):

mkazlauskas (dismissed):
Great work @AngelCastilloB! :rocket:

mirceahasegan:
This commit is not verified: [64513af](https://github.com/input-output-hk/cardano-js-sdk/pull/693/commits/64513af48c8a30e055c6fe83a45babf69a10532f)

mirceahasegan (approved):
Fantastic work, @AngelCastilloB :mage_man:

mkazlauskas (approved):
:rocket:

rhyslbw (approved):
Great work @AngelCastilloB :mechanical_arm:

PR #696, opened 2023-04-17
Feat/historical asset info

mirceahasegan:
# Context

Fetch assetInfo for all assets in the tx history. SingleAddressWallet already provides AssetInfo in assets$ observable, but only for assets that are currently owned by the wallet.

# Proposed Solution
- Rename ObservableWallet.assets$ into ObservableWallet.assetInfo$
- Update it to include all historical assets (tx history as source)
- Rename AssetInfo.quantity → AssetInfo.supply

See commits breaking changes

### DIFF
--- a/packages/cardano-services-client/src/AssetInfoProvider/assetInfoHttpProvider.ts
+++ b/packages/cardano-services-client/src/AssetInfoProvider/assetInfoHttpProvider.ts
@@ -1,5 +1,10 @@
-import { AssetProvider } from '@cardano-sdk/core';
-import { CreateHttpProviderConfig, HttpProviderConfigPaths, createHttpProvider } from '../HttpProvider';
+import { Asset, AssetProvider } from '@cardano-sdk/core';
+import {
+  CreateHttpProviderConfig,
+  HttpProviderConfig,
+  HttpProviderConfigPaths,
+  createHttpProvider
+} from '../HttpProvider';
 
 /**
  * The AssetProvider endpoint paths.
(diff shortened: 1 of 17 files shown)


mkazlauskas (approved):
Great work! :rocket: 

A possible optimization once we get historical tx inputs fully resolved (including `value`) is to filter inputs by wallet addresses + outputs by wallet addresses. This way it would only resolve AssetInfo for inputs and outputs relevant to the wallet, which (imho) are the only ones that the wallet should display anyway. Right now we are potentially resolving AssetInfo of mass NFT mint/drop transactions, where only 1 out of n (a lot) outputs is relevant to the wallet.

iccicci (approved):
LGTM

packages/cardano-services-client/src/AssetInfoProvider/assetInfoHttpProvider.ts
```diff
@@ -10,6 +15,24 @@ const paths: HttpProviderConfigPaths<AssetProvider> = {
   healthCheck: '/health'
 };
 
+const isAssetInfo = (assetInfo: Asset.AssetInfo | unknown): assetInfo is Asset.AssetInfo =>
```
mkazlauskas, line 18: That's probably equivalent
```suggestion
const isAssetInfo = (assetInfo: unknown): assetInfo is Asset.AssetInfo =>
```

mkazlauskas (dismissed):
Great work! :rocket:

iccicci (approved):

packages/cardano-services-client/src/AssetInfoProvider/assetInfoHttpProvider.ts
mirceahasegan: Done in [6e28df4](https://github.com/input-output-hk/cardano-js-sdk/pull/696/commits/6e28df412797974b8ce6f6deb0c3346ff5938a05)

iccicci (approved):

mkazlauskas (approved):

PR #699, opened 2023-04-19
refactor/ledger key agent upgrade

AngelCastilloB:
# Context

The existing implementation had quite a lot of technical debt due to the mapping back and forth through CML to be compatible, and this complexity has hidden bugs due to the large function that does not have adequate unit test coverage. It is also obscuring the fact we are lacking support for some more recent features of Cardano.

# Proposed Solution

We should extract the vendor-specific implementation to their own SDK packages. The benefits are:

- updates from one vendor do not impact the others, particularly useful for breaking change isolation.
- applications that are not implementing tree shaking get the benefit of cleaner separation
- Cleaner approach to rolling out specific vendor support in products.

- A new package with a new implementation of the ledger hardware wallet mappings was added.


rhyslbw (commented):
https://github.com/input-output-hk/cardano-js-sdk/pull/699/commits/01dfbbbe41171999cab4049857349044fe1969e8 and 429bc26cdacb82efe84dc002ba66a747461f2977 should be fixing up 7e64df8492874fed28bc40252c44b1a4ec6cbfd8

packages/hardware-ledger/src/transformers/assets.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    map.get(policyId)!.push({
+      amount: value,
+      assetNameHex: assetName
+    });
+  }
+
+  const tokenMapAssetsGroup = [];
+  for (const [key, value] of map.entries()) {
+    sortTokensCanonically(value);
```
mkazlauskas, line 30: This is fine as it's an internal implementation detail, just want to note that we prefer this style (typically don't do procedural)
```suggestion
    value.sort(compareCanonically);
```

mkazlauskas (changes requested):

mkazlauskas (dismissed):
Great work :rocket:! Please squash for final reviews

rhyslbw (changes requested):
Awesome work @AngelCastilloB. Some minor changes:

574f64817d7ae40502a3459aaa5024fbd9db417e should **not** be scoped to `core`

rhyslbw (approved):

rhyslbw (changes requested):
Have you run hardware integration tests? I'm getting a failures when running 

```console
yarn workspace @cardano-sdk/wallet test:hw
```

image

rhyslbw (changes requested):
Fixes are good @AngelCastilloB

I'm also wondering if we should move [this Ledger test](https://github.com/input-output-hk/cardano-js-sdk/blob/3df3a00309857ce871bdc539ee2b7cd514d55a39/packages/wallet/test/hardware/LedgerKeyAgent.test.ts) into the new package, as it's out of place in `wallet`. Thoughts @mkazlauskas ?

rhyslbw:
@iadmytro is going to review and also take a look at the issue of needing to run both Ledger and Trezor tests together, as that's the main issue I have now they're separated. I [edited my above comment ](https://github.com/input-output-hk/cardano-js-sdk/pull/699#pullrequestreview-1420018226) with a correction

packages/hardware-ledger/src/LedgerKeyAgent.ts
```diff
@@ -45,7 +56,106 @@ export interface CreateLedgerTransportProps {
...
+  ) {
+    return true;
+  }
+
+  return result;
+};
+
+/**
+ * Gets the mode in which we want to sign the transaction.
+ * Ledger has certain limitations due to which it cannot sign arbitrary combination of all transaction features.
```
iadmytro, line 121: How unsupported combinations handled?

rhyslbw: @AngelCastilloB can you add some tests that demonstrate the errors we expect?

rhyslbw (approved):
Nice work @AngelCastilloB. Please squash, then we can merge https://github.com/input-output-hk/cardano-js-sdk/pull/725, and @iadmytro can then run final tests with a Ledger locally

rhyslbw (approved):
Awesome work here @AngelCastilloB :rocket:

PR #700, opened 2023-04-19
feat: add CORS headers configuration to HTTP server

IvayloAndonov:
# Context
We need to add a CORS headers configuration to the HTTP server, similar to `cardano-graphql` server

- do not init `cors` middleware if `ALLOWED_ORIGINS` env OR `--allowed-origins` arg is not provided
- with initialized `cors` middleware, check what is the `origin` from the request headers, and if matches add it back into `Access-Control-Allow-Origin` response header, otherwise reject the request with appropriate status code and message.
- extend the centralized [express error handler](https://github.com/input-output-hk/cardano-js-sdk/pull/700/files#diff-db63771378e2c7af74fab64645861ab1f0556c3320ce46fabf60eedf796ab08eR117-R121) to return the correct status code in case of Provider error

# Note
Once this gets merged, the SRE team should set a proper `Origin` request header on readiness/liveness probes (polling) triggered by the orchestrator when determining the health status of the HTTP server. 
The header value should match some of the `ALLOWED_ORIGINS` values provided as a runtime config to the backend deployment. @rhyslbw


rhyslbw (dismissed):
Great work @IvayloAndonov! :tada:

packages/cardano-services/test/Http/HttpServer.test.ts
```diff
@@ -531,4 +532,78 @@ describe('HttpServer', () => {
...
+
+      const res = await axios.get(`${apiUrlBase}/health`, {
+        headers: { [CONTENT_TYPE]: 'application/json' }
+      });
+
+      expect(res.status).toBe(200);
+      expect(res.data).toBeDefined();
+    });
+
+    it('allows requests from whitelisted origins', async () => {
```
mkazlauskas, line 563:
```suggestion
    it('mirrors "origin" header in "access-control-allow-origin" if it matches a whitelisted origin', async () => {
```

packages/cardano-services/test/Http/HttpServer.test.ts
```diff
@@ -531,4 +532,78 @@ describe('HttpServer', () => {
...
+      await onHttpServer(apiUrlBase);
+
+      const res = await axios.get(`${apiUrlBase}/${ServiceNames.Utxo}/health`, {
+        headers: { [CONTENT_TYPE]: 'application/json', [ORIGIN]: allowedOrigin }
+      });
+
+      expect(res.headers['access-control-allow-origin']).toEqual(allowedOrigin);
+    });
+
+    it('rejects requests from non-whitelisted origins', async () => {
```
mkazlauskas, line 586: It's still allowing the requests. We could make it respond with 4xx to reject the request. WDYT @rhyslbw ?
```suggestion
    it('does not set "access-control-allow-origin" header for requests from non-whitelisted origins', async () => {
```

packages/cardano-services/test/cli.test.ts
```diff
@@ -65,6 +65,18 @@ const assertServiceHealthy = async (
...
 
+const assertCORSHeaders = async (apiUrl: string, origin: string) => {
+  expect.assertions(1);
+  await serverReady(apiUrl);
+  const headers = { 'Content-Type': 'application/json', Origin: origin };
+  try {
+    const res = await axios.get(`${apiUrl}/health`, { headers });
+    expect(res.headers['access-control-allow-origin']).toEqual(origin);
+  } catch (error) {
+    expect((error as AxiosError).response?.status).toBe(404);
```
mkazlauskas, line 76: We may get false positives with current implementation, as caller doesn't know which assertion was made
```suggestion
const assertCORSHeaders = async (apiUrl: string, origin: string, expectedFailureStatus?: number) => {
  expect.assertions(1);
  await serverReady(apiUrl);
  const headers = { 'Content-Type': 'application/json', Origin: origin };
  try {
    const res = await axios.get(`${apiUrl}/health`, { headers });
    expect(res.headers['access-control-allow-origin']).toEqual(origin);
  } catch (error) {
    expect((error as AxiosError).response!.status).toBe(expectedFailureStatus);
```

mkazlauskas (changes requested):

packages/cardano-services/test/Http/HttpServer.test.ts
IvayloAndonov: I think we should not reject requests/throw an error in this case. The clients will be restricted from accessing shared resources if the `access-control-allow-origin` response header is not set (mirrored by the requested origin) which is enough @mkazlauskas

mkazlauskas: > The clients will be restricted from accessing shared resources if the

I'm not sure it's a bad thing

packages/cardano-services/test/Http/HttpServer.test.ts
IvayloAndonov: 41789c949df6b3733e0f9a61ad34655614bd115e

packages/cardano-services/test/Http/HttpServer.test.ts
IvayloAndonov: > I'm not sure it's a bad thing

Didn't get that one

packages/cardano-services/test/Http/HttpServer.test.ts
```diff
@@ -531,4 +532,78 @@ describe('HttpServer', () => {
...
+
+      const res = await axios.get(`${apiUrlBase}/health`, {
+        headers: { [CONTENT_TYPE]: 'application/json' }
+      });
+
+      expect(res.status).toBe(200);
+      expect(res.data).toBeDefined();
+    });
+
+    it('mirrors "origin" header in "access-control-allow-origin" if it matches a whitelisted origin', async () => {
```
rhyslbw, line 563:
```suggestion
    it('mirrors "origin" header in "access-control-allow-origin" if it matches an allowed origin', async () => {
```

packages/cardano-services/test/Http/HttpServer.test.ts
```diff
@@ -531,4 +532,78 @@ describe('HttpServer', () => {
...
+      await onHttpServer(apiUrlBase);
+
+      const res = await axios.get(`${apiUrlBase}/${ServiceNames.Utxo}/health`, {
+        headers: { [CONTENT_TYPE]: 'application/json', [ORIGIN]: allowedOrigin }
+      });
+
+      expect(res.headers['access-control-allow-origin']).toEqual(allowedOrigin);
+    });
+
+    it('does not set "access-control-allow-origin" header for requests from non-whitelisted origins', async () => {
```
rhyslbw, line 586:
```suggestion
    it('does not set "access-control-allow-origin" header for requests from disallowed origins', async () => {
```

rhyslbw (changes requested):

packages/cardano-services/test/Http/HttpServer.test.ts
IvayloAndonov: Implemented in d51f27c55b83c23829e4d00b33e5de92a9c660b2 @mkazlauskas

packages/cardano-services/test/Http/HttpServer.test.ts
IvayloAndonov: d51f27c55b83c23829e4d00b33e5de92a9c660b2

mkazlauskas (approved):
Nice work! Please update the commit message to unscoped when squashing as it also includes a change in `core` package.

rhyslbw (dismissed):
Ready for squashing

mkazlauskas (approved):
:rocket:

gytis-ivaskevicius (approved):
Good stuff! :))

packages/cardano-services/src/Http/util.ts
```diff
@@ -0,0 +1,19 @@
...
+export const ORIGIN = 'Origin';
+export const CONTENT_TYPE = 'Content-Type';
+export const APPLICATION_JSON = 'application/json';
+
+export const getListen = (url: URL) => ({ host: url.hostname, port: url ? Number.parseInt(url.port) : undefined });
+
+type StaticOrigin = boolean | string | RegExp | (boolean | string | RegExp)[];
+
+export const corsOptions = (allowedOrigins: string[]) => ({
+  origin(requestOrigin: string | undefined, callback: (err: Error | null, options?: StaticOrigin) => void) {
```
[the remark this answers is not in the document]
IvayloAndonov: there is no defined type there: `...callback: (err: Error | null, origin?: StaticOrigin) => void)` in `@types/cors/index.d.ts`

packages/cardano-services/src/Http/util.ts
```diff
@@ -0,0 +1,19 @@
...
+    if (requestOrigin && allowedOrigins.includes(requestOrigin)) {
```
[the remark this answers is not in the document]
IvayloAndonov: If cors middleware is enabled (`ALLOWED_ORIGINS` is provided as runtime configuration ) then the provider server will expect all incoming requests to have an origin header, otherwise, they will be rejected as described in the PR notes

packages/cardano-services/src/Http/util.ts
```diff
@@ -0,0 +1,19 @@
+import { ProviderError, ProviderFailure } from '@cardano-sdk/core';
+
+export const ORIGIN = 'Origin';
+export const CONTENT_TYPE = 'Content-Type';
+export const APPLICATION_JSON = 'application/json';
+
+export const getListen = (url: URL) => ({ host: url.hostname, port: url ? Number.parseInt(url.port) : undefined });
```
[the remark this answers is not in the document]
IvayloAndonov: I haven't changed this function, just moved it into another `util` file.

packages/cardano-services/src/Http/util.ts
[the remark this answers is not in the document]
gytis-ivaskevicius: Sorry,  I misread it

packages/cardano-services/src/Http/util.ts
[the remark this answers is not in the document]
rhyslbw: From memory, it's not strictly required since we're just shadowing the express default, but yes it makes sense in case the underlying default changes.

packages/cardano-services/src/Http/util.ts
```diff
@@ -0,0 +1,19 @@
...
+
+type StaticOrigin = boolean | string | RegExp | (boolean | string | RegExp)[];
+
+export const corsOptions = (allowedOrigins: string[]) => ({
```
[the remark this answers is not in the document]
IvayloAndonov: Addressed with 19a3e760253516c6aea624587c7693189da1515a

gytis-ivaskevicius (dismissed):

rhyslbw (dismissed):
Can squash again

PR #702, opened 2023-04-20
feat!: rename confirmed$ to onChain$

mirceahasegan:
# Context

The transactions tracker confirmed$ observable name is a bit misleading. The transactions emitted by this observable are transactions that have been submitted with the SingleAddressWallet that were confirmed to be on the chain.

But these transactions are not confirmed as they could be rolled back.

A better name would be onChain$.

# Proposed Solution
The transactions emitted by this observable are transactions that have been submitted with the SingleAddressWallet that were confirmed to be on the chain.
But these transactions are not confirmed as they could be rolled back, so a more suitable name is onChain$ instead of confirmed$

BREAKING CHANGE:
- renamed `TransactionsTracker.outgoing.confirmed$` to `onChain$`
- renamed `TransactionReemitterProps.transactions.outgoing.confirmed$` to `onChain$`
- renamed web-extension `observableWalletProperties.transactions.outgoing.confirmed$` to `onChain$`
- rename ConfirmedTx to OutgoingOnChainTx
- renamed OutgoingOnChainTx.confirmedAt to `slot`

### DIFF
--- a/packages/e2e/test/wallet/SingleAddressWallet/delegation.test.ts
+++ b/packages/e2e/test/wallet/SingleAddressWallet/delegation.test.ts
@@ -33,10 +33,10 @@ const waitForTx = async (wallet: ObservableWallet, hash: Cardano.TransactionId) 
   await firstValueFromTimed(
     combineLatest([
       wallet.transactions.history$.pipe(filter((txs) => txs.some(({ id }) => id === hash))),
-      // test that confirmed$ works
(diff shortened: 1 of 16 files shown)


rhyslbw (changes requested):
This is a `refactor` commit type

rhyslbw (approved):
Nice work @mirceahasegan

mkazlauskas (approved):
Nice work! :rocket:

PR #703, opened 2023-04-24
More granularity in metrics buckets

iccicci:
# Context

The default buckets configuration of [express prometheus bundle](https://www.npmjs.com/package/express-prom-bundle) has not a good granularity.

# Proposed Solution

Provided to the module the option to give buckets the granularity we need.

### DIFF
--- a/packages/cardano-services/src/Http/HttpServer.ts
+++ b/packages/cardano-services/src/Http/HttpServer.ts
@@ -213,7 +213,7 @@ export class HttpServer extends RunnableModule {
     });
     this.app.use(
       expressPromBundle({
-        buckets: [0.003, 0.03, 0.1, 0.3, 1.5, 3, 5, 8, 12, 20, 30],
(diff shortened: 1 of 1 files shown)


packages/cardano-services/src/Http/HttpServer.ts
```diff
@@ -213,6 +213,7 @@ export class HttpServer extends RunnableModule {
     });
     this.app.use(
       expressPromBundle({
+        buckets: [0.003, 0.03, 0.1, 0.3, 1.5, 3, 5, 8, 12, 20, 30],
```
rhyslbw, line 216:
```suggestion
        buckets: [0.003, 0.03, 0.1, 0.3, 0.75, 1.5, 3, 5, 8, 12, 20, 30],
```
@blaggacao looking at this now, maybe we should this extra step

blaggacao: rather safe than sorry! :-)

I'm sure there's a cost, but confident it's negligible.

PR #704, opened 2023-04-24
fix(web-extension): proxy error responses

mirceahasegan:
# Context
errors from content script are not propagated 

# Proposed Solution
ContentScriptMessageProxy will catch and return errors.

### DIFF
--- a/packages/web-extension/src/messaging/runContentScriptMessageProxy.ts
+++ b/packages/web-extension/src/messaging/runContentScriptMessageProxy.ts
@@ -3,6 +3,7 @@
 import { Logger } from 'ts-log';
 import { MethodRequest, ResponseMessage } from './types';
 import { isRequestMessage } from './util';
+import { toSerializableObject } from '@cardano-sdk/util';
 
 export type AnyApi = any;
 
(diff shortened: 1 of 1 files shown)


packages/web-extension/src/messaging/types.ts
```diff
@@ -15,7 +15,8 @@ export interface RequestMessage extends AnyMessage {
 }
 
 export interface ResponseMessage<T = unknown> extends AnyMessage {
-  response: T;
+  response: T | undefined;
```
rhyslbw, line 18:
```suggestion
  response?: T;
```

rhyslbw (changes requested):

packages/web-extension/src/messaging/types.ts
mirceahasegan: this code was replaced

packages/web-extension/src/messaging/runContentScriptMessageProxy.ts
```diff
@@ -23,9 +24,16 @@ export const runContentScriptMessageProxy = (apis: AnyApi[], logger: Logger) =>
...
+    try {
+      response = await apiFunction(...data.request.args);
+    } catch (error) {
+      response = toSerializableObject(error);
+    }
+
     const responseMessage: ResponseMessage = {
       messageId: data.messageId,
-      response: await apiFunction(...data.request.args)
+      response
```
mkazlauskas, line 36: cip30 responses might not have any objects that need this right now, but it might save us from a sneaky bug in the future
```suggestion
      response: toSerializableObject(response)
```

mkazlauskas (changes requested):

packages/web-extension/src/messaging/runContentScriptMessageProxy.ts
mirceahasegan: fixed here: [34fa24d](https://github.com/input-output-hk/cardano-js-sdk/pull/704/commits/34fa24d0a3cdae21f89b41f4b85d9de34f2e1220)

rhyslbw (approved):
Nice work @mirceahasegan

mkazlauskas (dismissed):
Nice work!

rhyslbw (approved):

PR #707, opened 2023-04-25
feat: store and restore last activated wallet props

mkazlauskas:
# Context

WalletManagerWorker expects a call to activate(props) to initialize,
which is made from the UI script. When service worker starts in the
background (e.g. via cip30 content script), it will wait for UI to
open before activating the wallet.

# Proposed Solution

This feature resolves the issue by storing and restoring the `props`
object last passed to activate() and adding a new `initialize()`
method that application can call to activate last activated wallet.

See commit message

**BREAKING CHANGE:**

WalletManagerWorker now requires an extra dependency: managerStorage


rhyslbw (approved):
Nice work @mkazlauskas :raised_hands:

mrcnk (approved):

PR #711, opened 2023-04-27
fix: tx metadata memory leak

IvayloAndonov:
# Context
We observe many `RangeError: offset is out of bounds` errors related to Tx Metadata mappings.

# Proposed Solution
Add missed Managed freeable scope in a few places.


mkazlauskas (approved):
Nice catch! :detective:

rhyslbw (approved):
Great work @IvayloAndonov :european_castle:

PR #713, opened 2023-04-27
build(core): rm util-dev circular devDependency

mkazlauskas:
# Context

There is a circular dependency that `lerna` complains about

```
lerna WARN ECYCLE Dependency cycles detected, you should fix these!
lerna WARN ECYCLE @cardano-sdk/core -> @cardano-sdk/util-dev -> @cardano-sdk/core
```

# Proposed Solution

Remove `util-dev` dependency from `core` pacakge.json

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -51,7 +51,6 @@
     "prepack": "yarn build"
   },
   "devDependencies": {
-    "@cardano-sdk/util-dev": "^0.8.0",
     "@types/lodash": "^4.14.182",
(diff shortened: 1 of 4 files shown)


mirceahasegan:
Doing this will disable logging for the test. e.g. changing TL_LEVEL='debug' will have no effect and the test will not log anything

mirceahasegan (approved):
:rocket:

rhyslbw (approved):

PR #714, opened 2023-04-28
feat!: transaction body core type now includes all fields

AngelCastilloB:
# Context

Currently our Transaction body type is missing two fields, networkId (optional) and auxDataHash(optional). This is problematic when we deserialize TxBodies that come from a different sources.

Our TxBody should be compliant with the CDDL specification.

# Proposed Solution

- Add the two missing fields (networkId and auxDataHash).
- Auxiliary data was simplified, the hash was moved to the txBody and a new utility function to compute the auxiliary data hashes was introduced.


mkazlauskas (dismissed):
Nice work! There's a mistake in commit message
> - networkInfo is now included in the TxBody core type.

Should say "network**Id** is now included in the TxBody core type."

mkazlauskas (approved):

rhyslbw (approved):
🎉

PR #715, opened 2023-04-28
fix: cip30 getUtxos compliance

mkazlauskas:
# Context

1. "If amount is not undefined, this request shall be limited to just the UTXOs that are required to reach the combined ADA/multiasset value target specified in amount, and if this cannot be attained, null shall be returned". Currently null is never returned.
2. `getUtxos` 'paginate' argument doesn't work when used with 'amount' argument

# Proposed Solution

See commit log


rhyslbw (approved):
Nice work @mkazlauskas

PR #719, opened 2023-05-02
chore(key-management): rm extraneous util-dev dependency

mkazlauskas:
# Context

`key-management` has a dependency on `util-dev` but it's not being used

# Proposed Solution

- Remove it
- Add "browser" field to util-dev esm package.json to omit docker utils from browser builds


rhyslbw (approved):

mirceahasegan (dismissed):

rhyslbw (approved):

mirceahasegan (approved):
Nice catch! :+1:

PR #720, opened 2023-05-02
ci: rework releases

mkazlauskas:
# Context

- We need to update our release process to support pre-releases
- `nightly` releases are sometimes used as a deployment target, which isn't ideal as they aren't meant for that

# Proposed Solution

- Update package.json files to specify workspace inter-dependencies with `workspace:~` syntax
- As a result, releases should cause less (if any) merge conflicts, so we can make all releases from `master`:
- Add a new repository variable `RELEASE_MODE`, which can be used to control the state of SDK releases:
  - `stable` will bump package versions based on semantic commits
  - `rc` will bump package versions based on semantic commits **once** (assuming *major* bump) - subsequent `rc` releases will only bump the rc # (e.g. `package@1.0.0-rc.0` -> `package@1.0.0-rc.1`). This mode is intended to be used for:
    - Testing before major releases
    - Accumulating breaking changes to bundle in a major release
- Remove `nightly` releases and keep a single track. We can re-add nightly releases at any time if we need to do continuous integration.

- Replace `lerna` with `lerna-lite` - it has less dependencies, and also some useful features, such as better workspace support and `--dry-run`
- feat(util-dev)!: remove createStubLogger util (was causing circular deps: core->util->util-dev->core)
- chore: align package.json 'engines' field


package.json
```diff
@@ -30,6 +30,7 @@
     "prepare": "husky install",
     "pre-commit": "lint-staged",
     "version": "yarn install --mode=skip-build && git add yarn.lock yarn-project.nix",
+    "patch-sibling-deps-versions": "node scripts/patch-versions/index.js",
```
IvayloAndonov, line 33: This one has left from the original commit and should be removed since the script is not required

mkazlauskas: Nice catch, squashed to e83ac7f5bcf90aa3707076a87fc3516d295f6edd

mirceahasegan (approved):
Great work, @mkazlauskas! :1st_place_medal:

IvayloAndonov (dismissed):
Nice job!

mirceahasegan (approved):
Great work! :rocket:

PR #722, opened 2023-05-03
fix(cardano-services): token metadata core mapping

IvayloAndonov:
# Context
AssetProvider returns the wrong types for token metadata fetched from Cardano Token Registry.

# Proposed Solution
Fix the core type mappings

### DIFF
--- a/packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
+++ b/packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
@@ -19,7 +19,7 @@ describe('CardanoTokenRegistry', () => {
     it('complete attributes', () =>
       expect(
         toCoreTokenMetadata({
-          decimals: { value: 23 },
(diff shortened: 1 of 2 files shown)


packages/cardano-services/test/Asset/CardanoTokenRegistry.test.ts
```diff
@@ -19,7 +19,7 @@ describe('CardanoTokenRegistry', () => {
     it('complete attributes', () =>
       expect(
         toCoreTokenMetadata({
-          decimals: { value: 23 },
+          decimals: { value: 0 },
```
IvayloAndonov, line 22: The bug case was with a falsy decimals value so the test is updated accordingly to catch it

mkazlauskas (approved):
Nice work!

rhyslbw (approved):
Thanks @IvayloAndonov

PR #723, opened 2023-05-10
feat: remote method returning api

mkazlauskas:
# Context

Remote API can only return POJOs, so it can't be used for fluent interface-style APIs (e.g. tx builder)

# Proposed Solution

eacad41b5b8570ca4e5d0cc4178c4b5740569a4c

- cc0b3b0e10ef4eaf763bea085976669f8be72ed5
- 06939f65daf580b154aa98cc2162d5e1f399c2d7

### DIFF
--- a/packages/web-extension/src/messaging/remoteApi.ts
+++ b/packages/web-extension/src/messaging/remoteApi.ts
@@ -1,15 +1,19 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import {
   BindRequestHandlerOptions,
+  CompletionMessage,
+  ConsumeMessengerApiDependencies,
   ConsumeRemoteApiOptions,
   EmitMessage,
   ExposableRemoteApi,
   ExposeApiProps,
+  FactoryCallMessage,
   MessengerApiDependencies,
   MethodRequest,
   MethodRequestOptions,
-  ObservableCompletionMessage,
+  RemoteApiFactory,
   RemoteApiMethod,
+  RemoteApiProperties,
   RemoteApiPropertyType,
   RequestMessage,
   ResponseMessage
@@ -20,9 +24,12 @@ import {
   EmptyError,
   NEVER,
   Observable,
+  Subscription,
+  TeardownLogic,
   concat,
   filter,
   firstValueFrom,
+  from,
   isObservable,
   map,
   merge,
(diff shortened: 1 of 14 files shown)


packages/web-extension/src/messaging/remoteApi.ts
```diff
@@ -92,12 +106,66 @@ const consumeMethod =
...
+          subscribe: false
+        } as CompletionMessage)
+        .subscribe(() => {
+          postSubscription.unsubscribe();
+          apiMessenger.shutdown();
+        });
+    });
+
+    // Since it returns synchronously, we can't catch potential errors.
+    // If factory method throws on the remote side, it might be a good idea to "brick" the returned api object:
```
rhyslbw, line 158: What’s the alternative? Impact of not implementing? Sounds fine to do this though

packages/web-extension/test/remoteApi/remoteApi.integration.test.ts
```diff
@@ -474,4 +512,73 @@ describe('remoteApi integration', () => {
...
+              setTimeout(() => {
+                expect(areFactoryMessengersShutdown()).toBe(true);
+                // eslint-disable-next-line promise/no-callback-in-promise
+                done();
+              }, 1);
+            }, 1);
+          });
+        });
+
+        // Don't know how to test this
```
rhyslbw, line 576: Can it be covered with an integration test?

packages/web-extension/src/messaging/remoteApi.ts
mkazlauskas: This is the impact: https://github.com/input-output-hk/cardano-js-sdk/pull/723/files#diff-5250162cad8f88072beedc14e94ffdf4325136c0acb62e48dfaee349484b34b5R103

We can implement the bricking later if we find we need it

packages/web-extension/test/remoteApi/remoteApi.integration.test.ts
mkazlauskas: Could potentially be implemented as e2e test as we'd want it to run in the intended environment (web extension). I verified that it works manually as writing a test would take a lot of effort.

rhyslbw (approved):
Great work @mkazlauskas 🚀

mirceahasegan (approved):
Great work, @mkazlauskas ! :rocket:

PR #726, opened 2023-05-12
Feat/typeorm stakepool provider

IvayloAndonov:
# Context

# Proposed Solution


packages/cardano-services/src/StakePool/TypeormStakePoolProvider/TypeormStakePoolProvider.ts
```diff
@@ -0,0 +1,100 @@
...
+      dataSource
+        .createQueryBuilder()
+        .addSelect('count(*) as count, pool.status')
+        .from(StakePoolEntity, 'pool')
+        .groupBy('pool.status')
+        .getRawMany()
+    );
+
+    const stats = mapPoolStats(rawResult);
+    return { qty: stats };
```
IvayloAndonov, line 98: Love how much the querying is simplified with the new projection db schema even for complex stake pool interfaces. 💛
For comparison, this new provider has `100` lines of code, versus `403` hard-to-follow lines of code in the db-sync one (without the actual db queries)

packages/cardano-services/src/StakePool/TypeormStakePoolProvider/TypeormStakePoolProvider.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
```
IvayloAndonov, line 99: Love how much the querying is simplified with the new projection db schema and ORM even for complex interfaces such as stake pool search. 💛
For comparison, this new provider has `100` lines of code in total, versus `403` hard-to-follow lines of code for the DbSync one (without the actual raw db string queries which are `880` more lines of code)

packages/core/src/Cardano/util/estimateStakePoolAPY.ts
```diff
@@ -1,26 +0,0 @@
...
-const MILLISECONDS_PER_DAY = 1000 * 60 * 60 * 24;
-
-/**
- * Estimates annualized percentage yield given past stake pool rewards.
- * Assumes 365 day year.
- *
- * @param rewardsHistory The list of `StakePoolEpochRewards` to estimate the APY
- * @returns `null` if provided an empty `rewardsHistory` or the estimated APY otherwise
- */
-export const estimateStakePoolAPY = (rewardsHistory: StakePoolEpochRewards[]): Percent | null => {
```
mkazlauskas, line 12: This util might still be useful for someone, maybe let's keep it?

packages/cardano-services/src/StakePool/TypeormStakePoolProvider/mappers.ts
```diff
@@ -0,0 +1,116 @@
...
+import { Hash32ByteBase16 } from '@cardano-sdk/crypto';
+import { RelayModel, mapRelay } from '../DbSyncStakePoolProvider';
+
+export type Margin = {
+  numerator: number;
+  denominator: number;
+};
+
+export type PoolModel = {
+  pool_id: string;
```
mkazlauskas, line 11: This looks like something that should be handled by typeorm. Is there no way to tell typeorm to map to responses to entity classes when using query builder?

mkazlauskas (changes requested):
Looking good! We need to find a way to test the new provider, probably comparing the responses with DbSyncStakePoolProvider. Maybe set up projection database snapshot generation in e2e (together with db-sync snapshot). There might be a challenge synchronizing both projections (or stopping at the same block) though :thinking: 

Also, 686e6b777283d124b9b072ce4692905a754c8ac0 should be a `refactor` type since it doesn't affect functionality.

packages/cardano-services/src/Program/programs/providerServer.ts
```diff
@@ -86,6 +86,41 @@ const serviceMapFactory = (options: ServiceMapFactoryOptions) => {
...
+        epochMonitor: getEpochMonitor(dbPools.main),
+        genesisData,
+        logger,
+        metadataService: createHttpStakePoolMetadataService(logger)
+      }
+    );
+  }, ServiceNames.StakePool);
+
+  const getTypeormStakePoolProvider = () => {
+    const entities = [BlockEntity, PoolMetadataEntity, PoolRegistrationEntity, PoolRetirementEntity, StakePoolEntity];
```
iccicci, line 116: I wrote a similar line at least twice in some _pg-boss_ related previous PR: I think it's time to refactor this otherwise in the future when we need to change this array we need to apply the same change on some lines spread in the SDK source code.

I suggest to make this `Array` a common value: either hard coded or somehow extracted from [these configuration data](https://github.com/input-output-hk/cardano-js-sdk/blob/ad0a54c09b236034f2824138ffc83a3f6e81cbc8/packages/cardano-services/src/Projection/prepareTypeormProjection.ts#L26-L113).

packages/cardano-services/src/StakePool/TypeormStakePoolProvider/mappers.ts
```diff
@@ -0,0 +1,116 @@
...
+      homepage: metadata_homepage,
+      name: metadata_name,
+      ticker: metadata_ticker
+    };
+  }
+};
+
+// TODO: remove the static metrics once metrics projection is implemented
+const poolMetricsStaticData: Cardano.StakePoolMetrics = {
+  apy: Cardano.Percent(0.015),
```
iccicci, line 74: Wouldn't be better to set it as `null` (or `undefined`) since it is actually not currently supported?

iccicci (changes requested):
Looks awesome so far!

This also includes a lot of lovely refactoring I enjoyed reading!

Please let me ask for one more refactoring action about a code repetition (actually included by me, not by this PR)

packages/cardano-services/src/Program/programs/providerServer.ts
IvayloAndonov: Sounds good to move these "common entities" into a shared package and reuse them whenever it's possible + extend them if needed.

mkazlauskas: They are only common in context of stake pool services. I think the solution here is to resolve using [prepareTypeormProjection](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services/src/Projection/prepareTypeormProjection.ts) (or add a new util in that file, resolving just the entitites)

packages/cardano-services/src/StakePool/TypeormStakePoolProvider/mappers.ts
IvayloAndonov: If we were fetching the whole entities (`getMany` or `getOne()`) directly with all data in it the response type is matching entity class automatically. However, in our case, we use `getRawMany()` because we want to select some specified fields + some extra SQL computation such as `COUNT(*) OVER()` to get the total count with pagination.
When using raw data I couldn't find a way to say to typeorm to transform raw data fields into entities.
Another disadvantage when working with the query builder and entities is that if fetch the whole entity and over-fetch data we do not need, also at the end we should map again entity type to the Cardano core type as well.

packages/cardano-services/src/StakePool/TypeormStakePoolProvider/mappers.ts
IvayloAndonov: d7466f611a14c1f889af2aa5992409239122e1ec

packages/cardano-services/src/StakePool/TypeormStakePoolProvider/mappers.ts
mkazlauskas: Probably worth spending some more time on this to see if we can use typeorm internals to do the mapping for us.
> Another disadvantage when working with the query builder and entities is that if fetch the whole entity

Is it not possible to `SELECT field1, field2`?

IvayloAndonov: It's just one way or another based on the needs: https://orkhan.gitbook.io/typeorm/docs/select-query-builder#getting-raw-results
Although there is one opened [issue](https://github.com/typeorm/typeorm/issues/6803) "Mapping raw query to entities" which raises a similar case, it's exposed one solution with the using of [RawSqlResultsToEntityTransformer](https://github.com/typeorm/typeorm/blob/master/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts#L18) but it seems it drastically complicates the "mapping part" compared to the explicit and clear mapper method: "from db model to Cardano core type".

IvayloAndonov:
Rebased with master and squashed fixups for clean PR final review state. All conflicts are resolved.

iccicci (dismissed):

packages/core/src/Cardano/util/estimateStakePoolAPY.ts
mkazlauskas: We can keep the `StakePoolEpochRewards` type, just remove `StakePool.epochRewards`. What do you think @rhyslbw ?

rhyslbw: Yes, just dropping from the provider API seems like the right move here

iccicci: 48d72f92f should solve this

mkazlauskas: @iccicci please bring back estimateStakePoolApy with tests too

iccicci: 586eb3 should solve this last comment as well

packages/cardano-services/src/cli.ts
```diff
@@ -169,11 +170,14 @@ withCommonOptions(
...
-          .description('Start the Provider Server')
-          .argument('[serviceNames...]', `List of services to attach: ${Object.values(ServiceNames).toString()}`)
+      withPostgresOptions(
+        withRabbitMqOptions(
+          program
+            .command('start-provider-server')
+            .description('Start the Provider Server')
+            .argument('[serviceNames...]', `List of services to attach: ${Object.values(ServiceNames).toString()}`)
+        ),
+        'StakePool'
```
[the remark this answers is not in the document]
iccicci: In the end, no, because we are going to have more than one database: one for each projection groups

rhyslbw (approved):
Nice job @IvayloAndonov and @iccicci !

packages/cardano-services/test/StakePool/TypeormStakePoolProvider/TypeormStakePoolProvider.test.ts
```diff
@@ -0,0 +1,750 @@
...
+                values: [{ id: poolsInfo[0].id }, { id: poolsInfo[1].id }]
+              }
+            },
+            pagination
+          };
+          const response = await provider.queryStakePools(options);
+
+          expect(() => Cardano.PoolId(poolsInfo[0].id as unknown as string)).not.toThrow();
+          expect(() => Cardano.PoolIdHex(response.pageResults[0].hexId as unknown as string)).not.toThrow();
+          expect(() => Cardano.VrfVkHex(response.pageResults[0].vrf as unknown as string)).not.toThrow();
```
mkazlauskas, line 244: Would be great to add more assertions like that to ensure we're returning an object that matches StakePool type (including metrics & metadata)

mkazlauskas (changes requested):

mkazlauskas (approved):
Great work! :rocket:

PR #727, opened 2023-05-15
More db connections

iccicci:
# Context

In order to compute stake pool metrics on _db-sync_ data, the _pg-boss_ worker needs to access simultaneously two distinct databases.
Currently `cli.ts` is able to accept only one database connection set of parameters: replicating this set of command line parameters and all the `function`s accessing them may be source of a huge code repetition.

# Proposed Solution

- Made the `PosgresProgramOptions` a _generic type_.
- Changed the `withPostgresOptions` to parse command line parameters and environment variables depending on a specific connection config.


packages/cardano-services/src/Program/options/postgres.ts
```diff
@@ -32,66 +32,122 @@ export interface PosgresProgramOptions {
   postgresSslCaFile?: string;
 }
 
-export const withPostgresOptions = (command: Command) =>
-  command
+export type ConnectionNames = '' | 'Projection';
```
IvayloAndonov, line 35: How about `ConnectionNames = 'DbSync' | 'Projection'` rather than`'' | 'Projection'` ?

iccicci: The main reason is to make the lesser impact as possible.

Your proposal would require to change all the occurrences of `POSTGRES_DB` (and all other environment variables and command line option) in `POSTGRES_DB_DBSYNC`.

IvayloAndonov: Ah right, well that's an option to accept no suffix is equal to db-sync. Fair enough based on the impact otherwise.

iccicci: > well that's an option to accept "no suffix" is equal to "db-sync"

Exactly

packages/cardano-services/src/Program/services/postgres.ts
```diff
@@ -82,11 +82,13 @@ const mergeTlsOptions = (
 
 export const getConnectionConfig = (
   dnsResolver: DnsResolver,
-  options?: PosgresProgramOptions
+  options?: PosgresProgramOptions<'Projection'>
```
IvayloAndonov, line 85: If here we declare 'Projection' scope always then we should rename the function as well that works only for projection connection config, otherwise could be a generic one working with all types of config

packages/cardano-services/src/Program/options/postgres.ts
rhyslbw: Both reasonable points. I'll throw another one in the mix :laughing: 

`ConnectionNames = '' | 'DbSync';`

Why? Using the name "Projection" is a bit jarring to me, and it's not that specific since db-sync is also a projection. I think we can assign the non-specific variant to the database defined within the SDK. Yes it's a refactor now, but it will be less impactful when we fully replace db-sync, since we can just drop the config to be left with what we have now.

packages/cardano-services/src/Program/services/postgres.ts
IvayloAndonov: Is there a case where `getConnectionConfig` could be reused to provide config for all databases based on the input? If yes - then generic util would be more useful

packages/cardano-services/src/Program/options/postgres.ts
rhyslbw: Yes, breaking change is inevitable if we think that a future version will only need a single pg connection, so best to do it now while we're sub-1.0 :sunglasses: @mkazlauskas thoughts?

packages/cardano-services/src/Program/services/postgres.ts
iccicci: The only function which currently needs to be generic (i.e. used with both the suffixes) is `connectionStringFromArgs`.

I tried to avoid overengineering: when in the future some other function will need to be refactored being generic we can do that.

Another point is I thought that applying changes not strictly required (as making generic a function which doesn't need to be changed) is not a good time investment.

packages/cardano-services/src/Program/options/postgres.ts
iccicci: This last consideration was enough to convince me: I'll apply this change

mkazlauskas: I agree it's best to do it now. I would also rename the type to `ConnectionName`

mkazlauskas (changes requested):
Also need to update `cardano-services/README.md`

packages/cardano-services/src/Program/options/postgres.ts
iccicci: fae7aaff and subsequent fixes should solve this

.github/workflows/continuous-integration-e2e.yaml
```diff
@@ -65,8 +65,8 @@ jobs:
         CHAIN_HISTORY_PROVIDER: 'http'
         CHAIN_HISTORY_PROVIDER_PARAMS: '{"baseUrl":"http://localhost:4000/chain-history"}'
         DB_SYNC_CONNECTION_STRING: 'postgresql://postgres:doNoUseThisSecret!@localhost:5435/cexplorer'
-        PROJECTION_PG_CONNECTION_STRING: 'postgresql://postgres:doNoUseThisSecret!@localhost:5435/projection'
-        PROJECTION_TEST_PG_CONNECTION_STRING: 'postgresql://postgres:doNoUseThisSecret!@localhost:5435/projection_test'
+        STAKE_POOLS_CONNECTION_STRING: 'postgresql://postgres:doNoUseThisSecret!@localhost:5435/pools'
```
rhyslbw, line 68:
```suggestion
        STAKE_POOL_CONNECTION_STRING: 'postgresql://postgres:doNoUseThisSecret!@localhost:5435/stake_pool'
```
The convention here should be the full name of the provider (_stake pool_)

Dockerfile
```diff
@@ -86,34 +86,39 @@ FROM cardano-services as blockfrost-worker
...
 FROM cardano-services as projector
 RUN apt-get update && apt-get install -y --no-install-recommends jq postgresql-client
 ENV \
   API_URL="http://0.0.0.0:3002" \
-  POSTGRES_DB_FILE=/run/secrets/postgres_db_projection \
-  POSTGRES_HOST=postgres \
-  POSTGRES_PASSWORD_FILE=/run/secrets/postgres_password \
-  POSTGRES_PORT=5432 \
-  POSTGRES_USER_FILE=/run/secrets/postgres_user
+  POSTGRES_DB_FILE_STAKE_POOLS=/run/secrets/postgres_db_stake_pools \
```
rhyslbw, line 117:
```suggestion
  POSTGRES_DB_FILE_STAKE_POOL=/run/secrets/postgres_db_stake_pool \
```

compose/common.yml
```diff
@@ -188,10 +191,17 @@ services:
     restart: on-failure
 
 secrets:
+  # Replicates the db-sync one for historical reasons.
```
rhyslbw, line 194: This comment is a bit confusing. Let's discuss on call and I can help refine.

packages/cardano-services/README.md
```diff
@@ -144,12 +144,12 @@ RABBITMQ_SRV_SERVICE_NAME=some-domain-for-rabbitmq \
 ### Generating Projector Schema Migrations
 
 1. Start local projection dependencies (e.g. `yarn preprod:up cardano-node-ogmios postgres`).
-1. Initialize base schema by running `start-projector` with **all** projections enabled. *Hint: you can use `git` to checkout a previous commit if you already updated entities and skipped this step*
-2. Create new entities (or update existing ones).
-3. Include any new entities in `entities` object at `src/Projection/prepareTypeormProjection.ts`.
-4. Run `yarn generate-migration`
-5. Inspect the generated migration and add a static `entity` property. *Hint: you may need to split generated migration into multiple migrations manually if it affects more than 1 entity*
-6. Export the new migration(s) from `migrations` array at `src/Projections/migrations/index.ts`
+2. Initialize base schema by running `start-projector` with **all** projections enabled. _Hint: you can use `git` to checkout a previous commit if you already updated entities and skipped this step_
```
rhyslbw, line 147: :smile:

packages/cardano-services/src/Program/options/postgres.ts
```diff
@@ -32,66 +32,118 @@ export interface PosgresProgramOptions {
   postgresSslCaFile?: string;
 }
 
-export const withPostgresOptions = (command: Command) =>
-  command
+export type ConnectionNames = 'DbSync' | 'StakePools';
```
rhyslbw, line 35:
```suggestion
export type ConnectionNames = 'DbSync' | 'StakePool';
```

packages/cardano-services/src/Program/programs/providerServer.ts
```diff
@@ -231,6 +231,7 @@ export const loadProviderServer = async (
       },
       logger
     );
+  logger.fatal(args);
```
rhyslbw, line 234: :eyes:

rhyslbw (changes requested):

compose/common.yml
rhyslbw: Thanks for the chat @iccicci. OK how about using this as content, formatted as you please:

The upstream Docker image is missing environment variables that allow the consumer to mount into a non-standard path, so we must conform with the hard-coded secret name. As demonstrated here: https://docs.docker.com/compose/use-secrets/#advanced, we could extend the Docker image and define POSTGRES_DB_FILE as the custom path to remove this workaround. See https://github.com/input-output-hk/cardano-db-sync/blob/aed18d1be192bd58d054cdba23758b579dae9f4e/nix/docker.nix#L116

packages/cardano-services/README.md
iccicci: Sorry, @rhyslbw , I didn't understood what sparked your reaction.

Was it the repeated point 1 or something introduced by my change?

Thank you

rhyslbw: Sorry, that was unnecessary. I was smiling at the docs "bug" you found. Nice job

packages/cardano-services/src/Program/programs/providerServer.ts
rhyslbw: Was this committed by accident?

iccicci: Yep, development change refuse.
Even if your comment wasn't so specific, here I had no questions as the suggestion is clear enough as it is 😃

packages/cardano-services/src/Program/options/postgres.ts
iccicci: cba27e3cf should finalize the change requested her

.github/workflows/continuous-integration-e2e.yaml
iccicci: cba27e3cf should solve this

compose/common.yml
iccicci: cba27e3cf should solve this

packages/cardano-services/README.md
iccicci: As I understood, no actions are required here

packages/cardano-services/src/Program/options/postgres.ts
iccicci: cba27e3cf should solve this

packages/cardano-services/src/Program/programs/providerServer.ts
iccicci: removed the refuse with cba27e3cf

rhyslbw (approved):
Ready for squashing

mkazlauskas (approved):
Great work! `postgres_db_db_sync` reads a little weird but we can live with that :sweat_smile:

.github/workflows/continuous-integration-e2e.yaml
```diff
@@ -65,8 +65,8 @@ jobs:
         CHAIN_HISTORY_PROVIDER: 'http'
         CHAIN_HISTORY_PROVIDER_PARAMS: '{"baseUrl":"http://localhost:4000/chain-history"}'
         DB_SYNC_CONNECTION_STRING: 'postgresql://postgres:doNoUseThisSecret!@localhost:5435/cexplorer'
-        PROJECTION_PG_CONNECTION_STRING: 'postgresql://postgres:doNoUseThisSecret!@localhost:5435/projection'
-        PROJECTION_TEST_PG_CONNECTION_STRING: 'postgresql://postgres:doNoUseThisSecret!@localhost:5435/projection_test'
+        STAKE_POOL_CONNECTION_STRING: 'postgresql://postgres:doNoUseThisSecret!@localhost:5435/stake_pool'
```
IvayloAndonov, line 68: I think `STAKE_POOL_CONNECTION_STRING` (sounds like the conn string of the pool) -> `STAKE_POOL_DB_CONNECTION_STRING` would be one way better, however, it's not a requested change, just a preference

IvayloAndonov (approved):
Great job!

PR #729, opened 2023-05-16
[LW-6528] feat: the single address wallet now takes an additional dependency 'AddressDiscovery'

AngelCastilloB:
# Context

The SingleAddressWallet  must be update to support importing multi address wallets.

# Proposed Solution

This PR add the following to the current SingleAddressWallet implementation:

1. Address derivation & resource discovery
2. Deriving addresses sequentially (by payment key index AND stake key index)
3. Discovery of resources at those addresses and unique reward accounts
4. Spending resources form multiple addresses (HD wallets)

# Important Changes

1. The SingleAddressWallet now takes a AddressDiscovery strategy. A new HDSequentialDiscovery strategy was added that implements a look-ahead search of a specified number of addresses.
2. A new function was added in the AsyncKeyAgent to override the known address collection.
3. The TransactionTracker and UtxoTracker services were updated to fetch transaction history and UTXOs from all addresses by batching a group of addresses to be queried.
4. SingleAddressWallet was renamed to PersonalWallet


packages/util/src/arrays.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Splits the given array into sub arrays of the given size. If the number of item in the array
+ * is not a multiple of the subArrayCount. The last array will contain the remaining elements.
+ *
+ * @param array The array to split.
+ * @param subArrayCount The number of items per array.
+ */
+export const splitArray = <T>(array: Array<T>, subArrayCount: number): Array<Array<T>> => {
```
mkazlauskas, line 8: This looks similar to [lodash.chunk](https://lodash.com/docs/4.17.15#chunk)

mkazlauskas (changes requested):

packages/wallet/src/services/AddressDiscovery/MultiAddressDiscovery.ts
```diff
@@ -0,0 +1,80 @@
...
+ * - if there are some transactions, increase the payment credential index and set the gap count to 0.
+ * - if the gap count reaches the given lookAheadCount stop the discovery process.
+ *
+ * Please note that the algorithm works with the transaction history, not balances, so you can have an address with 0 total coins
+ * and the algorithm will still continue with discovery if the address was previously used.
+ *
+ * If the wallet hits gap limit of unused addresses in a row, it expects there are
+ * no used addresses beyond this point and stops searching the address chain.
+ */
+export class MultiAddressDiscovery implements AddressDiscovery {
```
rhyslbw, line 25: Looks to me this is a sequential HD?
```suggestion
export class HDSequentialDiscovery implements AddressDiscovery {
```
https://input-output-hk.github.io/cardano-wallet/concepts/address-derivation

rhyslbw (commented):
Nice work @AngelCastilloB!

packages/wallet/src/SingleAddressWallet/SingleAddressWallet.ts
```diff
@@ -255,13 +266,18 @@ export class SingleAddressWallet implements ObservableWallet {
         keyAgent.knownAddresses$.pipe(
           distinctUntilChanged(groupedAddressesEquals),
           tap(
-            // derive an address if none available
+            // derive an addresses if none available
```
mirceahasegan, line 269: Just a minor comment change:
```diff
-            // derive an addresses if none available
+           // derive addresses if none available

File: packages/e2e/test/wallet/SingleAddressWallet/multiAddress.test.ts [hunk 4]
```diff
@@ -0,0 +1,164 @@
...
+
+    const unsignedTx = await txBuilder.build();
+
+    assertTxIsValid(unsignedTx);
+
+    const signedTx = await unsignedTx.sign();
+    await signedTx.submit();
+
+    // Search chain history to see if the transaction is there.
+    const txFoundInHistory = await firstValueFrom(
```
mirceahasegan, line 85: I would prefer `firstValueFromTimed`. This would not hang the test in case the tx is not there.
Also, I think `expect`ing the txFoundInHistory toBeDefined is not necessary. If it's not there the test will timeout.

packages/e2e/test/wallet/SingleAddressWallet/multiAddress.test.ts
```diff
@@ -0,0 +1,164 @@
...
+      )
+    );
+
+    expect(txFoundInHistory).toBeDefined();
+    expect(txFoundInHistory.id).toEqual(signedTx.tx.id);
+    expect(normalizeTxBody(txFoundInHistory.body)).toEqual(normalizeTxBody(signedTx.tx.body));
+
+    // Create a new wallet using the mnemonics previously generated.
+    let mnemonicString = '';
+    for (const word of mnemonics) mnemonicString += `${word} `;
```
mirceahasegan, line 99: A slight simplification:
```diff
-    for (const word of mnemonics) mnemonicString += `${word} `;
+    mnemonicString = mnemonics.join(' ');

File: packages/wallet/src/services/AddressDiscovery/HDSequentialDiscovery.ts [hunk 6]
```diff
@@ -0,0 +1,86 @@
...
+
+      ++currentIndex;
+    }
+
+    const result = [...addresses.values()];
+
+    // We need to make sure the addresses are sorted since the wallet assumes that the first address
+    // in the list is the change address (payment cred 0 and stake cred 0).
+    result.sort((a, b) => a.index - b.index || a.stakeKeyDerivationPath!.index - b.stakeKeyDerivationPath!.index);
+    asyncKeyAgent.setKnownAddresses(result);
```
mkazlauskas, line 82: `discover` should be an atomic operation that keeps the KeyAgent in it's original state if it fails at any points (which is very possible when making multiple http requests). Since `deriveAddress` mutates the key agent, you could save knownAddresses at the start of the function and rolling back to it in a `catch`. However see my other comment about refactoring deriveAddress

packages/wallet/src/SingleAddressWallet/SingleAddressWallet.ts
```diff
@@ -255,13 +266,17 @@ export class SingleAddressWallet implements ObservableWallet {
...
-                  .deriveAddress({ index: 0, type: AddressType.External }, 0)
-                  .catch(() => this.#logger.error('Failed to derive address'));
+                this.#logger.debug('No addresses available; initiating address discovery process');
+
+                void this.#addressDiscovery
+                  .discover(keyAgent, this.chainHistoryProvider, this.#addressLookAheadSearch)
+                  // The key agent 'remembers' every address it derives, however, some derived addresses
+                  // we don't want to remember (I.E gap addresses). So we must override the known addresses
+                  // with the curated list of addresses returned by the discovery process.
+                  .catch(() => this.#logger.error('Failed to complete the address discovery process'));
```
mkazlauskas, line 279: If this fails, then the wallet is essentially bricked. We should consider a retry strategy (either here or as part of `discover` implementation).

packages/wallet/src/services/types.ts
```diff
@@ -35,6 +36,27 @@ export interface UtxoTracker extends TransactionalTracker<Cardano.Utxo[]> {
...
+   *
+   * @param keyAgent The key agent controlling the root key to be used to derive the addresses to be discovered.
+   * @param chainHistoryProvider A cardano chain history provider.
+   * @param lookAheadCount The number of addresses to look ahead in the search.
+   * @returns A promise that will be resolved into a GroupedAddress list containing the discovered addresses.
+   */
+  discover(
+    keyAgent: AsyncKeyAgent,
+    chainHistoryProvider: ChainHistoryProvider,
+    lookAheadCount: number
```
mkazlauskas, line 56: Those are specific to HDSequentialDiscovery implementation. I think it's unlikely that it would be needed to change those dynamically, so I would put them to the constructor of `new HDSequentialDiscovery({ addressLookAheadSearch, chainHistoryProvider })` rather than as argument of discover
```suggestion
```

packages/e2e/test/wallet/SingleAddressWallet/multiAddress.test.ts
```diff
@@ -0,0 +1,161 @@
...
+import { filter, map, take } from 'rxjs';
+import { getEnv, walletVariables } from '../../../src/environment';
+import { isNotNil } from '@cardano-sdk/util';
+
+const env = getEnv(walletVariables);
+const logger = createLogger();
+const PAYMENT_ADDRESSES_TO_GENERATE = 60;
+const COINS_PER_ADDRESS = 3_000_000n;
+
+describe('SingleAddressWallet/multiAddress', () => {
```
mkazlauskas, line 17: Yeah, we need to rename it :sweat_smile:

packages/key-management/src/types.ts
```diff
@@ -175,9 +177,11 @@ export interface KeyAgent {
 
 export type AsyncKeyAgent = Pick<KeyAgent, 'deriveAddress' | 'derivePublicKey' | 'signBlob' | 'signTransaction'> & {
   knownAddresses$: Observable<GroupedAddress[]>;
+  getKeyAgent(): Promise<KeyAgent>;
```
mkazlauskas, line 180: That won't work with our messaging implementation, because returned object has both async methods and sync properties - not supported. Seems like we need to refactor `deriveAddress` to support deriving without mutating the key agent, probably add an argument like `pure=false`
```suggestion
```

packages/key-management/src/types.ts
```diff
@@ -175,9 +177,11 @@ export interface KeyAgent {
...
   getChainId(): Promise<Cardano.ChainId>;
   getBip32Ed25519(): Promise<Crypto.Bip32Ed25519>;
   getExtendedAccountPublicKey(): Promise<Crypto.Bip32PublicKeyHex>;
+  setKnownAddresses(addresses: GroupedAddress[]): void;
```
mkazlauskas, line 184: All methods in **Async**KeyAgent must be async
```suggestion
  setKnownAddresses(addresses: GroupedAddress[]): Promise<void>;
```

mkazlauskas (dismissed):

packages/wallet/src/services/AddressDiscovery/HDSequentialDiscovery.ts
mkazlauskas: KeyAgentBase updates [knownAddresses](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/key-management/src/KeyAgentBase.ts#L107) so it doesn't matter whether you're using async or non-async variant

packages/wallet/src/SingleAddressWallet/SingleAddressWallet.ts
mkazlauskas: We typically use exponential backoff via [coldObservableProvider](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/wallet/test/services/util/coldObservableProvider.test.ts) util. You could either use that (by converting promise-observable-promise), or find something else. I'd say cut it in like 5 minutes?

packages/wallet/src/services/types.ts
mkazlauskas: In that case you could provide a new discovery strategy object? It just seems very specific to the HDSequentialDiscovery strategy, doesn't seem right on the AddressDiscovery interface.

packages/wallet/src/services/AddressDiscovery/HDSequentialDiscovery.ts
mkazlauskas: Would be great to add a test where the discovery fails, asserting that keyAgent.knownAddresses weren't changed

packages/wallet/src/SingleAddressWallet/SingleAddressWallet.ts
mkazlauskas: Looks good, but would be nice to validate with test(s): create a wallet with mock AddressDiscovery that fails once or twice before resolving.

packages/wallet/test/services/addressDiscovery/HDSequentialDiscovery.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        index: 0,
+        role: KeyRole.Stake
+      },
+      type: AddressType.External
+    };
+
+    await mockKeyAgent.setKnownAddresses([knownAddress]);
+
+    const discovery = new HDSequentialDiscovery(mockAlwaysFailChainHistoryProvider, 25);
+    await expect(async () => await discovery.discover(mockKeyAgent)).rejects.toThrow();
```
mkazlauskas, line 73: Just FIY this more compact syntax also works
```suggestion
    await expect(discovery.discover(mockKeyAgent)).rejects.toThrow();
```

rhyslbw (approved):
Great work @AngelCastilloB :rocket:

mirceahasegan (approved):
Amazing work @AngelCastilloB ! :mage_man: :rocket:

PR #730, opened 2023-05-16
fix(key-management): fixed ttl

iadmytro:
# Context

Incorrect TTL value prevented transaction from signing.


mkazlauskas (changes requested):
Please search the codebase for any other `toString` to verify that this is the only bug of this kind

mkazlauskas (approved):
Nice catch!

rhyslbw (approved):

PR #732, opened 2023-05-18
Feature/resolve handles using observable wallet tx builder

VanessaPC:
# Context
It is setting up `HandleProvider` into the `cardano-sdk/core`  package and setting up the initial work for it to be exposed to be integrated.

# Proposed Solution
- [x] Add the HandleProvider under the wallet packages with its unit test
- [x] Set up the initial modification in OutputBuilder to contemplate the integration in the builder. 
- [x] Set up the tests to test the integration against. 
- [x]  Set up the functionality in `txBuilder`.

Adds KoraLabsHandleProvider package to  the `Cardano-SDK/cardano-services-client` package.
Add functionality to `TxBuilder` to integrate the new `handleProvider`.


packages/tx-construction/src/tx-builder/OutputBuilder.ts
```diff
@@ -116,6 +120,21 @@ export class TxOutputBuilder implements OutputBuilder {
     return this;
   }
 
+  // make it take an output that has type AdaHandle
+  handle(_output: Handle): OutputBuilder {
+    // output/handle: $AdaHandle
```
VanessaPC, line 125: I imagine none of this set-up should be merged. So I can delete it, but I wanted to initially leave it, so we see how if I do the handle here, as opposed to txBuilder, I can continue chaining actions. But if I move it to txBuilder, then `.....handle('$AdaHandle').coin().` wouldn't work. 
Also,

packages/core/src/Provider/HandleProvider/types.ts
```diff
@@ -0,0 +1,23 @@
+import { ByronAddress, EnterpriseAddress, PaymentAddress, PointerAddress } from '../../Cardano/Address';
+import { Point } from '../..';
+
+export type Handle = string;
+
+type CardanoAddress = ByronAddress | PaymentAddress | EnterpriseAddress | PointerAddress;
+
+export interface HandleInfo {
```
mkazlauskas, line 8: Since it has `resolvedAt`, it's represents a handle resolved at some point in time - I think `HandleResolution` would communicate the purpose of the type a little better. Also, we would likely have more handle issuers in the future, so it would be great to add an `issuer` property, to not implicitly couple this with 1 implementation.

```suggestion
export enum HandleIssuer {
  KoraLabs = 'KoraLabs'
}

export interface HandleResolution {
  issuer: HandleIssuer;
```

packages/core/src/Provider/HandleProvider/types.ts
```diff
@@ -0,0 +1,23 @@
...
+  };
+  resolvedAt: Point;
+}
+
+export interface ResolveHandlesArgs {
+  handles: Handle[];
+}
+
+export interface HandleProvider {
+  resolveHandles(args: ResolveHandlesArgs): Promise<HandleInfo[] | null>;
```
mkazlauskas, line 22: Since it can be used to resolve multiple handles, user needs to be able to determine which handle was resolved and which wasn't.
```suggestion
  resolveHandles(args: ResolveHandlesArgs): Promise<Array<HandleInfo | null>>;
```

packages/tx-construction/src/tx-builder/OutputBuilder.ts
```diff
@@ -21,6 +21,8 @@ export interface OutputBuilderProps {
   txOut?: PartialTxOut;
   /** Logger */
   logger: Logger;
+  /** Koralabs Handle Provider for ADAHandle */
+  handleProvider?: KoraLabsHandleProvider;
```
mkazlauskas, line 25:
```suggestion
  handleProvider?: HandleProvider;
```

packages/tx-construction/src/tx-builder/OutputBuilder.ts
mkazlauskas: See

mkazlauskas (changes requested):

packages/core/src/Provider/HandleProvider/index.ts
```diff
@@ -0,0 +1,47 @@
...
+import { IHandle } from '@koralabs/handles-public-api-interfaces';
+import { toHandleInfo } from './utils';
+import axios, { AxiosInstance } from 'axios';
+
+export interface KoraLabsHandleProviderDeps {
+  serverUrl: string;
+  networkInfoProvider: NetworkInfoProvider;
+}
+
+export class KoraLabsHandleProvider implements HandleProvider {
```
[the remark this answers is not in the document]
VanessaPC: I believe `wallet` shouldn't be the one because it can be used by any other application than a wallet? so that would leave `cardano-services` @rhyslbw ?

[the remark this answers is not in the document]
VanessaPC: Ok, I will move it to `wallet`.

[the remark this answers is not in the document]
VanessaPC: when moving the provider to `wallet`, I find that the addresses I need from `core` aren't exported, and only used in `core`. It's ok to make them exported, or is it a particular reason they aren't?

`import { ByronAddress, EnterpriseAddress, PaymentAddress, PointerAddress } from '@cardano-sdk/core/src/Cardano/Address'; `

packages/core/src/Provider/index.ts
```diff
@@ -9,4 +9,3 @@ export * from './UtxoProvider';
 export * from './ChainHistoryProvider';
 export * from './providerFactory';
 export * from './types';
-export * from './HandleProvider';
```
mkazlauskas: Interface should remain in `core`, only the implemenation should be moved to `wallet`

mkazlauskas (changes requested):

packages/core/test/mocks/mockHandleProvider.ts
```diff
@@ -0,0 +1,11 @@
+import { Cardano } from '../../src';
+
+export const Handle = {
```
mkazlauskas, line 3: Once https://github.com/input-output-hk/cardano-js-sdk/pull/720 is merged, please hoist this to `util-dev/src/mockProviders` and also rename the variable
```suggestion
export const handleResolution = {
```

packages/wallet/src/services/ProviderTracker/HandleProvider.ts
```diff
@@ -0,0 +1,45 @@
+import { HandleProvider, HandleResolution, ResolveHandlesArgs } from '@cardano-sdk/core';
+import { HealthCheckResponse, NetworkInfoProvider, ProviderError, ProviderFailure } from '@cardano-sdk/core/src';
```
mkazlauskas, line 2: Imports from other SDK packages should be referencing the index. 

Also please hoist this file to `src/services`, as ProviderTracker has special meaning that doesn't match what this class does (used for SingleAddressWallet observable implementations)
```suggestion
import { HandleProvider, HandleResolution, ResolveHandlesArgs, HealthCheckResponse, NetworkInfoProvider, ProviderError, ProviderFailure } from '@cardano-sdk/core';
```

packages/wallet/src/services/ProviderTracker/HandleProvider.ts
```diff
@@ -0,0 +1,45 @@
...
+import { IHandle } from '@koralabs/handles-public-api-interfaces';
+import { toHandleInfo } from '../util';
+import axios, { AxiosInstance } from 'axios';
+
+export interface KoraLabsHandleProviderDeps {
+  serverUrl: string;
+  networkInfoProvider: NetworkInfoProvider;
```
mkazlauskas, line 9: This will probably be needed to make it work in the browser (see `cardano-services-client` implementations)
```suggestion
  networkInfoProvider: NetworkInfoProvider;
  axiosAdapter: AxiosAdapter;
```

packages/tx-construction/src/tx-builder/TxBuilder.ts
```diff
@@ -179,6 +183,7 @@ export class GenericTxBuilder implements TxBuilder {
             return {
               ctx: {
                 auxiliaryData,
+                handles: [],
```
VanessaPC, line 186: I have added the handles here, but I'm unsure how outputBuilder and txBuilder resolve the final response.. and I think that this needs to be added in more places.

VanessaPC:
As per this [conversation](https://input-output-rnd.slack.com/archives/C02BW8GUNF7/p1684503052923629?thread_ts=1684425295.799099&cid=C02BW8GUNF7) I have fixed the git history before having more issues with it, because I also needed to rebase master and I wanted to avoid further issues while delivering both commit history commits. I have made sure I added all the feedback from all the comments above, especially the last ones from Martynas on Friday.

packages/tx-construction/package.json
```diff
@@ -63,6 +63,8 @@
     "@cardano-sdk/input-selection": "workspace:~",
     "@cardano-sdk/key-management": "workspace:~",
     "@cardano-sdk/util": "workspace:~",
+    "@cardano-sdk/util-dev": "workspace:~",
```
iccicci, line 66: `util-dev` usually appears only in `devDependencies` (and it is already). Any reason to have it in `dependencies`?

packages/tx-construction/src/tx-builder/OutputBuilder.ts
```diff
@@ -21,6 +21,8 @@ export interface OutputBuilderProps {
   txOut?: PartialTxOut;
   /** Logger */
   logger: Logger;
+  /** Handle Provider for ADAHandle */
```
rhyslbw, line 24:
```suggestion
  /** Handle Provider for resolving an address from the supplied handle */
```
Let's avoid using the ADA Handle brand, since we're implementing the _Handle Standard_

packages/wallet/test/services/ProviderTracker/HandleProvider.test.ts
```diff
@@ -0,0 +1,105 @@
...
+      return res.end(JSON.stringify(result.body));
+    }
+    // No default
+  }
+
+  return res.end(JSON.stringify(result.body));
+});
+
+describe('HandleProvider', () => {
+  it('should resolve a handle', async () => {
```
rhyslbw, line 52: Best to nest each method within a `describe` block

``` typescript
describe('HandleProvider', () => {
   describe('resolveHandles', () => {
```

packages/wallet/test/services/ProviderTracker/HandleProvider.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    const { serverUrl, closeMock } = await mockServer(async () => ({
+      body: bobHandle
+    }));
+    const provider = new KoraLabsHandleProvider({
+      networkInfoProvider: mocks.mockNetworkInfoProvider2(100),
+      serverUrl
+    });
+    const tip = await mocks.mockNetworkInfoProvider2(100).ledgerTip();
+    const result = await provider.resolveHandles({ handles: ['bob'] });
```
rhyslbw, line 61: This is a good indication of the existing test style we're using:
https://github.com/input-output-hk/cardano-js-sdk/blob/6e28df412797974b8ce6f6deb0c3346ff5938a05/packages/cardano-services-client/test/AssetInfoProvider/assetInfoHttpProvider.test.ts#L49-L53

packages/core/src/Provider/HandleProvider/types.ts
```diff
@@ -0,0 +1,28 @@
+import { ByronAddress, EnterpriseAddress, PaymentAddress, PointerAddress } from '../../Cardano/Address';
+import { Point } from '../..';
+
+export type Handle = string;
+
+export type CardanoAddress = ByronAddress | PaymentAddress | EnterpriseAddress | PointerAddress;
+
+export enum HandleIssuer {
+  KoraLabs = 'KoraLabs'
```
rhyslbw, line 9:
```suggestion
  ADAHandles = 'ADAHandles'
```

packages/wallet/src/services/ProviderTracker/HandleProvider.ts
```diff
@@ -0,0 +1,75 @@
...
+export const toHandleResolution = ({
+  apiResponse,
+  tip
+}: {
+  apiResponse: IHandle;
+  tip: Cardano.Tip;
+}): HandleResolution => ({
+  handle: apiResponse.name,
+  hasDatum: apiResponse.hasDatum,
+  issuer: HandleIssuer.KoraLabs,
```
rhyslbw, line 31:
```suggestion
  issuer: HandleIssuer.AdaHandles,
```

packages/util-dev/src/mockProviders/mockHandleProvider.ts
```diff
@@ -0,0 +1,11 @@
+import { Cardano } from '@cardano-sdk/core';
+
+export const Handle = {
+  handle: 'AdaHandle',
```
rhyslbw, line 4:
```suggestion
  handle: 'mary',
```

packages/tx-construction/src/tx-builder/OutputBuilder.ts
rhyslbw, line 24:
```suggestion
  /** Handle Provider for resolving addresses */
```

packages/core/src/Provider/HandleProvider/types.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+export interface HandleResolution {
+  issuer: HandleIssuer;
```
rhyslbw, line 12:
```suggestion
  policyId: Cardano.PolicyId;
```
Thoughts on this @mkazlauskas? The extra construct is unnecessary and makes our implementation static when all we need here the policy ID to differentiate the issuer.

packages/tx-construction/src/tx-builder/TxBuilder.ts
rhyslbw: @mkazlauskas can you please pair up with @VanessaPC and flesh out the implementation detail?

rhyslbw (changes requested):

packages/tx-construction/src/tx-builder/TxBuilder.ts
rhyslbw: @mkazlauskas can you please pair up with @VanessaPC and flesh out the implementation detail?

packages/core/src/Provider/HandleProvider/types.ts
mkazlauskas: Good idea, agree

VanessaPC: So in the HandleProvider, where do I get the `policyId`? The `IHandle` API response, or the `tip`, from Cardano tip, con't have it, I believe.

mkazlauskas: It's static for cora labs provider: you could either determine network (e.g. preprod/mainnet) by URL, or just add a new `policyId` arg to the constructor

packages/wallet/package.json
```diff
@@ -85,6 +86,8 @@
     "@cardano-sdk/tx-construction": "workspace:~",
     "@cardano-sdk/util": "workspace:~",
     "@cardano-sdk/util-rxjs": "workspace:~",
+    "@koralabs/handles-public-api-interfaces": "1.6.6",
+    "axios": "^1.4.0",
```
mkazlauskas, line 90: This should be under `dependencies` since it's used in `src/`

packages/tx-construction/src/tx-builder/types.ts
```diff
@@ -11,7 +11,10 @@ import { OutputBuilderValidator } from './OutputBuilder';
 import { OutputValidation } from '../output-validation';
 
 export type PartialTxOut = Partial<
-  Pick<Cardano.TxOut, 'address' | 'datumHash' | 'datum' | 'scriptReference'> & { value: Partial<Cardano.Value> }
+  Pick<Cardano.TxOut, 'address' | 'datumHash' | 'datum' | 'scriptReference'> & {
+    value: Partial<Cardano.Value>;
+    handles: Array<Handle>;
```
mkazlauskas, line 16: A single TxOut should only refer to 1 handle, since it's to a single address
```suggestion
    handle: Handle;
```

packages/tx-construction/src/tx-builder/OutputBuilder.ts
```diff
@@ -116,6 +118,12 @@ export class TxOutputBuilder implements OutputBuilder {
     return this;
   }
 
+  handle(output: Handle): OutputBuilder {
+    // if there is more than one handle, then maybe we should just ask for all of it
```
mkazlauskas, line 122: See

mkazlauskas (changes requested):

packages/core/src/Provider/HandleProvider/types.ts
```diff
@@ -0,0 +1,28 @@
...
+  handle: Handle;
+  hasDatum: boolean;
+  resolvedAddresses: {
+    cardano: CardanoAddress;
+  };
+  resolvedAt: Point;
+  code?: number | undefined;
```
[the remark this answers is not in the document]
VanessaPC: A leftover of trying to fix some tests earlier. It's removed.

packages/tx-construction/src/tx-builder/types.ts
VanessaPC: `handle:` or `handles:`

packages/core/src/Provider/HandleProvider/types.ts
rhyslbw: > or just add a new policyId arg to the constructor

This. Best to keep the knowledge of the actual `policyId` at the configuration level

packages/tx-construction/src/tx-builder/OutputBuilder.ts
VanessaPC: [f293dc8313](https://github.com/input-output-hk/cardano-js-sdk/pull/732/commits/f293dc8313a0c07bc570e18d025412315abc45ef) should solve this

packages/tx-construction/src/tx-builder/types.ts
VanessaPC: [f293dc8313](https://github.com/input-output-hk/cardano-js-sdk/pull/732/commits/f293dc8313a0c07bc570e18d025412315abc45ef) should solve this

packages/wallet/package.json
VanessaPC: [f293dc8313](https://github.com/input-output-hk/cardano-js-sdk/pull/732/commits/f293dc8313a0c07bc570e18d025412315abc45ef) should solve this

packages/wallet/src/services/ProviderTracker/HandleProvider.ts
```diff
@@ -0,0 +1,75 @@
...
+  async resolveHandles(args: ResolveHandlesArgs): Promise<HandleResolution[]> {
+    try {
+      const tip = await this.networkInfoProvider.ledgerTip();
+      const response = await Promise.all(
+        args.handles.map((handle) => this.axiosClient.get<IHandle>(`/handles/${handle}`))
+      );
+      return response.map(({ data: apiResponse }) => toHandleResolution({ apiResponse, tip }));
+    } catch (error) {
+      if (axios.isAxiosError(error)) {
+        throw new ProviderError(ProviderFailure.Unhealthy, error, `Failed to resolve handles due to: ${error.message}`);
```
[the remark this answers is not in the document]
VanessaPC: [f293dc8313](https://github.com/input-output-hk/cardano-js-sdk/pull/732/commits/f293dc8313a0c07bc570e18d025412315abc45ef) should solve this

packages/wallet/src/services/ProviderTracker/HandleProvider.ts
```diff
@@ -0,0 +1,75 @@
...
+      }
+      throw error;
```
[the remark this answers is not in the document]
VanessaPC: [f293dc8313](https://github.com/input-output-hk/cardano-js-sdk/pull/732/commits/f293dc8313a0c07bc570e18d025412315abc45ef) should solve this

packages/wallet/src/services/ProviderTracker/HandleProvider.ts
VanessaPC: This was moved in the end to `cardano-services` [f293dc8313](https://github.com/input-output-hk/cardano-js-sdk/pull/732/commits/f293dc8313a0c07bc570e18d025412315abc45ef) should solve this

packages/core/src/Provider/HandleProvider/types.ts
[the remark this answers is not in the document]
VanessaPC: [f293dc8313](https://github.com/input-output-hk/cardano-js-sdk/pull/732/commits/f293dc8313a0c07bc570e18d025412315abc45ef) should solve this

packages/tx-construction/src/tx-builder/OutputBuilder.ts
```diff
@@ -116,7 +119,12 @@ export class TxOutputBuilder implements OutputBuilder {
     return this;
   }
 
-  async build(): Promise<Cardano.TxOut> {
+  handle(output: Handle): OutputBuilder {
```
mkazlauskas, line 122:
```suggestion
  handle(handle: Handle): OutputBuilder {
```

packages/tx-construction/src/tx-builder/OutputBuilder.ts
VanessaPC: e25a4e19c0746a361b37 this should solve this

packages/util-dev/src/mockProviders/mockHandleProvider.ts
VanessaPC: e25a4e19c0746a361b37 this should solve this

packages/tx-construction/src/tx-builder/types.ts
```diff
@@ -11,7 +11,10 @@ import { OutputBuilderValidator } from './OutputBuilder';
 import { OutputValidation } from '../output-validation';
 
 export type PartialTxOut = Partial<
-  Pick<Cardano.TxOut, 'address' | 'datumHash' | 'datum' | 'scriptReference'> & { value: Partial<Cardano.Value> }
+  Pick<Cardano.TxOut, 'address' | 'datumHash' | 'datum' | 'scriptReference'> & {
+    value: Partial<Cardano.Value>;
+    handle: Handle;
```
mkazlauskas, line 16: Not all TxOut have associated handles, especially not this partial type
```suggestion
    handle?: Handle;
```

mkazlauskas (changes requested):

packages/cardano-services-client/src/HandleProvider/HandleProvider.ts
```diff
@@ -0,0 +1,82 @@
+import {
```
mkazlauskas, line 1: Please rename this file to correspond to the class name: `KoraLabsHandleProvider`

packages/cardano-services-client/test/HandleProvider/handleProvider.test.ts
```diff
@@ -0,0 +1,74 @@
...
+import MockAdapter from 'axios-mock-adapter';
+import axios from 'axios';
+
+const config = {
+  networkInfoProvider: mocks.mockNetworkInfoProvider2(100),
+  policyId: Cardano.PolicyId('50fdcdbfa3154db86a87e4b5697ae30d272e0bbcfa8122efd3e301cb'),
+  serverUrl: 'http://some-hostname:3000'
+};
+
+describe('HandleProvider', () => {
```
mkazlauskas, line 15: Please add test cases for different kinds of errors it can throw

packages/cardano-services-client/package.json
```diff
@@ -69,6 +69,7 @@
   "dependencies": {
     "@cardano-sdk/core": "workspace:~",
     "@cardano-sdk/util": "workspace:~",
+    "@koralabs/handles-public-api-interfaces": "1.6.6",
```
mkazlauskas, line 72: This is only used for types that are not exposed (internal usage only), so it should probably be under `devDependencies`

packages/tx-construction/src/tx-builder/OutputBuilder.ts
```diff
@@ -1,9 +1,10 @@
-import { Cardano } from '@cardano-sdk/core';
```
mkazlauskas, line 1: Please add tests for the functional changes in `tx-construction` package

packages/util-dev/src/mockProviders/mockHandleProvider.ts
```diff
@@ -0,0 +1,77 @@
+import { Cardano } from '@cardano-sdk/core';
+
+export const getAliceHandleProviderResponse = {
```
mkazlauskas, line 3: Do we want to use those stub responses in Lace repo (for development/tests/storybook etc.)?
- If the answer is yes, then we should probably implement export a StubHandleProvider impl as well.
- If the answer is no, then we should probably hoist this to `cardano-services-client/test/util.ts`, as without the provider implementation I don't see other use cases for it. Are there any?

mkazlauskas (changes requested):

packages/cardano-services-client/package.json
VanessaPC: It is imported in the src/ folder, so if it's moved to the `devDependencies` the import won't work, and it will appear as listed on the wrong list.

packages/tx-construction/src/tx-builder/OutputBuilder.ts
VanessaPC: I fixed the tests in tx-constructions, according to the changes made in this PR. However, I don't think the context output is tested. I might be wrong.

mkazlauskas: The new functionality that we need to add tests for is:
- building outputs with a `handle`
- returning used handles in `ctx` together with signed transaction

packages/cardano-services-client/src/HandleProvider/KoraLabsHandleProvider.ts
```diff
@@ -0,0 +1,83 @@
...
+      );
+      return response.map(({ data: apiResponse }) => toHandleResolution({ apiResponse, policyId: this.policyId, tip }));
+    } catch (error) {
+      if (axios.isAxiosError(error)) {
+        if (error.request) {
+          throw new ProviderError(ProviderFailure.ConnectionFailure, error, error.code);
+        }
+        throw new ProviderError(ProviderFailure.Unhealthy, error, `Failed to resolve handles due to: ${error.message}`);
+      }
+      throw new ProviderError(ProviderFailure.Unknown, error, 'Failed to resolve handles');
```
iccicci, line 72: In the `try` block there is at least one call which can throw a `ProviderError`, a check as follows is required before the catch all error in order to avoid it produces an `Unknown` `ProviderError` with another `ProviderError` as its `innerError`.

```suggestion
      if (error instanceof ProviderError) throw error;
      throw new ProviderError(ProviderFailure.Unknown, error, 'Failed to resolve handles');
```

packages/core/src/Provider/HandleProvider/types.ts
```diff
@@ -0,0 +1,30 @@
...
+    cardano: CardanoAddress;
+  };
+  resolvedAt: Point;
+}
+
+export interface ResolveHandlesArgs {
+  handles: Handle[];
+}
+
+export interface HandleProvider {
```
iccicci, line 28:
```suggestion
export interface StakePoolProvider extends Provider {
```

This is how usually providers are defined in order to let them inherit the `healthCheck()` method.
Next it must be implemented as well, but I see the `KoraLabsHandleProvider` already correctly implements it.

packages/cardano-services-client/test/HandleProvider/handleProvider.test.ts
```diff
@@ -0,0 +1,89 @@
+/* eslint-disable no-magic-numbers */
```
iccicci, line 1: This file should be renamed to make clear it tests stuff from `KoraLabsHandleProvider.ts` - [ref](https://github.com/input-output-hk/cardano-js-sdk/pull/732#discussion_r1205665133)

VanessaPC: dde35b14797987 should address this

packages/core/src/Provider/HandleProvider/types.ts
VanessaPC: dde35b14797987 should address this

packages/cardano-services-client/src/HandleProvider/KoraLabsHandleProvider.ts
VanessaPC: dde35b14797987 should address this

packages/cardano-services-client/test/HandleProvider/koraLabsHandleProvider.test.ts
```diff
@@ -0,0 +1,94 @@
...
+import MockAdapter from 'axios-mock-adapter';
+import axios from 'axios';
+
+const config = {
+  networkInfoProvider: mocks.mockNetworkInfoProvider2(100),
+  policyId: Cardano.PolicyId('50fdcdbfa3154db86a87e4b5697ae30d272e0bbcfa8122efd3e301cb'),
+  serverUrl: 'http://some-hostname:3000'
+};
+
+describe('HandleProvider', () => {
```
mkazlauskas, line 16: Also this file name has inconsistent case `koraLabs...` -> `KoraLabs`
```suggestion
describe('KoraLabsHandleProvider', () => {
```

packages/tx-construction/test/tx-builder/TxBuilder.test.ts
```diff
@@ -302,6 +304,11 @@ describe('GenericTxBuilder', () => {
       expect(outputBuilder.toTxOut().datumHash).toEqual(datumHash);
     });
 
+    it('can set handle', () => {
```
mkazlauskas, line 307: I think we're missing a piece of essential functionality
```suggestion
    describe('build()', () => {
      it('resolves handle to an address');
      it('failing to resolve handle rejects with UnableToResolveHandleError');
    });
    it('can set handle', () => {
```

packages/util-dev/src/mockProviders/mockHandleProvider.ts
mkazlauskas: > I will move it to test/utils then.

Please do, thanks

mkazlauskas (changes requested):

packages/tx-construction/test/tx-builder/TxBuilder.test.ts
VanessaPC: There is no functionality for that yet. It will come in the next ticket so it makes more sense to add the tests when adding the actual TxBuilder handle resolution code.

packages/cardano-services-client/test/HandleProvider/koraLabsHandleProvider.test.ts
VanessaPC: 52d4659f5f146a8759de9fd65c45a9d8a2e8a961 should solve this

mkazlauskas: Please update the filename too `koraLabsHandleProvider.test.ts` -> `KoraLabsHandleProvider.test.ts`
`

packages/tx-construction/test/tx-builder/TxBuilder.test.ts
```diff
@@ -490,6 +497,18 @@ describe('GenericTxBuilder', () => {
...
 
+  it('returns a context with used handles along with the signed transaction', async () => {
+    const handle = '$adaHandle';
+    const tx = txBuilder
+      .addOutput({
+        ...mocks.utxo[0][1],
+        handle
+      })
+      .build();
+    const { ctx } = await tx.sign();
```
mkazlauskas, line 508:
```suggestion
    const { handles } = await tx.inspect();
    expect(handles).toEqual([handle]);
    const { ctx } = await tx.sign();
```

packages/tx-construction/test/tx-builder/TxBuilder.test.ts
VanessaPC: caf857dd9cbf3d89392f48f77bb91a1c55696edb should solve this

packages/cardano-services-client/test/HandleProvider/koraLabsHandleProvider.test.ts
VanessaPC: Sure, currently I followed the pattern in that folder, in which all the files containing the test, start in lower case.

packages/tx-construction/test/tx-builder/TxBuilder.test.ts
```diff
@@ -356,6 +393,23 @@ describe('GenericTxBuilder', () => {
...
+          .buildOutput()
+          .address(address)
+          .handle('alice')
+          .coin(output1Coin)
+          .build();
+
+        expect(outputBuilderTx.handle).toBe(resolvedHandle);
+      });
+
+      it('rejects with an error when handle resolution fails', async () => {
```
mkazlauskas, line 408: Missing a test for when handle provider resolves to `null` - no error, but also not resolved. Should probably create a new error type for that.
```suggestion
      it('rejects with an error when handle is not found')
      it('rejects with an error when handle resolution fails', async () => {
```

packages/wallet/src/services/ProviderTracker/TrackedHandleProvider.ts
```diff
@@ -0,0 +1,37 @@
...
+  reset() {
+    this.healthCheck$.next(CLEAN_FN_STATS);
+    this.resolveHandles$.next(CLEAN_FN_STATS);
+  }
+}
+
+/**
+ * Wraps a HandleProvider, tracking # of calls of each function
+ */
+export class TrackedHandleProvider extends ProviderTracker implements HandleProvider {
```
mkazlauskas, line 23: Missing tests for this. Although you could just remove it entirely and just use HandleProvider directly - tracked providers are used to determine wallet's sync state (when it settles etc.)

mkazlauskas (changes requested):

packages/cardano-services-client/test/HandleProvider/koraLabsHandleProvider.test.ts
mkazlauskas: Thanks, those existing test names are matching the exported function names

packages/wallet/src/services/ProviderTracker/TrackedHandleProvider.ts
mkazlauskas: It will be, but not the provider that's used in TxBuilder. Wrapped (tracked) provider should only be used for resolving wallet's state, and transaction that's being built is not part of it.

packages/tx-construction/src/tx-builder/OutputBuilder.ts
```diff
@@ -132,6 +134,14 @@ export class TxOutputBuilder implements OutputBuilder {
       throw outputValidation;
     }
 
+    if (this.#partialOutput.handle) {
+      const resolution = await this.#handleProvider.resolveHandles({ handles: [this.#partialOutput.handle] });
+
+      if (resolution[0] !== null) {
+        txOut.handle = resolution[0];
```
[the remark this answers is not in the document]
VanessaPC: I wasn't sure about this at the time and I wanted to ask today about it. The type of `resolvedAdresses.cardano` is:

```export type CardanoAddress =
  | Cardano.ByronAddress
  | Cardano.PaymentAddress
  | Cardano.EnterpriseAddress
  | Cardano.PointerAddress;
```

But the type for `txOut.address` is `export type PaymentAddress = OpaqueString<'PaymentAddress'>;`.
This breaks the typings since some of the `CardanoAdress` types are not strings, but classes with additional properties. 

Do we need to try and check the type and attempt to resolve it to a string?  Or is the type of `HandleResolution.resolvedAddresses.cardano` actually just `PaymentAddress`?

packages/tx-construction/test/tx-builder/TxBuilder.test.ts
VanessaPC: I think the type was wrong and the provider will either return an array of `HandleResolution` or throw an error. At least that is how it is in its current implementation.

packages/tx-construction/src/tx-builder/OutputBuilder.ts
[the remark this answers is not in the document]
mkazlauskas: > Or is the type of HandleResolution.resolvedAddresses.cardano actually just PaymentAddress

That's correct, let's update the type.

packages/tx-construction/test/tx-builder/TxBuilder.test.ts
mkazlauskas: No, the type intentionally returns an array of either `null | HandleResolution`, where `null` means that it couldn't be resolved. This enables the user to interpret the result better, e.g.
```typescript
resolveHandles(['handle1', 'handle2']) // => [null, {...}]
```

packages/wallet/src/services/ProviderTracker/TrackedHandleProvider.ts
VanessaPC: ea7c3ca3009ec6aaf161505165af8c9a2481bbb2 should address this

packages/tx-construction/src/tx-builder/OutputBuilder.ts
```diff
@@ -23,11 +23,11 @@ export interface OutputBuilderProps {
   /** Logger */
   logger: Logger;
   /** Handle Provider for resolving addresses */
-  handleProvider?: HandleProvider;
+  handleProvider: HandleProvider;
```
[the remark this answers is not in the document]
VanessaPC: ea7c3ca3009ec6aaf161505165af8c9a2481bbb2 should address this

packages/e2e/src/factories.ts
```diff
@@ -160,6 +164,21 @@ utxoProviderFactory.register(HTTP_PROVIDER, async (params: any, logger: Logger):
   });
 });
 
+handleProviderFactory.register(HANDLE_PROVIDER, async (params: any): Promise<HandleProvider> => {
+  if (params.serverUrl === undefined) throw new Error(`${KoraLabsHandleProvider.name}: ${MISSING_URL_PARAM}`);
+
+  return new Promise<HandleProvider>(async (resolve) => {
+    resolve(
+      new KoraLabsHandleProvider({
```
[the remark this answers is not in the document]
VanessaPC: ea7c3ca3009ec6aaf161505165af8c9a2481bbb2 should address this

packages/tx-construction/test/tx-builder/TxBuilder.test.ts
```diff
@@ -349,6 +405,25 @@ describe('GenericTxBuilder', () => {
...
+        const txOut = await txBuilder.buildOutput().handle('alice').coin(output1Coin).build();
+
+        expect(txOut.handle).toBe(resolvedHandle);
+        expect(txOut.address).toBe(resolvedHandle.resolvedAddresses.cardano);
+      });
+
+      it('rejects with an error when a handle is not found', async () => {
+        await expect(
+          txBuilderWithNullHandles.buildOutput().handle('alice').coin(output1Coin).build()
+        ).rejects.toThrowError(OutputValidationMissingRequiredError);
```
mkazlauskas, line 419: Maybe let's rename this to better communicate what's actually happening
```suggestion
        ).rejects.toThrowError(HandleNotFoundError);
```

packages/tx-construction/test/tx-builder/TxBuilder.test.ts
```diff
@@ -349,6 +405,25 @@ describe('GenericTxBuilder', () => {
...
+      });
+
+      it('rejects with an error when handle is not found', async () => {
```
mkazlauskas, line 422:
```suggestion
      it('rejects with an error when handle provider fails to resolve', async () => {
```

packages/tx-construction/test/tx-builder/TxBuilder.test.ts
```diff
@@ -302,6 +353,11 @@ describe('GenericTxBuilder', () => {
       expect(outputBuilder.toTxOut().datumHash).toEqual(datumHash);
     });
 
+    it('can set handle', () => {
+      outputBuilder.handle(handle);
+      expect(outputBuilder.toTxOut().handle).toEqual(handle);
+    });
+
```
mkazlauskas, line 360: Missing a test case for this new error
```suggestion
  it('attempting to set handle without handleProvider throws an error');
```

packages/core/src/Provider/HandleProvider/types.ts
```diff
@@ -0,0 +1,42 @@
...
+export type Handle = string;
+
+/**
+ * Type of addresses that can resolve from a Handle
+ */
+export type CardanoAddress =
+  | Cardano.ByronAddress
+  | Cardano.PaymentAddress
+  | Cardano.EnterpriseAddress
+  | Cardano.PointerAddress;
```
mkazlauskas, line 12: This should no longer be used, please delete
```suggestion
```

mkazlauskas (changes requested):

packages/core/src/Provider/HandleProvider/types.ts
VanessaPC: aa5eaf7df693d418c3cdb800b0e524e2e316ae28 should fix this

packages/tx-construction/test/tx-builder/TxBuilder.test.ts
VanessaPC: aa5eaf7df693d418c3cdb800b0e524e2e316ae28 should fix this

packages/cardano-services-client/src/HandleProvider/KoraLabsHandleProvider.ts
```diff
@@ -0,0 +1,102 @@
...
+      adapter,
+      baseURL: serverUrl
+    });
+    this.policyId = policyId;
+  }
+
+  async resolveHandles(args: ResolveHandlesArgs): Promise<Array<HandleResolution | null>> {
+    try {
+      const tip = await this.networkInfoProvider.ledgerTip();
+      const results = await Promise.all(
```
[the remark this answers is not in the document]
rhyslbw: Good catch @VanessaPC 💪 

> the moment a handle comes as **null**, it won't resolve further handles.

Do you mean _Not found_, i.e a `404` HTTP response, which `axios` translates to rejecting with an error? If so, how does that not:

> ...affect the implementation on the txBuilder or outputBuilder.

... if a single error would result in a rejection of the batch? Maybe you can walk me through this on a call.

packages/core/src/Provider/HandleProvider/types.ts
rhyslbw: @mkazlauskas can you point me to any prior discussion/justification on this decision? The `CardanoAddress` type is a domain expression

packages/cardano-services-client/src/HandleProvider/KoraLabsHandleProvider.ts
```diff
@@ -0,0 +1,102 @@
...
+  HandleProvider,
+  HandleResolution,
+  HealthCheckResponse,
+  NetworkInfoProvider,
+  ProviderError,
+  ProviderFailure,
+  ResolveHandlesArgs
+} from '@cardano-sdk/core';
+
+// eslint-disable-next-line import/no-extraneous-dependencies
```
rhyslbw, line 12: This eslint disable is curious. What's the reason for it?

packages/cardano-services-client/test/util.ts
```diff
@@ -22,3 +22,79 @@ export const healthCheckResponseWithState = {
...
+
+export const getAliceHandleAPIResponse = {
+  background: 'zrljm7nskakjydxlr450ktsj08zuw6aktvgfkmmyw9semrkrezryq3yd',
+  characters: 'rljm7n/23455',
+  created_slot_number: 33,
+  default_in_wallet: 'alice_default_hndle',
+  hasDatum: false,
+  hex: '0f3abbc8fc19c2e61bab6059bf8a466e6e754833a08a62a6c56fe0e78f19d9d5',
+  holder_address:
+    'addr_test1qzrljm7nskakjydxlr450ktsj08zuw6aktvgfkmmyw9semrkrezryq3ydtmkg0e7e2jvzg443h0ffzfwd09wpcxy2fuql2erwd',
```
rhyslbw, line 62:
```suggestion
    'stake1uyehkck0lajq8gr28t9uxnuvgcqrc6070x3k9r8048z8y5gh6ffgw',
```
Holder address is a _stake address_.

https://api.handle.me/swagger/#/handles/get_handles__handle_

packages/cardano-services-client/test/util.ts
```diff
@@ -22,3 +22,79 @@ export const healthCheckResponseWithState = {
...
+
+export const getBobHandleAPIResponse = {
+  background: 'zrljm7nskakjydxlr450ktsj08zuw6aktvgfkmmyw9semrkrezryq3yd',
+  characters: 'rljm7n/23455',
+  created_slot_number: 33,
+  default_in_wallet: 'bob_default_handle',
+  hasDatum: false,
+  hex: '0f3abbc8fc19c2e61bab6059bf8a466e6e754833a08a62a6c56fe0e78f19d9d5',
+  holder_address:
+    'addr_test1qzrljm7nskakjydxlr450ktsj08zuw6aktvgfkmmyw9semrkrezryq3ydtmkg0e7e2jvzg443h0ffzfwd09wpcxy2fuql2erwd',
```
rhyslbw, line 86:
```suggestion
    'stake178phkx6acpnf78fuvxn0mkew3l0fd058hzquvz7w36x4gtcccycj5',
```

packages/e2e/test/web-extension/.env.example
```diff
@@ -16,6 +16,8 @@ UTXO_PROVIDER=http
 UTXO_PROVIDER_PARAMS='{"baseUrl":"http://localhost:4000/utxo"}'
 STAKE_POOL_PROVIDER=stub
 STAKE_POOL_PROVIDER_PARAMS='{"baseUrl":"http://localhost:4000/stake-pool"}'
+HANDLE_PROVIDER=handleProvider
```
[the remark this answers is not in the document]
rhyslbw: This commit seems to be on another branch?

rhyslbw (changes requested):
Excellent work @VanessaPC 🚀 

Some minor feedback, but looks close now

packages/cardano-services-client/src/HandleProvider/KoraLabsHandleProvider.ts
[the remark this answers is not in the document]
rhyslbw: Oh, I'm guessing it's because we're only resolving a single handle at a time. If that's the case, I agree we can defer this. Please add a ticket to the backlog.

packages/cardano-services-client/src/HandleProvider/KoraLabsHandleProvider.ts
VanessaPC: https://github.com/input-output-hk/cardano-js-sdk/pull/732#discussion_r1206350277 
For this to happen, I can only disable the error.

packages/cardano-services-client/src/HandleProvider/KoraLabsHandleProvider.ts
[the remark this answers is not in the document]
VanessaPC: Yea, currently we are only using it to resolve a single handle in the OutputBuilder, so for now it's not going to break anything. But since the HandleProvider is in a separate package, it may potentially use elsewhere and it wouldn't work because it's implementation doesn't handle batch requests, where one of the handles may return a 404, but the rest should still resolve, also, because this is what we say we do on the interface, ultimately, we return an array of resolved handles. 

Ticket: https://input-output.atlassian.net/browse/LW-6778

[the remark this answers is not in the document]
rhyslbw: Thanks! The follow-up task is not critical, but we should aim to cover it off in the next sprint, since it is a bug. BTW, as this is an open source repo, please only post a reference to Jira tickets i.e `LW-6778`

packages/cardano-services-client/test/util.ts
VanessaPC: 4ea340ef522bc4fccefec25455597bc3e5b79f5d should solve this

packages/core/src/Provider/HandleProvider/types.ts
mkazlauskas: Created a placeholder ticket LW-6784 so we can get back to this later

.github/workflows/continuous-integration-e2e.yaml
```diff
@@ -6,78 +6,78 @@ env:
 
 on:
   pull_request:
-    branches: ['master']
+    branches: [ "master" ]
```
mkazlauskas: This commit should probably go to a separate PR that sets up some tool/linter that enforces consistent yaml format, as well as formats all yaml files in the repository.

mkazlauskas (changes requested):
Great work!

.github/workflows/continuous-integration-e2e.yaml
VanessaPC: It's already been fixed, I pushed one last commit. But I can create a PR with a rule.

VanessaPC: Create ticket: LW-6786

rhyslbw (approved):

iccicci (approved):
LGTM

PR #735, opened 2023-05-22
Feat/lw 6582 wallet reports delegation portfolio as cip17 pool objects

mirceahasegan:
# Context

Communicate the wallet’s stake distribution as CIP-17 pool objects.

# Proposed Solution
SDK to expose a new observable in the ObservableWallet interface: delegation.portfolio$ which emits a subset of the CIP-17 specification + stakingAddress and percentage fields.
Some of the properties may be defined as they are part of the stake pool metadata.

id: Base16 poolId
name?: string
ticker?: string
**Additional properties:**
percentage: number
pool: Cardano.StakePool
coins: actual staked value
**Removed properties from CIP-17 object**
weight

see commit history

### DIFF
--- a/packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
+++ b/packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
@@ -1,5 +1,6 @@
+import { AsyncKeyAgent } from '@cardano-sdk/key-management';
 import { Cardano, ChainHistoryProvider, EraSummary, SlotEpochCalc, createSlotEpochCalc } from '@cardano-sdk/core';
-import { DelegationTracker, TransactionsTracker } from '../types';
+import { DelegationTracker, TransactionsTracker, UtxoTracker } from '../types';
 import { Logger } from 'ts-log';
 import { Observable, combineLatest, map, tap } from 'rxjs';
 import {
@@ -17,6 +18,7 @@ import { TrackerSubject } from '@cardano-sdk/util-rxjs';
 import { TxWithEpoch } from './types';
 import { WalletStores } from '../../persistence';
 import { coldObservableProvider } from '../util';
+import { createDelegationDistributionTracker } from './DelegationDistributionTracker';
 import { transactionsWithCertificates } from './transactionCertificates';
 
 export const createBlockEpochProvider =
@@ -42,6 +44,8 @@ export interface DelegationTrackerProps {
   epoch$: Observable<Cardano.EpochNo>;
   transactionsTracker: TransactionsTracker;
   retryBackoffConfig: RetryBackoffConfig;
+  utxoTracker: UtxoTracker;
+  knownAddresses$: AsyncKeyAgent['knownAddresses$'];
   stores: WalletStores;
   internals?: {
     queryStakePoolsProvider?: ObservableStakePoolProvider;
@@ -67,7 +71,6 @@ export const certificateTransactionsWithEpochs = (
       transactions.map((tx) => ({ epoch: slotEpochCalc(tx.blockHeader.slot), tx }))
     )
   );
-
 export const createDelegationTracker = ({
   rewardAccountAddresses$,
   epoch$,
@@ -76,6 +79,8 @@ export const createDelegationTracker = ({
   transactionsTracker,
   eraSummaries$,
   stakePoolProvider,
+  knownAddresses$,
+  utxoTracker,
   stores,
   logger,
   onFatalError,
@@ -128,7 +133,11 @@ export const createDelegationTracker = ({
       transactionsInFlight$: transactionsTracker.outgoing.inFlight$
     })
   );
+  const distribution$ = new TrackerSubject(
+    createDelegationDistributionTracker({ knownAddresses$, rewardAccounts$, utxoTracker })
+  );
   return {
+    distribution$,
     rewardAccounts$,
     rewardsHistory$,
     shutdown: () => {
--- a/packages/wallet/src/services/types.ts
+++ b/packages/wallet/src/services/types.ts
@@ -1,6 +1,7 @@
 import { AsyncKeyAgent, GroupedAddress } from '@cardano-sdk/key-management';
 import { Cardano, CardanoNodeErrors, EpochRewards, TxCBOR } from '@cardano-sdk/core';
 import { Observable } from 'rxjs';
+import { Percent } from '@cardano-sdk/util';
 
 export enum TransactionFailure {
   InvalidTransaction = 'INVALID_TRANSACTION',
@@ -98,7 +99,19 @@ export interface RewardsHistory {
   lifetimeRewards: Cardano.Lovelace;
 }
 
+/** Wallet delegated stake by pool. Multiple reward accounts could be delegated to the same pool */
+export interface DelegatedStake {
+  pool: Cardano.StakePool;
+  /** How much from the total delegated stake is delegated to this pool */
+  percentage: Percent;
+  /** Absolute stake value */
+  stake: bigint;
+  /** Reward accounts delegated to this pool */
+  rewardAccounts: Cardano.RewardAccount[];
+}
+
 export interface DelegationTracker {
   rewardsHistory$: Observable<RewardsHistory>;
   rewardAccounts$: Observable<Cardano.RewardAccountInfo[]>;
+  distribution$: Observable<Map<Cardano.PoolId, DelegatedStake>>;
 }
--- a/packages/e2e/test/wallet/PersonalWallet/delegationDistribution.test.ts
+++ b/packages/e2e/test/wallet/PersonalWallet/delegationDistribution.test.ts
@@ -1,0 +1,242 @@
+import { Cardano } from '@cardano-sdk/core';
+import { DelegatedStake, PersonalWallet, createUtxoBalanceByAddressTracker } from '@cardano-sdk/wallet';
+import { MINUTE, getWallet } from '../../../src';
+import { Observable, filter, firstValueFrom, map, tap } from 'rxjs';
+import { Percent } from '@cardano-sdk/util';
+import { createLogger } from '@cardano-sdk/util-dev';
+import { firstValueFromTimed, submitAndConfirm, walletReady } from '../../util';
+import { getEnv, walletVariables } from '../../../src/environment';
+
+const env = getEnv(walletVariables);
+const logger = createLogger();
+const TEST_FUNDS = 1_000_000_000n;
+
+/** Distribute the wallet funds evenly across all its addresses */
+const distributeFunds = async (wallet: PersonalWallet) => {
+  await walletReady(wallet, 0n);
+  const addresses = await firstValueFrom(wallet.addresses$);
+  expect(addresses.length).toBeGreaterThan(1);
+
+  // Check that we have enough funds. Otherwise, fund it from wallet account at index 0
+  let { coins: totalCoins } = await firstValueFrom(wallet.balance.utxo.available$);
+
+  const coinDeficit = TEST_FUNDS - totalCoins;
+  if (coinDeficit > 10_000_000n) {
+    logger.debug(
+      `Insufficient funds in wallet account index 1. Missing ${coinDeficit}. Transferring from wallet account index 0`
+    );
+    const fundingWallet = (await getWallet({ env, idx: 0, logger, name: 'WalletAcct0', polling: { interval: 50 } }))
+      .wallet;
+    await walletReady(fundingWallet);
+    const fundingTxBuilder = fundingWallet.createTxBuilder();
+    const { tx } = await fundingTxBuilder
+      .addOutput(fundingTxBuilder.buildOutput().address(addresses[0].address).coin(coinDeficit).toTxOut())
+      .build()
+      .sign();
+    await submitAndConfirm(fundingWallet, tx);
+    await walletReady(wallet);
+    totalCoins = (await firstValueFrom(wallet.balance.utxo.available$)).coins;
+  }
+
+  const coinsPerAddress = totalCoins / BigInt(addresses.length);
+
+  const txBuilder = wallet.createTxBuilder();
+
+  logger.debug(`Sending ${coinsPerAddress} to the ${addresses.length - 1} derived addresses`);
+  // The first one was generated when the wallet was created.
+  for (let i = 1; i < addresses.length; ++i) {
+    const derivedAddress = addresses[i];
+    txBuilder.addOutput(txBuilder.buildOutput().address(derivedAddress.address).coin(coinsPerAddress).toTxOut());
+  }
+
+  const { tx: signedTx } = await txBuilder.build().sign();
+  await submitAndConfirm(wallet, signedTx);
+};
+
+/** await for rewardAccounts$ to be registered, unregistered, as defined in states   */
+const rewardAccountStatuses = async (
+  rewardAccounts$: Observable<Cardano.RewardAccountInfo[]>,
+  statuses: Cardano.StakeKeyStatus[]
+) =>
+  firstValueFromTimed(
+    rewardAccounts$.pipe(
+      tap((accts) => accts.map(({ address, keyStatus }) => logger.debug(address, keyStatus))),
+      map((accts) => accts.map(({ keyStatus }) => keyStatus)),
+      filter((statusArr) => statusArr.every((s) => statuses.includes(s)))
+    ),
+    `Timeout waiting for all reward accounts stake keys to be one of ${statuses.join('|')}`,
+    MINUTE
+  );
+
+/** Create stakeKey deregistration transaction for all reward accounts */
+const deregisterAllStakeKeys = async (wallet: PersonalWallet): Promise<void> => {
+  const txBuilder = wallet.createTxBuilder();
+  txBuilder.delegate();
+  const { tx: deregTx } = await txBuilder.build().sign();
+  await submitAndConfirm(wallet, deregTx);
+
+  await rewardAccountStatuses(wallet.delegation.rewardAccounts$, [
+    Cardano.StakeKeyStatus.Unregistered,
+    Cardano.StakeKeyStatus.Unregistering
+  ]);
+  logger.debug('Deregistered all stake keys');
+};
+
+const createStakeKeyRegistrationCert = (rewardAccount: Cardano.RewardAccount): Cardano.Certificate => ({
+  __typename: Cardano.CertificateType.StakeKeyRegistration,
+  stakeKeyHash: Cardano.RewardAccount.toHash(rewardAccount)
+});
+
+const createDelegationCert = (rewardAccount: Cardano.RewardAccount, poolId: Cardano.PoolId): Cardano.Certificate => ({
+  __typename: Cardano.CertificateType.StakeDelegation,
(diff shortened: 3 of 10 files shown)


mirceahasegan:
I forgot to remove "WIP" from commit message [6f05aa7](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/6f05aa717ae28a6333f307d8039b148d655085a8)

packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
```diff
@@ -68,6 +90,92 @@ export const certificateTransactionsWithEpochs = (
...
+  rewardAccounts$,
+  knownAddresses$,
+  utxoTracker
+}: DelegationPortfolioTrackerProps): Observable<Cip17Pool[]> => {
+  const balanceAllUtxosTracker = createBalanceByAddressTracker(utxoTracker);
+
+  const delegatedAccounts$ = rewardAccounts$.pipe(
+    map((rewardsAccounts) =>
+      rewardsAccounts.filter(
+        (account) => account.keyStatus === Cardano.StakeKeyStatus.Registered && account.delegatee?.nextNextEpoch
```
mirceahasegan, line 103: REVIEW: as far as I could understand, nextNextEpoch will always reflect the latest delegation changes, even if they will be applied 2 epochs from now.
I chose this property as the source of truth when building the cip17 object

packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
```diff
@@ -68,6 +90,92 @@ export const certificateTransactionsWithEpochs = (
...
+    ),
+    // Merge rewardAccounts delegating to the same pool
+    map((cip17Pools) =>
+      _map(
+        _groupBy(cip17Pools, ({ id }) => id),
+        (pools): Cip17Pool =>
+          pools.reduce((mergedPool, pool) => ({
+            ...pool,
+            coins: pool.coins + mergedPool.coins || 0n,
+            percentage: pool.percentage + mergedPool.percentage || 0
```
mirceahasegan, line 162: I just realized i should delete this line. Percentages are calculated in the next step

packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
mirceahasegan: The most flexible solution seems to be #3.

packages/util/src/Percentage.ts
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Dividing bigints is always rounded down, creating this problem:
+ * sum([...percentage(part)]) != percentage(sum([...parts]))
+ * This function will be distribute the difference to the highest percentage value.
+ *
+ * @param parts array bigint values
+ * @param total optional param to allow sum(parts) to be smaller than the total
+ * @returns array of integer percentages
+ */
+export const calcRoundedPercentages = (parts: bigint[], total: bigint = BigIntMath.sum(parts)): number[] => {
```
mkazlauskas, line 15: Would be great to be consistent when representing percentage. This type represents percentage as floating point though.
```suggestion
export const calcRoundedPercentages = (parts: bigint[], total: bigint = BigIntMath.sum(parts)): Cardano.Percent[] => {
```

packages/util/src/Percentage.ts
```diff
@@ -0,0 +1,48 @@
+import { BigIntMath } from './BigIntMath';
+
+/**
+ * Calculates the rounded percentages for each part from {@link total}.
+ * When total is omitted, it is assumed that the total is the sum of the parts.
+ *
+ * Dividing bigints is always rounded down, creating this problem:
+ * sum([...percentage(part)]) != percentage(sum([...parts]))
+ * This function will be distribute the difference to the highest percentage value.
```
mkazlauskas, line 9: Consider changing the strategy here: for an edge case where only **some** coins are delegated I would rather see 1% than 0%.
```suggestion
 * This function will be distribute the difference to the lowest percentage value.
```

packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
```diff
@@ -68,6 +90,92 @@ export const certificateTransactionsWithEpochs = (
...
+
+  const hydratedDelegatedAccounts$ = delegatedAccounts$.pipe(
+    withLatestFrom(knownAddresses$),
+    map(([delegatedAccounts, knownAddresses]) =>
+      delegatedAccounts.map((delegatedAccount) => {
+        const groupedAddress = knownAddresses.find(
+          (knownAddr) => knownAddr.rewardAccount === delegatedAccount.address
+        )!;
+        return {
+          balance: createBalanceByAddressTracker(utxoTracker, groupedAddress.address),
```
mirceahasegan, line 116: createBalanceByAddressTracker calculates only the utxo balance.
The rewards must be included in this calculation

packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
mirceahasegan: [4a1e70c](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/4a1e70c437f0f710b0f098e7831f52bfb542c58c)

packages/util/src/Percentage.ts
mirceahasegan: Updated code to use floating point percentages in:
- [98115e8](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/98115e844dfaac0fa1128b9c7b84acb6d9d53c90)
- [73358ab](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/73358ab247f8b3c4ffe5dff72116e433f3503825)
- [4a1e70c](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/4a1e70c437f0f710b0f098e7831f52bfb542c58c)
- [a2b9155](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/a2b9155d0207557807c732290fc428826f1a7bd6)

packages/util/src/Percentage.ts
mirceahasegan: No longer valid. Percentages are now floating point numbers between 0 and 1, of type `Percent`

packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
mirceahasegan: Done here: [da7fc70](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/da7fc70df3a90dfce49211ecb21ae740b30fed39)

packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
mirceahasegan: Leave it as is for now

mkazlauskas (dismissed):
Looking good, please squash for final reviews

packages/util/test/Percent.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  it('one part is zero, total is implicitly zero, so it takes 100% of the total', () => {
+    expect(calcPercentages([0])).toEqual([1]);
+  });
+
+  it('multiple parts are zero, total is implicitly zero, percent is distributed evenly', () => {
+    expect(calcPercentages([0, 0, 0, 0])).toEqual([0.25, 0.25, 0.25, 0.25]);
+  });
+
+  it('total is adjusted to equal at least as much as the sum of the parts', () => {
```
mkazlauskas, line 31: This seems a little odd to me. Is there a use case? If not, perhaps throwing would be better?

packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
```diff
@@ -68,6 +90,95 @@ export const certificateTransactionsWithEpochs = (
...
```
mkazlauskas, line 116: There could potentially be multiple addresses for a single reward account
```suggestion
        const groupedAddresses = knownAddresses.filter(
          (knownAddr) => knownAddr.rewardAccount === delegatedAccount.address
        )!;
        return {
          balance: createBalanceByAddressesTracker(utxoTracker, groupedAddresses),
```
Alternatively could do `createBalanceByRewardAccountTracker`

packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
```diff
@@ -68,6 +90,95 @@ export const certificateTransactionsWithEpochs = (
...
+  const delegatedAccounts$ = rewardAccounts$.pipe(
+    map((rewardsAccounts) =>
+      rewardsAccounts.filter(
+        (account) => account.keyStatus === Cardano.StakeKeyStatus.Registered && account.delegatee?.nextNextEpoch
+      )
+    )
+  );
+
+  const hydratedDelegatedAccounts$ = delegatedAccounts$.pipe(
+    withLatestFrom(knownAddresses$),
```
mkazlauskas, line 109: I think this is technically a race condition
```suggestion
    blockingWithLatestFrom(knownAddresses$),
```

packages/wallet/test/services/BalanceTracker.test.ts
```diff
@@ -47,4 +47,46 @@ describe('createBalanceTracker', () => {
...
+  it('createBalanceByAddressTracker returns balance filtered by address', () => {
+    const address = utxo[0][1].address;
+    const utxoMultipleAddresses = utxo.slice(0, 3);
+
+    // Change the first utxo address so it's filtered out
+    utxoMultipleAddresses[0][1].address = Cardano.PaymentAddress('addr_test1qzs0umu0s2ammmpw0hea0w2crtcymdjvvlqngpgqy76gpfnuzcjqw982pcftgx53fu5527z2cj2tkx2h8ux2vxsg475qp3y3vz');
+    createTestScheduler().run(({ hot, expectObservable }) => {
+      const available$ = hot(       'a', { a: utxoMultipleAddresses }) as unknown as BehaviorObservable<Cardano.Utxo[]>;
+      const total$ = hot(           'a', { a: utxoMultipleAddresses }) as unknown as BehaviorObservable<Cardano.Utxo[]>;
+      const unspendable$ = hot(     'a', { a: utxoMultipleAddresses }) as unknown as BehaviorObservable<Cardano.Utxo[]>;
```
mkazlauskas, line 60: A distinctUntilChanged is probably a good idea here, so that it doesn't emit if irrelevant utxo change
```suggestion
      const available$ = hot(       'aa', { a: utxoMultipleAddresses }) as unknown as BehaviorObservable<Cardano.Utxo[]>;
      const total$ = hot(           'aa', { a: utxoMultipleAddresses }) as unknown as BehaviorObservable<Cardano.Utxo[]>;
      const unspendable$ = hot(     'aa', { a: utxoMultipleAddresses }) as unknown as BehaviorObservable<Cardano.Utxo[]>;
```

mkazlauskas (changes requested):

packages/tx-construction/src/types.ts
```diff
@@ -36,3 +37,14 @@ export interface InitializeTxProps {
 export interface InitializeTxPropsValidationResult {
   minimumCoinQuantities: MinimumCoinQuantityPerOutput;
 }
+
+/** Pool delegation profile inspired from CIP-17. */
```
rhyslbw, line 41: I'd prefer to see a precise type, as think of the refactor that will be required when it comes time to implement an exporting function, which is a critical aspect of the standard. How about this:

``` typescript
export interface Cip17Pool {
  id: Cardano.PoolId;
  name?: string;
  ticker?: Cardano.StakePoolMetadata['ticker'];
  /** Value between 0 and 1. Example: 0.3344 */
  weight: number;
}
```
 ... which is then transformed to 
 
 ``` typescript
 export interface DelegationProfilePool extends Cip17Pool {
  pool: Cardano.StakePool;
  /** Value between 0 and 1. Example: 0.3344 is 33.44% */
  percentage: Percent;
  coins: bigint;
}
```

`Cip17Pool` shouldn't be scoped to transaction construction, but `DelegationProfilePool` can be if that's the only relevant context.

**Review by rhyslbw — CHANGES_REQUESTED** [review]
Looks great @mirceahasegan 🚀

File: packages/util/test/Percent.test.ts (hunk 6 above)
Comment (mirceahasegan) [reply packages/util/test/Percent.test.ts:31]: No use case comes to mind. I added it to guard against abnormal inputs.
It could happen if the parts balance calculator emits before the total calculator, depending on how the observables are combined, that there is an invocation to calculate percentages where the total hasn't caught up with the balances.
If that's alright, I would be a bit defensive in this case.

File: packages/wallet/src/services/DelegationTracker/DelegationTracker.ts (hunk 7 above)
Comment (mirceahasegan) [reply packages/wallet/src/services/DelegationTracker/DelegationTracker.ts:116]: great catch! thanks

File: packages/wallet/src/services/DelegationTracker/DelegationTracker.ts (hunk 8 above)
Comment (mirceahasegan) [reply packages/wallet/src/services/DelegationTracker/DelegationTracker.ts:109]: I was counting on knownAddresses to be a TrackingSubject, which would make it repeat any previous value it emitted

File: packages/wallet/src/services/DelegationTracker/DelegationTracker.ts (hunk 7 above)
Comment (mirceahasegan) [reply packages/wallet/src/services/DelegationTracker/DelegationTracker.ts]: Fixed in [4bb58e4](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/4bb58e415d7e58c437939064515414896695ba6a) & [b59a6c0](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/b59a6c077c5e4b9dfb7f77526cbd6b97b8204706)

File: packages/wallet/test/services/BalanceTracker.test.ts (hunk 9 above)
Comment (mirceahasegan) [reply packages/wallet/test/services/BalanceTracker.test.ts]: Fixed in [4bb58e4](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/4bb58e415d7e58c437939064515414896695ba6a)

File: packages/wallet/src/services/DelegationTracker/DelegationTracker.ts (hunk 8 above)
Comment (mkazlauskas) [reply packages/wallet/src/services/DelegationTracker/DelegationTracker.ts:109]: Right, rewardAccounts$ won't emit until there are knownAddresses$. That's an implementation detail, that's what I meant by 'technically'. Fine to leave as-is.

File: packages/tx-construction/src/types.ts (hunk 10 above)
Comment (mkazlauskas) [reply packages/tx-construction/src/types.ts:41]: I think the problem here is that we're trying to use a single type for 2 different things:
1. cip17 is delegation preferences - provided by the application, as configured by the user.
2. PersonalWallet reports current delegation portfolio state - this is where `weights` don't work and we should use `coins` instead

File: packages/wallet/src/services/DelegationTracker/DelegationTracker.ts (hunk 8 above)
Comment (mirceahasegan) [reply packages/wallet/src/services/DelegationTracker/DelegationTracker.ts:109]: Todo: do not export `createDelegationPortfolioTracker`

**Comment by mirceahasegan** [comment]
Commits up to [d268466](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/d2684660f4c57f94784601dfbfcedc7cda31cb54) are preexisting, but were rebased on master.
New commits:
- [d268466](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/d2684660f4c57f94784601dfbfcedc7cda31cb54) & [a9da69b](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/a9da69bb59c9452f9aa35ac7266c3848b1a9ae0f) Updates to fix conflicts with new code from `master`
- [78fb90f](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/78fb90f4eb7a044097ea5f68fb734731283a8b6d) & [63e7597](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/63e7597dea79efdd27dcf138d894f2b6801d4276): remove Cip17 type and add `StakeByPool` and `Personal.wallet.delegation.stakeDistribution$: Observable<StakeByPool[]>`

File: packages/tx-construction/src/types.ts (hunk 10 above)
Comment (mirceahasegan) [reply packages/tx-construction/src/types.ts]: [78fb90f](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/78fb90f4eb7a044097ea5f68fb734731283a8b6d) & [63e7597](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/63e7597dea79efdd27dcf138d894f2b6801d4276): `Cip17Pool` is replaced by `StakeByPool`

File: packages/wallet/src/services/DelegationTracker/DelegationTracker.ts (hunk 8 above)
Comment (mirceahasegan) [reply packages/wallet/src/services/DelegationTracker/DelegationTracker.ts]: [78fb90f](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/78fb90f4eb7a044097ea5f68fb734731283a8b6d) & [63e7597](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/63e7597dea79efdd27dcf138d894f2b6801d4276): symbol is now exported only for unit testing.

File: packages/wallet/test/services/DelegationTracker/DelegationTracker.test.ts [hunk 11]
```diff
@@ -135,7 +131,7 @@ describe('DelegationTracker', () => {
     });
   });
 
-  describe('createDelegationPortfolioTracker', () => {
+  describe('createStakeDistributionTracker', () => {
```
mkazlauskas, line 134: Let's move this to `StakeDistributionTracker.test.ts` to match `src/` structure

mkazlauskas (changes requested):

packages/wallet/src/services/types.ts
```diff
@@ -98,7 +99,19 @@ export interface RewardsHistory {
   lifetimeRewards: Cardano.Lovelace;
 }
 
+/** Wallet delegated stake by pool. Multiple reward accounts could be delegated to the same pool */
+export interface StakeByPool {
+  pool: Cardano.StakePool;
+  /** How much from the total delegated stake is delegated to this pool */
+  percentage: Percent;
+  /** Absolute stake value */
+  value: bigint;
```
mkazlauskas, line 108: `Value` has a specific meaning - structure that has both coins and asset
```suggestion
  stake: bigint;
```

mkazlauskas (changes requested):

packages/wallet/src/services/types.ts
mirceahasegan: Updated here: [d4e79ed](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/d4e79ed6e4d7cc95b1ceaeb19ad830bfe2c058c1)

packages/tx-construction/src/types.ts
mirceahasegan: Updated types here: [d4e79ed](https://github.com/input-output-hk/cardano-js-sdk/pull/735/commits/d4e79ed6e4d7cc95b1ceaeb19ad830bfe2c058c1)

mkazlauskas (dismissed):
Great work! :rocket:

rhyslbw (approved):
Amazing work @mirceahasegan 🧙

mkazlauskas (approved):
:rocket:

PR #737, opened 2023-05-23
refactor: restore `AssetInfo.quantity` as depreciated field for backwards compat

rhyslbw:
# Context
A compatibility mapping was made in 6e28df412797974b8ce6f6deb0c3346ff5938a05, however it's only sufficient for newer clients connecting to the older API, but was a breaking change, and didn't cover backwards compatibility with older clients.

# Proposed Solution
- Restore the field as depreciated
- Remove the client-side compatibility mapping, which is no longer required.


iccicci (approved):
LGTM

mkazlauskas (approved):

PR #734, opened 2023-05-22
feat(hardware-trezor): add dedicated Trezor package

DominikGuzei:
# Context

JIRA card: https://input-output.atlassian.net/browse/LW-6514

This PR extracts Trezor integration from `key-management` into dedicated `hardware-trezor` package similar to the existing `hardware-ledger`, with fine-grained, unit-tested data transformers.

# Proposed Solution

- [x] New package called @cardano-sdk/hardware-trezor is created
- [x] A new `TrezorKeyAgent` is implemented in the new package using the fine-grained data transformers
- [x] All new Trezor data transformers are implemented and covered by unit tests
- [ ] ~apply bug fix similar to https://github.com/input-output-hk/cardano-js-sdk/pull/745~ descoped
- [x] The new `TrezorKeyAgent` is covered by unit and integration tests
- [x] The new `TrezorKeyAgent` is tested in Lace

PR #742, opened 2023-05-26
chore(hardware-trezor): finish trezor tx input mappers

DominikGuzei:
# Context

This PR finishes the implementation of mapping Cardano tx inputs to Trezor inputs.


packages/hardware-trezor/test/transformers/txIn.test.ts
```diff
@@ -20,4 +20,21 @@ describe('tx-inputs', () => {
...
+  describe('mapTxIns', () => {
+    it('can map a a set of TxIns', async () => {
+      const txIns = await mapTxIns([txIn, txIn, txIn], CONTEXT_WITH_KNOWN_ADDRESSES);
+
+      expect(txIns.length).toEqual(3);
+
+      for (const input of txIns) {
+        expect(input).toEqual({
+          path: knownAddressKeyPath,
+          prev_hash: Buffer.from(txIn.txId).toString('hex'),
```
mkazlauskas, line 32: Isn't this already correct format?
```suggestion
          prev_hash: txIn.txId,
```

packages/hardware-trezor/src/transformers/tx.ts
```diff
@@ -12,9 +13,10 @@ import { util as depricated } from '@cardano-sdk/key-management';
  */
 const ledgerTxTransformer = async (
```
mkazlauskas, line 14:
```suggestion
const trezorTxTransformer = async (
```

mkazlauskas (changes requested):

packages/hardware-trezor/test/transformers/txIn.test.ts
DominikGuzei: thanks @mkazlauskas 🙏  I still need to learn a lot about the correct formats in the SDK ;)

packages/hardware-trezor/test/transformers/txIn.test.ts
```diff
@@ -20,4 +20,21 @@ describe('tx-inputs', () => {
...
     });
   });
+  describe('mapTxIns', () => {
+    it('can map a a set of TxIns', async () => {
+      const txIns = await mapTxIns([txIn, txIn, txIn], CONTEXT_WITH_KNOWN_ADDRESSES);
+
+      expect(txIns.length).toEqual(3);
+
+      for (const input of txIns) {
+        expect(input).toEqual({
```
marcin-mazurek, line 30: Just a thought, can we have one assertion for entire array instead of going through the txIns in a for-of loop? It could make the test a bit easier to read (it will show the exact structure the function will output)

marcin-mazurek (changes requested):

packages/hardware-trezor/src/transformers/txIn.ts
```diff
@@ -8,11 +8,21 @@ import { resolvePaymentKeyPathForTxIn } from './keyPaths';
...
  * input transaction format using the given trezor input resolver.
  */
-export const toTrezorTxIn: Transform<Cardano.TxIn, Promise<trezor.CardanoInput>, TrezorTxTransformerContext> = async (
+export const toTrezorTxIn: Transform<Cardano.TxIn, Promise<Trezor.CardanoInput>, TrezorTxTransformerContext> = async (
   txIn,
   context?
 ) => ({
   path: (await resolvePaymentKeyPathForTxIn(txIn, context)) ?? undefined,
-  prev_hash: Buffer.from(txIn.txId).toString('hex'),
+  prev_hash: txIn.txId,
```
tomislavhoracek, line 16: I'm not sure this is correct

mkazlauskas: I suggest applying [this change](https://github.com/input-output-hk/cardano-js-sdk/pull/734#discussion_r1213916043) and then running a test that signs any tx with TrezorKeyAgent. If this is not correct then it'll throw

DominikGuzei: @mkazlauskas @tomislavhoracek im getting this error in the "unit tests" for trezor in the "wallet" package, could this be related?
grafik

DominikGuzei: hm ok no, getting the same error when switching back to the old format

DominikGuzei:
All tests in `wallet` package are now passing after I refactored the new `TrezorKeyAgent` to extend the old one and only override the minimal surface for us to hook-in the new mappers. This way we can ensure that the new code behaves exactly like the old.
grafik

tomislavhoracek (commented):
@DominikGuzei Looks good, I left just one minor comment

packages/hardware-trezor/src/TrezorKeyAgent.ts
```diff
@@ -151,10 +34,6 @@ export class TrezorKeyAgent extends KeyAgentBase {
       }
 
       const signedData = result.payload;
-      if (signedData.hash !== tx.hash) {
-        throw new errors.HwMappingError('Trezor computed a different transaction id');
-      }
```
marcin-mazurek, line 156: Just saw this comment https://github.com/input-output-hk/cardano-js-sdk/pull/734#discussion_r1213916043 - did you remove this on purpose?

DominikGuzei: @marcin-mazurek yes, because if you add it everything is broken, but this is not "our" fault in this PR (the existing Trezor implementation is broken)

marcin-mazurek: Understood, as discussed on Slack - let's raise a separate task for it

mkazlauskas: Consider skipping the tests instead (keeping the tx hash check)

PR #743, opened 2023-05-31
Feat/lw 6704 add greedy input selector

AngelCastilloB:
# Context

We need to add an input selection strategy that allow us to partition the existing UTXO set of a wallet given a set of proportions.

# Proposed Solution

Implement a greedy input selection that will use up all the outputs in the UTXO set to account for outputs and distribute the remaining balance following the given set of proportions. If no outputs are given, the input selection will allocate 100% of the balance to the change distribution.

To splits the change proportionally between the given addresses. This algorithm makes the best effort to be as accurate as possible in distributing the amounts, however, due to rounding there may be a small error in the final distribution, I.E 8 lovelace divided in three equal parts will yield 3, 3, 2 lovelace with an error of 33% in the last change output as lovelaces can't be further subdivided (The error is marginal for large/normal amounts of lovelace).

While lovelaces will be split according to the given distribution, native assets will use a different heuristic. The algorithm will try to add all native assets to the UTXO with the most coins in the change outputs, if they don't 'fit', it will spill over the extra native assets to the next change output (and so on). 

REMARK: This this algorithm always selects all UTXOs, if the wallet has too many UTXOs to fit into a single transaction, the final selection will yield an invalid transaction as it will exceed that maximum allowed transaction size.

This algorithm wont work if a wallet:
- Have too many UTXOs to fit into a single transaction.
- Is at max capacity for holding native assets due to minCoinRequired per UTXO restrictions (this also applies to regular input selection under the same circumstances if the user tries to send only ADA and doesn't remove any native asset tho).

- The input selectors now return change as UTXOs rather than value.
- RoundRobinRandomImprove now takes a callback that returns a promise with the change address to be used.
- Added a new input selection algorithm: GreedySelector.


rhyslbw (commented):
Really nice work @AngelCastilloB 😎 

> Have too many UTXOs to fit into a single transaction.

What's the mitigation for this scenario? Have we considered capping the inputs, so the result produces a valid, but incomplete distribution, which is at least not a failure. With this approach the whole set could be redistributed with consecutive transactions until all inputs can fit into a single tx.

rhyslbw:
> splitting the UTXO set into several subsets that can fit into transactions that then would need to be submitted 

If you're confident this can be easily managed by the callee, then I'm happy to see an integration tests added later, otherwise, consider adding a unit test to demonstrate this scenario.

packages/tx-construction/src/tx-builder/initializeTx.ts
```diff
@@ -11,24 +11,23 @@ import { firstValueFrom } from 'rxjs';
...
+  const [tip, genesisParameters, protocolParameters, addresses, rewardAccounts, utxo] = await Promise.all([
+    txBuilderProviders.tip(),
+    txBuilderProviders.genesisParameters(),
+    txBuilderProviders.protocolParameters(),
+    firstValueFrom(keyAgent.knownAddresses$),
+    txBuilderProviders.rewardAccounts(),
+    txBuilderProviders.utxoAvailable()
+  ]);
+
+  inputSelector = inputSelector
```
mirceahasegan, line 25: Alternative way to write this with nullish coalescing:
```diff
-  inputSelector = inputSelector
-    ? inputSelector
-    : roundRobinRandomImprove({
-        getChangeAddress: async () => addresses[0].address
-      });
+  inputSelector = inputSelector ?? roundRobinRandomImprove({
+        getChangeAddress: async () => addresses[0].address
+      })

**Review by mirceahasegan — APPROVED** [review]
Fantastic work, @AngelCastilloB ! :1st_place_medal:

File: packages/input-selection/src/GreedySelection/GreedyInputSelector.ts [hunk 2]
```diff
@@ -0,0 +1,192 @@
+/* eslint-disable max-params */
+import { AssetId } from '@cardano-sdk/core/dist/cjs/Cardano';
```
mkazlauskas, line 2: Use `Cardano.AssetId` instead
```suggestion
```

mkazlauskas (approved):
Great work! :muscle: Please squash for final reviews

mkazlauskas (approved):
Great work! :muscle:

rhyslbw (approved):
Great work @AngelCastilloB 🧙 !

PR #745, opened 2023-06-02
Lw 6822 in memory key agent should take into account required signers

AngelCastilloB:
# Context

The InMemoryKeyAgent is ignoring the requireSigners field of the transaction when determining which key to use to sign the transaction.

# Proposed Solution

Make the InMemoryKeyAgent search for the keys specified in the requireSigners field.

- Fixed an issue in the InMemoryKeyAgent which was preventing added signers to the requireSigners to be consider for signing.
- Fixed an issue in the LedgerKeyAgent that was preventing stake key hashes in the requiredSigners field to be map correctly

### DIFF
--- a/packages/key-management/src/util/ownSignatureKeyPaths.ts
+++ b/packages/key-management/src/util/ownSignatureKeyPaths.ts
@@ -1,8 +1,11 @@
+import * as Crypto from '@cardano-sdk/crypto';
 import { AccountKeyDerivationPath, GroupedAddress } from '../types';
 import { Cardano } from '@cardano-sdk/core';
 import { isNotNil } from '@cardano-sdk/util';
+import isEqual from 'lodash/isEqual';
 import uniq from 'lodash/uniq';
 import uniqBy from 'lodash/uniqBy';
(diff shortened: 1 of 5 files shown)


packages/key-management/src/util/ownSignatureKeyPaths.ts
```diff
@@ -82,5 +116,9 @@ export const ownSignatureKeyPaths = async (
     ).filter(isNotNil)
   ).map(({ type, index }) => ({ index, role: Number(type) }));
 
-  return [...paymentKeyPaths, ...getStakingKeyPaths(knownAddresses, txBody)];
+  return [
```
mkazlauskas, line 119: Might be already returning required signers key paths as payment or stake keys
```suggestion
  return uniqBy([
```

mkazlauskas (changes requested):

mkazlauskas (approved):
Nice work! :detective:

rhyslbw (approved):
Nice work @AngelCastilloB 💪

PR #750, opened 2023-06-05
Handle server and client providers

iccicci:
# Context

We need to make the SDK able to serve ADA handles data.

# Proposed Solution

- Added `handleHttpProvider`.
- Added `TypeOrmHandleProvider`.
- Added `HandleServiceDependencies`.

- Improved the error middleware to log unknown errors (required to detect `oneApi` related errors).
- Refactored `copy-assets.sh` to copy all the `OpenApi` spec files.
- Added env vars to give the ability to override `cardano-node` log in docker compose infrastructures.
- Fixed dev docker compose with new projector services.
- Improved `getEntities` to get all entities traversing entities dependencies.
- Some refactors in _provider server_ to generalize `TypeOrmProviders` initialization.
- Added the `handle` database to _server provider_.
- Updated the fixtures generation script to reflect the new _one db for each projector_ structure.
- Refactor on `HttpProviderConfigPaths` and hoisted it into `core` package to enforce the types when using `Provider`s both in client and server sides.

# Important notes

The `TypeOrmHandleProvider` integration test is not complete because the fixtures generation procedure still doesn't produces the data required to run it (i.e. some ADA handles projected as records in `handle` table). Follow up: LW-7051

### DIFF
--- a/packages/cardano-services-client/test/HandleProvider/handleHttpProvider.test.ts
+++ b/packages/cardano-services-client/test/HandleProvider/handleHttpProvider.test.ts
@@ -1,0 +1,29 @@
+import { handleHttpProvider } from '../../src';
+import { logger } from '@cardano-sdk/util-dev';
+import MockAdapter from 'axios-mock-adapter';
+import axios from 'axios';
+
+const config = { baseUrl: 'http://some-hostname:3000/handle', logger };
+
+describe('handleHttpProvider', () => {
+  let axiosMock: MockAdapter;
+
+  beforeAll(() => {
+    axiosMock = new MockAdapter(axios);
+  });
+
+  afterEach(() => {
+    axiosMock.reset();
+  });
+
+  afterAll(() => {
+    axiosMock.restore();
+  });
+
+  test("resolve handles doesn't throw when called with an empty array of handles", async () => {
+    const provider = handleHttpProvider(config);
+
+    axiosMock.onPost().replyOnce(200, []);
+    await expect(provider.resolveHandles({ handles: [] })).resolves.toEqual([]);
+  });
+});
--- a/packages/cardano-services/copy-assets.sh
+++ b/packages/cardano-services/copy-assets.sh
@@ -1,25 +1,20 @@
- #!/usr/bin/env bash
+#!/usr/bin/env bash
 
 # TODO: run this with 'shx' or replace this with some cross-platform script
 
+# TODO: uncomment esm lines when ESM builds are enabled for this package
+
 cp ./package.json ./dist/cjs/original-package.json
-
-cp ./src/Http/openApi.json ./dist/cjs/Http/openApi.json
-cp ./src/Asset/openApi.json ./dist/cjs/Asset/openApi.json
-cp ./src/ChainHistory/openApi.json ./dist/cjs/ChainHistory/openApi.json
-cp ./src/NetworkInfo/openApi.json ./dist/cjs/NetworkInfo/openApi.json
-cp ./src/Rewards/openApi.json ./dist/cjs/Rewards/openApi.json
-cp ./src/StakePool/openApi.json ./dist/cjs/StakePool/openApi.json
-cp ./src/TxSubmit/openApi.json ./dist/cjs/TxSubmit/openApi.json
-cp ./src/Utxo/openApi.json ./dist/cjs/Utxo/openApi.json
-cp -R ./src/StakePool/HttpStakePoolMetadata/schemas ./dist/cjs/StakePool/HttpStakePoolMetadata/schemas
-
-# TODO: uncomment when ESM builds are enabled for this package
-# cp ./src/StakePool/openApi.json ./dist/esm/StakePool/openApi.json
-# cp ./src/TxSubmit/openApi.json ./dist/esm/TxSubmit/openApi.json
-# cp ./src/Utxo/openApi.json ./dist/esm/Utxo/openApi.json
-# cp ./src/ChainHistory/openApi.json ./dist/esm/ChainHistory/openApi.json
-# cp ./src/NetworkInfo/openApi.json ./dist/esm/NetworkInfo/openApi.json
-# cp ./src/Rewards/openApi.json ./dist/esm/Rewards/openApi.json
-
 # cp ./package.json ./dist/esm/original-package.json
+
+for i in `ls ./src` ; do
+  SRC=./src/$i/openApi.json
+
+  if [ -f $SRC ] ; then
+    cp $SRC ./dist/cjs/$i/openApi.json
+    # cp $SRC ./dist/esm/$i/openApi.json
+  fi
+done
+
+cp -R ./src/StakePool/HttpStakePoolMetadata/schemas ./dist/cjs/StakePool/HttpStakePoolMetadata/schemas
+# cp -R ./src/StakePool/HttpStakePoolMetadata/schemas ./dist/esm/StakePool/HttpStakePoolMetadata/schemas
--- a/packages/cardano-services/src/Handle/HandleHttpService.ts
+++ b/packages/cardano-services/src/Handle/HandleHttpService.ts
@@ -1,4 +1,4 @@
-import { HandleProvider } from '@cardano-sdk/core';
+import { HandleProvider, handleProviderPaths } from '@cardano-sdk/core';
 import { HttpService } from '../Http';
 import { Logger } from 'ts-log';
 import { Router } from 'express';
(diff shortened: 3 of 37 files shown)


packages/cardano-services-client/test/HandleProvider/handleProvider.test.test.ts
```diff
@@ -0,0 +1,29 @@
+import { handleHttpProvider } from '../../src';
```
VanessaPC, line 1: I think the file extension has one too many `.test`

iccicci: 3e3a6b1a1 should solve this

packages/cardano-services/src/Handle/TypeOrmHandleProvider.ts
```diff
@@ -0,0 +1,46 @@
...
+import { TypeormProvider, TypeormProviderDependencies } from '../util/TypeormProvider';
+
+export type TypeOrmHandleProviderDependencies = TypeormProviderDependencies;
+
+const handleFields = ['cardanoAddress', 'handle', 'hasDatum', 'policyId'] as const;
+
+type HandleFields = typeof handleFields[number];
+type PartialHandleEntity = Required<Pick<HandleEntity, HandleFields>>;
+
+const handleSelect = <{ [k in HandleFields]: true }>Object.fromEntries(handleFields.map((_) => [_, true]));
```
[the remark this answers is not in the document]
iccicci: About `PartialHandleEntity`, c7b790c02 should solve this

About `handleSelect`:
- splitting it in multiple lines breaks some SDK eslint rules
- the return type of `fromEntries<true>()` is `{ [k: string]: true; }` it would require a type cast anyway

I searched for some ways to make the return type of `fromEntries()` a bit more strict, but all the solutions I found require some complex type operations... I thought that a simple type cast is a good compromise between readability, simplicity and strictness.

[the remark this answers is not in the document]
VanessaPC: yea, it's a good balance then

packages/cardano-services/test/Handle/TypeOrmHandleProvider.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    await provider.initialize();
+    await provider.start();
+  });
+
+  afterAll(async () => {
+    jest.restoreAllMocks();
+
+    await provider.shutdown();
+  });
+
```
mkazlauskas, line 25: From adahandle docs:
> When querying a Handle, you should never allow an empty string. An empty string will return a valid address, but it is not a valid Handle.
```suggestion
  it('does not resolve an empty string handle');
```

mkazlauskas (changes requested):

rhyslbw (commented):
Ignoring the changes I know are coming from the other branch, this is looking good @iccicci

packages/cardano-services/test/Handle/TypeOrmHandleProvider.test.ts
iccicci: 8e60fbeea should solve this

packages/cardano-services-client/test/HandleProvider/handleHttpProvider.test.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  afterEach(() => {
+    axiosMock.reset();
+  });
+
+  afterAll(() => {
+    axiosMock.restore();
+  });
+
+  test("resolve handles doesn't throw", async () => {
```
VanessaPC, line 23: 'Resolve handles doesn't throw an error when handles come empty'

packages/cardano-services/copy-assets.sh
```diff
@@ -1,25 +1,20 @@
...
-cp ./src/Http/openApi.json ./dist/cjs/Http/openApi.json
-cp ./src/Asset/openApi.json ./dist/cjs/Asset/openApi.json
-cp ./src/ChainHistory/openApi.json ./dist/cjs/ChainHistory/openApi.json
-cp ./src/NetworkInfo/openApi.json ./dist/cjs/NetworkInfo/openApi.json
-cp ./src/Rewards/openApi.json ./dist/cjs/Rewards/openApi.json
-cp ./src/StakePool/openApi.json ./dist/cjs/StakePool/openApi.json
-cp ./src/TxSubmit/openApi.json ./dist/cjs/TxSubmit/openApi.json
-cp ./src/Utxo/openApi.json ./dist/cjs/Utxo/openApi.json
-cp -R ./src/StakePool/HttpStakePoolMetadata/schemas ./dist/cjs/StakePool/HttpStakePoolMetadata/schemas
+for i in `ls ./src` ; do
```
VanessaPC, line 10: Nice cleanup.

packages/cardano-services/src/Handle/HandleHttpService.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export class HandleHttpService extends HttpService {
+  constructor({ logger, handleProvider }: HandleServiceDependencies, router: Router = Router()) {
+    super(ServiceNames.Handle, handleProvider, router, logger);
+
+    useOpenApi(__dirname, router);
+
+    router.post(
+      '/resolve',
```
VanessaPC, line 20: In case we have more routes in the future, and maybe for consistency, should we clean up the routes? 

`router.post(`${routes.resolve}, ...`)`

packages/cardano-services/src/Handle/TypeOrmHandleProvider.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  async resolveHandles(args: ResolveHandlesArgs) {
+    const { handles } = args;
+
+    for (const handle of handles) if (!handle) throw emptyStringHandleResolutionRequestError();
+
+    return this.withDataSource(async (dataSource) => {
+      const queryRunner = dataSource.createQueryRunner();
+      const query = 'SELECT hash, slot FROM block WHERE slot = (SELECT MAX(slot) FROM block)';
```
VanessaPC, line 39: Should this query be saved to a queries file within this folder or another appropriate one?

rhyslbw (dismissed):
Great work @iccicci :muscle:

packages/cardano-services/src/Handle/TypeOrmHandleProvider.ts
```diff
@@ -0,0 +1,58 @@
...
+
+        const { cardanoAddress, handle, hasDatum, policyId } = entity;
+
+        return { handle, hasDatum, policyId, resolvedAddresses: { cardano: cardanoAddress }, resolvedAt };
+      };
+
+      const findOptions = { select: handleSelect, where: { handle: In(handles) } };
+      const entities = await dataSource.getRepository<PartialHandleEntity>(HandleEntity).find(findOptions);
+
+      return handles.map((handle) => mapEntity(entities.find((_) => _.handle === handle)));
```
VanessaPC, line 55: instead of `_` I'd probably rename it to 'handle' which is more common syntax..

VanessaPC (commented):
Great work, I just left a few small comments.

packages/cardano-services/src/Handle/TypeOrmHandleProvider.ts
rhyslbw: This is our existing standard, so it's best to not  deviate here, but rather in a separate complete refactor commit. Having said that, small queries like this work well inline as we have the context in the same file, plus there's not a demand for this query to be reused.

iccicci: From @rhyslbw 's comment I would say this doesn't need to be changed

packages/cardano-services/src/Handle/HandleHttpService.ts
iccicci: This required some more refactoring...

anyway 4532e13d1 should solve this

packages/cardano-services-client/test/HandleProvider/handleHttpProvider.test.ts
iccicci: fde4064eb should solve this

rhyslbw (approved):
Ready for squashing

VanessaPC (approved):
LGTM!

packages/cardano-services/src/Handle/TypeOrmHandleProvider.ts
VanessaPC: Yea, I agree.

PR #751, opened 2023-06-05
Feature/LW-6653 add context to txSubmit

VanessaPC:
# Context
We want to continue the implementation to return and work with standard handles. 
LW-6653

# Proposed Solution
- [x] Resolve the context with handles to return HandleResolution in `SubmitTxArgs`
- [ ] ~~Replace `SignedTx` with `SubmitTxArgs`~~
- [x] add signedTx or cbor to SignedTx type
- [x] Refactor `ObservableWallet.submitTx` to take the appropriate arguments  and implement it in `PersonalWallet`
- [x] Implement the handles from context in PersonalWallet
- [x] Add tests showcasing building and submitting a transaction that sends coins 

- Wallet package: 
PersonalWallet now has the handles implementation. 
ObservableWallet now takes handles. 

- Tx-construction: 
~~Replaces signedTx with SubmitTxArgs~~
Adds Cbor to SignedTx

### DIFF
--- a/packages/core/src/Provider/TxSubmitProvider/types.ts
+++ b/packages/core/src/Provider/TxSubmitProvider/types.ts
@@ -1,11 +1,12 @@
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
-import Cardano, { CardanoNodeErrors, Provider } from '../..';
+import Cardano, { CardanoNodeErrors, HandleResolution, Provider } from '../..';
 
 type SerializedTransaction = Cardano.util.HexBlob;
 
 export interface SubmitTxArgs {
   signedTransaction: SerializedTransaction;
+  context?: { handles: HandleResolution[] };
 }
 
 export interface TxSubmitProvider extends Provider {
--- a/packages/tx-construction/src/tx-builder/OutputBuilder.ts
+++ b/packages/tx-construction/src/tx-builder/OutputBuilder.ts
@@ -140,11 +140,6 @@ export class TxOutputBuilder implements OutputBuilder {
   async build(): Promise<OutputBuilderTxOut> {
     const txOut = this.toTxOut();
 
-    const outputValidation = toOutputValidationError(txOut, await this.#outputValidator.validateOutput(txOut));
-    if (outputValidation) {
-      throw outputValidation;
-    }
-
     if (this.#partialOutput.handle && this.#handleProvider) {
       const resolution = await this.#handleProvider.resolveHandles({ handles: [this.#partialOutput.handle] });
 
@@ -156,6 +151,11 @@ export class TxOutputBuilder implements OutputBuilder {
         // an address for the transaction.
         throw new HandleNotFoundError(this.#partialOutput);
       }
+    }
+
+    const outputValidation = toOutputValidationError(txOut, await this.#outputValidator.validateOutput(txOut));
+    if (outputValidation) {
+      throw outputValidation;
     }
 
     return txOut;
--- a/packages/tx-construction/src/tx-builder/finalizeTx.ts
+++ b/packages/tx-construction/src/tx-builder/finalizeTx.ts
@@ -4,7 +4,7 @@ import {
   TransactionSigner,
   util as keyManagementUtil
 } from '@cardano-sdk/key-management';
-import { Cardano } from '@cardano-sdk/core';
+import { Cardano, TxCBOR } from '@cardano-sdk/core';
 import { FinalizeTxDependencies, SignedTx, TxContext } from './types';
 
 const getSignatures = async (
@@ -41,19 +41,22 @@ export const finalizeTx = async (
       )
     : await getSignatures(keyAgent, tx, witness?.extraSigners, signingOptions);
 
+  const transaction = {
+    auxiliaryData,
+    body: tx.body,
+    id: tx.hash,
+    isValid,
+    witness: {
+      ...witness,
+      signatures: new Map([...signatures.entries(), ...(witness?.signatures?.entries() || [])])
+    }
+  };
+
   return {
-    ctx: {
+    cbor: TxCBOR.serialize(transaction),
+    context: {
       handles: handles ?? []
     },
-    tx: {
-      auxiliaryData,
-      body: tx.body,
-      id: tx.hash,
-      isValid,
-      witness: {
-        ...witness,
-        signatures: new Map([...signatures.entries(), ...(witness?.signatures?.entries() || [])])
-      }
-    }
+    tx: transaction
   };
 };
--- a/packages/tx-construction/src/tx-builder/types.ts
+++ b/packages/tx-construction/src/tx-builder/types.ts
@@ -1,4 +1,4 @@
-import { Cardano, Handle, HandleProvider, HandleResolution } from '@cardano-sdk/core';
+import { Cardano, Handle, HandleProvider, HandleResolution, TxCBOR } from '@cardano-sdk/core';
 import { CustomError } from 'ts-custom-error';
 
 import { InputSelectionError, InputSelector, SelectionSkeleton } from '@cardano-sdk/input-selection';
@@ -124,8 +124,9 @@ export type TxInspection = Cardano.TxBodyWithHash &
   };
 
 export interface SignedTx {
+  cbor: TxCBOR;
   tx: Cardano.Tx;
-  ctx: {
+  context: {
     handles: HandleResolution[];
   };
 }
--- a/packages/wallet/src/PersonalWallet/PersonalWallet.ts
+++ b/packages/wallet/src/PersonalWallet/PersonalWallet.ts
@@ -88,6 +88,7 @@ import {
   InitializeTxProps,
   InitializeTxResult,
   InvalidConfigurationError,
+  SignedTx,
   TxBuilderDependencies,
   finalizeTx,
   initializeTx
@@ -145,10 +146,12 @@ export const DEFAULT_LOOK_AHEAD_SEARCH = 20;
 // Ideally we should calculate this based on the activeSlotsCoeff and probability of a single block per epoch.
 const BLOCK_SLOT_GAP_MULTIPLIER = 2.5;
 
-const isOutgoingTx = (input: Cardano.Tx | TxCBOR | OutgoingTx): input is OutgoingTx =>
+const isOutgoingTx = (input: Cardano.Tx | TxCBOR | OutgoingTx | SignedTx): input is OutgoingTx =>
   typeof input === 'object' && 'cbor' in input;
-const isTxCBOR = (input: Cardano.Tx | TxCBOR | OutgoingTx): input is TxCBOR => typeof input === 'string';
-const processOutgoingTx = (input: Cardano.Tx | TxCBOR | OutgoingTx): OutgoingTx => {
+const isTxCBOR = (input: Cardano.Tx | TxCBOR | OutgoingTx | SignedTx): input is TxCBOR => typeof input === 'string';
+const isSignedTx = (input: Cardano.Tx | TxCBOR | OutgoingTx | SignedTx): input is SignedTx =>
+  typeof input === 'object' && 'context' in input;
+const processOutgoingTx = (input: Cardano.Tx | TxCBOR | OutgoingTx | SignedTx): OutgoingTx => {
   // TxCbor
   if (isTxCBOR(input)) {
     return {
@@ -156,6 +159,16 @@ const processOutgoingTx = (input: Cardano.Tx | TxCBOR | OutgoingTx): OutgoingTx 
       cbor: input,
       // Do not re-serialize transaction body to compute transaction id
       id: Cardano.TransactionId.fromTxBodyCbor(TxBodyCBOR.fromTxCBOR(input))
+    };
+  }
+  // SignedTx
+  if (isSignedTx(input)) {
+    return {
+      body: input.tx.body,
+      cbor: input.cbor,
+      context: input.context,
+      // Do not re-serialize transaction body to compute transaction id
+      id: input.tx.id
     };
   }
   // OutgoingTx (resubmitted)
@@ -509,7 +522,7 @@ export class PersonalWallet implements ObservableWallet {
   }
 
   async submitTx(
-    input: Cardano.Tx | TxCBOR | OutgoingTx,
+    input: Cardano.Tx | TxCBOR | OutgoingTx | SignedTx,
     { mightBeAlreadySubmitted }: SubmitTxOptions = {}
   ): Promise<Cardano.TransactionId> {
     const outgoingTx = processOutgoingTx(input);
@@ -517,6 +530,7 @@ export class PersonalWallet implements ObservableWallet {
     this.#newTransactions.submitting$.next(outgoingTx);
     try {
       await this.txSubmitProvider.submitTx({
+        context: outgoingTx.context,
         signedTransaction: outgoingTx.cbor
       });
       const { slot: submittedAt } = await firstValueFrom(this.tip$);
--- a/packages/wallet/src/services/types.ts
+++ b/packages/wallet/src/services/types.ts
@@ -2,6 +2,7 @@ import { AsyncKeyAgent, GroupedAddress } from '@cardano-sdk/key-management';
 import { Cardano, CardanoNodeErrors, EpochRewards, TxCBOR } from '@cardano-sdk/core';
 import { Observable } from 'rxjs';
 import { Percent } from '@cardano-sdk/util';
+import { SignedTx } from '@cardano-sdk/tx-construction';
 
 export enum TransactionFailure {
   InvalidTransaction = 'INVALID_TRANSACTION',
@@ -65,6 +66,7 @@ export interface OutgoingTx {
   cbor: TxCBOR;
   body: Cardano.TxBody;
   id: Cardano.TransactionId;
+  context?: SignedTx['context'];
 }
 
 export interface FailedTx extends OutgoingTx {
--- a/packages/wallet/test/PersonalWallet/methods.test.ts
+++ b/packages/wallet/test/PersonalWallet/methods.test.ts
@@ -57,10 +57,12 @@ describe('PersonalWallet methods', () => {
     txSubmitProvider = mocks.mockTxSubmitProvider();
     networkInfoProvider = mocks.mockNetworkInfoProvider();
     utxoProvider = mocks.mockUtxoProvider();
+
     const assetProvider = mocks.mockAssetProvider();
     const stakePoolProvider = createStubStakePoolProvider();
     const rewardsProvider = mockRewardsProvider();
     const chainHistoryProvider = mockChainHistoryProvider();
+    const handleProvider = mocks.mockHandleProvider();
     const groupedAddress: GroupedAddress = {
       accountIndex: 0,
       address,
@@ -83,6 +85,7 @@ describe('PersonalWallet methods', () => {
           {
             assetProvider,
             chainHistoryProvider,
+            handleProvider,
             keyAgent,
             logger,
             networkInfoProvider,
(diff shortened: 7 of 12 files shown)


packages/core/src/Provider/TxSubmitProvider/types.ts
```diff
@@ -1,11 +1,12 @@
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
-import Cardano, { CardanoNodeErrors, Provider } from '../..';
+import Cardano, { CardanoNodeErrors, HandleResolution, Provider } from '../..';
 
 type SerializedTransaction = Cardano.util.HexBlob;
 
 export interface SubmitTxArgs {
   signedTransaction: SerializedTransaction;
+  context?: { handles?: HandleResolution[] | [] };
```
mkazlauskas, line 9: It's already optional, so if we we're not setting the context we can do just `{signedTransaction}` or `{signedTransaction, context: {}}`
```suggestion
  context?: { handles?: HandleResolution[] };
```

packages/tx-construction/src/tx-builder/types.ts
```diff
@@ -139,7 +139,7 @@ export interface SignedTx {
  */
 export interface UnsignedTx {
   inspect(): Promise<TxInspection>;
-  sign(): Promise<SignedTx>;
```
mkazlauskas, line 142: Let's also delete the type that's no longer used.

packages/wallet/src/services/types.ts
```diff
@@ -64,6 +64,7 @@ export interface OutgoingTx {
   cbor: TxCBOR;
   body: Cardano.TxBody;
   id: Cardano.TransactionId;
+  context?: HandleResolution[] | [];
```
mkazlauskas, line 67: Same as
```suggestion
  context?: HandleResolution[];
```

packages/wallet/src/PersonalWallet/PersonalWallet.ts
```diff
@@ -137,19 +138,32 @@ export const DEFAULT_LOOK_AHEAD_SEARCH = 20;
...
       cbor: input,
       // Do not re-serialize transaction body to compute transaction id
       id: Cardano.TransactionId.fromTxBodyCbor(TxBodyCBOR.fromTxCBOR(input))
     };
   }
+  // SubmitTxArgs
+  if (isSubmitTxArgs(input)) {
+    return {
+      body: input.signedTransaction.body,
+      cbor: TxCBOR.serialize(input.signedTransaction),
```
mkazlauskas, line 161: Let's refactor `SubmitTxArgs` to `{ cbor: TxCBOR }` as we have 2 types that represent the same thing. Delete the old `SerializedTransaction` type that was used there. Please do that in a separate commit.
```suggestion
      cbor: input.signedTransaction,
```

packages/wallet/src/PersonalWallet/PersonalWallet.ts
```diff
@@ -137,19 +138,32 @@ export const DEFAULT_LOOK_AHEAD_SEARCH = 20;
...
+      body: getSerializedTxCBOR(input),
       cbor: input,
       // Do not re-serialize transaction body to compute transaction id
       id: Cardano.TransactionId.fromTxBodyCbor(TxBodyCBOR.fromTxCBOR(input))
     };
   }
+  // SubmitTxArgs
+  if (isSubmitTxArgs(input)) {
+    return {
+      body: input.signedTransaction.body,
```
mkazlauskas, line 160: This shouldn't be working as `signedTransaction` should be an opaque string here :thinking: 
```suggestion
   const deserializedTx = TxCBOR.deserialize(input.signedTransaction);
   ...
      body: deserializedTx.body,
```

packages/wallet/src/PersonalWallet/PersonalWallet.ts
```diff
@@ -137,19 +138,32 @@ export const DEFAULT_LOOK_AHEAD_SEARCH = 20;
...
+      cbor: TxCBOR.serialize(input.signedTransaction),
+      context: input?.context?.handles ?? [],
+      // Do not re-serialize transaction body to compute transaction id
+      id: input.signedTransaction.id
```
mkazlauskas, line 164:
```suggestion
      id: deserializedTx.id
```

packages/wallet/src/PersonalWallet/PersonalWallet.ts
```diff
@@ -480,18 +494,20 @@ export class PersonalWallet implements ObservableWallet {
...
-    input: Cardano.Tx | TxCBOR | OutgoingTx,
+    input: Cardano.Tx | TxCBOR | OutgoingTx | SubmitTxArgs,
     { mightBeAlreadySubmitted }: SubmitTxOptions = {}
   ): Promise<Cardano.TransactionId> {
     const outgoingTx = processOutgoingTx(input);
     this.#logger.debug(`Submitting transaction ${outgoingTx.id}`);
     this.#newTransactions.submitting$.next(outgoingTx);
     try {
       await this.txSubmitProvider.submitTx({
+        ...(outgoingTx.context && { context: { handles: outgoingTx.context } }),
```
mkazlauskas, line 505: This reads a little weird, how about
```suggestion
        context: outgoingTx.context,
```

packages/wallet/src/PersonalWallet/PersonalWallet.ts
```diff
@@ -480,18 +494,20 @@ export class PersonalWallet implements ObservableWallet {
...
         signedTransaction: outgoingTx.cbor
       });
       const { slot: submittedAt } = await firstValueFrom(this.tip$);
       this.#logger.debug(`Submitted transaction ${outgoingTx.id} at slot ${submittedAt}`);
+      this.#logger.debug(outgoingTx);
```
mkazlauskas, line 510: Let's keep a single log per submission
```suggestion
      this.#logger.debug(`Submitted transaction ${outgoingTx.id} at slot ${submittedAt}; # handles: ${outgoingTx.context?.handles?.length || 0}`);
```

packages/wallet/test/PersonalWallet/methods.test.ts
```diff
@@ -268,6 +271,20 @@ describe('PersonalWallet methods', () => {
...
+        const txOutput = await txBuilder.buildOutput().handle('alice').coin(1_000_000n).build();
+
+        const txOut = await txBuilder.addOutput(txOutput).build().sign();
+
+        const txPending = firstValueFrom(wallet.transactions.outgoing.pending$);
+        await wallet.submitTx(txOut);
+
+        const txPendingResult = await txPending;
+
+        expect(txPendingResult.body.outputs[0].address).toEqual(mocks.resolvedHandle.resolvedAddresses.cardano);
```
mkazlauskas, line 286:
```suggestion
        expect(txPendingResult.body.outputs[0].address).toEqual(mocks.resolvedHandle.resolvedAddresses.cardano);
        // expect txSubmitProvider.submitTx to be called with handle in context
```

mkazlauskas (changes requested):

packages/wallet/src/PersonalWallet/PersonalWallet.ts
VanessaPC: I thought we wanted to make it optional... so if it didn't exist, we just wouldn't pass it, so we wouldn't have to modify a lot of the other transactions that consume the outgoingTx. Or that's what I understood you and Daniele agreed in the call?

mkazlauskas: Optional means `T | undefined`, setting it to `undefined` explicitly is fine. I understand there's a slight difference between `{}` and `{context: undefined}`, but both matches the type which is the important thing.

packages/tx-construction/src/tx-builder/types.ts
VanessaPC: Is this no longer used? I think it is, but we now pass the `SubmitTxArgs` to the sign method as opposed to the `SignedArgs` to it.  So all the signedArgs types have been removed, and we now pass different ones. We could rename the method from sign to submit, but TxBuilder still uses a sign, and I think we need a sign (or submit) to get a txOut.

mkazlauskas: I think we no longer use https://github.com/input-output-hk/cardano-js-sdk/pull/751/files#diff-d2eebb6c7502396b4d246766bf4f31551ecdef4f174f4be4f9784c69bed987e3R126 type. If that's true, then there's no reason to keep it.

VanessaPC: Ah yes. I'll remove the interface.

packages/wallet/src/PersonalWallet/PersonalWallet.ts
VanessaPC: While trying this refactor, I found that interestingly this breaks quite a few things. The code for the `SubmitTxArgs` has `signedTransaction` set to `SerializedTransaction` which for me is resolving to type `any` because there is no `HexBlob` type in the current import path. As a result there seem to be some places that are doing unexpected things with the `signedTransaction` assuming its actually an object. 
I think this does need to be refactored, but I think that it should be it's own ticket after having a look at it. 

Perhaps it would be better to untangle this in a separate ticket?

packages/wallet/test/PersonalWallet/methods.test.ts
VanessaPC: 2d88f9ab41c0bf941b8126409e8d597fa2713303  should address this

packages/wallet/src/PersonalWallet/PersonalWallet.ts
VanessaPC: 2d88f9ab41c0bf941b8126409e8d597fa2713303  should address this

packages/wallet/src/PersonalWallet/PersonalWallet.ts
VanessaPC: 2d88f9ab41c0bf941b8126409e8d597fa2713303  should address this

packages/wallet/src/PersonalWallet/PersonalWallet.ts
VanessaPC: 2d88f9ab41c0bf941b8126409e8d597fa2713303  should address this

packages/wallet/src/PersonalWallet/PersonalWallet.ts
VanessaPC: 2d88f9ab41c0bf941b8126409e8d597fa2713303  should address this

packages/wallet/src/services/types.ts
VanessaPC: 2d88f9ab41c0bf941b8126409e8d597fa2713303  should address this

packages/tx-construction/src/tx-builder/types.ts
VanessaPC: 2d88f9ab41c0bf941b8126409e8d597fa2713303  should address this

packages/core/src/Provider/TxSubmitProvider/types.ts
VanessaPC: 2d88f9ab41c0bf941b8126409e8d597fa2713303  should address this

mkazlauskas (approved):
Looking good! Please update the failing tests and squash down to a single commit for final reviews

packages/tx-construction/src/tx-builder/finalizeTx.ts
```diff
@@ -30,7 +30,7 @@ export const finalizeTx = async (
   { ownAddresses, witness, signingOptions, auxiliaryData, isValid, handles }: TxContext,
   { inputResolver, keyAgent }: FinalizeTxDependencies,
   stubSign = false
-): Promise<SignedTx> => {
+): Promise<SubmitTxArgs> => {
```
VanessaPC, line 33: We're changing this type to `SubmitTxArgs` but the type of `SubmitTxArgs['signedTransaction']` is meant to be an `OpaqueString`, here we're returning an object of type `Cardano.Tx`. Are we meant to be serializing the `signedTransaction` here, or is the type incorrect? @mkazlauskas

VanessaPC: b15261eddb78a02ef40bd9688f6f5792babe75e9 should fix this.

packages/wallet/src/PersonalWallet/PersonalWallet.ts
```diff
@@ -145,19 +146,33 @@ export const DEFAULT_LOOK_AHEAD_SEARCH = 20;
 // Ideally we should calculate this based on the activeSlotsCoeff and probability of a single block per epoch.
 const BLOCK_SLOT_GAP_MULTIPLIER = 2.5;
 
-const isOutgoingTx = (input: Cardano.Tx | TxCBOR | OutgoingTx): input is OutgoingTx =>
+const getSerializedTxCBOR = (input: TxCBOR) => TxCBOR.deserialize(input).body;
```
mkazlauskas, line 149:
```suggestion
const getSerializedTxBody = (input: TxCBOR) => TxCBOR.deserialize(input).body;
```

mkazlauskas (changes requested):

yarn-project.nix
```diff
@@ -1327,6 +1327,7 @@ cacheEntries = {
 "fs-minipass@npm:3.0.0" = { filename = "fs-minipass-npm-3.0.0-3692c14b65-b72e9fe426.zip"; sha512 = "b72e9fe426e39f05b35bf237c8218b7ab3f68a65f325725ad7b4e431ff5a10725946fc62883b78446c07515ab938d25fdde3d08fb5ac8693f7f9eb9990da21f0"; };
 "fs.realpath@npm:1.0.0" = { filename = "fs.realpath-npm-1.0.0-c8f05d8126-99ddea01a7.zip"; sha512 = "99ddea01a7e75aa276c250a04eedeffe5662bce66c65c07164ad6264f9de18fb21be9433ead460e54cff20e31721c811f4fb5d70591799df5f85dce6d6746fd0"; };
 "fsevents@npm:2.3.2" = { filename = "fsevents-npm-2.3.2-a881d6ac9f-97ade64e75.zip"; sha512 = "97ade64e75091afee5265e6956cb72ba34db7819b4c3e94c431d4be2b19b8bb7a2d4116da417950c3425f17c8fe693d25e20212cac583ac1521ad066b77ae31f"; };
+"fsevents@patch:fsevents@npm%3A2.3.2#~builtin<compat/fsevents>::version=2.3.2&hash=18f3a7" = { filename = "fsevents-patch-3340e2eb10-8.zip"; sha512 = "edbd0fd80be379c14409605f77e52fdc78a119e17f875e8b90a220c3e5b29e54a1477c21d91fd30b957ea4866406dc3ff87b61432d2840ff8866b309e5866140"; };
```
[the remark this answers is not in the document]
mkazlauskas: I would start with LW-6816 `Fail PR check when running `yarn` updates lockfiles`

packages/tx-construction/src/tx-builder/OutputBuilder.ts
```diff
@@ -158,6 +153,11 @@ export class TxOutputBuilder implements OutputBuilder {
       }
     }
 
+    const outputValidation = toOutputValidationError(txOut, await this.#outputValidator.validateOutput(txOut));
```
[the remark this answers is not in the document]
VanessaPC: This was originally moved to have access to a resolved address but It's not needed now. 
883b14af431d6c40765f8350a71fa6995236fae6 should solve this

packages/wallet/src/PersonalWallet/PersonalWallet.ts
VanessaPC: 883b14af431d6c40765f8350a71fa6995236fae6 should solve this

packages/wallet/src/PersonalWallet/PersonalWallet.ts
```diff
@@ -145,19 +146,33 @@ export const DEFAULT_LOOK_AHEAD_SEARCH = 20;
...
-      body: TxCBOR.deserialize(input).body,
+      body: getSerializedTxCBOR(input),
       cbor: input,
       // Do not re-serialize transaction body to compute transaction id
       id: Cardano.TransactionId.fromTxBodyCbor(TxBodyCBOR.fromTxCBOR(input))
     };
   }
+  // SignedTx
+  if (isSignedTx(input)) {
+    const deserializedTx = TxCBOR.deserialize(input.cbor);
```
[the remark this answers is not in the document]
VanessaPC: 883b14af431d6c40765f8350a71fa6995236fae6 should solve this

packages/wallet/src/PersonalWallet/PersonalWallet.ts
```diff
@@ -145,19 +146,33 @@ export const DEFAULT_LOOK_AHEAD_SEARCH = 20;
...
+    return {
+      body: deserializedTx.body,
+      cbor: TxCBOR.serialize(input.tx),
```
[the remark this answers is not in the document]
VanessaPC: 883b14af431d6c40765f8350a71fa6995236fae6 should solve this

yarn-project.nix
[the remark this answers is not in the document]
VanessaPC: This change came when I got the latest from master, when I was going to update the PR for one last round of reviews... and It doesn't go now.. I can checkout the file from master on the last rebase

mkazlauskas (dismissed):
Nice work! Please squash for final reviews

yarn-project.nix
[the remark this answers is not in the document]
mkazlauskas: I just tried running `yarn` on master and it does not flip. Must be dependent on environment (maybe os).
> https://github.com/input-output-hk/cardano-js-sdk/pull/751#discussion_r1221773728

Please do, thanks

> we should backlog this for DevX to look into

Probably needs to be fixed on nix plugin, as it's only flipping the nix lockfile (not `yarn.lock`)

packages/tx-construction/src/tx-builder/OutputBuilder.ts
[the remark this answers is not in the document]
VanessaPC: @martynas when I run `yarn test:build:verify` I didn't see any error, but test failed after I pushed, so this code does need to be where it is in this line. 
However, we have a test for it, because on the tests I added we check whether the tx has a cardano.address, and if it doesn't it fails.  Which is the reason we moved it. 
I'll leave it like this, do you think it needs it further testing?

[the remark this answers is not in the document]
mkazlauskas: If it's a `tsc` error otherwise then no need to do any further testing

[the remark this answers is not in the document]
VanessaPC: No, no it's not a `tsc` error, it's an actual data validation error, when there's no address, it fails.
Screenshot 2023-06-08 at 08 17 39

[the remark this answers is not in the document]
mkazlauskas: Ah I see, that's good too, seems to be covered by test already - nothing to do

yarn-project.nix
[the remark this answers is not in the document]
VanessaPC: d8f957d8e412600b7cdfbf0e53d5422d184bce5f <- I have just got it from master and it looks like that

mkazlauskas (approved):
Nice work! :rocket:

yarn-project.nix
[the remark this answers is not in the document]
rhyslbw: > Probably needs to be fixed on nix plugin, as it's only flipping the nix lockfile

Yes, but it's still in the DevX domain. I'll keep this conversation open until I've captured the work items

rhyslbw (approved):
Great work @VanessaPC 🚀

PR #752, opened 2023-06-06
feat: discover own handles in PersonalWallet

mkazlauskas:
# Context

The [PersonalWallet](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/wallet/src/PersonalWallet/PersonalWallet.ts) (previously known as SingleAddressWallet) can discover it’s own registered handles.

LW-6693

# Proposed Solution

- add `ObservableWallet.handles$`
- PersonalWallet has a new optional prop in the constructor: `handlePolicyIds`

- Add some handle-related data to `mockProviders` in `util-dev`, update affected tests
- Rename util `arrayEquals`->`sameArrayItems`


rhyslbw (approved):
Looks great @mkazlauskas 🚀

packages/wallet/test/services/util/equals.test.ts
```diff
@@ -26,10 +26,10 @@ describe('equals', () => {
...
   });
 
-  test('arrayEquals', () => {
-    expect(arrayEquals([], [], strictEquals)).toBe(true);
-    expect(arrayEquals(['a'], ['a', 'b'], strictEquals)).toBe(false);
-    expect(arrayEquals(['a', 'b'], ['a', 'b'], strictEquals)).toBe(true);
+  test('sameArrayItems', () => {
+    expect(sameArrayItems([], [], strictEquals)).toBe(true);
+    expect(sameArrayItems(['a'], ['a', 'b'], strictEquals)).toBe(false);
+    expect(sameArrayItems(['a', 'b'], ['a', 'b'], strictEquals)).toBe(true);
```
mirceahasegan, line 32: This request is out of scope. It would be nice to have a test for:
```diff
+ expect(sameArrayItems(['a', 'b'], ['b', 'a'], strictEquals)).toBe(true);
```

packages/wallet/src/PersonalWallet/PersonalWallet.ts
```diff
@@ -451,6 +461,17 @@ export class PersonalWallet implements ObservableWallet {
...
+
+    this.handles$ = handlePolicyIds?.length
+      ? createHandlesTracker({
+          assetInfo$: this.assetInfo$,
+          handlePolicyIds,
+          logger: contextLogger(this.#logger, 'handles$'),
+          tip$: this.tip$,
+          utxo$: this.utxo.total$
+        })
+      : throwError(() => new InvalidConfigurationError('Missing handlePolicyIds option in PersonalWallet'));
```
mirceahasegan, line 473: Does this mean that specifying `handlePolicyIds` in the constructor is mandatory?

packages/wallet/src/services/HandlesTracker.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export interface HandlesTrackerProps {
+  utxo$: Observable<Cardano.Utxo[]>;
+  assetInfo$: Observable<Assets>;
+  tip$: Observable<Cardano.Tip>;
+  handlePolicyIds: Cardano.PolicyId[];
+  logger: Logger;
+}
+
+const handleInfoEquals = (a: HandleInfo, b: HandleInfo) =>
```
mirceahasegan, line 27: This could be part of 'util' (package or wallet/services/util/equals), and be unit tested separately.
But if you think it only makes sense here, to keep things encapsulated, then it's ok.

mkazlauskas: I'd prefer to keep it there and not export it, not very likely to be used elsewhere. It's tested indirectly with https://github.com/input-output-hk/cardano-js-sdk/pull/752/files#diff-85aa868305ebb433a1a6444ac1039908afa8ca085dc0889066e9ebc44d2afb72R70

packages/wallet/src/services/HandlesTracker.ts
```diff
@@ -0,0 +1,89 @@
...
+            resolvedAddresses: {
+              cardano: txOut.address
+            },
+            resolvedAt: tip
+          };
+        })
+        .filter(isNotNil);
+      if (utxo.length > 0 && handlesWithAssetInfo.length === 0) {
+        // AssetInfo is still resolving
+        return NEVER;
```
mirceahasegan, line 75: I think this will create a leaky/never completing observable every time this branch is reached.
```diff
-        return NEVER;
+        return NEVER;
```

mkazlauskas: Nice catch! :mage: Don't know how to add a test for it but fixed it in ed43d33504bc5591954a035b20abf6f1a396d61e

packages/wallet/test/services/util/equals.test.ts
mkazlauskas: ce2df04debff48b03aa9a0d75309cfa2ca8b0163

mirceahasegan (approved):
Great work, @mkazlauskas 🚀

rhyslbw (approved):

PR #753, opened 2023-06-06
test(e2e): handle discovery in PersonalWallet

iadmytro:
# Context

E2E tests for minting and sending 'handles'.

### DIFF
--- a/packages/e2e/test/wallet/PersonalWallet/handle.test.ts
+++ b/packages/e2e/test/wallet/PersonalWallet/handle.test.ts
@@ -1,0 +1,271 @@
+/* eslint-disable sonarjs/no-duplicate-string */
+import { Cardano, metadatum, nativeScriptPolicyId } from '@cardano-sdk/core';
+import { FinalizeTxProps, PersonalWallet } from '@cardano-sdk/wallet';
+import { InitializeTxProps } from '@cardano-sdk/tx-construction';
+import { KeyRole, TransactionSigner, util } from '@cardano-sdk/key-management';
+import { Metadatum, TokenMap } from '@cardano-sdk/core/dist/cjs/Cardano';
+import { burnTokens, createStandaloneKeyAgent, submitAndConfirm, txConfirmed, walletReady } from '../../util';
+import { createLogger } from '@cardano-sdk/util-dev';
+import { firstValueFrom } from 'rxjs';
+import { getEnv, getWallet, walletVariables } from '../../../src';
+
+const env = getEnv(walletVariables);
+const logger = createLogger();
+
+type HandleMetadata = {
+  [policyId: string]: {
+    [handleName: string]: {
+      augmentations: [];
+      core: {
+        handleEncoding: string;
+        og: number;
+        prefix: string;
+        termsofuse: string;
+        version: number;
+      };
+      description: string;
+      image: string;
+      name: string;
+      website: string;
+    };
+  };
+};
+
+const createHandleMetadata = (handlePolicyId: string, handleNames: string[]): HandleMetadata => {
+  const result: HandleMetadata[0] = {};
+  for (const key of handleNames) {
+    result[key] = {
+      augmentations: [],
+      core: {
+        handleEncoding: 'utf-8',
+        og: 0,
+        prefix: '$',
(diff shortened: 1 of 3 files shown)


mkazlauskas (commented):
Nice work! Should probably add a few assertions on `handles$` observable after #752 is merged

packages/e2e/test/wallet/PersonalWallet/handle.test.ts
```diff
@@ -0,0 +1,252 @@
...
+      wallet: receivingWallet
+    });
+  });
+
+  afterAll(async () => {
+    wallet.shutdown();
+    receivingWallet.shutdown();
+  });
+
+  it('mint and send handle', async () => {
```
rhyslbw, line 191:
```suggestion
  it('PersonalWallet discovers it's own handles', async () => {
```
We're not testing the minting of handles

rhyslbw (changes requested):

packages/e2e/test/wallet/PersonalWallet/handle.test.ts
```diff
@@ -0,0 +1,254 @@
...
+    expect(receivingUtxo.assets?.size).toEqual(1);
+    expect(receivingUtxo.assets?.keys().next().value).toEqual(assetIds[0]);
+    handles = await firstValueFrom(wallet.handles$);
+    receivingHandles = await firstValueFrom(receivingWallet.handles$);
+    expect(handles.length).toEqual(1);
+    expect(receivingHandles.length).toEqual(1);
+  });
+
+  describe('double mint handling', async () => {
+    it('mint same handle again', async () => {
```
rhyslbw, line 229:
```suggestion
    it('filters out double mints in separate transactions', async () => {
```

rhyslbw (changes requested):
There's actually one more test required in the `double mint handling`: burning the second mint to correct the error.

packages/e2e/test/wallet/PersonalWallet/handle.test.ts
```diff
@@ -0,0 +1,254 @@
...
+      await getWallet({
+        env,
+        handlePolicyIds: [policyId],
+        idx: 1,
+        logger,
+        name: 'Receiving Wallet',
+        polling: { interval: 50 }
+      })
+    ).wallet;
+    await walletReady(wallet, coins);
```
iccicci, line 167: Here probably we want to wait for `receivingWallet` to be ready rather than for `wallet`.

One more suggestions: grouping the two calls as follows should speed up a bit the test
```
await Promise.all([walletReady(wallet, coins), walletReady(receivingWallet, coins)]);
```

packages/e2e/test/wallet/PersonalWallet/handle.test.ts
rhyslbw: This hasn't been addressed

packages/e2e/test/wallet/PersonalWallet/handle.test.ts
```diff
@@ -0,0 +1,271 @@
...
+        polling: { interval: 50 }
+      })
+    ).wallet;
+    await Promise.all([walletReady(wallet, coins), walletReady(receivingWallet, 0n)]);
+    assetIds = [Cardano.AssetId(`${policyId}${assetNames[0]}`), Cardano.AssetId(`${policyId}${assetNames[1]}`)];
+    walletAddress = (await firstValueFrom(wallet.addresses$))[0].address;
+  });
+
+  afterEach(async () => {
+    await burnTokens({
```
rhyslbw, line 172: Burning the tokens after the test makes the state unusable for integration test requirements, so best to not do it in this suite.

rhyslbw: After discussing this on a call, it's better to separate the demands of the integration tests with the specific scenarios expressed here and just mint handles as part of setting up the initial state.

https://github.com/input-output-hk/cardano-js-sdk/tree/master/packages/e2e/local-network/scripts

rhyslbw (approved):
Great work @iadmytro. I updated the commit message to what we spoke about in the call

iccicci (approved):
LGTM

PR #755, opened 2023-06-06
Feat/handle projection integration to cardano services

lucas-barros:
# Context

# Proposed Solution


rhyslbw (dismissed):
LGTM!

packages/cardano-services/src/Projection/prepareTypeormProjection.ts
```diff
@@ -91,6 +116,7 @@ const entityInterDependencies: Partial<Record<EntityName, EntityName[]>> = {
   asset: ['block'],
   blockData: ['block'],
   currentPoolMetrics: ['stakePool'],
+  handle: ['block', 'tokens', 'asset', 'output'],
```
iccicci, line 119: Looking at the migrations I would say that `handle` depends only on `tokens` while `tokens` depends on `asset` and `output`.
Should we make this reflect the dependencies in the migrations or is there something I'm missing?

mkazlauskas: Nice catch, the entity actually only depends on AssetEntity, but `storeHandles` depends on all those other ones. Made a fixup to make it more accurate 1507e0361a2b1f7e0a217d45def76c01f8950b6f (the end result is the same though)

rhyslbw (approved):
Ready for squashing

mkazlauskas (approved):
:rocket:

iccicci (approved):
LGTM

PR #756, opened 2023-06-06
cd to dev-preview

blaggacao:
# WIP


michalrus (commented):
I configured the self-hosted runners, but now:

```
         … while fetching the input 'github:input-output-hk/cardano-js-sdk/0676c61e2e4f561e2778b15403c25a9bead0f6f9'
  
         error: unable to download 'https://api.github.com/repos/input-output-hk/cardano-js-sdk/tarball/0676c61e2e4f561e2778b15403c25a9bead0f6f9': HTTP error 401
```

1. I'm not sure why Nix is trying to fetch the tgz of this repository, instead of using the checked out source?

2. I don’t know why it's a 401 Unauthorized (something off with netrc tokens?), because this repository is public, see:

```
$ curl -s -L -I https://api.github.com/repos/input-output-hk/cardano-js-sdk/tarball/0676c61e2e4f561e2778b15403c25a9bead0f6f9 | grep -E '^(HTTP|location)'
HTTP/2 302 
location: https://codeload.github.com/input-output-hk/cardano-js-sdk/legacy.tar.gz/0676c61e2e4f561e2778b15403c25a9bead0f6f9

HTTP/2 200 
```

blaggacao:
@michalrus Thanks for taking care of this!

> 1. I'm not sure why Nix is trying to fetch the tgz of this repository, instead of using the checked out source?

I found _where_ it does it. Not sure why. I'd infer that the goal of this is a feature and safety mechanism so that users don't accidentally mutate the checkout, make it dirty, and by doing so, for example invalidate image tag revisions in inadverted ways that would be only detected extemporaneously (if at all).
```bash
  # will be a list of actions
  EVAL=$(
    $nix eval --show-trace --json \
      "$flake_url#__std.ci.$system"
  )
```

> 2. I don’t know why it's a 401 Unauthorized (something off with netrc tokens?), because this repository is public, see:

I'm pretty certain that this is a configuration error. I'm pretty certain that curl-lib always uses the token when interacting with the GH api (through nix), regardless of if the target repository is public or not.

So in this case, it might just have the wrong credentials. I'll have a look and report back.

blaggacao:
Superseded by https://github.com/input-output-hk/cardano-js-sdk/pull/864 & https://github.com/input-output-hk/cardano-js-sdk/pull/865

PR #757, opened 2023-06-07
Test flakiness

iccicci:
# Context

Some unit tests sometimes fails...

# Proposed Solution

- changed a single fixed sleep of 100ms into a loop... if the loop condition is never satisfied the tests will fail for timeout
- added some log lines to try to identify another source of flakiness


VanessaPC (approved):
Looks good to me!

rhyslbw (approved):

PR #761, opened 2023-06-12
feat: verify handles in ogmiosTxProvider

VanessaPC:
# Context
Adding the verification of the handles making sure it resolves to the same address in the OgmiosTxSubmitProvider.
LW-6543

# Proposed Solution
- [x] Pass the `KoraLabsProvider` into 'OgmiosTxSubmitProvider`
- [x] In the `OgmiosTxSubmitProvider` verify the current `context.handles` against the handles from Koralabs.
- [x] Set up a new error `409` to throw when it no longer verifies.
- [x] Set up a pre-submission error in `@cardano-sdk/core` for HandleProviderVerification.
- [x] Add tests to make sure the verification in `OgmiosTxSubmitProvider` is working accordingly. 
- [x] Make sure the `cardano-services` tests pass since there are modifications. 
- [x] Make sure the `ogmios` package tests pass since there are modifications. 

- Now the `OgmiosTxSubmitProvider` takes a `HandleProvider`.


packages/cardano-services/src/TxSubmit/TxSubmitHttpService.ts
```diff
@@ -46,7 +46,9 @@ export class TxSubmitHttpService extends HttpService {
           } catch {
             isHealthy = false;
           }
-
+          if (error instanceof ProviderError) {
```
VanessaPC, line 49: Where does the error type come from? when I add this to the CardanoNodeErrors, I eventually run into an issue they come from another library and they aren't added to a mapper but it doesn't seem like the mapper would accept it either.

packages/cardano-services/test/Program/services/ogmios.test.ts
```diff
@@ -28,6 +28,21 @@ import { types } from 'util';
 import axios from 'axios';
 import http from 'http';
 
+export const getAliceHandleProviderResponse: HandleResolution = {
```
VanessaPC, line 31: This is unavailable because it's in a utils file for a test folder. I think it should be moved, as utils for the KoralabsHandleProvider? or somewhere else where it's available.

rhyslbw: i.e this is a duplicate because the other mock object isn't able to be imported? Did you try adding a path to the tsconfig?

packages/cardano-services/src/TxSubmit/TxSubmitHttpService.ts
rhyslbw: We’re currently only expecting _submission_ errors, the ones returned by Cardano Node, to be thrown by the provider, but we should change that to expect a new type of error `PreSubmissionErrors` in addition. These errors can be defined in the `core` package`

packages/core/src/CardanoNode/types/CardanoNodeErrors.ts
```diff
@@ -67,6 +67,8 @@ export const TxSubmissionErrors = {
   ExtraRedeemersError: ogmiosSubmissionErrors.ExtraRedeemers.Error,
   ExtraScriptWitnessesError: ogmiosSubmissionErrors.ExtraScriptWitnesses.Error,
   FeeTooSmallError: ogmiosSubmissionErrors.FeeTooSmall.Error,
+  // Temporary addition
+  HandleConflictError: Error,
```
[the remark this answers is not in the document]
rhyslbw: https://github.com/input-output-hk/cardano-js-sdk/pull/761#discussion_r1226310296

packages/cardano-services/test/Program/services/ogmios.test.ts
```diff
@@ -21,13 +21,29 @@ import { SrvRecord } from 'dns';
 import { bufferToHexString } from '@cardano-sdk/util';
 import { clearDbPools, createHealthyMockOgmiosServer, ogmiosServerReady } from '../../util';
 import { createMockOgmiosServer } from '../../../../ogmios/test/mocks/mockOgmiosServer';
+// import { getAliceHandleProviderResponse } from '@cardano-sdk/cardano-services-client';
```
rhyslbw, line 24:
```suggestion
// import { getAliceHandleProviderResponse } from '../../../../cardano-services-client/test';
```
Might need to narrow the path further

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
```diff
@@ -63,10 +66,28 @@ export class OgmiosTxSubmitProvider extends RunnableModule implements TxSubmitPr
     }
   }
 
-  async submitTx({ signedTransaction }: SubmitTxArgs): Promise<void> {
+  async submitTx({ signedTransaction, context }: SubmitTxArgs, handleProvider?: HandleProvider): Promise<void> {
     if (this.state !== 'running') {
       throw new CardanoNodeErrors.NotInitializedError('submitTx', this.name);
     }
+
+    if (context && context.handles && context.handles.length > 0) {
```
rhyslbw, line 74:
```suggestion
    if (context?.handles?.length > 0) {
```

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
```diff
@@ -63,10 +66,28 @@ export class OgmiosTxSubmitProvider extends RunnableModule implements TxSubmitPr
...
+      if (!handleProvider) {
```
rhyslbw, line 75: Better to handle this in the constructor

rhyslbw (changes requested):

packages/cardano-services/test/Program/services/ogmios.test.ts
VanessaPC: I did try. But it wont' work, because the `getAliceHandleProviderResponse` gets exported from a `test/utils` file. So in order to get the import to work, the path should be something like: `@cardano-sdk/cardano-services-client/test/utils` and I believe this is something we want to avoid. 
Either the 'utils' file should be moved in a 'utils' folder with an index file. And even so, I'm not sure we export utils from tests to other folders, I think they are just for testing purposes. 
Or we then move it to another exported folder altogether.

packages/cardano-services/test/Program/services/ogmios.test.ts
VanessaPC: I thought we didn't want to narrow paths.. but in this case, yea, I can do that and then we don't have to move it anywhere.

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
VanessaPC: Well, the handleProvider is passed into the `submitTx`, not into the constructor. And it's not an error until I'm using handles.. so I thought this the best way to pass it for now. 
If we want to pass it to the constructor and down to `submitTx` then I change how things are instantiated.

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
VanessaPC: If I do this, the ternary can return an `undefined`, but handles don't take `undefined`.. it takes a `handle: HandleResolution[]` or an empty `[]`. So the first check would return a `null` if context and context handles don't exist. Which is more suited to this situation? Otherwise, I need to modify handles to take undefined, but in previous work it was specified not to do it this way to better control the data we have in different scenarios..

packages/core/src/CardanoNode/types/CardanoNodeErrors.ts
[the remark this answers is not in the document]
VanessaPC: 0bc4fd1a9ab89f5e229feace62c4b0e3d1485b9c adds the presubmission errors 
 ebc0237c005b9b050f038dcac487ca1e9a5875e6 deletes this line.

packages/cardano-services/src/TxSubmit/TxSubmitHttpService.ts
VanessaPC: 0bc4fd1a9ab89f5e229feace62c4b0e3d1485b9c adds the presubmission errors 
 ebc0237c005b9b050f038dcac487ca1e9a5875e6 addresses this line.

packages/cardano-services/test/Program/services/ogmios.test.ts
VanessaPC: ebc0237c005b9b050f038dcac487ca1e9a5875e6 should solve  this

packages/cardano-services/test/Program/services/ogmios.test.ts
VanessaPC: ebc0237c005b9b050f038dcac487ca1e9a5875e6 should solve  this

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
mkazlauskas: Agree with @rhyslbw here: let's remove it from the core interface and add as a dependency in OgmiosTxSubmitProvider constructor

packages/core/src/errors.ts
```diff
@@ -31,6 +34,16 @@ export class ProviderError<InnerError = unknown> extends ComposableError<InnerEr
   }
 }
 
+export class HandleOwnerChangeError extends CustomError {
+  constructor(
+    public handle: Handle,
+    public expectedAddress: Cardano.PaymentAddress,
+    public actualAddress: Cardano.PaymentAddress | null
+  ) {
+    super('Results are different between expectedAddress and actualAddress for the handle.');
```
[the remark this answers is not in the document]
mkazlauskas: Current message is fine, addesses and handle are available as public fields

packages/core/src/Provider/TxSubmitProvider/types.ts
```diff
@@ -14,5 +14,5 @@ export interface TxSubmitProvider extends Provider {
    * @param signedTransaction signed and serialized cbor
    * @throws {CardanoNodeErrors.TxSubmissionError}
    */
-  submitTx: (args: SubmitTxArgs) => Promise<void>;
+  submitTx: (args: SubmitTxArgs, handleProvider?: HandleProvider) => Promise<void>;
```
mkazlauskas, line 17: Let's not change the signature of this, because this interface is used in different contexts (client-side as well). HandleProvider is an implementation detail of a specific TxSubmitProvider
```suggestion
  submitTx: (args: SubmitTxArgs) => Promise<void>;
```

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
```diff
@@ -82,4 +91,27 @@ export class OgmiosTxSubmitProvider extends RunnableModule implements TxSubmitPr
   async startImpl(): Promise<void> {
     return Promise.resolve();
   }
+
+  private async validateHandles(context: SubmitTxArgs['context']) {
+    if (context && context.handles && context.handles.length > 0) {
+      if (!this.#handleProvider) {
+        throw new Error('No handleProvider passed to submitTx, but is needed to validate handles used in transaction.');
```
mkazlauskas, line 98:
```suggestion
        throw new ProviderError(ProviderFailure.NotImplemented, undefined, 'No handleProvider passed to submitTx, but is needed to validate handles used in transaction.');
```

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
```diff
@@ -82,4 +91,27 @@ export class OgmiosTxSubmitProvider extends RunnableModule implements TxSubmitPr
...
+      }
+
+      const handleInfoList = await this.#handleProvider.resolveHandles({
+        handles: context.handles.map((handle) => handle.handle)
+      });
+
+      for (const [index, handleInfo] of handleInfoList.entries()) {
+        if (!handleInfo || handleInfo.resolvedAddresses.cardano !== context.handles[index].handle) {
```
mkazlauskas, line 106: This looks like a bug. Please add a test that catches it before fixing.
```suggestion
        if (!handleInfo || handleInfo.resolvedAddresses.cardano !== context.handles[index].resolvedAddresses.cardano) {
```

packages/ogmios/test/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.test.ts
```diff
@@ -101,6 +116,61 @@ describe('OgmiosTxSubmitProvider', () => {
...
+      await expect(
+        provider.submitTx({
+          context: {
+            handles: [handleResolution]
+          },
+          signedTransaction: emptyUintArrayAsHexString
+        })
+      ).rejects.toThrowError(/no handleprovider passed/i);
+    });
+
```
mkazlauskas, line 138:
```suggestion
    it('does not throw an error if handles resolve to same addresses as in context', async () => {
```

mkazlauskas (changes requested):
A couple things missing:
1. Hydrate `HandleOwnerChangeError` in `cardano-services-client` with object prototype
2. Pass an instance of `HandleProvide`r to `OgmiosTxSubmitProvider` in `cardano-services`. Currently we'd be hitting https://github.com/input-output-hk/cardano-js-sdk/pull/761/files#diff-b7119a8f737f72567611a7b9e26aba55cab3472da2b8446ce73a6623a0b2ce93R98 if we deployed it

packages/ogmios/test/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.test.ts
VanessaPC: 0dfd234d683942a8f490be4b630f098e8d0fa823 should solve this

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
VanessaPC: 0dfd234d683942a8f490be4b630f098e8d0fa823 should solve this

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
VanessaPC: 0dfd234d683942a8f490be4b630f098e8d0fa823 should solve this

packages/core/src/Provider/TxSubmitProvider/types.ts
VanessaPC: 0dfd234d683942a8f490be4b630f098e8d0fa823 should solve this

packages/cardano-services/src/Program/programs/providerServer.ts
```diff
@@ -222,12 +223,33 @@ const serviceMapFactory = (options: ServiceMapFactoryOptions) => {
...
+        },
+        cardanoNode,
+        dbPools,
+        epochMonitor: getEpochMonitor(dbPools.main),
+        genesisData,
+        logger
+      });
+      const handleProvider = new KoraLabsHandleProvider({
+        networkInfoProvider,
+        policyId: '', // <--------
```
VanessaPC, line 245: @mkazlauskas  I am not sure where the policyId or serverUrl would come from in this step.

packages/cardano-services/src/Program/programs/providerServer.ts
```diff
@@ -222,12 +223,33 @@ const serviceMapFactory = (options: ServiceMapFactoryOptions) => {
...
       return new NetworkInfoHttpService({ logger, networkInfoProvider });
     }, ServiceNames.NetworkInfo),
-    [ServiceNames.TxSubmit]: async () => {
+    [ServiceNames.TxSubmit]: withDbSyncProvider(async (dbPools, cardanoNode) => {
+      if (!genesisData)
+        throw new MissingProgramOption(
+          ServiceNames.NetworkInfo,
+          ProviderServerOptionDescriptions.CardanoNodeConfigPath
+        );
+      const networkInfoProvider = new DbSyncNetworkInfoProvider({
```
VanessaPC, line 232: @mkazlauskas I think this is the right thing to do, for set up requirements, but it might not be.

packages/cardano-services/src/Program/programs/providerServer.ts
mkazlauskas: Need to add a new arg/variable for provider server in `cli.ts`

packages/ogmios/test/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.test.ts
```diff
@@ -101,6 +116,80 @@ describe('OgmiosTxSubmitProvider', () => {
...
+      await provider.start();
+
+      await expect(
+        provider.submitTx({
+          context: {
+            handles: [handleResolution]
+          },
+          signedTransaction: emptyUintArrayAsHexString
+        })
+      ).not.toThrowError(/no handleprovider passed/i);
```
mkazlauskas, line 155: Would be best to loosen this assertion to prevent potential false positives
```suggestion
      ).not.toThrow();
```

packages/cardano-services/src/Program/programs/providerServer.ts
mkazlauskas: A single instance of DbSyncNetworkInfoProvider should be reused for both:
- here as a dependency for kora labs provider
- [ServiceNames.NetworkInfo]

It should be created lazily, so that it's not instantiated when neither `ServiceNames.TxSubmit` nor `ServiceNames.NetworkInfo` is enabled

mkazlauskas (changes requested):

packages/cardano-services/src/Program/programs/providerServer.ts
VanessaPC: 9a0c19b5dc62fd220c9d1414e8cc59b646a50a97 should fix this

packages/cardano-services/src/Program/programs/providerServer.ts
VanessaPC: 9a0c19b5dc62fd220c9d1414e8cc59b646a50a97 should fix this

packages/cardano-services/src/Program/options/common.ts
```diff
@@ -91,4 +100,19 @@ export const withCommonOptions = (command: Command, defaults: { apiUrl: URL }) =
         .env('SERVICE_DISCOVERY_TIMEOUT')
         .default(SERVICE_DISCOVERY_TIMEOUT_DEFAULT)
         .argParser((interval) => Number.parseInt(interval, 10))
+    )
+    .addOption(
+      new Option('--handle-provider-policy-id <policyId>', CommonOptionDescriptions.PolicyId)
```
mkazlauskas, line 105: The new options shouldn't be in 'common' options, but only for the "provider server". "Common" options are the ones applicable to all services: provider server, workers, projector

Also, let's align the naming with [the existing projector option](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services/src/cli.ts#L162), and accept an array of policy IDs:
- you can just throw a NotImplementedError if >1 is passed for now
- would be nice to extract a util like `withHandlePolicyIds`

packages/ogmios/test/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.test.ts
mkazlauskas: Thanks, that'll do it. Just FIY (no change required): `toThrow` will also accept non-errors thrown, while `toThrowError` expects `instanceof Error`

mkazlauskas (changes requested):

packages/cardano-services/src/Program/options/common.ts
VanessaPC: 3da615fdee6e0139e13a22d9b39ec1fb8ea9499f should fix this (also, I'm aware I rebased, but I wanted the latest from master because the cli test were breaking on stuff I hadn't touched ... not sure the rebase fixed it or not, but after that I could see the breakages from my changes).

packages/cardano-services/src/Program/options/policyIds.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const withHandlePolicyIdsOptions = (command: Command, maxPolicyIdsSupported?: number) =>
+  command.addOption(
+    new Option('--handle-policy-ids <handlePolicyIds>', HandlePolicyIdsOptionDescriptions.HandlePolicyIds)
+      .env('HANDLE_POLICY_IDS')
+      .argParser((policyIds: string) => {
+        const policyIdsArray = handlePolicyIdsParser(policyIds);
+
+        if (maxPolicyIdsSupported && policyIdsArray.length > maxPolicyIdsSupported) {
+          throw new NotImplementedError('Multiple policyIds not supported');
```
rhyslbw, line 22:
```suggestion
          throw new NotImplementedError(`${policyIdsArray.length} policyIds are not supported`);
```

packages/cardano-services/src/Program/programs/types.ts
```diff
@@ -49,13 +50,16 @@ export enum ProviderServerOptionDescriptions {
   UseTypeOrmStakePoolProvider = 'Enables the TypeORM Stake Pool Provider',
   UseBlockfrost = 'Enables Blockfrost cached data DB',
   UseQueue = 'Enables RabbitMQ',
-  PaginationPageSizeLimit = 'Pagination page size limit shared across all providers'
+  PaginationPageSizeLimit = 'Pagination page size limit shared across all providers',
+  HandlePolicyIds = 'Cardano Policy ID',
```
rhyslbw, line 54:
```suggestion
  HandlePolicyIds = 'Minting Policy ID',
```

packages/cardano-services/src/Program/programs/types.ts
```diff
@@ -49,13 +50,16 @@ export enum ProviderServerOptionDescriptions {
...
+  HandleProviderServerUrl = 'Server url for the Kora Labs handle provider'
```
rhyslbw, line 55:
```suggestion
  HandleProviderServerUrl = 'URL for the Handle provider server'
```
We'll be refactoring to use https://github.com/input-output-hk/cardano-js-sdk/pull/750/ soon enough, so can use naming and descriptions with this in mind.

packages/cardano-services/test/Program/services/ogmios.test.ts
```diff
@@ -400,6 +404,36 @@ describe('Service dependency abstractions', () => {
       await provider.shutdown();
     });
 
+    it('should execute a provider error from the ogmios, and be intercepted by txSubmit', async () => {
```
rhyslbw, line 407:
```suggestion
    it('throws a provider error if the submitted transaction does not contain addresses that can be resolved from the included context', async () => {
```
Is this it?

packages/cardano-services/test/Program/services/ogmios.test.ts
```diff
@@ -400,6 +404,36 @@ describe('Service dependency abstractions', () => {
...
+      await provider.start();
+      expect(
+        provider.submitTx({
+          context: { handles: [getAliceHandleProviderResponse] },
+          signedTransaction: bufferToHexString(Buffer.from(new Uint8Array([])))
+        })
+      ).resolves.toEqual({ context: { handles: [getAliceHandleProviderResponse] } });
+    });
+
+    it('should execute a provider error from the ogmios provider and be intercepted by txSubmit', async () => {
```
rhyslbw, line 422: Try rewording to precisely describe the test, using my suggestion above as a guide.

packages/core/src/errors.ts
rhyslbw, line 43:
```suggestion
    super(`Expected: ${expectedAddress} for handle $${handle}. Actual: ${actualAddress}`);
```

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
```diff
@@ -82,4 +91,31 @@ export class OgmiosTxSubmitProvider extends RunnableModule implements TxSubmitPr
...
     return Promise.resolve();
   }
+
+  private async validateHandles(context: SubmitTxArgs['context']) {
+    if (context && context.handles && context.handles.length > 0) {
+      if (!this.#handleProvider) {
+        throw new ProviderError(
+          ProviderFailure.NotImplemented,
+          undefined,
+          'No handleProvider passed to submitTx, but is needed to validate handles used in transaction.'
```
rhyslbw, line 101:
```suggestion
          'No HandleProvider was set during construction.'
```

rhyslbw (changes requested):

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
```diff
@@ -82,4 +91,31 @@ export class OgmiosTxSubmitProvider extends RunnableModule implements TxSubmitPr
   async startImpl(): Promise<void> {
     return Promise.resolve();
   }
+
+  private async validateHandles(context: SubmitTxArgs['context']) {
```
rhyslbw, line 95:
```suggestion
  private async validateHandles(context: SubmitTxArgs['context']): Promise<boolean> {
```

packages/cardano-services/test/Program/services/ogmios.test.ts
rhyslbw: @mkazlauskas please check this and advise. We've have issues with test imports in the past, so just want to ensure we're not tripping over this issue again.

mkazlauskas: I don't know how that even works `'@cardano-sdk/cardano-services-client/test/util';` :thinking: 

I suggest moving `getAliceHandleProviderResponse` to `util-dev`. `'../../../../cardano-services-client/test'` is also acceptable because it's the pattern already in use, but need to be careful with this to make sure that the dependency test project is in `test/tsconfig.ts`

packages/cardano-services/test/Program/services/ogmios.test.ts
VanessaPC: It is

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
VanessaPC: aa8096909c8ee4b312f8d2d9395b28766ebd4ff2 should solve this

packages/core/src/errors.ts
VanessaPC: aa8096909c8ee4b312f8d2d9395b28766ebd4ff2 should solve this

packages/cardano-services/test/Program/services/ogmios.test.ts
VanessaPC: aa8096909c8ee4b312f8d2d9395b28766ebd4ff2 should solve this

packages/cardano-services/src/Program/programs/types.ts
VanessaPC: aa8096909c8ee4b312f8d2d9395b28766ebd4ff2 should solve this

packages/cardano-services/src/Program/options/policyIds.ts
VanessaPC: aa8096909c8ee4b312f8d2d9395b28766ebd4ff2 and af7ef863a0e74c879dd5daba73879a75a9f60662 should solve this

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
rhyslbw, line 95:
```suggestion
  private async validateHandles(context: SubmitTxArgs['context']): Promise<void> {
```

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
rhyslbw: Superseded by: https://github.com/input-output-hk/cardano-js-sdk/pull/761#discussion_r1230742567 and https://github.com/input-output-hk/cardano-js-sdk/pull/761#discussion_r1230742819

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
```diff
@@ -82,4 +91,31 @@ export class OgmiosTxSubmitProvider extends RunnableModule implements TxSubmitPr
...
+    if (context && context.handles && context.handles.length > 0) {
```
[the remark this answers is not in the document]
VanessaPC: df1aee5 should solve this

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
VanessaPC: df1aee5 should solve this

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
```diff
@@ -82,4 +95,31 @@ export class OgmiosTxSubmitProvider extends RunnableModule implements TxSubmitPr
...
+          'No HandleProvider was set during construction.'
+        );
+      }
+
+      const handleInfoList = await this.#handleProvider.resolveHandles({
+        handles: context.handles.map((handle) => handle.handle)
+      });
+
+      for (const [index, handleInfo] of handleInfoList.entries()) {
+        if (!handleInfo || handleInfo.cardanoAddress !== context.handles[index].resolvedAddresses.cardano) {
```
rhyslbw, line 114:
```suggestion
        if (!handleInfo || handleInfo.cardanoAddress !== context.handles[index].cardanoAddress) {
```
?

rhyslbw (commented):
There are quite a few places where the legacy `resolvedAddresses: { cardanoAddress }` structure is being used. This could be the source of some test failures?

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
VanessaPC: The `context.handles[index].resolvedAddresses.cardano` is correct, because it has the previous data type. 
This is what I tried to say earlier, and the reason I had started the refactor to convert the type to the new type, so that we use `cardano.addresses` instead of `resolvedAddresses` everywhere, because otherwise, it does not work. 447670e499528e136f2ca3888e2e7264b685efd7  
However, the tests that are breaking now, are only the timeout ones that have to do with Ogmios and not with the data types ( thought these should have broken as well as the ts linter ). 
- a refactor to use the new type everywhere and clean up the code accordingly.

VanessaPC: Edit: 
Ticket: LW-7172
PR: https://github.com/input-output-hk/cardano-js-sdk/pull/780

rhyslbw (approved):

rhyslbw (changes requested):

packages/cardano-services/package.json
```diff
@@ -56,6 +56,7 @@
   },
   "devDependencies": {
     "@cardano-ogmios/client": "5.6.0",
+    "@cardano-sdk/cardano-services-client": "workspace:~",
```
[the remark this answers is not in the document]
VanessaPC: 0708ad5 should fix it

rhyslbw (approved):
Ready for squashing

packages/cardano-services/src/Program/programs/types.ts
```diff
@@ -52,14 +53,17 @@ export enum ProviderServerOptionDescriptions {
   UseTypeOrmStakePoolProvider = 'Enables the TypeORM Stake Pool Provider',
   UseBlockfrost = 'Enables Blockfrost cached data DB',
   UseQueue = 'Enables RabbitMQ',
-  PaginationPageSizeLimit = 'Pagination page size limit shared across all providers'
+  PaginationPageSizeLimit = 'Pagination page size limit shared across all providers',
+  HandlePolicyIds = 'Minting Policy ID',
```
iccicci, line 57: `HandlePolicyIds` is repeated three times always with the same meaning.

Better to replace any use of `ProjectorOptionDescriptions.HandlePolicyIds` and `ProviderServerOptionDescriptions.HandlePolicyIds` with the (correctly) newly introduced `HandlePolicyIdsOptionDescriptions.HandlePolicyIds`, and remove the first two.

packages/cardano-services/src/cli.ts
```diff
@@ -107,17 +107,19 @@ const runServer = async (message: string, loadServer: () => Promise<HttpServer>)
...
-      program
-        .command('start-projector')
-        .description('Start a projector')
-        .argument(
-          '[projectionNames...]',
-          `List of projections to start: ${Object.values(ProjectionName).toString()}`,
-          projectionNameParser
-        ),
-      ''
+  withHandlePolicyIdsOptions(
```
iccicci, line 110: Minor minor minor minor: adding `withHandlePolicyIdsOptions` as inner _option modifier function_ would have a lesser impact on the diff making it a bit more clear.

packages/cardano-services-client/src/TxSubmitProvider/txSubmitHttpProvider.ts
```diff
@@ -62,6 +63,10 @@ export const txSubmitHttpProvider = (config: CreateHttpProviderConfig<TxSubmitPr
                   : ProviderFailure.BadRequest;
               throw new ProviderError(failure, txSubmissionError);
             }
+
+            if (error.name === 'HandleOwnerChangeError') {
+              Object.setPrototypeOf(error, HandleOwnerChangeError);
```
mkazlauskas, line 68: Please add a test for this

packages/cardano-services/src/cli.ts
```diff
@@ -175,24 +172,28 @@ withCommonOptions(
   );
 
 withCommonOptions(
-  withOgmiosOptions(
-    withPostgresOptions(
+  withHandlePolicyIdsOptions(
```
iccicci, line 175: As minor as above, but here is even more true

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
```diff
@@ -28,16 +32,22 @@ export class OgmiosTxSubmitProvider extends RunnableModule implements TxSubmitPr
...
   /**
    * @param {ConnectionConfig} connectionConfig Ogmios connection configuration
    * @param {Logger} logger object implementing the Logger abstract class
    * @throws {TxSubmission.errors}
    */
-  constructor(connectionConfig: ConnectionConfig, { logger }: ProviderDependencies) {
+  constructor(
+    connectionConfig: ConnectionConfig,
+    { logger }: ProviderDependencies,
+    handleProvider?: KoraLabsHandleProvider
```
mkazlauskas, line 45: Is there a reason to depend on specific implementation here?
```suggestion
    handleProvider?: HandleProvider
```

iccicci (changes requested):
Excellent!

I like so much the introduction of `withHandlePolicyIdsOptions` and `getNetworkInfoProvider`!

I just suggest a few minor changes.

mkazlauskas (changes requested):

packages/ogmios/src/Provider/TxSubmitProvider/OgmiosTxSubmitProvider.ts
VanessaPC: af17ac42b57672c085e8d89d7f2565a979aaad0c should fix this

packages/cardano-services/src/cli.ts
VanessaPC: af17ac42b57672c085e8d89d7f2565a979aaad0c should fix this

packages/cardano-services/src/cli.ts
VanessaPC: af17ac42b57672c085e8d89d7f2565a979aaad0c should fix this

packages/cardano-services/src/Program/programs/types.ts
VanessaPC: af17ac42b57672c085e8d89d7f2565a979aaad0c should fix this

packages/cardano-services-client/src/TxSubmitProvider/txSubmitHttpProvider.ts
VanessaPC:
```
it('wraps HandleOwnerChangeErrors', async () => {
          axiosMock.onPost().replyOnce(() => {
            throw handleProviderMocks.axiosError(
              new HandleOwnerChangeError(
                '$adaHandle',
                'addr1' as Cardano.PaymentAddress,
                'addr2' as Cardano.PaymentAddress
              )
            );
          });
          const provider = txSubmitHttpProvider(config);
          try {
            await provider.submitTx({ signedTransaction: emptyUintArrayAsHexString });
          } catch (error: any) {
            expect(error.innerError).toBeInstanceOf(HandleOwnerChangeError);
          }
        });
```
A test like this would still expect a `ProviderError` not the `HandleOwnerChangeError` .. so I am not sure how would you test it. Is there an example somewhere?

rhyslbw: Let's decouple this from the current PR

rhyslbw (approved):
Great work @VanessaPC 🚀

iccicci (approved):

mkazlauskas (approved):
Nice work! :rocket:

rhyslbw (approved):

PR #763, opened 2023-06-12
fix: correct ledger mapping canonical asset and asset group ordering

mkazlauskas:
# Context

Some Ledger transactions are failing with

```
"Error: L: Transport failure: Ledger transport failed due to
 L: invalid asset group in multiasset token bundle - incorrect ordering of tokens
    at chrome-extension://gafhhkghbfjjkeiendhlofajokpaflmk/js/353.js:1:177432
    at async chrome-extension://gafhhkghbfjjkeiendhlofajokpaflmk/js/343.js:101:98823"
```

# Proposed Solution

Correct canonical ordering


rhyslbw (approved):
💪

PR #767, opened 2023-06-13
feat: cip 67 utils

lucas-barros:
# Context

Util functions to suppport CIP 67

# Proposed Solution


packages/core/src/Asset/cip67.ts
```diff
@@ -0,0 +1,73 @@
...
+import crc8 from './crc8';
+
+const ASSET_LABEL_LENGTH = 8;
+const ASSET_LABEL_BRACKET = '0';
+
+export enum AssetNameLabelNum {
+  ReferenceNFT = 100,
+  SubStandard222 = 222,
+  SubStandard333 = 333,
+  SubStandard444 = 444
```
mkazlauskas, line 12: Let's use more descriptive names. These labels are better explained in [cip68](https://cips.cardano.org/cips/cip68/)
```suggestion
  UserNFT = 222,
  UserFT = 333,
  UserRFT = 444
```

packages/core/src/Asset/cip67.ts
```diff
@@ -0,0 +1,73 @@
...
+    isInvalidBracket(labelLeadBracket, labelEndBracket) ||
+    isInvalidChecksum(labelNumHex, labelChecksum)
+  ) {
+    return null;
+  }
+
+  return AssetNameLabel(Number.parseInt(labelNumHex, 16));
+};
+
+AssetNameLabel.decode = (assetId: Cardano.AssetId): DecodedAssetName => {
```
mkazlauskas, line 57: I think we should scope this util to work with just the `AssetName`.
Also maybe let's return `null` when it fails to decode (and change the DecodedAssetName.label to non-nullable), because that indicates that it failed to decode - so there's no 'decoded asset name' in that case.
```suggestion
AssetNameLabel.decode = (assetName: Cardano.AssetName): DecodedAssetName | null => {
```

mkazlauskas (changes requested):

packages/core/src/Asset/index.ts
```diff
@@ -1,2 +1,3 @@
 export * as util from './util';
 export * from './types';
+export * as cip67 from './cip67';
```
mkazlauskas, line 3: I take back my previous suggestion about the namespace, I think it's too nested when importing from '@cardano-sdk/core': `Asset.cip67.AssetNameLabel.decode` -> `Asset.AssetNameLabel.decode`
```suggestion
export * from './cip67';
```

lucas-barros: Fixed: https://github.com/input-output-hk/cardano-js-sdk/pull/767/commits/1032606789f812ae16dc643c27e091c58a8d0c09

rhyslbw (approved):
Great work @lucas-barros. Really nice job with the isolated fixups and links left in context, this helps a lot with follow-up reviews 🙌

mkazlauskas (approved):
Great work! :rocket:

PR #770, opened 2023-06-15
Feat/lw 6854 portfolio state maintenance investigation

AngelCastilloB:
# Context

We need to conduct an investigate on how input selection affect the user's stake distribution across `n` stake keys during extended operation. 

Given this investigation we should have some idea of the constraints we will face when a user is transacting and delegating to multiple stake pools.

# Proposed Solution

Create a tool that emulates user operation of a wallet over time and tracks several metrics, transaction fees, UTXO set health and drift of delegation portfolio.

multi-delegation-data-gen


packages/e2e/src/tools/multi-delegation-data-gen/example.json
```diff
@@ -0,0 +1,31 @@
...
+      "min": 1000000000,
+      "max": 1000000000
+    },
+    "count": {
+      "min": 1,
+      "max": 1
+    },
+    "period": 1
+  },
+  "inputs": {
```
mkazlauskas, line 19: This got me confused a little bit, how about:
```suggestion
  "utxoIn": {
    "amount": {
      "min": 1000000000,
      "max": 1000000000
    },
    "count": {
      "min": 1,
      "max": 1
    },
    "period": 1
  },
  "utxoOut": {
```

mkazlauskas (dismissed):
Looking good @AngelCastilloB! One concern is that this is also shipped as a library (`cardano-sdk/e2e` package), and exported (moved) utils don't have direct test coverage. I'd be ok with not having good coverage for this one package, just wanted to point this out so that we can all agree to this.

packages/e2e/src/tools/multi-delegation-data-gen/utils/utils.ts
```diff
@@ -0,0 +1,438 @@
...
+  stakeKeyHash: Cardano.RewardAccount.toHash(rewardAccount)
+});
+
+/**
+ * Creates a delegation certificate for the given reward account and pool id.
+ *
+ * @param rewardAccount The reward account that is delegating.
+ * @param poolId The pool id of the pool we are delegating to.
+ */
+const createDelegationCert = (rewardAccount: Cardano.RewardAccount, poolId: Cardano.PoolId): Cardano.Certificate => ({
```
rhyslbw, line 66: This and `createStakeKeyRegistrationCert` seem like they should be part of a more generic utility, maybe within `tx-construction` :thinking:

rhyslbw (commented):
Looks good @AngelCastilloB, well done

packages/e2e/src/tools/multi-delegation-data-gen/utils/utils.ts
mirceahasegan: Delegate certificate is [already there](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/tx-construction/src/tx-builder/TxBuilder.ts#LL290C32-L290C32), but it's internal private

mirceahasegan (approved):
Fantastic work, @AngelCastilloB! :mage_man:

mirceahasegan (approved):

packages/e2e/src/tools/multi-delegation-data-gen/utils/utils.ts
rhyslbw: Yeah, even better

mirceahasegan (approved):

rhyslbw (approved):
Ready for squashing

PR #771, opened 2023-06-19
fix: the transaction id computation now accounts for serialization round trip errors

AngelCastilloB:
# Context

Because of how CBOR works. There are multiple ways in CBOR to represent the same data. Unfortunately right now the CML  doesn't maintain the serialization format upon using from_bytes so if cardano-cli uses non-canonical CBOR this will result in different hashes.

# Proposed Solution

Maintain the original the serialized bytes of the body, and use that to compute the hash instead of re serializing (when possible).

Added a new method to the transaction class 'getId' which always compute the hash of the transaction body using the original bytes found in the transaction (if available). Please note that if changes to the transaction body are made, this cache will be invalidated.

### DIFF
--- a/packages/core/test/Serialization/Transaction.test.ts
+++ b/packages/core/test/Serialization/Transaction.test.ts
@@ -121,13 +121,19 @@ describe('Transaction', () => {
     expect(scope.manage(tx3.body()).to_bytes()).not.toEqual(scope.manage(tx.body()).to_bytes());
   });
 
-  it('can compute the right TX Id', () => {
+  it('can compute the right Tx ID', () => {
     expect(Transaction.fromCbor(HexBlob(TX)).getId()).toEqual(
       '856c8bc6ce3725188b496d62fa389f2beff2f701e6d35af39d3f3464bbce0cec'
     );
 
-    // This transaction was created with the CLI and will cause a round trip different in the body.
-    // If we can compute the right TX Id it means we are handling the round trip difference correctly.
+    // This transaction was created with the CLI and will cause a round trip difference in the body.
+    // If we can compute the right Tx ID it means we are handling the round trip difference correctly.
+    //
+    // The correct hash can be verified using the cardano-cli as follows:
+    //
+    // cardano-cli transaction txid --tx-file ./tx  <-- tx contains the CBOR of the transaction.
+    //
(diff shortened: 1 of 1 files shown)


packages/core/test/Serialization/Transaction.test.ts
```diff
@@ -117,4 +120,16 @@ describe('Transaction', () => {
     expect(scope.manage(tx3.body()).to_bytes()).toEqual(scope.manage(referenceTx.body()).to_bytes());
     expect(scope.manage(tx3.body()).to_bytes()).not.toEqual(scope.manage(tx.body()).to_bytes());
   });
+
+  it('can compute the right TX Id', () => {
```
rhyslbw, line 124:
```suggestion
  it('can compute the right Tx ID', () => {
```

packages/core/test/Serialization/Transaction.test.ts
```diff
@@ -117,4 +120,16 @@ describe('Transaction', () => {
...
+    expect(Transaction.fromCbor(HexBlob(TX)).getId()).toEqual(
+      '856c8bc6ce3725188b496d62fa389f2beff2f701e6d35af39d3f3464bbce0cec'
+    );
+
+    // This transaction was created with the CLI and will cause a round trip different in the body.
+    // If we can compute the right TX Id it means we are handling the round trip difference correctly.
```
rhyslbw, line 130:
```suggestion
    // If we can compute the right Tx ID it means we are handling the round trip difference correctly.
```

packages/core/test/Serialization/Transaction.test.ts
```diff
@@ -117,4 +120,16 @@ describe('Transaction', () => {
...
+    expect(Transaction.fromCbor(HexBlob(CLI_TX)).getId()).toEqual(
+      '2d7f290c815e061fb7c27e91d2a898bd7b454a71c9b7a26660e2257ac31ebe32'
```
rhyslbw, line 132: Please add a comment here to explain how this value can be verified

rhyslbw (changes requested):

rhyslbw (approved):

mirceahasegan (approved):
Nice! :rocket:

iadmytro (approved):

PR #772, opened 2023-06-19
Refactor: removes networkInfoProvider dependency from KoralabHandleProvider

VanessaPC:
# Context
The current implementation of KoralabsHandleProvider depends on the network info provider to return a `resolvedAt` property. However, this is not returned by KoralabsHandleProvider, so we are making that property option but not returning it as the API doesn't have it. 

# Proposed Solution
- [x] remove the network info provider dependency from KoralabsHandleProvider
- [x] make `resolvedAt` property, optional. 

N/A


iccicci (approved):
LGTM

rhyslbw (commented):
Looks good. Let's just wait to rebase once #750 is merged

rhyslbw (approved):
Ready for squashing

iccicci (approved):

PR #773, opened 2023-06-20
feat: map out optional properties from Koralab api response

greatertomi:
# Context

Mapped out background image and profile picture from koralab api response

# Proposed Solution

### DIFF
--- a/packages/cardano-services-client/src/HandleProvider/KoraLabsHandleProvider.ts
+++ b/packages/cardano-services-client/src/HandleProvider/KoraLabsHandleProvider.ts
@@ -44,10 +44,12 @@ export const toHandleResolution = ({
   apiResponse: IHandle;
   policyId: Cardano.PolicyId;
 }): HandleResolution => ({
+  backgroundImage: Asset.Uri(`ipfs://${apiResponse.background}`),
   cardanoAddress: Cardano.PaymentAddress(apiResponse.resolved_addresses.ada),
   handle: apiResponse.name,
   hasDatum: apiResponse.hasDatum,
-  policyId
+  policyId,
+  profilePic: Asset.Uri(`ipfs://${apiResponse.profile_pic}`)
 });
 
 /**
--- a/packages/cardano-services-client/test/util.ts
+++ b/packages/cardano-services-client/test/util.ts
@@ -24,23 +24,27 @@ export const healthCheckResponseWithState = {
 };
 
 export const getAliceHandleProviderResponse = {
+  backgroundImage: 'ipfs://zrljm7nskakjydxlr450ktsj08zuw6aktvgfkmmyw9semrkrezryq3bd',
   cardanoAddress:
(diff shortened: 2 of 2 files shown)


rhyslbw (approved):
Thanks @greatertomi

packages/cardano-services-client/src/HandleProvider/KoraLabsHandleProvider.ts
```diff
@@ -44,10 +44,12 @@ export const toHandleResolution = ({
   apiResponse: IHandle;
   policyId: Cardano.PolicyId;
 }): HandleResolution => ({
+  backgroundImage: Asset.Uri(`ipfs://${apiResponse.background}`),
```
VanessaPC, line 47: I have noticed that the interface added for HandleResolution here, at the top of the file matches the HandleResolution from the HandleProvider, the native one... but not the KoralabsType in the `core/provider/handleProvider/types.ts` .. so I am unsure this works... I think that they do return images though, but the interface may need to change the type here in that package, and then we can map the properties to return the same names as on the other provider if we want.. 
[KoraLabs](https://github.com/koralabs/handles-public-api-interfaces/blob/master/index.ts#LL94C1-L94C1) interface

VanessaPC (approved):
LGTM!

packages/cardano-services-client/src/HandleProvider/KoraLabsHandleProvider.ts
lucas-barros, line 47: The handle image is returned with the ipfs protocol prefix. It would make sense that `profilePic` and `backgroundImage` would too.  Do you know if this could result in` ipfs://ipfs://...`?

lucas-barros (changes requested):

packages/cardano-services-client/src/HandleProvider/KoraLabsHandleProvider.ts
rhyslbw: @lucas-barros there's a different between `image`, `backgroundImage` and `profilePic`, as the latter two are from the custom metadata, whereas the former is the standard NFT metadata.

lucas-barros: @rhyslbw makes sense. I see the scheme on public API specifies it as only the ipfs hash.

lucas-barros (approved):

packages/cardano-services-client/src/HandleProvider/KoraLabsHandleProvider.ts
greatertomi: @rhyslbw Actually the API response was already in the codebase [here](https://github.com/input-output-hk/cardano-js-sdk/blob/1efd9862686136fd5c106ddc56e126e8b0da8868/packages/cardano-services-client/test/util.ts#L46-L67). I confirmed that it matches the data in the api [docs](https://koralabs.github.io/handles-public-api/#/handles/get_handles__handle_).

PR #775, opened 2023-06-21
Test/fixtures

iadmytro:
# Context

Updated test fixtures


packages/cardano-services/test/Asset/DbSyncAssetProvider.test.ts
```diff
@@ -104,10 +104,10 @@ describe('DbSyncAssetProvider', () => {
 
     expect(asset.history).toEqual(history);
     // TODO: review test data, this is false positive right now
-    // expect(asset.nftMetadata).toBeTruthy();
+    expect(asset.nftMetadata).toBeTruthy();
     expect(asset.nftMetadata).toStrictEqual(assets[0].metadata);
     expect(asset.tokenMetadata).toStrictEqual({
-      assetId: '728847c7898b06f180de05c80b37d38bf77a9ea22bd1e222b8014d964e46542d66696c6573',
+      assetId: '17ebe33f8aeee1fe9a7277fe0dc02261531a896a8b89457895fe60294349502d303032352d7632',
```
iccicci, line 110: It is better to read this `assetId` from the DB in the program, otherwise next time we generate new fixtures this will fail again.

iccicci (changes requested):
Good job! Just one more minor improvement request

packages/core/src/Provider/providerUtil.ts
```diff
@@ -15,6 +15,7 @@ export const withProviderErrors = <T extends {}>(providerImplementation: T, toPr
   }, {} as any) as T;
 
 const tryParseBigIntKey = (key: string) => {
+  if (key.startsWith('0x')) return key.slice(2);
```
iccicci, line 18: Better to do a separate commit for this and adding an inline comment to explain why; if I remember correctly to support CIP-005 v2

iccicci (changes requested):
Sorry, one more comment

iccicci (approved):
GJ!

packages/cardano-services/test/Asset/DbSyncAssetProvider.test.ts
rhyslbw: Why is this resolved @iccicci ? Seems like legitimate feedback

rhyslbw: Oh, it's mocked :+1:

packages/cardano-services/test/ChainHistory/ChainHistoryHttpService.test.ts
```diff
@@ -429,7 +429,7 @@ describe('ChainHistoryHttpService', () => {
 
       it('includes transactions only in specified block range', async () => {
         const { addresses, blockRange, txInRangeCount } = await fixtureBuilder.getAddressesWithSomeInBlockRange(2, {
-          lowerBound: Cardano.BlockNo(200),
+          lowerBound: Cardano.BlockNo(600),
```
rhyslbw, line 432: This could be handled dynamically with a fixture, but let's not get bogged down here.

rhyslbw (approved):
Thanks @iadmytro

PR #776, opened 2023-06-22
Test/lw 6702 e2e webextension test multi delegation

mirceahasegan:
# Context
Add web-extension e2e test for multi delegation

image
image


rhyslbw (approved):
Nice work scoping the commits @mirceahasegan :star_struck:

packages/e2e/test/web-extension/specs/wallet.spec.ts
```diff
@@ -105,11 +114,49 @@ describe('wallet', () => {
...
       });
+
+      it('can delegate to multiple pools', async () => {
+        (await $(btnDelegate)).click();
+
+        // There should be 3 pools available
+        await browser.waitUntil(async () => {
+          try {
+            const poolIds = await $(spanPoolIds).getText();
+            return poolIds.split(' ').length > 1;
```
mkazlauskas, line 132: We're expecting wallet to be delegating to exactly 3 pools later, so it would be better to fail early here when this is the cause. No need to change it, just an idea.
```suggestion
            return poolIds.split(' ').length === NUM_POOLS;
```

mkazlauskas (approved):
Great work! :muscle:

packages/e2e/test/web-extension/specs/wallet.spec.ts
rhyslbw: Are you planning to implement this @mirceahasegan? If not, feel free to merge

mirceahasegan: Yeah, I implemented it.I'll force push as it's a small change :smile:  and I want to avoid 2 CI cycles

rhyslbw (approved):

PR #777, opened 2023-06-23
fix: unsupported character bug in handles projection

greatertomi:
# Context

Update the projection not return handles with unsupported characters

# Proposed Solution

### DIFF
--- a/packages/projection-typeorm/test/operators/storeHandles.test.ts
+++ b/packages/projection-typeorm/test/operators/storeHandles.test.ts
@@ -1,3 +1,4 @@
+import { Asset, Cardano, ChainSyncEventType } from '@cardano-sdk/core';
 import {
   AssetEntity,
   BlockDataEntity,
@@ -14,7 +15,6 @@ import {
   withTypeormTransaction
 } from '../../src';
 import { BaseProjectionEvent, Bootstrap, Mappers, ProjectionEvent, requestNext } from '@cardano-sdk/projection';
-import { Cardano, ChainSyncEventType } from '@cardano-sdk/core';
 import { ChainSyncDataSet, chainSyncData, logger, mockProviders } from '@cardano-sdk/util-dev';
 import { Observable, defer, firstValueFrom, from } from 'rxjs';
 import { QueryRunner } from 'typeorm';
@@ -373,6 +373,44 @@ describe('storeHandles', () => {
         cardanoAddress: maryAddress,
         handle
       });
+    });
+
+    it('ignores assets minted under the policy with a cip67 UserFT label', async () => {
+      const mintTxId = Cardano.TransactionId('0000000000000000000000000000000000000000000000000000000000000000');
+      const repository = queryRunner.manager.getRepository(HandleEntity);
+      const invalidAssetName = Asset.AssetNameLabel.encode(Cardano.AssetName('ace'), Asset.AssetNameLabelNum.UserFT);
+      const invalidAssetId = Cardano.AssetId.fromParts(mockProviders.handlePolicyId, invalidAssetName);
+      const testAddress = Cardano.PaymentAddress(
+        'addr_test1qz690wvatwqgzt5u85hfzjxa8qqzthqwtp7xq8t3wh6ttc98hqtvlesvrpvln3srklcvhu2r9z22fdhaxvh2m2pg3nuq0n8gf2'
+      );
+      const newHandle = 'test_me_0001';
+
+      const source$ = createMultiTxProjectionSource([
+        {
+          body: {
+            fee: 111n,
+            inputs: [],
+            mint: new Map([[invalidAssetId, 1n]]),
+            outputs: [
+              {
+                address: testAddress,
+                value: {
+                  assets: new Map([[invalidAssetId, 1n]]),
+                  coins: 123n
+                }
+              }
+            ]
+          },
+          id: mintTxId
+        }
+      ]);
+      await firstValueFrom(source$.pipe(applyOperators()));
+      expect(
+        await repository.findOne({
+          select: { cardanoAddress: true, handle: true },
+          where: { handle: newHandle }
+        })
+      ).toEqual(null);
     });
   });
 });
--- a/packages/projection/src/operators/Mappers/withHandles.ts
+++ b/packages/projection/src/operators/Mappers/withHandles.ts
@@ -1,5 +1,8 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import { Cardano } from '@cardano-sdk/core';
 import { FilterByPolicyIds } from './types';
+import { HandleParsingError, HexBlob } from '@cardano-sdk/util';
+import { Logger, dummyLogger } from 'ts-log';
 import { ProjectionOperator } from '../../types';
 import { map } from 'rxjs';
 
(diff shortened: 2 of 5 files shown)


packages/projection-typeorm/test/operators/storeHandles.test.ts
```diff
@@ -374,5 +374,44 @@ describe('storeHandles', () => {
...
+          id: mintTxId
+        }
+      ]);
+      await firstValueFrom(source$.pipe(applyOperators()));
+      expect(
+        await repository.findOne({
+          select: { cardanoAddress: true, handle: true },
+          where: { handle: newHandle }
+        })
+      ).toEqual(null);
```
lucas-barros, line 414: I believe the test case from the task should actually be saved in the projection, but the asset name should be decoded first.
You can use the decode function from packages/core/src/Asset/cip67.ts to extract the asset name content and validate against that value.

greatertomi: I raised this issue with @rhyslbw, but he didn't agree.

@rhyslbw thoughts?

greatertomi: This is exactly the approach I suggested. :)

packages/projection/src/operators/Mappers/withHandles.ts
```diff
@@ -25,21 +27,34 @@ export interface WithHandles {
 const handleFromAssetId = (assetId: Cardano.AssetId) =>
   Buffer.from(Cardano.AssetId.getAssetName(assetId), 'hex').toString('utf8');
 
+const logger: Logger = dummyLogger;
+
+const isValidHandle = (handle: string) => {
+  const pattern = /^[\w.-]+$/;
```
rhyslbw, line 33: Supported characters are `-._ 0-9 a-z`

packages/projection/test/operators/Mappers/withHandles.test.ts
```diff
@@ -214,4 +214,105 @@ describe('withHandles', () => {
       expect(handles[0].latestOwnerAddress).toBe(null);
     });
   });
+
+  describe('assets with invalid asset names', () => {
+    const invalidAssetId = Cardano.AssetId(
```
rhyslbw, line 219: Same feedback as here

packages/projection/src/operators/Mappers/withHandles.ts
```diff
@@ -25,21 +27,34 @@ export interface WithHandles {
...
+
+        handles[handle] = {
+          assetId,
+          datum,
+          handle,
+          latestOwnerAddress: address,
+          policyId
+        };
+      } catch {
+        logger.warn('Error parsing handle from assetId', { assetId });
```
rhyslbw, line 56:
```suggestion
        logger.error(error.message);
```
Now that I see the implementation I think we can just log it as an error, since it's unexpected for the protocol.

packages/projection/src/operators/Mappers/withHandles.ts
```diff
@@ -53,13 +68,19 @@ const getBurnedHandles = (mint: Cardano.TokenMap | undefined, policyIds: FilterB
...
       if (!policyIds.includes(policyId)) continue;
-      const handle = handleFromAssetId(assetId);
-      handles[handle] = {
-        assetId,
-        handle,
-        latestOwnerAddress: null,
-        policyId
-      };
+      try {
+        const handle = handleFromAssetId(assetId);
```
rhyslbw, line 72: Seems we have some code duplication here. Please hoist to DRY the code

rhyslbw (changes requested):

packages/projection-typeorm/test/operators/storeHandles.test.ts
rhyslbw: The idea is not to add support for cip67 NFTs (yet) but solidify the existing implementation to be resilient to unexpected `assetNames`. I've [suggested a change](https://github.com/input-output-hk/cardano-js-sdk/pull/777#discussion_r1239699877) to make the test relevant once we add support for the new format though

lucas-barros: Ok, understood.

packages/projection/src/operators/Mappers/withHandles.ts
greatertomi: Yes. That regex covers these cases. This is Eslint optimized version. 😅

packages/projection-typeorm/test/operators/storeHandles.test.ts
```diff
@@ -374,5 +374,44 @@ describe('storeHandles', () => {
         handle
       });
     });
+
+    it('handles error gracefully when minting an assetName in cip67 format', async () => {
+      const mintTxId = Cardano.TransactionId('0000000000000000000000000000000000000000000000000000000000000000');
+      const repository = queryRunner.manager.getRepository(HandleEntity);
+      const invalidAssetId = Cardano.AssetId(
```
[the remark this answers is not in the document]
greatertomi: Fixed on ff85341

packages/projection/test/operators/Mappers/withHandles.test.ts
greatertomi: Fixed on ff85341

packages/projection-typeorm/test/operators/storeHandles.test.ts
```diff
@@ -374,5 +374,44 @@ describe('storeHandles', () => {
         handle
       });
     });
+
+    it('handles error gracefully when minting an assetName in cip67 format', async () => {
```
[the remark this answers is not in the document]
greatertomi: Fixed on e9954fc

packages/projection/src/operators/Mappers/withHandles.ts
greatertomi: Fixed on e9954fc

packages/projection/src/operators/Mappers/withHandles.ts
greatertomi: Fixed on 225763b

packages/util/src/errors.ts
```diff
@@ -101,3 +101,9 @@ export class InvalidArgumentError extends CustomError {
     super(`Invalid argument '${argName}': ${message}`);
   }
 }
+
+export class HandleParsingError extends CustomError {
```
mkazlauskas, line 105: This seems out of place, should be in `projection` package. I suggest defining it in the same file where it's thrown.

packages/projection/test/operators/Mappers/withHandles.test.ts
```diff
@@ -214,4 +215,104 @@ describe('withHandles', () => {
...
+          withHandles({
+            policyIds: [handlePolicyId]
+          })
+        )
+      );
+
+      expect(handles.length).toBe(1);
+    });
+
+    it('it returns only valid handles when over multiple outputs ', async () => {
```
mkazlauskas, line 289:
```suggestion
    it('it returns only valid handles when block multiple transactions with handle outputs', async () => {
```

packages/projection/src/operators/Mappers/withHandles.ts
```diff
@@ -25,21 +28,43 @@ export interface WithHandles {
...
-        datum,
-        handle,
-        latestOwnerAddress: address,
-        policyId
-      };
+      try {
+        const handleData = mapHandleToData(assetId, { address, datum, policyId });
+        handles[handleData.handle] = handleData;
+      } catch (error: any) {
+        logger.error(error.message);
```
mkazlauskas, line 66: This is a good practice because technically you can throw anything in js. Same in `getBurnedHandles` - probably a good idea to create a new `logError` util function.
```suggestion
      } catch (error: unknown) {
        const message = isError(error) ? error.message : 'Internal error: ' + error;
        logger.error(message);
```

Might no longer need a file-level `/* eslint-disable @typescript-eslint/no-explicit-any */` too then

mkazlauskas (changes requested):

packages/util/src/errors.ts
greatertomi: Fixed on 6c65203

packages/projection/test/operators/Mappers/withHandles.test.ts
greatertomi: Fixed on 6c65203

packages/projection/src/operators/Mappers/withHandles.ts
greatertomi: Fixed on 6c65203

mkazlauskas (approved):
Nice work!

lucas-barros (approved):

rhyslbw (approved):
Great work @greatertomi :muscle:

PR #778, opened 2023-06-23
Test/lw 6857 projector ogmios reconnection

iccicci:
# Context

We need to be sure the projection service is able to reconnect to ogmios in case of problem of the latter.

# Proposed Solution

Added some test cases to e2e tests suite

Added a bind volume to our local network docker compose infrastructure to let containers share files between them and with the extern.

# Important note

Not all tests pass. Some follow up is required.

### DIFF
--- a/packages/e2e/test/long-running/cache-invalidation.test.ts
+++ b/packages/e2e/test/long-running/cache-invalidation.test.ts
@@ -47,7 +47,7 @@ describe('cache invalidation', () => {
 
   afterAll(() => wallet1.wallet.shutdown());
 
-  test('cache is invalidated on epoch rollover', async () => {
+  test.skip('cache is invalidated on epoch rollover', async () => {
     const wallet = wallet1.wallet;
 
     await walletReady(wallet);
--- a/packages/e2e/.env.example
+++ b/packages/e2e/.env.example
@@ -47,3 +47,6 @@ WORKER_PARALLEL_TRANSACTION=3
 
 # Required by test:providers
 STAKE_POOL_PROVIDER_URL="http://localhost:4000/stake-pool"
+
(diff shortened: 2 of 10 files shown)


packages/util-dev/src/util.ts
```diff
@@ -1,2 +1,4 @@
 export const flushPromises = (setImmediate = global.setImmediate) =>
   new Promise((resolve) => setImmediate(resolve, void 0));
+
+export const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
```
mkazlauskas, line 4: We're already using [delay](https://www.npmjs.com/package/delay) package for that

packages/e2e/test/long-running/cache-invalidation.test.ts
```diff
@@ -47,7 +47,7 @@ describe('cache invalidation', () => {
 
   afterAll(() => wallet1.wallet.shutdown());
 
-  test('cache is invalidated on epoch rollover', async () => {
+  test.skip('cache is invalidated on epoch rollover', async () => {
```
mkazlauskas, line 50: Why do we need to skip an existing test without any code changes? :thinking:

mkazlauskas (changes requested):

packages/util-dev/src/util.ts
iccicci: f3e0b8296 and de03eca3c should fix this

VanessaPC (approved):

VanessaPC:
LGTM!

rhyslbw (approved):
Nice work @iccicci

iadmytro (approved):

PR #781, opened 2023-06-26
Fix/software version source

rhyslbw:
# Context
- Software version source doesn't need to be separately maintained
- [The comment](https://github.com/input-output-hk/cardano-js-sdk/compare/fix/software-version-source?expand=1#diff-927718184f485156899e2d7eab22b7b4b1e7b47d5bb4d766cf3a75bfcdad9ce3L1) uses an absolute path to identify the generating script

# Proposed Solution
- Use the package.json version as a source
- Hard-code the script path

- Rename the version.ts export
- Copies `./cardano-services-client/package.json` into the `dist`

### DIFF
--- a/packages/cardano-services-client/scripts/createVersionSource.js
+++ b/packages/cardano-services-client/scripts/createVersionSource.js
@@ -1,14 +1,10 @@
 const fs = require('fs');
 const path = require('path');
 
-const packageJson = require('../package.json');
 const openApiSpec = require('../../cardano-services/src/Http/openApi.json');
 
 const contents = `// auto-generated using ../scripts/createVersionSource.js
-export const version = {
-  api: '${openApiSpec.info.version}',
-  software: '${packageJson.version}'
-};
+export const apiVersion = '${openApiSpec.info.version}';
 `;
 
 fs.writeFileSync(path.join(__dirname, '../src/version.ts'), contents, { encoding: 'utf8', flag: 'w' });
--- a/packages/cardano-services-client/scripts/createVersionSource.js
+++ b/packages/cardano-services-client/scripts/createVersionSource.js
@@ -4,7 +4,7 @@ const path = require('path');
 const packageJson = require('../package.json');
 const openApiSpec = require('../../cardano-services/src/Http/openApi.json');
 
-const contents = `// auto-generated using ${__filename}
+const contents = `// auto-generated using ../scripts/createVersionSource.js
 export const version = {
   api: '${openApiSpec.info.version}',
   software: '${packageJson.version}'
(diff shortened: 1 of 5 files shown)


packages/cardano-services-client/scripts/createVersionSource.js
```diff
@@ -1,14 +1,10 @@
...
-const packageJson = require('../package.json');
 const openApiSpec = require('../../cardano-services/src/Http/openApi.json');
 
-const contents = `// auto-generated using ${__filename}
-export const version = {
-  api: '${openApiSpec.info.version}',
-  software: '${packageJson.version}'
-};
+const contents = `// auto-generated using ../scripts/createVersionSource.js
+export const apiVersion = '${openApiSpec.info.version}';
```
rhyslbw, line 7: Note this will likely change to use individual provider versions, as this is currently just using the top level HTTP server schema, but it's out of scope for this PR.

packages/cardano-services-client/src/HttpProvider.ts
```diff
@@ -1,9 +1,12 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { HttpProviderConfigPaths, Provider, ProviderError, ProviderFailure } from '@cardano-sdk/core';
 import { Logger } from 'ts-log';
+import { apiVersion } from './version';
 import { fromSerializableObject, toSerializableObject } from '@cardano-sdk/util';
-import { version } from './version';
 import axios, { AxiosAdapter, AxiosRequestConfig, AxiosResponseTransformer } from 'axios';
+import path from 'path';
+
+const packageJson = require(path.join(__dirname, '..', 'package.json'));
```
mkazlauskas, line 9: I think that won't work with built package, because it's more nested: `dist/cjs/HttpProvider.js` would resolve to `dist/cjs`

mkazlauskas (changes requested):

packages/cardano-services-client/src/version.ts
```diff
@@ -1,5 +1,2 @@
-// auto-generated using /home/rhys/code/iohk/cardano-js-sdk/packages/cardano-services-client/scripts/createVersionSource.js
-export const version = {
-  api: '1.0.0',
-  software: '0.9.9'
-};
+// auto-generated using ../scripts/createVersionSource.js
+export const apiVersion = '1.0.0';
```
mkazlauskas, line 2: Since we're using package.json as source for software version, perhaps it would be better to patch package.json and have versions contained in 1 file?

rhyslbw: Let's do this in the next iteration? I'll need to update this implementation to consider all provider versions.

packages/cardano-services-client/src/HttpProvider.ts
rhyslbw: `path.join(__dirname, '..', 'package.json')` resolves to `dist/package.json`

mkazlauskas: It's different when it's running with `ts-node` from `src/HttpProvider.ts`

mirceahasegan (approved):

mkazlauskas (approved):

PR #784, opened 2023-06-27
Handle policy id to projector - step 1

iccicci:
# Context

We need to make the handle projector in the local testnet able to project handles minted in network startup.

# Proposed Solution

- Added an option to the cli to read the handle policy ids from a file
- Applied some changes to the local network to propagate the handle policy id to the handle projector

# Important Notes

This PR introduced some merge conflicts with #761 , since the changes in the latter are better organized, I strongly suggest to merge the other PR before this one.


rhyslbw (dismissed):
Nice solution @iccicci 🚀

packages/cardano-services/src/cli.ts
```diff
@@ -166,7 +166,8 @@ withCommonOptions(
         ...args,
         postgresConnectionString: connectionStringFromArgs(args, ''),
         // Setting the projection names via env variable takes preference over command line argument
-        projectionNames: args.projectionNames ? args.projectionNames : projectionNames
+        projectionNames: args.projectionNames ? args.projectionNames : projectionNames,
+        ...handlePolicyIdsFromFile(args)
```
lucas-barros, line 170: It seems `handlePolicyIdsFromFile` is called here and also inside `loadProjector` function. Should it be called just once? Also since it returns a promise instead of modified args, using spread operator might have unexpected result.

lucas-barros (changes requested):

packages/cardano-services/src/cli.ts
iccicci: Yep, it's a refuse. Thank you

iccicci: 2def8b682 should solve this

rhyslbw:
Ready for squashing @iccicci

rhyslbw (approved):

PR #787, opened 2023-06-28
lw-6694: implement dynamic change allocation

AngelCastilloB:
# Context
Currently a user’s delegation preferences do not last very long if they start spending frequently from the wallet. 

We need to update the round robin input-selection algorithm that maintains the user’s stake distribution over a longer time frame while the user is spending from the wallet (spending can be separated into two scenarios (1.) only spending out of the wallet (2.) spending out of the wallet and receiving to the wallet)

# Proposed Solution

Add an extension point to the Round Robin Input selection to select the strategy for change output distribution among stake addresses.

Distributing change is related to a class of problems called "multiway number partitioning", which is NP-hard. If the problem
size is small (such as is our case), more exact methods such as dynamic programming or integer programming could be used,
but since this is still a POC, for simplicity’s sake, we are implementing a simplified and iterative approach to
solve this:

1. Sort the change list in descending order. The reason to sort is to allocate the largest change outputs first to the 'buckets' with the largest gap, and the smaller outputs to the buckets with smaller gaps, reducing wasted change (over flow).
2. For each change output, add it to the bucket that currently has the largest gap and is not overflowed.
3. Repeat until all change outputs are in a bucket.

This will yield a reasonable approximation to the optimal distribution of change, giving priority to buckets with larger
gaps, but at the same time making sure we are not 'wasting' change by overflowing a single bucket too much.

1. Round robin input selection now takes a ChangeAddressResolver object to update change outputs target addresses.
2. The multi-delegation-data-gen tool can now be configured to use the static change address resolver or the new dynamic change address resolver.

### DIFF
--- a/packages/e2e/package.json
+++ b/packages/e2e/package.json
@@ -86,6 +86,7 @@
     "@cardano-sdk/core": "workspace:~",
     "@cardano-sdk/crypto": "workspace:~",
     "@cardano-sdk/hardware-ledger": "workspace:~",
+    "@cardano-sdk/input-selection": "workspace:~",
     "@cardano-sdk/key-management": "workspace:~",
     "@cardano-sdk/ogmios": "workspace:~",
     "@cardano-sdk/tx-construction": "workspace:~",
--- a/packages/e2e/src/tools/multi-delegation-data-gen/example.json
+++ b/packages/e2e/src/tools/multi-delegation-data-gen/example.json
@@ -1,4 +1,5 @@
 {
+  "changeAddressResolver": "static",
   "startingFunds": 1000000000000,
   "stakeDistribution": [
     0.2,
--- a/packages/e2e/src/tools/multi-delegation-data-gen/index.ts
+++ b/packages/e2e/src/tools/multi-delegation-data-gen/index.ts
@@ -18,7 +18,8 @@ import {
 } from './utils';
 
 import { Cardano } from '@cardano-sdk/core';
-import { PersonalWallet } from '@cardano-sdk/wallet';
+import { DynamicChangeAddressResolver, PersonalWallet } from '@cardano-sdk/wallet';
+import { roundRobinRandomImprove } from '@cardano-sdk/input-selection';
 import { walletReady } from '../../';
 import chalk from 'chalk';
 
@@ -71,9 +72,15 @@ const monitor = new TerminalProgressMonitor();
 
     const stakeAddresses = await generateStakeAddresses(delegationWallet, config.stakeDistribution.length, monitor);
 
-    await delegateToMultiplePools(delegationWallet, stakeAddresses, monitor);
-
-    await distributeStake(delegationWallet, config.startingFunds, config.stakeDistribution, stakeAddresses, monitor);
+    const pools = await delegateToMultiplePools(delegationWallet, stakeAddresses, monitor);
+    const portfolio = await distributeStake(
+      delegationWallet,
+      config.startingFunds,
+      config.stakeDistribution,
+      stakeAddresses,
+      monitor,
+      pools
+    );
 
     monitor.startTask('Waiting for delegation to be updated on the wallet.');
     await rewardAccountStatuses(delegationWallet.delegation.rewardAccounts$, [
@@ -81,6 +88,21 @@ const monitor = new TerminalProgressMonitor();
       Cardano.StakeKeyStatus.Registered
     ]);
     monitor.endTask('Delegation updated.', TaskResult.Success);
+
+    if (config.changeAddressResolver === 'dynamic') {
+      monitor.logInfo(`Using dynamic change address resolution: Portfolio: ${JSON.stringify(portfolio)}`);
+      const selector = roundRobinRandomImprove({
+        changeAddressResolver: new DynamicChangeAddressResolver(
+          delegationWallet.addresses$,
+          delegationWallet.delegation.distribution$,
+          async () => portfolio,
+          console
+        )
+      });
+      delegationWallet.setInputSelector(selector);
+    } else {
+      monitor.logInfo('Using static change address resolution');
+    }
 
     monitor.logInfo('Setup phase ended. Starting interactions...');
 
--- a/packages/e2e/src/tools/multi-delegation-data-gen/utils/config.ts
+++ b/packages/e2e/src/tools/multi-delegation-data-gen/utils/config.ts
@@ -23,6 +23,14 @@ convict.addFormat({
   }
 });
 
+convict.addFormat({
+  name: 'changeAddressResolver',
+  validate(val) {
+    if (val !== 'static' && val !== 'dynamic')
+      throw new TypeError('changeAddressResolver must be set to either static or dynamic');
+  }
+});
+
 export type ValueTransferConfig = {
   amount: { max: number; min: number };
   count: { max: number; min: number };
@@ -30,6 +38,11 @@ export type ValueTransferConfig = {
 };
 
 export const configLoader = convict({
+  changeAddressResolver: {
+    default: 'static',
+    doc: 'Whether to use the static or dynamic strategy to resolve change addresses.',
+    format: 'changeAddressResolver'
+  },
   iterations: {
     default: 10,
     doc: 'How many iterations will be executed before stopping the test.',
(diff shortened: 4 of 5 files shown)


packages/input-selection/src/ChangeAddress/StaticChangeAddressResolver.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export class StaticChangeAddressResolver implements ChangeAddressResolver {
+  readonly #addresses$: Observable<GroupedAddress[]>;
+
+  /**
+   * Initializes a new instance of the StaticChangeAddressResolver.
+   *
+   * @param addresses$ The observable with the known addresses.
+   */
+  constructor(addresses$: Observable<GroupedAddress[]>) {
```
mkazlauskas, line 18: `input-selection` doesn't have a dependency on `rxjs` and it would be great to keep it that way
```suggestion
  constructor(getAddresses: () => Promise<GroupedAddress[]>) {
```

packages/wallet/src/services/ChangeAddress/DynamicChangeAddressResolver.ts
```diff
@@ -0,0 +1,369 @@
...
+  /**
+   * Initializes a new instance of the StakeDistributionChangeAddressResolver.
+   *
+   * @param addresses$ The wallet known addresses.
+   * @param delegationTracker The wallet delegation tracker.
+   * @param getDelegationPortfolio The current delegation portfolio.
+   */
+  constructor(
+    addresses$: Observable<GroupedAddress[]>,
+    delegationTracker: DelegationTracker,
```
mkazlauskas, line 300: Consider narrowing this down
```suggestion
    distribution$: DelegationTracker['distribution$'],
```

mkazlauskas (changes requested):
Looks great! :muscle:

52e7cc87dbab1feafb86098f3735da03adfe4bd6 is a breaking change due to updating `RoundRobinRandomImproveOptions`

packages/input-selection/src/ChangeAddress/StaticChangeAddressResolver.ts
```diff
@@ -0,0 +1,40 @@
...
+    this.#addresses$ = addresses$;
+  }
+
+  /**
+   * Always resolves to the same address.
+   */
+  async resolve(selection: Selection): Promise<Array<Cardano.TxOut>> {
+    const updatedSelection = { ...selection };
```
mirceahasegan, line 26: I don't think that creating the `updatedSelection` is needed. You only need to calculate the change based on the input `selection.change`:
```diff
-    const updatedSelection = { ...selection };
    const groupedAddresses = await firstValueFrom(this.#addresses$);
....
+ return updatedSelection.change.map((txOut) => ({...txout, address}));
```

packages/input-selection/test/util/index.ts
```diff
@@ -7,3 +8,13 @@ export * as SelectionConstraints from './selectionConstraints';
 export const asAssetId = (x: string): Cardano.AssetId => x as unknown as Cardano.AssetId;
 export const asPaymentAddress = (x: string): Cardano.PaymentAddress => x as unknown as Cardano.PaymentAddress;
 export const asTokenMap = (elements: Iterable<[Cardano.AssetId, bigint]>) => new Map<Cardano.AssetId, bigint>(elements);
+
+export class MockChangeAddressResolver implements ChangeAddressResolver {
+  async resolve(selection: Selection) {
+    return selection.change.map((txOut) => {
+      txOut.address =
```
mirceahasegan, line 15: This also updates the input param `selection.change` as a side effect. Is this the intended behavior?

packages/tx-construction/test/tx-builder/mocks.ts
```diff
@@ -0,0 +1,12 @@
+import { Cardano } from '@cardano-sdk/core';
```
mirceahasegan, line 1: I think we can import the similar mock defined in input-selection/test/util and avoid redefining here

packages/wallet/src/services/ChangeAddress/DynamicChangeAddressResolver.ts
```diff
@@ -0,0 +1,369 @@
...
+ *
+ * Buckets are considered filled if their current balance + allocated change is => than
+ * their capacity and unfilled if their current balance + allocated change is < than their capacity.
+ *
+ * The bucket capacity is defined as the % (as specified in the delegation portfolio) of total staked balance.
+ */
+type Bucket = {
+  address: Cardano.PaymentAddress;
+  change: Array<Cardano.TxOut>;
+  capacity: bigint;
```
mirceahasegan, line 24: The comment says capacity is a %. If so, `capacity: number` would be more appropriate.

packages/wallet/src/services/ChangeAddress/DynamicChangeAddressResolver.ts
```diff
@@ -0,0 +1,369 @@
...
+  filledAmount: bigint;
+};
+
+/**
+ * Gets the coalesced coin amount from the given TxOut array.
+ *
+ * @param txOuts The TxOuts to coalesce.
+ * @returns The coin value coalesced.
+ */
+const getBalanceInTxOuts = (txOuts: Array<Cardano.TxOut>) => {
```
mirceahasegan, line 34: You could use [coalesceValueQuantities](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/core/src/util/coalesceValueQuantities.ts#L8) instead

mirceahasegan: or make use of it in this function. Although it's simple enough, so it's not a hard requirement :sweat_smile:

packages/wallet/src/services/ChangeAddress/DynamicChangeAddressResolver.ts
```diff
@@ -0,0 +1,369 @@
...
+
+/**
+ * Aggregates all the coin balance in the input set from any input in the selection that belongs to us.
+ *
+ * @param knownAddresses The wallet know addresses.
+ * @param inputs The inputs.
+ */
+const getSpentInInputs = (knownAddresses: Array<GroupedAddress>, inputs: Array<Cardano.Utxo>) =>
+  inputs
+    .filter((utxo) => knownAddresses.find((groupedAddress) => groupedAddress.address === utxo[0].address))
```
mirceahasegan, line 52: `.some` would be more suitable here.
```diff
-    .filter((utxo) => knownAddresses.find((groupedAddress) => groupedAddress.address === utxo[0].address))
+    .filter((utxo) => knownAddresses.some((groupedAddress) => groupedAddress.address === utxo[0].address))
```

packages/wallet/src/services/ChangeAddress/DynamicChangeAddressResolver.ts
```diff
@@ -0,0 +1,369 @@
...
+    const groupedAddress = getAddressForRewardAccount(knownAddresses, account);
+
+    if (!groupedAddress) throw new InvalidStateError(`Reward account '${account}' unknown.`);
+
+    const adjustedStake =
+      delegated.stake - getSpentFromRewardAccount(account, inputs) + getDepositToRewardAccount(account, outputs);
+
+    const percentageForPool = weightsAsPercent.get(delegated.pool.hexId);
+
+    if (!percentageForPool) throw new InvalidStateError(`Pool '${delegated.pool.id}' bot found in the portfolio.`); // Shouldn't happen.
```
mirceahasegan, line 165: typo
```diff
 -   if (!percentageForPool) throw new InvalidStateError(`Pool '${delegated.pool.id}' bot found in the portfolio.`); // Shouldn't happen.
+    if (!percentageForPool) throw new InvalidStateError(`Pool '${delegated.pool.id}' not found in the portfolio.`); // Shouldn't happen.
```

mirceahasegan (dismissed):
Fantastic work, @AngelCastilloB :rocket:

mkazlauskas (approved):
:rocket:

mirceahasegan (approved):
Great work, @AngelCastilloB 🧙‍♂️

PR #788, opened 2023-06-28
Feat/lw 7051 handle policy id to projector

iccicci:
# Context

Now that in the local network the projector projects handle we mint, we need to produce relative fixtures to also implement the `TypeOrmHandleProvider` integration test.

# Proposed Solution

Regenerated the fixtures and implemented the missing test.

- Heavy refactored the script to generate the fixtures
  - it stops the `cardano-node-ogmios` container rather than the projectors ones to produce consistent data between the dumped databases
  - it dumps all the databases in our secrets directory
  - changed the output to be compliant with native postgres docker image init procedure
- Refactored the procedure to load fixtures during the tests startup
  - should make the test startup procedure a bit faster
  - allows to remove some code making the maintenance more easy
- Refactored `TypeormStakePoolProvider` tests
  - reworked some tests to make them more resilient against fixture re-generation
  - fixed the db connection initialization: this should solve the `No pools found` issue, cause of flakiness in our tests


rhyslbw (commented):
Amazing work here @iccicci, particularly in solving the Docker connection issue that's been plaguing us for so long 😍 

> Refactored `TypeormStakePoolProvider`

I can only see changes in the `/test` dir 🤔 . Is the description wrong? The commit should also be typed as `test` if this is intended.

iccicci:
> I can only see changes in the `/test` dir 🤔 . Is the description wrong?

Yes, sorry: I forgot the _tests_ word.

rhyslbw (approved):

packages/projection-typeorm/test/jest-setup/jest-setup.ts
```diff
@@ -1,10 +1,14 @@
...
 
 module.exports = async () => {
+  const fixturesPath = path.join(__dirname, '..', '..', '..', 'cardano-services', 'test', 'jest-setup', 'snapshots');
+
   await DockerUtil.setupPostgresContainer(
     connectionConfig.username,
     connectionConfig.password,
-    connectionConfig.port.toString()
+    connectionConfig.port.toString(),
+    [`${fixturesPath}:/docker-entrypoint-initdb.d`]
```
mkazlauskas, line 12: `projection-typeorm` is only testing the **writes** so I don't think we need to import data here

packages/projection-typeorm/test/util.ts
```diff
@@ -3,7 +3,7 @@ import { createDatabase } from 'typeorm-extension';
 import { logger } from '@cardano-sdk/util-dev';
 
 export const connectionConfig = {
-  database: 'projection',
+  database: 'stake_pool',
```
mkazlauskas, line 6: This is a generic util, having `stake_pool` as database name for all projection tests isn't great

mkazlauskas (changes requested):

packages/projection-typeorm/test/jest-setup/jest-setup.ts
iccicci: 4f6d8619f , 4f6d8619f and c64200743 should solve this

packages/projection-typeorm/test/util.ts
iccicci: 4f6d8619f , 4f6d8619f and c64200743 should solve this

rhyslbw (approved):

PR #796, opened 2023-07-03
Refactor cache invalidation test

iccicci:
# Context

Recently we changed the `DISABLE_DB_CACHE` setting of our test _local-network_ to `true` while the cache invalidation tests requires it to `false`: it needs a cache to check it is invalidated.

# Proposed Solution

Changed the test in order to start its own provider-server with the setting configured as it needs.

Also refactored the main `Dockerfile` to take more advantage from docker build cache


rhyslbw (changes requested):
Looks good, although there's an out of place fixup commit 
Screenshot 2023-07-03 at 13 32 11

mkazlauskas (approved):

rhyslbw (approved):

PR #799, opened 2023-07-04
fix(tx-construction): builder now awaits for non-empty knownAddresses before building the tx

mkazlauskas:
# Context

`KeyAgent` doesn't wait for wallet to settle before taking the current value of `knownAddresses$` when building the transaction.

# Proposed Solution

KeyAgent now awaits for non-empty knownAddresses$ for all it's uses

### DIFF
--- a/packages/tx-construction/package.json
+++ b/packages/tx-construction/package.json
@@ -40,6 +40,7 @@
   "devDependencies": {
     "@cardano-sdk/util-dev": "workspace:~",
     "@types/lodash": "^4.14.182",
+    "delay": "^5.0.0",
     "eslint": "^7.32.0",
     "jest": "^28.1.3",
     "madge": "^5.0.1",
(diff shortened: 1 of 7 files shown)


packages/tx-construction/package.json
```diff
@@ -56,6 +56,7 @@
     "@cardano-sdk/util": "workspace:~",
     "@cardano-sdk/util-rxjs": "workspace:~",
     "@cardano-sdk/wallet": "workspace:~",
+    "delay": "^5.0.0",
```
mirceahasegan, line 59: it looks like delay should be part of devDependencies, since it's being used only in the test

mkazlauskas: Nice catch, I intended to add it do devDependencies but failed :grin: 1523fcebe953a529e076434bebf74c020a5485fd

rhyslbw (approved):

mirceahasegan (dismissed):
Awesome find, @mkazlauskas ! 🕵️‍♂️

rhyslbw (approved):

mirceahasegan (approved):
🚀

rhyslbw (approved):

PR #801, opened 2023-07-05
LW-7114 Remove faucet from local network

iccicci:
# Context

The faucet is no longer used in our _local-network_ and e2e tests.

# Proposed Solution

Removed


rhyslbw (changes requested):
Looks good @iccicci , just a typo in this commit message
image

rhyslbw (approved):

PR #804, opened 2023-07-05
fix(tx-construction): rm redundant circular dependency on 'wallet' package

mkazlauskas:
# Context

`tx-construction` depends on `wallet` but does not use it. It creates a circular dependency messing up installed modules for the users


mirceahasegan (approved):
😱 🥇

rhyslbw (approved):
Nice catch @mkazlauskas

PR #805, opened 2023-07-05
feat: PlutusData to NftMetadata

mkazlauskas:
# Context

In order to support cip68/222 NFT metadata we need to have a mapper from our core PlutusData type

# Proposed Solution

See commits


packages/core/src/Asset/NftMetadata/fromMetadatum.ts
```diff
@@ -101,11 +95,10 @@ const getAssetMetadata = (policy: Cardano.MetadatumMap, assetName: Cardano.Asset
       })()
   );
 
-// TODO: consider hoisting this function together with cip25 types to core or a new cip25 package
 /**
  * @returns {NftMetadata | null} CIP-0025 NFT metadata
```
rhyslbw, line 99: Is this docblock stale?

packages/core/src/Asset/NftMetadata/index.ts
```diff
@@ -0,0 +1,11 @@
+import { fromMetadatum } from './fromMetadatum';
+
+import { NftMetadata as _NftMetadata } from './types';
+
+export * from './types';
+
+// This seems to only be working when both type and value are defined in the same file
```
rhyslbw, line 9: What "seems" to be working?

packages/core/src/Asset/NftMetadata/fromMetadatum.ts
mkazlauskas: This is part of LW-6909, no need to have this inline comment

packages/core/src/Asset/NftMetadata/index.ts
mkazlauskas: I couldn't
`export { NftMetadata } from './types'`
+
`export const NftMetadata = {`,
but it works as expected (can use as type or as value) when exporting both from the same file

packages/core/src/Asset/NftMetadata/fromMetadatum.ts
rhyslbw: I'm referring to the docblock that's reading "CIP-0025 NFT metadata" but the commit indicates it's being generalised, so should it be updated?

mkazlauskas: Updated doc comments in d87885f66f26df44543c8059cd9c6fe936c9cf1d and cb282ec5acb64f37f87ebab961b25e0d519edf08

packages/core/src/Asset/NftMetadata/index.ts
rhyslbw:
```suggestion
// tsc requires the type and object, sharing a name, to be exported from the same module
```

mkazlauskas: Thanks! 933ac784f612c15daa46184640c92f1befb66ff6

rhyslbw (approved):
Great work @mkazlauskas :rocket:

PR #807, opened 2023-07-07
test(e2e): mint cip-68(222) handle

iadmytro:
# Context

Added minting CIP-68(222) handle during local network startup.

### DIFF
--- a/packages/e2e/local-network/scripts/mint-handles.sh
+++ b/packages/e2e/local-network/scripts/mint-handles.sh
@@ -89,3 +89,67 @@ cardano-cli transaction sign \
 
 cardano-cli transaction submit --testnet-magic 888 --tx-file handle-tx.signed
 wait_tx_complete $utxo
+
+# CIP-68 Handle
+utxo=$(cardano-cli query utxo --address "$addr" --testnet-magic 888 | awk 'NR == 3 {printf("%s#%s", $1, $2)}')
+# {"constructor": 0,
+#   "fields": [
+#     {"map": [
+#       {"k": {"bytes": "core"}, "v": {"map": [
+#         {"k": {"bytes": "og"}, "v": {"int": 0}},
+#         {"k": {"bytes": "prefix"}, "v": {"bytes": "24"}},
+#         {"k": {"bytes": "version"}, "v": {"int": 0}},
+#         {"k": {"bytes": "termsofuse"}, "v": {"bytes": "https://cardanofoundation.org/en/terms-and-conditions/"}},
+#         {"k": {"bytes": "handleEncoding"}, "v": {"bytes": "utf-8"}}
+#       ]}},
+#       {"k": {"bytes": "name"}, "v": {"bytes": "(100)handle68"}},
+#       {"k": {"bytes": "image"}, "v": {"bytes": "ipfs://some-hash"}},
+#       {"k": {"bytes": "website"}, "v": {"bytes": "https://cardano.org/"}},
+#       {"k": {"bytes": "description"}, "v": {"bytes": "The Handle Standard"}},
+#       {"k": {"bytes": "augmentations"}, "v": {"list": []}}]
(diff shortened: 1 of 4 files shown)


packages/e2e/local-network/scripts/mint-handles.sh
```diff
@@ -51,3 +51,45 @@ cardano-cli transaction sign \
...
+      {"k": {"bytes": "76657273696f6e"}, "v": {"int": 0}},
+      {"k": {"bytes": "7465726d736f66757365"}, "v": {"bytes": "68747470733a2f2f63617264616e6f666f756e646174696f6e2e6f72672f656e2f7465726d732d616e642d636f6e646974696f6e732f"}},
+      {"k": {"bytes": "68616e646c65456e636f64696e67"}, "v": {"bytes": "7574662d38"}}
+    ]}},
+    {"k": {"bytes": "6e616d65"}, "v": {"bytes": "283130302968616e646c653638"}},
+    {"k": {"bytes": "696d616765"}, "v": {"bytes": "697066733a2f2f736f6d652d68617368"}},
+    {"k": {"bytes": "77656273697465"}, "v": {"bytes": "68747470733a2f2f63617264616e6f2e6f72672f"}},
+    {"k": {"bytes": "6465736372697074696f6e"}, "v": {"bytes": "5468652048616e646c65205374616e64617264"}},
+    {"k": {"bytes": "6175676d656e746174696f6e73"}, "v": {"list": []}}]},
+    {"int": 1}
```
mkazlauskas, line 73: Seems like we're missing the 3rd field:
```
The version number is at the second field of this constructor. The third field allows for arbitrary plutus data.
```
```
  ; Custom user defined plutus data.
    ; Setting data is optional, but the field is required
    ; and needs to be at least Unit/Void: #6.121([])
    extra = plutus_data
```

mkazlauskas (changes requested):

packages/e2e/local-network/scripts/mint-handles.sh
iadmytro: added in [e3bea15](https://github.com/input-output-hk/cardano-js-sdk/pull/807/commits/e3bea157c1d241db8de7a2505d33d5c1b579e153)

packages/e2e/local-network/scripts/mint-handles.sh
```diff
@@ -70,3 +70,63 @@ cardano-cli transaction sign \
   --out-file handle-tx.signed
 
 cardano-cli transaction submit --testnet-magic 888 --tx-file handle-tx.signed
+
+sleep 10
```
mkazlauskas, line 74: Why is this needed and why 10? :sweat_smile:

mkazlauskas (approved):
Looks good! Just 1 question below

packages/e2e/local-network/scripts/mint-handles.sh
iadmytro: it will take outdated utxo othervise

mkazlauskas: Sounds like we could see some flakiness due to this. Can we implement loop with sleep 1 that checks for state change instead?

mkazlauskas (approved):
Great work! :rocket:

rhyslbw (changes requested):
Looks good @iadmytro. It would be better to reorder the commits so abc19d75a01cd32e73f3763221bc93c8fffb07d4 comes first

rhyslbw (approved):
Well done @iadmytro. Sorry I didn't realise [this line](https://github.com/input-output-hk/cardano-js-sdk/pull/807/files#diff-c245fa097162abd676a16c2a811730268a0542e4b802b5d06ea42271bd7d6f94L36) was existing, which was the basis for me requesting the commit reorder since it appeared to be added in the prior commit

PR #811, opened 2023-07-11
Feat/lw 7373 improvements to greedy input selection

AngelCastilloB:
# Context

The greedy input selector currently generates on output per stake id. Ideally we want a more robust and varied UTXO set so our strategy for maintaining multi delegation stake distribution is more consistent in active wallets.

# Proposed Solution

Split then change outputs following an exponential distribution. For example 100000 will yield:

`[ 50000n, 25000n, 12500n, 6250n, 3125n, 3125n ]`

The exponential distribution (n**2),was the best compromise in granularity and amount of UTXOs generated. We don't want too many UTXOs as this could make the transaction exceed the max allow TX size, but at the same time we want a diverse and big enough amount of UTXOs to reasonably build any TX with a fairly small amount of inputs.

Using an exponential distribution also guarantees that the number of UTXOs generated will be more or less the same
regardless of the amount of total available lovelace, for example:

```
10      => [ 5n,      2n,      1n,      1n,     1n             ]
100     => [ 50n,     25n,     12n,     6n,     3n,     4n     ]
1000    => [ 500n,    250n,    125n,    62n,    31n,    32n    ]
10000   => [ 5000n,   2500n,   1250n,   625n,   312n,   313n   ]
100000  => [ 50000n,  25000n,  12500n,  6250n,  3125n,  3125n  ]
1000000 => [ 500000n, 250000n, 125000n, 62500n, 31250n, 31250n ]
```

The greedy input selector now generates more than one change output per stake address.

### DIFF
--- a/packages/e2e/src/tools/multi-delegation-data-gen/index.ts
+++ b/packages/e2e/src/tools/multi-delegation-data-gen/index.ts
@@ -5,9 +5,7 @@ import {
   TaskResult,
   TerminalProgressMonitor,
   createDelegationWallet,
-  delegateToMultiplePools,
   distributeStake,
-  generateStakeAddresses,
   getOutputPathName,
   loadConfiguration,
   logState,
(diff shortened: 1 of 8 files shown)


mkazlauskas (approved):
Excellent work! :muscle:

rhyslbw (approved):
Fantastic work @AngelCastilloB 💥

PR #812, opened 2023-07-11
ci/build and publish images

blaggacao:
- ci: implement packaging layers for lace
- env: add basic dev env with
- env: add std to reproduce ci locally

# Context

This PR comes without Github Issue but implements LW-6842.

In order to shorten the SDLC cycle time for this repository's Provider Server that serves the Lace Wallet, we need to incorporate (the currently employed) CI/CD within the scope of this repositoy.

# Proposed Solution

This PR impletents the first three layers of what we've been calling the ["4 Layers of Packaging"](https://std.divnix.com/patterns/four-packaging-layers.html) with the exclusion of the scheduler chart layer.

This last layer requires further testing and iteration and will be implemented in a follow up PR. However,  this PR already enables shorter iterations cycles on this follow up which is why it merits indepentent consideration.

The layers are implemented employing the Standard DevOps library and its
["DevOps type system"](https://std.divnix.com/ARCHITECTURE.html#standards-block-types-devops-type-system) for maximized re-use of code and best practices.

- Implements Layers: Packaging, Operable & OCI-Image
- A unified "smart" CI for them (only does work when something significant has changed)
- Add minimal repruducible dev environment to handle the associated toolchain soomthly and automagically.

# Pending for Review / Discussion

- TODO items in the code (BUILD_INFO & Cardano Config)
- Recommended: documenting toolchain use in the readme (if desired)
- To aid situational awareness during operation, defaulting heuristics should be extremely easy to discover and clearly isolated from business logic. In this spirit of "operable layer", we'd need to consolidate all default values into a clear and simple-to-read file that is directly associated with the operable layer (e.g. a code comment linking it or similar).


.github/std-unexploded.yml
```diff
@@ -0,0 +1,107 @@
...
+          mask-aws-account-id: false
+      - name: Login to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v1
+      # TODO: uncomment when nixbuild works well together with
+      #       nix daemon mode on hosted runners
+      # - uses: nixbuild/nixbuild-action@v17
+      #   with:
+      #     nixbuild_ssh_key: ${{ secrets.SSH_PRIVATE_KEY }}
+      #     generate_summary_for: job
```
gytis-ivaskevicius, line 51: Why do we need nixbuild.net on discovery? Discovery is not doing much and CI runs are likely to be faster without nixbuild.net since it introduces network overhead

.github/std-unexploded.yml
```diff
@@ -0,0 +1,107 @@
...
+      - uses: divnix/std-action/discover@main
+        with: {ffBuildInstructions: true}
```
gytis-ivaskevicius, line 53: Out of curiosity: 
Can we do this as a multiline? One project of mines would benefit from passing json like that but I thought its not supported

.github/std-unexploded.yml
```diff
@@ -0,0 +1,107 @@
...
+    needs: discover
+    name: ${{ matrix.target.jobName }}
+    runs-on: ubuntu-latest
+    if: fromJSON(needs.discover.outputs.hits).packages.build != '{}'
+    strategy:
+      matrix:
+        target: ${{ fromJSON(needs.discover.outputs.hits).packages.build }}
+    steps:
+      - uses: nixbuild/nix-quick-install-action@v25
+      - uses: nixbuild/nixbuild-action@v17
```
gytis-ivaskevicius, line 66: It might be worth benchmarking without nixbuild.net after my recent build speedup patch

.github/std-unexploded.yml
gytis-ivaskevicius: Also what does this do?

flake.nix
```diff
@@ -1,124 +1,26 @@
...
-          cp -r ${production-deps}/libexec/$sourceRoot/packages/cardano-services/config $out/libexec/$sourceRoot/packages/cardano-services/config
-
-          cd "$out/libexec/$sourceRoot"
+  inputs = {
+    std.url = "github:divnix/std"; # keep minor version in lockstep with GH std-action
+    # std.inputs.nixpkgs.follows = "nixpkgs";
+    std.inputs.n2c.follows = "n2c";
+    haumea.url = "github:nix-community/haumea/v0.2.2";
+    n2c.url = "github:nlewo/nix2container";
+    n2c.inputs.nixpkgs.follows = "std/nixpkgs";
```
gytis-ivaskevicius, line 8: Let's have separate nixpkgs entry and set other inputs to follow it in hopes of reducing developers mental overhead

flake.nix
```diff
@@ -1,124 +1,26 @@
...
-          #!${pkgs.bash}/bin/bash
-          exec "${pkgs.nodejs}/bin/node" "$out/libexec/$sourceRoot/packages/cardano-services/dist/cjs/cli.js" "\$@"
-          EOF
-          chmod a+x $out/bin/cli
-        '';
-        meta.mainProgram = "cli";
-        passthru.nodejs = pkgs.nodejs;
-        passthru.production-deps = production-deps;
-      });
+  inputs.nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
```
gytis-ivaskevicius, line 11: I think this input got lost. Move it up there and use this as main nixpkgs for all inputs

flake.nix
```diff
@@ -1,124 +1,26 @@
...
-            git subrepo pull packages/cardano-services/config --message='chore: git subrepo pull packages/cardano-services/config' "$@"
-          '';
-        };
-      };
-    });
-  };
+  outputs = {std, ...} @ inputs:
+    std.growOn {
+      inherit inputs;
+      cellsFrom = ./ops;
```
gytis-ivaskevicius, line 16: Consider changing this folder to `nix` or something else. As a third-party contributor it would take me a little to figure out that all this magic is hiding in this folder

local/flake.nix
```diff
@@ -0,0 +1,51 @@
+{
```
gytis-ivaskevicius, line 1: Why does this subflake exist?

local/flake.nix
```diff
@@ -0,0 +1,51 @@
...
+      {
+        self,
+        nixpkgs,
+        devshell,
+        std,
+        ...
+      }:
+        with nixpkgs.legacyPackages;
+        with nixpkgs.legacyPackages.nodePackages;
+        with devshell.legacyPackages; let
```
gytis-ivaskevicius, line 22: Let's avoid this antipattern

local/flake.nix
```diff
@@ -0,0 +1,51 @@
...
+          checkMod = {
+            commands = [{package = treefmt;}];
+            packages = [
+              alejandra
+              shfmt
+              prettier
+              prettier-plugin-toml
+            ];
+            devshell.startup.nodejs-setuphook = noDepEntry ''
+              export NODE_PATH=${prettier-plugin-toml}/lib/node_modules:$NODE_PATH
```
gytis-ivaskevicius, line 33:
```suggestion
              export NODE_PATH=${prettier-plugin-toml}/lib/node_modules:$NODE_PATH
              export PATH=$PATH:$PRJ_ROOT/node_modules/.bin
```

local/flake.nix
```diff
@@ -0,0 +1,51 @@
...
+        nixpkgs,
+        devshell,
+        std,
+        ...
+      }:
+        with nixpkgs.legacyPackages;
+        with nixpkgs.legacyPackages.nodePackages;
+        with devshell.legacyPackages; let
+          inherit (lib.stringsWithDeps) noDepEntry;
+          checkMod = {
```
gytis-ivaskevicius, line 24: I see what it contains but I have no idea what does this name mean. Please rename it

local/flake.nix
gytis-ivaskevicius: Please remove it to reduce this project complexity

ops/app/oci-images.nix
```diff
@@ -0,0 +1,24 @@
+let
+  inherit (inputs) std self;
+
+  # TODO: express this as OCI labels (what they are for)
+  BUILD_INFO = builtins.toJSON {
```
gytis-ivaskevicius, line 5: This is consumed by the application which is why I believe that we can not use labels for it

Also, please make the variable name camelCase

ops/app/oci-images.nix
```diff
@@ -0,0 +1,24 @@
...
+    extra = {
+      sourceInfo = self;
+      narHash = self.narHash;
+      path = self.outPath;
+    };
+  };
+in {
+  server = std.lib.ops.mkStandardOCI {
+    # TODO: set up Repo w/ registry, and test
+    name = "926093910549.dkr.ecr.us-east-1.amazonaws.com/provider-server";
```
gytis-ivaskevicius, line 17: Please fix `server` and `provider-server` naming. This image may not be necessary `provider-server`. Shall we call it `cardano-services`?

ops/app/operables.nix
```diff
@@ -0,0 +1,12 @@
+let
+  inherit (inputs) std;
+  inherit (inputs.nixpkgs) lib;
+  # TODO: understand if CARDANO_NODE_CONFIG_PATH has appropriate defaults (depending on network)
+  # see also epic about organizing defaults in a single place
+in {
+  server = std.lib.ops.mkOperable {
+    package = cell.packages.server;
+    runtimeScript = "${cell.packages.server}/bin/cli";
```
gytis-ivaskevicius, line 9: **NOT READY TO BE MERGED!**

ops/app/packages.nix
```diff
@@ -0,0 +1,98 @@
...
+    project = nixpkgs.callPackage (self + /yarn-project.nix) {} {inherit src;};
+
+    replaceLine = regex: replacement: s: let
+      m = builtins.match "(.*\n)${regex}(\n.*)" s;
+    in
+      builtins.concatStringsSep "" [
+        (builtins.elemAt m 0)
+        replacement
+        (builtins.elemAt m 1)
+      ];
```
gytis-ivaskevicius, line 29: Where is this used?

ops/app/packages.nix
```diff
@@ -0,0 +1,98 @@
...
+    production-deps = project.overrideAttrs (oldAttrs: {
```
gytis-ivaskevicius, line 30: I'm pretty sure that you can pass `overrideAttrs` as an argument. Please do that

ops/app/packages.nix
```diff
@@ -0,0 +1,98 @@
...
+      name = "cardano-sdk-production-deps";
+      configurePhase =
+        builtins.replaceStrings
+        ["yarn install --immutable --immutable-cache"]
+        ["yarn workspaces focus --all --production"]
+        oldAttrs.configurePhase;
+      # A bunch of deps build binaries using node-gyp that requires Python
+      PYTHON = "${nixpkgs.python3}/bin/python3";
```
gytis-ivaskevicius, line 38:
```suggestion
      PYTHON = lib.getExe nixpkgs.python3;
```
Same with others. `rg '/bin/' $PRJ_ROOT`

ops/app/packages.nix
```diff
@@ -0,0 +1,98 @@
...
+        mkdir -p $out/libexec/$sourceRoot $out/bin
+        yarn workspace @cardano-sdk/cardano-services-client run build:version
+
+        cp -r ${production-deps}/libexec/$sourceRoot/node_modules $out/libexec/$sourceRoot/node_modules
+        cp -r scripts $out/libexec/$sourceRoot/scripts
+        for p in cardano-services cardano-services-client core ogmios util crypto projection projection-typeorm util-rxjs; do
+          mkdir -p $out/libexec/$sourceRoot/packages/$p
+          cp -r packages/$p/dist $out/libexec/$sourceRoot/packages/$p/dist
+          cp -r packages/$p/package.json $out/libexec/$sourceRoot/packages/$p/package.json
+        done
```
gytis-ivaskevicius, line 78: I am very questioning if this is necessary

ops/app/packages.nix
```diff
@@ -0,0 +1,98 @@
...
+        cd "$out/libexec/$sourceRoot"
+
+        runHook postInstall
+      '';
+      # add a bin script that should be used to run cardano-services CLI
+      postInstall = ''
+        cat > $out/bin/cli <<EOF
+        #!${nixpkgs.bash}/bin/bash
+        exec "${nixpkgs.nodejs}/bin/node" "$out/libexec/$sourceRoot/packages/cardano-services/dist/cjs/cli.js" "\$@"
+        EOF
```
gytis-ivaskevicius, line 90: `cli.js` already has nodejs as an interpreter. Not necessary

treefmt.toml
```diff
@@ -0,0 +1,17 @@
...
+# TODO: would this conflict or just work (tm)?
+# [formatter.prettier]
+# command = "prettier"
+# includes = ["*.md", "*.toml", "*.yaml", "*.json"]
+# options = ["--plugin", "prettier-plugin-toml", "--write"]
+#
+# [formatter.shell]
+# command = "shfmt"
+# includes = ["*.sh"]
+# options = ["-i", "2", "-s", "-w"]
```
gytis-ivaskevicius, line 17: Not ready to be merged

gytis-ivaskevicius (changes requested):
Please resolve the comments and write down your testing scenarios

I'm moving this PR to `draft` until you actually test it

flake.nix
rhyslbw: It was discussed as an option to better capture the intent but that was a different context. Here we have more than just ops code, as the lower packaging layers are present, so it would be more aligned as `nix`

ops/app/oci-images.nix
rhyslbw: If there's just a single image for all services then I agree

.github/std-unexploded.yml
blaggacao: This decisively depends on the amount of Import From Derivation. My recollectoion is that we still haven't reduced the amount of IFD in this repository which is why the goal to mitigate it with remote builds remains on the agenda as soon as we have compatibility.

.github/std-unexploded.yml
blaggacao: Nixbuild here is a bit of a "double use" service. We also use it for its persistent build caching on the account.

So even if there may be independent optimizations, that cache still buys us huge discounts in cycle time

local/flake.nix
blaggacao: If I understand this suggestion properly, this would move the responsibility to procure the prettier-plugin-toml binary to `node_modules` folder and I don't know if it's there, but even if it is "outsourcing" that contract (without actually having established one) seems a bit wonky in terms of maintenance.

ops/app/oci-images.nix
blaggacao: Ok, this is due to the re-architecture. What's the name, then? `cardano-sdk-server`?

ops/app/operables.nix
blaggacao: Yes, absolutely. The questions here is how do we inject `CARDANO_NODE_CONFIG_PATH` and the corresponding data?

Since it's sort of in the category a dockerfile's `ADD config.json` and quite a crucial source of configuraition data, this would best moved here for full operational transparency visibility.

Like on the other TODO, we don't need to expand the scope, but I'd like to get the discussion started with @rhyslbw and the team.

ops/app/packages.nix
blaggacao: I pretty much moved these sections from the former `flake.nix` without intent to refactor. Since this is a classical "green-to-green", even if not reflected in git history, let's scope out "Revise and Refactor Package Layer" into a separate task/PR.

ops/app/packages.nix
blaggacao: I'd stack that onto https://github.com/input-output-hk/cardano-js-sdk/pull/812#discussion_r1262641927, too.

ops/app/packages.nix
blaggacao: https://github.com/input-output-hk/cardano-js-sdk/pull/812#discussion_r1262641927, too :smile:

treefmt.toml
blaggacao: @rhyslbw can you maybe help to answer the question of the TODO? Maybe point around on how this is currently wired.

blaggacao:
>  write down your testing scenarios

This is fully covered by CI/CD. The critical path is the operable which needs to be aligned downstream. I suggest that we may deploy the further testing downstream, therefore?

ops/app/packages.nix
rhyslbw: Reasonable. Migrate then refactor

treefmt.toml
rhyslbw: We currently have no linting covering those files, so it should just work, although when enabling we should also commit it's effect

ops/app/oci-images.nix
rhyslbw: If this image is being used to run instances of `provider-server`, `projector`, `pgboss-worker` etc, it's the appropriate to call it by the package name: `cardano-services`

ops/app/operables.nix
rhyslbw: > how do we inject CARDANO_NODE_CONFIG_PATH and the corresponding data?

Please elaborate, the issue is not clear to me.

gytis-ivaskevicius: Please just export the variable. You can copy the logic from `wallet-world`

ops/app/oci-images.nix
gytis-ivaskevicius: Oh, I just realized that since we embed build info into OCI it will get rebuilt each time, even if the code has not changed

.github/std-unexploded.yml
gytis-ivaskevicius: As far as i am aware we are doing exactly zero IFD in this project

flake.nix
blaggacao: Addressed in 93cd451b3bf08ad132da0e18839473efda672658

flake.nix
blaggacao: Addressed in 93cd451b3bf08ad132da0e18839473efda672658

flake.nix
blaggacao: Addressed in 93cd451b3bf08ad132da0e18839473efda672658

local/flake.nix
blaggacao: Addressed in 93cd451b3bf08ad132da0e18839473efda672658

local/flake.nix
blaggacao: Addressed in 93cd451b3bf08ad132da0e18839473efda672658

local/flake.nix
gytis-ivaskevicius: The idea here was to make the better developer experience, we can 'outsource' it or have 2 of them installed

.github/std-unexploded.yml
gytis-ivaskevicius: Lets remove this comment and we can resolve this conversation

gytis-ivaskevicius (changes requested):
Mainly whats missing is testing out the OCI image. 
I'd like if buildInfo was to be moved as part of deployment configuration

local/flake.nix
blaggacao: Got ya! I'd say let's do a closer integration in a next step, once this proposal has landed and people are a bit more familiar with "what this is".

treefmt.toml
blaggacao: Attenuated (w/o `*.json`) & addressed in e4f6ed1e2e2c0b252d6ee62c70ae64c5276446a9

.github/std-unexploded.yml
blaggacao: > it seems that will result in performance bump

I can copy that statement, just yet. I know you've shared some initial observations, but I haven't verified them yet.

We can easily run a proper experiment a bit later and compare the results, e.g. making sure speed increases are not due to caching effects, etc. Ensuring that isn't trivial and I'd like to keep the focus on this PR on the migration.

I propose to postpone this improvement.

.github/std-unexploded.yml
blaggacao: It's a bit linked to leaving https://github.com/input-output-hk/cardano-js-sdk/pull/812#discussion_r1268714723. I intent to follow up on this, later and as discussed today in our standup, hence the TODO.

ops/app/oci-images.nix
blaggacao: @rhyslbw did you backlog this? Removing this and adopting best practices through using OCI labels probably needs some changes to how we think of today's `/meta`.

ops/app/oci-images.nix
blaggacao: Implemented in 0b0b91f153c7d232ac10a797ad646f2f26950e23

blaggacao:
> Mainly whats missing is testing out the OCI image.

That (+ eventual fine tuning) will be the objective of the [sister PR](https://github.com/input-output-hk/cardano-js-sdk/pull/756) or some variant of it.

> I'd like if buildInfo was to be moved as part of deployment configuration

This is where typically the [well-known OCI label annocations](https://github.com/opencontainers/image-spec/blob/main/annotations.md) provide the best levels of tooling and ecosystem integration. (cc @rhyslbw for backlogging this conversation to be held).

ops/app/operables.nix
blaggacao: Added in aa80281462df5fc6dc04448bd9471735e3cdcfb5. It's ok-ish. Conceptually, I'd rather like to inject these configuration invariants during runtime and also directly from the cardano _source of truth_ rather than a second or third order replica.

Let's see how we fare with this, it's not likely to cause any problems, and we can revisit later.

nix/app/operables.nix
blaggacao, line 1: @mkazlauskas @iccicci fysa, the expectation about this file is that it will be the primary ops-interface, alongside `packages/cardano-services/src/cli.ts`, ofc.

So the idea is that if configuration changes ("drifts" ehem :smile:) it would surface through a PR review either here or in `packages/cardano-services/src/cli.ts` where all @input-output-hk/lace-sre are actively and _conscientously_ involved.

That may safe us misunderstandings down the line and keep us informed on the critical changes / improvements to the configuration.

You can think of this file as a transparent `nix`-handle to `packages/cardano-services/src/cli.ts`. :handshake: 

Please let me know if you have any questions or concerns! The idea is to streamline the flow of critical information as close as possible to the code.

blaggacao: Note to self: amend `CODOWNERS` in follow up PR after some more discussion and understanding on the workflows involved.

local/flake.nix
blaggacao: Changed in 69d33b662eb5ef0df1326c2a013d9d062e7e79e1 to align with `std` (for now).

iccicci:
Apart from a CI/CD failure, I can see a new formatter was added and it seems it is driven by `treefmt.toml` file: how that formatter is run? Is it run by some automation r should be run by hand?

nix/app/operables.nix
```diff
@@ -0,0 +1,21 @@
...
+  mainnet  preprod  preprod_p2p  preview  preview_p2p  testnet  vasil-dev  vasil-dev_p2p  vasil-qa
+  */
+  networkConfig = inputs.self + /packages/cardano-services/config/network;
+in {
+  server = std.lib.ops.mkOperable {
+    package = cell.packages.server;
+    # find all configuration options in: packages/cardano-services/src/cli.ts
+    runtimeScript = ''
+      ${cell.packages.server}/bin/cli --cardano-node-config-path \
+        "${networkConfig}/''${NETWORK-mainnet}/cardano-node/config.json"
```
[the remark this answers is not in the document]
gytis-ivaskevicius: I just realized that you did exactly zero testing on this. It is missing `$@` at the very end which means that this OCI image is constantly failing

michalrus:
> I am not a big fan of all those formatting tweaks in this PR, maybe we should have separated it

Agreed, formatting changes should happen separately from important changes, as they make it easier for something to slip through by distracting the reader

local/flake.nix
gytis-ivaskevicius: flake.nix still exists, the only difference is that I have no idea what it does now

tshaynik:
As I'm hacking on this PR directly (starting with removing the formatting changes) I've made a separate branch to reflect the state before I started messing with it: https://github.com/input-output-hk/cardano-js-sdk/tree/ci/build-and-publish-images-old

nix/local/flake.nix
```diff
@@ -0,0 +1,10 @@
+{
```
gytis-ivaskevicius, line 1: I dont think this does anything, can we remove it?

nix/app/operables.nix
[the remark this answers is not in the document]
blaggacao: > I just realized that you did exactly zero testing on this. It is missing $@ at the very end which means that this OCI image is constantly failing

https://github.com/input-output-hk/cardano-js-sdk/pull/756 - blocked by this PR

> This last layer requires further testing and iteration and will be implemented in a follow up PR. However, this PR already enables shorter iterations cycles on this follow up which is why it merits indepentent consideration.

nix/local/flake.nix
blaggacao: > I dont think this does anything

[Here](https://github.com/paisano-nix/core/blob/10270dc46532c947de473ee88c8f5a3346a396fb/grow/default.nix#L92C8-L95) is the implementation of the effectuator overloading or adding extra inputs (which in turn don't need to be added / updated c/o the main flake).

blaggacao:
> Please copy OCI logic from wallet-world. It's almost identical anyway

The initial intent was to proof this implementation against https://github.com/input-output-hk/cardano-js-sdk/pull/756. The only reason this was separated out is to reduce the cycle-time and risk of build breakage (which in the meantime happened due to https://github.com/input-output-hk/cardano-js-sdk/commit/94c27bbdbab7447487987c3ec1fbf95c10ed797f.

Until that PR is finalized I don't expect that this PR is fully functional and fully tested, but I expect it to reduce cycle times and the risk of build failure for #756, instead.

Dealing with long running CI cycle times during development is a bit nerve wrecking, so I appreciate if we don't move the goal post too much and honor this intent.

nix/local/flake.nix
gytis-ivaskevicius: Can we remove it and use only top-level inputs? I think it's too much to expect from stakeholders to understand how it works and at the end of the day if anything it adds extra storage usage and extra evaluation time (since its unlikely that dependencies will be the same as top-level flake)

.github/std-unexploded.yml
```diff
@@ -0,0 +1,107 @@
+# yq '. | explode(.)' std-unexploded.yml > workflows/std.yml
+name: CI/CD
+
+on:
+  pull_request:
```
gytis-ivaskevicius, line 5:
```suggestion
  workflow_trigger:
  pull_request:
```
This will likely to come in handy (please double check if that's how workflow trigger event calls)

nix/app/operables.nix
```diff
@@ -0,0 +1,30 @@
...
+
+  $ ls packages/cardano-services/config/network/
+  mainnet  preprod  preprod_p2p  preview  preview_p2p  testnet  vasil-dev  vasil-dev_p2p  vasil-qa
+  */
+  cardanoServicesPath = pkg: let
+    sourceRoot = "source";
+  in "${pkg}/libexec/${sourceRoot}/packages/cardano-services";
+  runCardanoServices = pkg: "${lib.getExe pkg.nodejs} ${cardanoServicesPath pkg}";
+in {
+  server = std.lib.ops.mkOperable rec {
```
gytis-ivaskevicius, line 15:
```suggestion
  cardano-services = std.lib.ops.mkOperable rec {
```

gytis-ivaskevicius (approved):
Now that I am reasonably confident that the OCI image is working I am happy to merge this PR, there are a couple of improvements that I'd like to be addressed in this PR but we can address them in follow-up PR's

- Remove nixbuild.net for performance reasons
- Remove separate job for publishing images for performance reasons
- Remove subflake to reduce mental overhead and improve performance/storage efficiency
- Test OCI Image (just copy over string reference to deployment configuration and deploy it on ops env)

@rhyslbw @tshaynik Can you guys review this? Either approve it or 'request changes' based on the above mentioned minor issues since fixes should be quick and easy

blaggacao:
> Remove nixbuild.net for performance reasons

This seems to be a misconception or misunderstanding. Nixbuild significantly improves the Cycle Time. To leverage it fully, there's just a little tweak required to the nixbuild action that I'll follow-up with shortly.

gytis-ivaskevicius:
Okay, let's fix it up and run it for a week or two then we can disable it for a week and revert if necessary then we can decide if its worth our while

flake.nix
```diff
@@ -1,124 +1,28 @@
...
-          exec "${pkgs.nodejs}/bin/node" "$out/libexec/$sourceRoot/packages/cardano-services/dist/cjs/cli.js" "\$@"
-          EOF
-          chmod a+x $out/bin/cli
-        '';
-        meta.mainProgram = "cli";
-        passthru.nodejs = pkgs.nodejs;
-        passthru.production-deps = production-deps;
-      });
-
-    apps = flake-utils.lib.eachDefaultSystemMap (system: let
```
tshaynik, line 110: I think we need to re-add this functionality that this PR currently removes and run the script in CI before building the container image.

tshaynik:
I'm working on a solution to this error, caused by missing modules.
```
node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module '/nix/store/vnx7dqcj8wq9a0bivw5f4axs72f2pz7s-cardano-sdk/libexec/source/packages/cardano-services/dist/cjs/cli.js'
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:77:12)
    at node:internal/main/run_main_module:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```

gytis-ivaskevicius:
I think now OCI image should be working. For some reason, cli.js path has changed

Also I ran `nix eval --no-allow-import-from-derivation --json .\#__std.ci.x86_64-linux` and it succeeded. I have no idea why discovery takes so long

gytis-ivaskevicius:
@rhyslbw @tshaynik I think this is ready to be merged
Applied few tweaks and tested OCI image, everything seems to be good :)) As a sanity check ill test the latest image again just in case

gytis-ivaskevicius (approved):
Confirmed to be working
Tested:
- Devshell
- OCI image in ops-preview-1

rhyslbw (approved):

nix/local/envs.nix
```diff
@@ -6,6 +6,8 @@ let
     alejandra
     shfmt
     yq
+    git
+    git-subrepo
```
gytis-ivaskevicius, line 10: Can we leave this impure?

gytis-ivaskevicius (changes requested):

nix/local/envs.nix
tshaynik: I guess. Git probably isn't necessary. It was more for the git-subrepo, which most people won't have installed.

gytis-ivaskevicius (approved):

tshaynik (approved):

PR #813, opened 2023-07-12
fix: mainnet projection issues

mkazlauskas:
# Context

When running projection services on mainnet, we ran into several issues:
1. Projector stops due to local state queries being unavailable in Byron era
2. Fails to insert some pool registrations, where `cost` or `pledge` exceeds postgres `bigint` (signed 64bit integer)

# Proposed Solution

See commits


rhyslbw (dismissed):
Nice work @mkazlauskas 🦾

iadmytro (approved):

rhyslbw:
Please provide a progress update on the local _mainnet_ sync @mkazlauskas

rhyslbw (approved):

PR #815, opened 2023-07-13
fix(wallet): cip30 getUsedAddresses now returns all known addresses

AngelCastilloB:
# Context

The API introduced in CIP-30 getUsedAddresses was implemented with the single address wallet in mind, but this is now returning an invalid response when used in conjunction with a multi-address wallet (either HD compatibility or multi-delegation)

# Proposed Solution

Update getUsedAddresses so it return all known addresses.


mkazlauskas (approved):
:rocket:

rhyslbw (approved):
Thanks @AngelCastilloB, well done

PR #816, opened 2023-07-13
fix: wallet finalizeTx and CIP30 now wait for at least one known address

AngelCastilloB:
# Context

With the introduction of HD wallet discovery, is taking longer for the known addresses array to be populated, this is causing issue when trying to sign transaction before the known addresses have been resolved.

# Proposed Solution

Wait in the relevant places for the addresses to be resolved.


mkazlauskas (approved):
Nice work!

rhyslbw (approved):
Nice work @AngelCastilloB!

PR #818, opened 2023-07-14
Test/e2e handle

iadmytro:
# Context

Added E2E test which sends funds using handle

# Important changes

Removed resolved handle from addOutput in TxBuilder

### DIFF
--- a/packages/tx-construction/src/tx-builder/TxBuilder.ts
+++ b/packages/tx-construction/src/tx-builder/TxBuilder.ts
@@ -26,6 +26,7 @@ import { filter, firstValueFrom, lastValueFrom } from 'rxjs';
 import { finalizeTx } from './finalizeTx';
 import { initializeTx } from './initializeTx';
 import minBy from 'lodash/minBy';
+import omit from 'lodash/omit';
 
 type BuiltTx = {
   tx: Cardano.TxBodyWithHash;
@@ -117,12 +118,11 @@ export class GenericTxBuilder implements TxBuilder {
   }
 
   addOutput(txOut: OutputBuilderTxOut): TxBuilder {
-    this.partialTxBody = { ...this.partialTxBody, outputs: [...(this.partialTxBody.outputs || []), txOut] };
-
     if (txOut.handle) {
       this.#handles = [...this.#handles, txOut.handle];
     }
-
+    const txOutNoHandle = omit(txOut, 'handle');
+    this.partialTxBody = { ...this.partialTxBody, outputs: [...(this.partialTxBody.outputs || []), txOutNoHandle] };
     return this;
   }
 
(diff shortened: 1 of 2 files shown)


rhyslbw (changes requested):
e5c8e9cb89604505b9e9467a345eed85a581c7b6 is a `refactor` type, also it's a breaking change. Otherwise, looks good!

packages/tx-construction/src/tx-builder/TxBuilder.ts
```diff
@@ -117,17 +118,16 @@ export class GenericTxBuilder implements TxBuilder {
...
   addOutput(txOut: OutputBuilderTxOut): TxBuilder {
-    this.partialTxBody = { ...this.partialTxBody, outputs: [...(this.partialTxBody.outputs || []), txOut] };
-
-    if (txOut.handle) {
-      this.#handles = [...this.#handles, txOut.handle];
+    if (txOut.handleResolution) {
+      this.#handleResolutions = [...this.#handleResolutions, txOut.handleResolution];
     }
-
+    const txOutNoHandle = omit(txOut, 'handle');
```
mkazlauskas, line 124: This was updated in original commit
```suggestion
    const txOutNoHandle = omit(txOut, 'handleResolution');
```

packages/core/src/Provider/TxSubmitProvider/types.ts
```diff
@@ -6,7 +6,7 @@ type SerializedTransaction = Cardano.util.HexBlob;
 
 export interface SubmitTxArgs {
   signedTransaction: SerializedTransaction;
-  context?: { handles: HandleResolution[] };
+  context?: { handleResolutions: HandleResolution[] };
```
mkazlauskas, line 9: @rhyslbw this refactor makes sense but we need to be careful with deployments: since the name changed, old backend will not validate when you pass it as `handles:`. Also I noticed we don't have `context` specified in `openApi.json` - should we add it in this PR?

mkazlauskas (changes requested):
Nice refactor @iadmytro :rocket: 

When squashing please remove the scope from refactor commit message (it affects multiple packages).

packages/core/src/Provider/TxSubmitProvider/types.ts
rhyslbw: Good points.

1. The breaking change will need to be reverted (deferred)
2. Yes, adding the missing OpenAPI spec content is a good idea here

iadmytro: Ok, refactoring will be moved to separate PR

mkazlauskas (approved):
Great work! :rocket:

packages/core/src/Provider/TxSubmitProvider/types.ts
mkazlauskas: Captured openApi.json task in LW-7535

rhyslbw (approved):

PR #819, opened 2023-07-17
refactor!: renamed field handle to handleResolutions

iadmytro:
# Context

Where necessary field `handle` renamed to `handleResolution` to avoid confusion.


rhyslbw (changes requested):
Also need to bump the major version here since it's an API breaking change

https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services/src/TxSubmit/openApi.json#L9

mkazlauskas:
Maybe let's target `feat/plutus-data-serialization` to batch breaking changes

iccicci (approved):
LGTM

mkazlauskas (approved):
:rocket:

PR #821, opened 2023-07-18
Fix/provider server cfg fixes

iccicci:
# Context

Provider servers do not work in _mainnet_ / _preprod_ docker compose infrastructures.

# Proposed Solution

- Added the missing `poolMetadata` entity to `DataSource` of `TypeormStakePoolProvider` to fix the `stake-pool-provider-server` container.
- Fixed the `docker-compose` in order to forward the `HANDLE_POLICY_IDS` environment variable to `provider-server` container.

### DIFF
--- a/packages/cardano-services/docker-compose.yml
+++ b/packages/cardano-services/docker-compose.yml
@@ -49,6 +49,10 @@ services:
     volumes:
       - postgres-data:/var/lib/postgresql/data
 
+  provider-server:
+    environment:
+      HANDLE_POLICY_IDS: $HANDLE_POLICY_IDS
+
 secrets:
   blockfrost_key:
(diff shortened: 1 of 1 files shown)


gytis-ivaskevicius (approved):
I really like these extra quick fixes, good job :+1:

rhyslbw (commented):
I'm curious to know how this wasn't caught in our e2e or integration tests 🤔 

Also, you can drop the second `fix` (after the commit type).
Screenshot 2023-07-18 at 15 24 46

rhyslbw (approved):
Great work @iccicci. Clearly a gap in test coverage but, as you've pointed out in a back-channel, it's better to defer to an SDET since the suite run against the microservice is failing suspiciously.

PR #823, opened 2023-07-19
fix(input-selection): greedy input selector now produces a correct selection under all conditions

AngelCastilloB:
# Context

After the last round of changes to the greedy input selector, some of the property tests were failing when testing some edge cases.

# Proposed Solution

Fix the greedy input selector so it can handle all edge cases correctly.

### DIFF
--- a/packages/input-selection/test/GreedySelection/util.test.ts
+++ b/packages/input-selection/test/GreedySelection/util.test.ts
@@ -16,7 +16,7 @@ describe('splitChange', () => {
       undefined,
       () => 10n,
       () => false,
-      2_000_000n
+      0n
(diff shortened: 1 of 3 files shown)


iadmytro (approved):

packages/input-selection/test/GreedySelection/util.test.ts
```diff
@@ -16,7 +16,7 @@ describe('splitChange', () => {
       undefined,
       () => 10n,
       () => false,
-      2_000_000n
+      0n
```
mkazlauskas, line 19: Do we need a test where this is not 0?

PR #824, opened 2023-07-20
feat(input-selection): greedy input selection now distributes native assets evenly

AngelCastilloB:
# Context

Greedy input selection currently allocates all native assets at the first UTXO. This can create some problems for the user when doing some transactions (I.E interacting with DeFi). The algorithm should spread out the native assets as best as it can.

# Proposed Solution

The algorithm now spread assets evenly among available UTXOs.

### DIFF
--- a/packages/input-selection/src/GreedySelection/util.ts
+++ b/packages/input-selection/src/GreedySelection/util.ts
@@ -1,20 +1,20 @@
 /* eslint-disable func-style, max-params */
 import { BigNumber } from 'bignumber.js';
-import { Cardano } from '@cardano-sdk/core';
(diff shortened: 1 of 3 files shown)


iadmytro (commented):
It would be good to add property test for checking that:
 - no assets lost, all assets from input are present in output
 - assets evenly distributed, in output any (utxoA.assets.length - utxoB.assets.length) <= 1

iadmytro (dismissed):

mkazlauskas (dismissed):
Looks good!

mkazlauskas (approved):

iadmytro (approved):

PR #828, opened 2023-07-25
fix(web-extension): lw-7563 throw RemoteApiShutdownError on disconnec…

mirceahasegan:
# Context

Handle the case where port disconnects while waiting for remote method execution result.
Currently, the caller (api consumer) will hang waiting for a message with the execution result.

# Proposed Solution
Messengers have a new `disconnect$` observable that emits when on port disconnect.
The remoteApi consumer promise calls will return rejected promise in case of disconnect$ event.

### DIFF
--- a/packages/e2e/test/web-extension/extension/background/services.ts
+++ b/packages/e2e/test/web-extension/extension/background/services.ts
@@ -1,16 +1,18 @@
 // Expose any additional services to be shared with UIs
-import { BackgroundServices, adaPriceProperties, env, logger } from '../util';
+import { BackgroundServices, adaPriceProperties, disconnectPortTestObjProperties, env, logger } from '../util';
 import { Cardano } from '@cardano-sdk/core';
+import { Subject, of } from 'rxjs';
 import { adaPriceServiceChannel, walletName } from '../const';
 import { authenticator } from './authenticator';
-import { exposeApi, exposeSupplyDistributionTracker } from '@cardano-sdk/web-extension';
-import { of } from 'rxjs';
+import { consumeRemoteApi, exposeApi, exposeSupplyDistributionTracker } from '@cardano-sdk/web-extension';
 import { runtime } from 'webextension-polyfill';
 import { stakePoolProviderFactory } from '../../../../src';
 import { supplyDistributionTrackerReady } from './supplyDistributionTracker';
 
+const apiDisconnectResult$ = new Subject<string>();
 const priceService: BackgroundServices = {
   adaUsd$: of(2.99),
+  apiDisconnectResult$,
   clearAllowList: authenticator.clear.bind(authenticator),
   getPoolIds: async (count: number): Promise<Cardano.StakePool[]> => {
     const stakePoolProvider = await stakePoolProviderFactory.create(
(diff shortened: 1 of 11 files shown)


packages/web-extension/src/messaging/NonBackgroundMessenger.ts
```diff
@@ -102,6 +109,8 @@ export const createNonBackgroundMessenger = (
       }
       return messenger;
     },
+    // TODO: implement
```
mirceahasegan, line 112: Oups. Forgot to remove this

rhyslbw: That's a nice line to remove 😎

rhyslbw (approved):
Great work @mirceahasegan and @mkazlauskas 🐛

packages/e2e/test/web-extension/extension/background/services.ts
```diff
@@ -36,6 +38,28 @@ exposeApi(
   { logger, runtime }
 );
 
+// Dummy exposeApi object that closes the port as soon as it gets a message.
+// UI promise call should reject as a result of this
+// UI consumes API -> BG exposes fake API that closes port
+runtime.onConnect.addListener((port) => {
```
mkazlauskas, line 44: Previously adding multiple `onConnect` listeners didn't work reliably. Do the web-extension tests still pass with the same with similar consistency? We might want to use messenger for it - a full (even though typed as partial) port object should emitted on `connect$`.

packages/web-extension/src/messaging/types.ts
```diff
@@ -159,11 +159,17 @@ export interface DeriveChannelOptions {
...
+  disconnected: MinimalPort;
+  remaining: MinimalPort[];
+}
+
 export interface Messenger extends Shutdown {
   channel: ChannelName;
   connect$: Observable<MinimalPort>;
   postMessage(message: unknown): Observable<void>;
   message$: Observable<PortMessage>;
+  disconnect$: Observable<DisconnectEvent>;
```
mkazlauskas, line 172: When squashing, please mark the commit as breaking, as other possible implementations of the interface will be broken, also worth pointing it out in the commit body notes.

packages/e2e/test/web-extension/extension/ui.ts
```diff
@@ -244,3 +251,27 @@ document.querySelector('#buildAndSignTx')!.addEventListener('click', async () =>
   const { tx: signedTx } = await builtTx.sign();
   setSignature(signedTx.witness.signatures.values().next().value);
 });
+
+// Code below tests that a disconnected port in background script will result in the consumed API method call promise to reject
+// UI consumes API -> BG exposes fake API that closes port
+const disconnectPortTestObj = consumeRemoteApi(
```
mkazlauskas, line 257: Probably time to refactor ui.ts/ui.html to group by what it's testing. It's ok to create a separate task for it.

packages/web-extension/src/messaging/types.ts
mirceahasegan: Yeah, I missed this detail

packages/e2e/test/web-extension/extension/background/services.ts
mirceahasegan: The web-extension e2e test suite is passing consistently.

packages/e2e/test/web-extension/extension/ui.ts
mirceahasegan: added a task for it

mkazlauskas (approved):
Nice work! :rocket:

PR #832, opened 2023-07-27
refactor: transform handlePolicyId to an observable

greatertomi:
# Context

- PersonalWallet should always use the same policyId(s) as the HandleProvider, so it should take it from the HandleProvider rather than having a separate property/option for it.

# Proposed Solution
- create a getPolicyId function in HandleProvider
- Store handle information in pouchDB wallet stores


packages/cardano-services/src/Handle/TypeOrmHandleProvider.ts
```diff
@@ -55,4 +56,16 @@ export class TypeOrmHandleProvider extends TypeormProvider implements HandleProv
       return handles.map((handle) => mapEntity(entities.find((entity) => entity.handle === handle)));
     });
   }
+
+  async getPolicyIds(): Promise<Cardano.PolicyId[]> {
+    return this.withDataSource(async (dataSource) => {
```
mkazlauskas, line 61: If >1 policyId is found, we should cache it to not repeat the query for every request

mkazlauskas (dismissed):

packages/cardano-services/src/Handle/TypeOrmHandleProvider.ts
greatertomi: Fixed on 5b9717e

rhyslbw: @mkazlauskas won't this cache be invalid if a projector, configured with multiple policies, only has handles from one policy early in the chain history and the API is called?

rhyslbw: I'm OK to leave this query optimisation in place, but maybe we should consider maintaining a policy IDs meta table on projector startup as future work?

mkazlauskas: Yes you're right @rhyslbw , but shouldn't be an issue for now. We could also just not cache this (or short duration cache) if not at the tip, but I think for initial implementation this is fine.

rhyslbw: LW-7725

rhyslbw (changes requested):
`chore` commit types are not intended for modifying production code, so please rename `chore: transform handlePolicyId to an observable` to be of type `refactor`

https://stackoverflow.com/questions/26944762/when-to-use-chore-as-type-of-commit-message

There's also a lint warning to resolve

rhyslbw (changes requested):
https://github.com/input-output-hk/cardano-js-sdk/pull/832/commits/dfac4379ce4fac91ab20f9e8b0d284940e763a54 should also be a `fixup!` of one of the earlier commits

packages/cardano-services/src/Handle/TypeOrmHandleProvider.ts
```diff
@@ -55,4 +60,33 @@ export class TypeOrmHandleProvider extends TypeormProvider implements HandleProv
...
+      const fetchPolicyIdsFromDB = async () => {
+        const distinctPolicyIds = await dataSource
+          .getRepository(HandleEntity)
+          .createQueryBuilder('handle')
+          .select(`DISTINCT ${columnName}`, columnName)
+          .getRawMany();
+        const policyIds = distinctPolicyIds.map((row) => row[columnName]);
+
+        if (policyIds.length === 0) {
+          throw new Error('Value from the database is 0');
```
iccicci, line 76: Can't we directly return `[]` from here?

mkazlauskas: I don't think so, that would cache the `[]`

iccicci: Ok, thank you

packages/cardano-services/test/Handle/TypeOrmHandleProvider.test.ts
```diff
@@ -38,4 +38,10 @@ describe('TypeOrmHandleProvider', () => {
 
     expect({ handle, hasDatum }).toEqual({ handle: 'TestHandle', hasDatum: false });
   });
+
+  it('fetches all distinct policy ids', async () => {
+    const result = await provider.getPolicyIds();
+    expect(result.length).toBe(1);
+    expect(result).toEqual(['2c74874ae85a596ad88adc391c5fd7fdb8d8f85e22ced233ee85539d']);
```
rhyslbw, line 45: This assertion is a little weak as we can't see the source. Where is this value defined and why do we expect it here?

greatertomi: This data is inserted into the database [here](https://github.com/input-output-hk/cardano-js-sdk/blob/98262e0f2861b9e32956c0852c1d23bea3ae9b64/packages/cardano-services/test/jest-setup/snapshots/handle.sql#L1257-L1260). Since they both have the same policyId, the assertion expects just one to be selected.

rhyslbw: OK, here's two changes that can help improve the assertion:

1. Make a comment: `Sourced from https://github.com/input-output-hk/cardano-js-sdk/blob/98262e0f2861b9e32956c0852c1d23bea3ae9b64/packages/e2e/local-network/scripts/mint-handles.sh#L14`

2. Since the ID isn't part of the test intent:`expect(result.length).toBe(1);`

greatertomi: Do you mean I should remove the second assertion `expect(result).toEqual(['2c74874ae85a596ad88adc391c5fd7fdb8d8f85e22ced233ee85539d'])` and leave only `expect(result.length).toBe(1)`

rhyslbw: The policy ID can't be verified just by looking at the test or source code, so just an assertion on result length is enough IMO. Alternatively you could independently determine the policy ID from the same keys, which would make the test transparent and self-verifying, but linking to the source showing we're only minting under a single policy seems reasonable. Thoughts @mkazlauskas @iadmytro?

mkazlauskas: Agree, verifying length is sufficient here. IF we want to assert the policy id then we can also do this:
```ts
expect(typeof result[0]).toBe('string');
expect(result[0]).toHaveLength(56);
```

rhyslbw: Agree with both suggestions

greatertomi: Fixed on 9cac32d

rhyslbw: I was referencing the originating source, which is a script in the `e2e` package, but this is fine too.

compose/common.yml
```diff
@@ -235,6 +235,20 @@ services:
...
+      - *logging
+      - *provider-server
+      - *with-postgres
+    environment:
+      <<:
+        - *sdk-environment
+        - *provider-server-environment
+      SERVICE_NAMES: handle
+    ports:
+      - ${HANDLE_API_PORT:-4020}:3000
```
iccicci, line 250: I would use a progressive port number here: 4011 rather than 4020

But moreover `handle-provider-server` specifications are missing in `packages/e2e/docker-compose.yml` and in `packages/cardano-services/docker-compose.yml`.
They are required to provide the handle policyIds from the env.

packages/cardano-services/src/Handle/TypeOrmHandleProvider.ts
```diff
@@ -55,4 +61,33 @@ export class TypeOrmHandleProvider extends TypeormProvider implements HandleProv
...
+        }
+
+        return policyIds;
+      };
+
+      try {
+        return this.inMemoryCache.get('policyIds', fetchPolicyIdsFromDB);
+      } catch (error: any) {
+        if (error.message === 'Value from the database is 0') {
```
iccicci, line 86:
```suggestion
        if (error instanceof Error && error.message === 'Value from the database is 0') {
```

This should allow to not disable `@typescript-eslint/no-explicit-any` eslint rule

iccicci (commented):
I'm not strong enough to review the core of the change... I did what I can

compose/common.yml
greatertomi: Fixed on 5eef2a2

iccicci (dismissed):
From my point of view, it's good

rhyslbw (approved):
Good job getting this over the line @greatertomi 🤝

PR #833, opened 2023-07-28
LW-6950 projected blocks buffer

iccicci:
# Context

Our projector needs some speed improvement

# Proposed Solution

Implemented a blocks buffer between ogmios and the projector, parallelizing more requests to ogmios.

### DIFF
--- a/packages/core/src/CardanoNode/util/bufferChainSyncEvent.ts
+++ b/packages/core/src/CardanoNode/util/bufferChainSyncEvent.ts
@@ -1,0 +1,65 @@
+import { Observable } from 'rxjs';
+import { RequestNext, WithRequestNext } from '../types';
+
+export const bufferChainSyncEvent =
+  <T extends WithRequestNext>(length: number) =>
+  // eslint-disable-next-line sonarjs/cognitive-complexity
+  (source$: Observable<T>) =>
+    new Observable<T>((subscriber) => {
+      const buffer: WithRequestNext[] = [];
+      let lastPull: RequestNext | undefined;
+      let unsubscribed = false;
+      let subscriberReady = true;
+      let sourceCompleted = false;
+      let sourceError: unknown | undefined;
+
+      const next = () => {
+        if (unsubscribed) return;
+
+        if (buffer.length < length && lastPull) {
+          lastPull();
+          lastPull = undefined;
+        }
+
+        if (!subscriberReady) return;
+
+        if (buffer.length === 0) {
+          if (sourceCompleted) return subscriber.complete();
+          if (sourceError) return subscriber.error(sourceError);
+
+          return;
+        }
+
+        const value = buffer.shift()!;
+
+        subscriberReady = false;
+        subscriber.next({
+          ...value,
+          requestNext: () => {
+            subscriberReady = true;
+            next();
+          }
+        } as T);
+      };
+
+      const subscription = source$.subscribe({
+        complete: () => {
+          sourceCompleted = true;
+          next();
+        },
+        error: (error) => {
+          sourceError = error;
+          next();
+        },
+        next: (value) => {
+          lastPull = value.requestNext;
+          buffer.push(value);
+          next();
+        }
+      });
+
+      return () => {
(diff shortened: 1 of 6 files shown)


greatertomi (approved):
LGTM

packages/core/src/CardanoNode/util/bufferChainSyncEvent.ts
```diff
@@ -0,0 +1,50 @@
+import { Observable } from 'rxjs';
+import { RequestNext } from '../types';
+
+export interface WithRequestNext {
```
mkazlauskas, line 4: This already exists

packages/core/src/CardanoNode/util/bufferChainSyncEvent.ts
```diff
@@ -0,0 +1,50 @@
...
+        complete: subscriber.complete.bind(subscriber),
+        error: subscriber.error.bind(subscriber),
+        next: (value) => {
+          lastPull = value.requestNext;
+          buffer.push(value);
+          next();
+        }
+      });
+
+      return () => (unsubscribed = true);
```
mkazlauskas, line 49:
```suggestion
      const subscription = source$.subscribe({
        complete: subscriber.complete.bind(subscriber),
        error: subscriber.error.bind(subscriber),
        next: (value) => {
          lastPull = value.requestNext;
          buffer.push(value);
          next();
        }
      });

      return () => {
         return subscription.unsubscribe();
         unsubscribed = true;
       };
```

packages/core/src/CardanoNode/util/bufferChainSyncEvent.ts
```diff
@@ -0,0 +1,50 @@
...
+        } as T);
+      };
+
+      source$.subscribe({
+        complete: subscriber.complete.bind(subscriber),
+        error: subscriber.error.bind(subscriber),
+        next: (value) => {
+          lastPull = value.requestNext;
+          buffer.push(value);
+          next();
```
mkazlauskas, line 45: This looks a little odd: are we triggering the emission when source$ emits? It should only do that when buffer is empty.

packages/cardano-services/src/Projection/createTypeormProjection.ts
```diff
@@ -118,6 +118,7 @@ export const createTypeormProjection = ({
   });
   return projectionSource$.pipe(
     applyMappers(mappers),
+    CardanoNodeUtil.bufferChainSyncEvent<ProjectionEvent>(10),
```
mkazlauskas, line 121: This should be inferred from source$
```suggestion
    CardanoNodeUtil.bufferChainSyncEvent(10),
```

mkazlauskas (changes requested):

packages/core/src/CardanoNode/util/bufferChainSyncEvent.ts
```diff
@@ -0,0 +1,50 @@
...
+
+      const next = () => {
+        if (unsubscribed) return;
+
+        if (buffer.length < length && lastPull) {
+          lastPull();
+          lastPull = undefined;
+        }
+
+        if (buffer.length === 0 || !subscriberReady) return;
```
mkazlauskas, line 25: I may be wrong but if I remember correctly observable chain sync client emits 1 event initially on it's own, without calling requestNext. This may be why it's not projecting.

packages/core/src/CardanoNode/util/bufferChainSyncEvent.ts
iccicci: fd7ef848c and 7328a70cc should solve this

packages/cardano-services/src/Projection/createTypeormProjection.ts
iccicci: fd7ef848c and 3aeb79ca1 should solve this

rhyslbw (approved):
Nice job @iccicci 💥

iccicci:
Replaced by #848

PR #838, opened 2023-08-01
refactor: move isValidHandle util function into core package

greatertomi:
# Context

Move `isValidHandle` function from being used locally into `core` package and added tests

# Proposed Solution


mkazlauskas (dismissed):
LGTM!

packages/core/test/Asset/util/isValidHandle.test.ts
```diff
@@ -0,0 +1,26 @@
...
+  'handle_name'
+];
+
+const invalidHandles = ['bob!', '$wallet', 'alice@', 'test*handle#2', 'lace&bob', 'ada%1_test'];
+
+describe('isValidHandle', () => {
+  test('returns false for empty string', () => {
+    expect(Asset.util.isValidHandle('')).toBe(false);
+  });
+  test.each(validHandles)('returns true for valid handles', (handle) => {
```
iccicci, line 20:
```suggestion
  test.each(validHandles)('returns true for valid handle "%s"', (handle) => {
```

This would make the output of `yarn test --verbose` a bit more clear as the failing tests description will immediately report the failing _handle name_.

iccicci (commented):
Nice! Just a minor improvement on tests output

packages/core/test/Asset/util/isValidHandle.test.ts
greatertomi: 06e824b should resolve this

rhyslbw (changes requested):
This is a `feat` commit type

rhyslbw (approved):
Ah sorry, I was getting confused with tickets. All good! Well done @greatertomi

PR #841, opened 2023-08-02
lw-6517: replace legacy trezor connect with dedicated node and web packages

tomislavhoracek:
# Context

Trezor introduced dedicated packages for the web and node implementation. The scope of this task was to split the legacy `trezor-connect` package with `@trezor/connect` and `@trezor/connect-web` packages.

# Proposed Solution

Split the legacy `trezor-connect` package with `@trezor/connect` and `@trezor/connect-web` packages and extend the code with `communicationType` so the user can communicate with the specific Trezor  package depending on the environment he needs
 
1. BREAKING CHANGE: `communicationType` param is the required param for all TrezorKeyAgent methods.
By defining `communicationType` user is able to communicate with the proper Trezor web or node package
2. Minimal node version updated to v16 in all packages

### DIFF
--- a/packages/cardano-services-client/package.json
+++ b/packages/cardano-services-client/package.json
@@ -3,7 +3,7 @@
   "version": "0.10.6",
   "description": "Cardano Services Client",
   "engines": {
-    "node": ">=14.20.1"
+    "node": ">=16.20.1"
   },
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
--- a/packages/cardano-services/package.json
+++ b/packages/cardano-services/package.json
@@ -3,7 +3,7 @@
   "version": "0.14.8",
   "description": "Cardano GraphQL Services",
   "engines": {
-    "node": ">=14.20.1"
+    "node": ">=16.20.1"
   },
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -3,7 +3,7 @@
   "version": "0.15.4",
   "description": "Core types and libraries for Cardano",
   "engines": {
-    "node": ">=14.20.1"
+    "node": ">=16.20.1"
   },
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
--- a/packages/crypto/package.json
+++ b/packages/crypto/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.9",
   "description": "Cryptographic types and functions for Cardano",
   "engines": {
-    "node": ">=14.20.1"
+    "node": ">=16.20.1"
   },
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
--- a/packages/dapp-connector/package.json
+++ b/packages/dapp-connector/package.json
@@ -3,7 +3,7 @@
   "version": "0.9.11",
   "description": "TypeScript definitions for the dApp Connector standard CIP30",
   "engines": {
-    "node": ">=14.20.1"
+    "node": ">=16.20.1"
   },
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
--- a/packages/e2e/package.json
+++ b/packages/e2e/package.json
@@ -3,7 +3,7 @@
   "version": "0.17.1",
   "description": "End to end tests for the cardano-js-sdk packages.",
   "engines": {
-    "node": ">=16.0"
+    "node": ">=16.20.1"
   },
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
--- a/packages/golden-test-generator/package.json
+++ b/packages/golden-test-generator/package.json
@@ -3,7 +3,7 @@
   "version": "0.7.19",
   "description": "Generate golden test files for a range of Cardano concepts",
   "engines": {
-    "node": ">=14.20.1"
+    "node": ">=16.20.1"
   },
   "bin": "dist/cjs/index.js",
   "main": "dist/cjs/index.js",
--- a/packages/governance/package.json
+++ b/packages/governance/package.json
@@ -3,7 +3,7 @@
   "version": "0.7.14",
   "description": "Governance types and utilities for Cardano",
   "engines": {
-    "node": ">=14.20.1"
+    "node": ">=16.20.1"
   },
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
--- a/packages/hardware-ledger/package.json
+++ b/packages/hardware-ledger/package.json
@@ -3,7 +3,7 @@
   "version": "0.2.17",
   "description": "Mappings and integration with Ledger hardware",
   "engines": {
-    "node": ">=14.20.1"
+    "node": ">=16.20.1"
   },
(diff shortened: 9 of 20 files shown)


rhyslbw (commented):
Thanks @tomislavhoracek. Node 16 is under maintenance so we could make the min version [16.20.1](https://nodejs.org/download/release/latest-v16.x/)

rhyslbw (dismissed):
LGTM

DominikGuzei (approved):
Code LGTM 👍 left one small comment

packages/key-management/src/util/mapHardwareSigningData.ts
```diff
@@ -163,7 +163,7 @@ const prepareTrezorOutputs = (
       const destination = ownAddress
         ? {
             addressParameters: {
-              addressType: trezor.CardanoAddressType.BASE,
+              addressType: trezor.PROTO.CardanoAddressType.BASE,
```
marcin-mazurek, line 166: Nitpick: just wondering if we should make `trezor` capitalized. We seem to apply this convention in SDK packages eg. `KeyManagement.TrezorKeyAgent.initializeTrezorTransport`.

marcin-mazurek (commented):
Hey Tomo, well done 👍
Left two questions.

packages/key-management/src/TrezorKeyAgent.ts
```diff
@@ -52,11 +61,12 @@ export class TrezorKeyAgent extends KeyAgentBase {
     silentMode = false,
     lazyLoad = false
   }: TrezorConfig): Promise<boolean> {
+    const trezorConnect = getTrezorConnectInstance(communicationType);
```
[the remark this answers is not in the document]
mkazlauskas: I don't think they have to be static, but one thing to have in mind is that connection to the device probably has to be initiated synchronously on user gesture, so a separate function to connect can be convenient. I'm sure we can improve this API, but need to review how it's being used in Lace.

packages/key-management/src/util/mapHardwareSigningData.ts
tomislavhoracek: @marcin-mazurek I kept it as is since the `ledger` is also not capitalized (https://github.com/input-output-hk/cardano-js-sdk/blob/047aed18216d3b03abf7c06d0931015a40d10ee2/packages/key-management/src/util/mapHardwareSigningData.ts#L4). 
It is changed in new dedicated `hardware-trezor` package https://github.com/input-output-hk/cardano-js-sdk/blob/12d8422fadb3ced2d0ffc1d00c611d3728713870/packages/hardware-trezor/src/TrezorKeyAgent.ts#L3

packages/key-management/src/TrezorKeyAgent.ts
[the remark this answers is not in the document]
tomislavhoracek: @marcin-mazurek 
1. yeah instance is always initialized with `new` but in this case, I had a dilemma and I rather used an instance keyword to say what package you will use instead of saying e.g. `getTrezorConnectPackage` or similar. 
2. `static` methods: That is an idea from the very first implementation and it is quite useful when you want to instantiate ledger or trezor communication on your own and without instantiating TrezorKeyAgent. For example, you just want to check the device connection or you want to create device connection and pass it to `createWithDevice` [here](https://github.com/input-output-hk/cardano-js-sdk/blob/12d8422fadb3ced2d0ffc1d00c611d3728713870/packages/hardware-ledger/src/LedgerKeyAgent.ts#L266C53-L266C69) . In that case, you are able to communicate with a device on your own without KeyAgent instances and without installing ledger/trezor packages in the app. Let's say that you have more open hands in that way

packages/key-management/src/util/mapHardwareSigningData.ts
```diff
@@ -44,7 +44,7 @@ export interface TrezorMintBundle {
 
 export interface TrezorCertificates {
   certs: trezor.CardanoCertificate[];
-  signingMode?: trezor.CardanoTxSigningMode;
+  signingMode?: trezor.PROTO.CardanoTxSigningMode;
```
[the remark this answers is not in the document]
tomislavhoracek: yes it is https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/constants/index.ts#L6
It was exported from `'@trezor/transport/lib/types/messages';` before

packages/key-management/src/util/mapHardwareSigningData.ts
marcin-mazurek: So `Trezor` used to be PascalCase before your changes. Lowercase `ledger` feels inconsistent with internal SDK types too:

image

Thoughts @AngelCastilloB / @mkazlauskas / @rhyslbw?

marcin-mazurek: Sorry - it was TrezorConnect that was pascal case. But still something I'd like to consider :)

mkazlauskas: We might have some inconsistencies but in general yes, PascalCase is used for namespaces

rhyslbw: `ledger` is an import so doesn't impact the API, but yeah should probably be `Ledger`

packages/key-management/src/TrezorKeyAgent.ts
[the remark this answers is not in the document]
marcin-mazurek: @tomislavhoracek 

1. Two proposals not using `instance` keyword: how about `getTrezorConnect` or `getTrezorConnectByCommunicationType`?
2. OK - thanks for the explanation, considering the fact that Ledger is also done this way and any changes would result in a lot of breaking changes, it'd probably make sense to stick to the current approach

packages/key-management/src/util/mapHardwareSigningData.ts
marcin-mazurek: Ok, @tomislavhoracek explained to me on a call that this file will be removed, so it's totally fine to keep the diff to minimum :)

packages/key-management/src/TrezorKeyAgent.ts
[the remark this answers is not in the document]
tomislavhoracek: We discussed this in a call and we decided to use `getTrezorConnect` instead of `getTrezorConnectInstance`
cc @rhyslbw @mkazlauskas

DominikGuzei (dismissed):
Still great as before :D 🚢

marcin-mazurek (approved):

DominikGuzei (approved):

marcin-mazurek (approved):

rhyslbw (approved):

tomislavhoracek:
> 

sure, thanks @rhyslbw

tomislavhoracek:
> Sorry @tomislavhoracek, I just noticed [6efc12d](https://github.com/input-output-hk/cardano-js-sdk/commit/6efc12d237be64df31847245f988f977fcc8d295) is marked as breaking, which will increment the package versions unnecessarily. Please commit it without the delineator

@rhyslbw ✅

rhyslbw (approved):

PR #848, opened 2023-08-07
LW-6950 buffer operator

iccicci:
# Context

We need to optimize the sync phase of the projector

# Proposed Solution

Added a buffer operator after reading blocks from ogmios


packages/ogmios/src/CardanoNode/OgmiosObservableCardanoNode/OgmiosObservableCardanoNode.ts
```diff
@@ -149,7 +149,7 @@ export class OgmiosObservableCardanoNode implements ObservableCardanoNode {
                   chainSync$: createObservableChainSyncClient(
                     { intersectionPoint: intersection.point },
                     { interactionContext$: this.#interactionContext$ }
-                  ),
+                  ).pipe(ObservableCardanoNode.bufferChainSyncEvent(10)),
```
mkazlauskas, line 152: This should be configurable. Adding it as argument to `findIntersect` doesn't seem right, so maybe it's not the best place to apply this operator. `fromCardanoNode` seems like a better place.
```suggestion
                  ).pipe(ObservableCardanoNode.bufferChainSyncEvent(bufferSize)),
```

mkazlauskas (changes requested):

packages/ogmios/src/CardanoNode/OgmiosObservableCardanoNode/OgmiosObservableCardanoNode.ts
iccicci: Sorry @mkazlauskas , I can't understand where...
Could you please tell me, precisely, where in `fromCardanoNode` should it be moved?
Thank you

mkazlauskas: I think we can add buffer size prop [here](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/projection/src/Bootstrap/fromCardanoNode.ts#L125-L127) and apply the operator [here before withRolledBackBlock](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/projection/src/Bootstrap/fromCardanoNode.ts#L44)

iccicci: 0a7e9f0dc and 05e82468f should solve this

mkazlauskas: `tsc` error in `projection-typeorm` tests (it's using `fromCardanoNode`). I suggest setting a default in there as well, or just make it optional and `passthrough()` instead of the buffer if not set

iccicci: Why it is not caught by `yarn test:build:verify`? Checking...

packages/core/src/CardanoNode/util/bufferChainSyncEvent.ts
```diff
@@ -0,0 +1,65 @@
+import { Observable } from 'rxjs';
+import { RequestNext, WithRequestNext } from '../types';
+
+export const bufferChainSyncEvent =
+  <T extends WithRequestNext>(length: number) =>
+  // eslint-disable-next-line sonarjs/cognitive-complexity
```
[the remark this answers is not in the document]
iccicci: I tried, but it is a set of nested closures... every _simplification attempt_ I tried resulted in something harder to read...

packages/ogmios/src/CardanoNode/OgmiosObservableCardanoNode/OgmiosObservableCardanoNode.ts
iccicci: The change to `tsconfig.json` in dd11eab11 solves the `yarn test:build:verify` problem as well

packages/cardano-services/src/Program/programs/projector.ts
```diff
@@ -61,6 +63,7 @@ const createProjectionHttpService = async (options: ProjectionMapFactoryOptions)
       handlePolicyIds
     },
     projectionSource$: Bootstrap.fromCardanoNode({
+      blocksBufferLength,
```
mkazlauskas, line 66: Question: are we sure this is mapped correctly from the cli arg/env?

packages/projection-typeorm/test/tsconfig.json
```diff
@@ -3,9 +3,7 @@
   "compilerOptions": {
     "baseUrl": "."
   },
-  "include": [
-    "../../golden-test-generator/src"
-  ],
+  "include": [".", "../../golden-test-generator/src"],
```
mkazlauskas, line 6: Please extract this to a separate commit and apply the same change to tsconfig in other packages where applicable

mkazlauskas (changes requested):

packages/cardano-services/src/Program/programs/projector.ts
iccicci: Actually we do not: I haven't tested this, nice catch!
The only idea I have is to log it from the projector and check the logs in `cli.test.ts`.
Does it sounds as a good test or do you have some better idea?
Thank you

mkazlauskas: Yeah that would be fine, it's probably really difficult to set up a proper test, need to abstract away the cli :sweat_smile:

iccicci: 1d07bf077 should solve this

mkazlauskas (approved):
Great work! :muscle:

rhyslbw (approved):
Great work @iccicci 🚀

PR #852, opened 2023-08-08
LW-7733 Added custom classes for certificate serialization

AngelCastilloB:
# Context

# Proposed Solution

- Added Certificate serialization classes.
- Update MIR certificate core type


packages/core/src/Serialization/Certificates/CertificateKind.ts
```diff
@@ -0,0 +1,47 @@
+/**
+ * Certificates are used to register, update, or deregister stake pools, and delegate stake.
```
mkazlauskas, line 2: I had a question why not consolidate this with `Cardano.CertificateType`, so probably worth adding a comment to explain
```suggestion
 * Certificates are used to register, update, or deregister stake pools, and delegate stake.
 * Values are used for serialization.
```

mkazlauskas (changes requested):

mkazlauskas (approved):
Great work! Appreciate the extensive doc comments :muscle:

mirceahasegan (approved):
Impressive 🧙‍♂️ 👏

PR #855, opened 2023-08-09
chore: treefmt formatting

blaggacao:
- chore: enable shfmt and prettier with treefmt
- style: treefmt
- chore: ignore formatting commit in git blame

# Context

Same as #825, but with ignoring the formatting commit during blame.

# Proposed Solution

- Isolates the reformatting into a single commit based on an already fine-tuned `treefmt.toml` (from #825)
- Adds that commit to the ingore rev for blame file


.git-blame-ignore-revs
```diff
@@ -0,0 +1,4 @@
+# To apply this file locally, run: `git config blame.ignoreRevsFile .git-blame-ignore-revs`
+
+# applied treefmt for the first time
+3391af283a72871b3afe23dd721e35c52daba241
```
gytis-ivaskevicius, line 4: Not sure if it's all that helpful and probably not worth keeping it around. CC @rhyslbw

blaggacao: This is common recommended practice and leverages the corresponding [GitHub feature](https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/).

rhyslbw: Oh that's a cool feature! Looks good here

blaggacao:
Adopted into #825 after consultation with @tshaynik .

PR #856, opened 2023-08-09
ci optimize cycle time

blaggacao:
- fix(local/envs): yq is provided by yq-go package - oversight
- ci: speed up cycle time via the use of remote store

# Context

Nix builds can use remote stores (on the cli seen as something like `--store ssh-ng://...`).

Their use significantly reduces the amount of back and forth copying when building on a remote builder and thereby notably reduces the cycle time.

# Proposed Solution

Newer versions of `std-action` (on current `main` branch) have support for this feature, so we enable it here.

The other fixes are incidential.

### DIFF
--- a/.github/workflows/std.yml
+++ b/.github/workflows/std.yml
@@ -1,6 +1,7 @@
+# yq -y < std-unexploded.yml > workflows/std.yml
 name: STD
-'on':
-  workflow_dispatch: null
+on:
+  workflow_dispatch:
   pull_request:
     branches:
       - develop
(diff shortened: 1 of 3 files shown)


.github/workflows/std.yml
```diff
@@ -24,22 +26,26 @@ jobs:
...
           mask-aws-account-id: false
       - name: Login to Amazon ECR
         id: login-ecr
         uses: aws-actions/amazon-ecr-login@v1
+      # TODO: uncomment when nixbuild works well together with
+      #       nix daemon mode on hosted runners
+      # - uses: nixbuild/nixbuild-action@v17
+      #   with:
+      #     nixbuild_ssh_key: ${{ secrets.SSH_PRIVATE_KEY }}
+      #     generate_summary_for: job
```
gytis-ivaskevicius, line 46:
```suggestion
```

gytis-ivaskevicius (approved):
Looks good, I'd like to see some benchmarks before merging this but I guess we can just merge it and run this repo for a couple of weeks

.github/workflows/std.yml
blaggacao: This file is autogenerated from `../std-unexploded.yaml` so this alone wouldn't have any lasting effect.

tshaynik (approved):
LGTM

blaggacao:
The failing test is necessary flaky since this PR doesn't even remotely touch that code path.

PR #859, opened 2023-08-10
test(cardano-services): removed healthcheck from typeorm stake pool provider test

iadmytro:
# Context

Health-check in stake pool provider test caused race condition.

# Proposed Solution
Check only status code of started service.


mkazlauskas (approved):
Thanks @iadmytro

iccicci (approved):
As good as easy

PR #862, opened 2023-08-11
feat(wallet): filter out duplicate utxo

mkazlauskas:
# Context

Some bug in transactions tracker state machine or  backend utxo queries might result in duplicate utxo

# Proposed Solution

Filter out the duplicates before emitting, but keep an error log message


mirceahasegan (approved):
💪

rhyslbw (approved):

PR #864, opened 2023-08-14
ci: implement 4th packaging layer & continuous deployment

blaggacao:
Deployment to `dev-preview` only.

# Context

To shorten the feedback loop ...

# Proposed Solution

... this PR implements CD to `dev-preview`.

- The source of truth for deployment configuration is now `./nix/cardano-services/deployments/`
- Kustomize is used for specialization (see `./nix/cardano-services/deployments/kustomization.yaml`)
- Current Image builds are overlayed from the build DAG via `./nix/cardano-services/deployments.nix`


.github/std-unexploded.yml
```diff
@@ -106,3 +106,48 @@ jobs:
...
+      matrix:
+        target: ${{ fromJSON(needs.discover.outputs.hits).deployments.apply }}
+    steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@main
+        with:
+          role-to-assume: ${{ env.AWS_ROLE_ARN }}
+          aws-region: ${{ env.AWS_REGION }}
+
+      - uses: nixbuild/nix-quick-install-action@v25
```
tshaynik, line 133: So nixbuild is working well now? Cross-referencing with the other PR where removing Nixbuild seemed to result in a significant build time improvement.

rhyslbw (commented):
This looks like a nice improvement from a configuration standpoint. We're still carrying the "backend" term, which is used in place of `provider-server`, but I can appreciate that's a separate issue.

nix/cardano-services/deployments.nix
```diff
@@ -0,0 +1,57 @@
...
+  baseline = with haumea.lib;
+    load {
+      src = ./deployments;
+      loader = [(matchers.regex ''^.+\.(yaml|yml)'' loadYaml)];
+    };
+in {
+  dev-preview = dmerge baseline {
+    meta.description = "Development Environment on the Cardano Preview Chain (Frankfurt)";
+    backend-deployment = {
+      spec.template.spec.containers = dmerge.updateOn "name" [
```
tshaynik, line 17: My big questions going forward here are:
- How will we manage this code as we move to more complex overrides?
- What are the boundaries between dmerge and Kustomize, if both are going to be used together?

Right now this is effectively reimplementing Kustomize functionality, but without the useful domain-specific abstractions for common Kubernetes use cases, like setting a container image version, namespace, or applying common labels.

nix/cardano-services/deployments/kustomization.yaml
```diff
@@ -0,0 +1 @@
+namespace: dev-preview
```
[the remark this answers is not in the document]
blaggacao: It did appear to work just with `namespace`. But better have this well-specced. Thx!

nix/cardano-services/deployments/backend-deployment.yaml
```diff
@@ -0,0 +1,108 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: backend
+    network: preview
+    release: dev-preview-cardanojs
```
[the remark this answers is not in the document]
blaggacao: I've done a bit of research about well known labels and otherwise best practices and it appears to me that the adoption of best practices requires a revision of our current set up that is more than just move then into `commonLabels`. For example, `release` is not a well-known label and also doesn't really have much of a semantic meaning (what does "`release: dev-preview-cardanojs`" even refer to?). `network` on the other hand should probably be `cardano.iog.io/network` as labels and annotations appear to always refer to fully qualified names rather than shorthands.

I've backlogged this important work as (internal ticket) [LW-7969](https://input-output.atlassian.net/browse/LW-7969).

nix/cardano-services/deployments.nix
blaggacao: > What are the boundaries between dmerge and Kustomize, if both are going to be used together?

This is easy. If the nix-plane is the source of a configuration datum, use it to ensure cheap dynamic value updates.  If not, keep that datum self-contained "within the YAML" (including `Kustomization` for ease of maintenance). If a value is overwhelmingly a constant, but technically emanates from the nix-plane, then make an educated judgement with propensity to prefer Kustomization over nix-plane sourcing of that nearly constant configuration datum.

> How will we manage this code as we move to more complex overrides?

I don't expect more than one layer of overrides in this scenario, especially since we can leverage Kustomize for many things. If, after gathering more feedback, we conclude that multi-layer / nested overlays are unavoidable we can do so by duck-chaining overlays with the `dmerge.chainable` facility that was design precisely for more complex overlays.

But I can't yet fathom that this may be even necessary, given how much of kubernete's accidential complexity due to flat data structures is already mitigated by `Kustomization`.

nix/cardano-services/deployments.nix
```diff
@@ -0,0 +1,57 @@
...
+    };
+    handle-projector-deployment = {
+      spec.template.spec.containers = dmerge.updateOn "name" [
+        {
+          name = "handle-projector";
+          image = cell.oci-images.cardano-services.image.name;
+        }
+      ];
+    };
+    pgboss-deployment = {
```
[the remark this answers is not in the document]
blaggacao: This is still a mess and I captured improvement as part of our (internal ticket) [LW-7969](https://input-output.atlassian.net/browse/LW-7969).

In this particular case the `pgboss-deployment` key matches `pgboss-deployment.yaml` object (key is filname less extension). But if you look at the actual manifest, we observe naming like:

```yaml
    app: pg-boss-worker
    [ ... ]
          name: pg-boss-worker
```

To increase consistency, here too, I'd defer to LW-7969

.github/std-unexploded.yml
```diff
@@ -106,3 +106,48 @@ jobs:
...
+    environment:
+      name: dev-preview
+      url: https://backend.dev-preview.eks.lw.iog.io
+    if: fromJSON(needs.discover.outputs.hits).deployments.apply != '{}'
+    strategy:
+      matrix:
+        target: ${{ fromJSON(needs.discover.outputs.hits).deployments.apply }}
+    steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@main
```
gytis-ivaskevicius, line 128: Once we stabilize this repo Ill expect for all these inputs to be pinned

.github/std-unexploded.yml
```diff
@@ -106,3 +106,48 @@ jobs:
       - uses: divnix/std-action/run@main
         with: {ffBuildInstructions: true, remoteStore: "ssh-ng://eu.nixbuild.net"}
 
+  deploy-to-eu:
```
gytis-ivaskevicius, line 109: Why is it region-specific?

nix/cardano-services/deployments.nix
```diff
@@ -0,0 +1,57 @@
+let
+  inherit (inputs.std) dmerge;
+  inherit (inputs.std.inputs) haumea;
+  inherit (inputs.std.lib.ops) readYAML;
+
+  loadYaml = _: _: readYAML;
```
gytis-ivaskevicius, line 6: :raised_eyebrow:

nix/cardano-services/deployments.nix
```diff
@@ -0,0 +1,57 @@
...
+    };
+    pgboss-deployment = {
+      spec.template.spec.containers = dmerge.updateOn "name" [
+        {
+          name = "pg-boss-worker";
+          image = cell.oci-images.cardano-services.image.name;
+        }
+      ];
+    };
+    stake-pool-projector-deployment = {
```
gytis-ivaskevicius, line 48: Missing stake-pool provider and blockfrost services

nix/cardano-services/deployments.nix
```diff
@@ -0,0 +1,57 @@
...
+      ];
+    };
+    coingecko-proxy-deployment = {
+      spec.template.spec.containers = dmerge.updateOn "name" [
+        {
+          name = "coingecko-proxy";
+          image = cell.oci-images.cardano-services.image.name;
+        }
+      ];
+    };
```
gytis-ivaskevicius, line 31: I don't think that this should be here

nix/cardano-services/deployments/backend-deployment.yaml
```diff
@@ -0,0 +1,108 @@
...
+        - name: dockerconfigjson
+      volumes:
+        - csi:
+            driver: csi.cert-manager.io
+            readOnly: true
+            volumeAttributes:
+              csi.cert-manager.io/issuer-kind: ClusterIssuer
+              csi.cert-manager.io/issuer-name: root-ca
+              csi.cert-manager.io/key-usages: client auth
+          name: tls
```
gytis-ivaskevicius, line 108: I am not a fan to have this set up for single environment only but I guess we can merge it and fix it later

I was curious how it would look like if this yaml was formatted properly. 108 lines turned into 50 lines. In case of cardano-stack we would also have to remove `resources`
```suggestion
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: backend
spec:
  selector:
    matchLabels:
      app: backend
  template:
    metadata:
      labels:
        app: backend
    spec:
      containers:
        - args:
            - start-provider-server
          livenessProbe:
            httpGet:
              path: /health
              port: 3000
            timeoutSeconds: 5
          name: backend
          ports:
            - containerPort: 3000
              name: http
          resources:
            limits:
              cpu: 1500m
              memory: 512Mi
            requests:
              cpu: 1000m
              memory: 350Mi
          securityContext:
            runAsGroup: 0
            runAsUser: 0
          volumeMounts:
            - mountPath: /tls
              name: tls
      imagePullSecrets:
        - name: dockerconfigjson
      volumes:
        - csi:
            driver: csi.cert-manager.io
            readOnly: true
            volumeAttributes:
              csi.cert-manager.io/issuer-kind: ClusterIssuer
              csi.cert-manager.io/issuer-name: root-ca
              csi.cert-manager.io/key-usages: client auth
          name: tls
```

nix/cardano-services/deployments/stake-pool-provider-deployment.yaml
```diff
@@ -0,0 +1,101 @@
...
+                secretKeyRef:
+                  key: username
+                  name: stakepool-owner-user.dev-preview-dbsync-db.credentials.postgresql.acid.zalan.do
+            - name: SERVICE_NAMES
+              value: stake-pool
+            - name: TOKEN_METADATA_SERVER_URL
+              value: http://dev-preview-cardano-stack-metadata.dev-preview.svc.cluster.local
+            - name: USE_TYPEORM_STAKE_POOL_PROVIDER
+              value: "true"
+          image: 926093910549.dkr.ecr.us-east-1.amazonaws.com/cardano-services-server:aana7a352vdaz0rs41yhfh3s85inw004
```
gytis-ivaskevicius, line 68: :roll_eyes:

gytis-ivaskevicius (changes requested):
Please write down testing scenarios and also please prepare PR to remove dev-preview deployment from the other repo

nix/cardano-services/deployments/stake-pool-provider-deployment.yaml
gytis-ivaskevicius: This thing exists everywhere. Should be removed

.github/std-unexploded.yml
blaggacao: That's an oversight, `v2.0.0` had been broken for a short while back then.

.github/std-unexploded.yml
blaggacao: The kubecontext is acquired by the CI role from the region variable and is also only valid for that region, unfortunately.

It is subobtimal. I have investigated k8s region topology just a bit in an attempt to find better long term answers. Another intermediate one would be to simply parallelise GH actions jobs. In any case, this needs more investigation, still.

nix/cardano-services/deployments.nix
blaggacao: I kept the function signature that implements the `haumea` contract. So this is to say: "it's not an error, it's here so that the reader can identify and understand the haumea contract. In-lining makes this easily overlooked.

And you stumbling over it, it seems to have fulfilled its purpose :smile:

nix/cardano-services/deployments.nix
blaggacao: Indeed not. :smile: And the coingecko proxy should move into this repo consequentially, too. At least judging by the architectural criterion. I backlogged with due urgency [LW-8023](https://input-output.atlassian.net/browse/LW-8023) for follow up.

nix/cardano-services/deployments.nix
gytis-ivaskevicius: Can you just drop the type definition comment or something so contract would be clear?

.github/std-unexploded.yml
blaggacao: Addressed in [force-push](https://github.com/input-output-hk/cardano-js-sdk/compare/10167f20f9c7d8e8870f84755ab2e0bd1a902e2f..47ce6e3ec1f74f224a253a6c05706323d0dd3ba9)

nix/cardano-services/deployments.nix
blaggacao: Partially addressed in [force-push](https://github.com/input-output-hk/cardano-js-sdk/compare/10167f20f9c7d8e8870f84755ab2e0bd1a902e2f..47ce6e3ec1f74f224a253a6c05706323d0dd3ba9)

nix/cardano-services/deployments.nix
blaggacao: Addressed (removed only) in [force-push](https://github.com/input-output-hk/cardano-js-sdk/compare/10167f20f9c7d8e8870f84755ab2e0bd1a902e2f..47ce6e3ec1f74f224a253a6c05706323d0dd3ba9)

nix/cardano-services/deployments/stake-pool-provider-deployment.yaml
blaggacao: Originally I was hoping that `ghcr.io` would have worked and we would have been able to publish a `latest` tag, here. But alas!

blaggacao: I've backlogged the original intent in [LW-8024](https://input-output.atlassian.net/browse/LW-8024)

blaggacao: This would be my [working suggestion](https://github.com/input-output-hk/cardano-js-sdk/pull/864/commits/25beff27b46b1b81ee01b0f1bc2a9fa9991cd8e8) for now.

blaggacao: Yeah, it's not good. The only other alternative I could come up with would be to set a fake name `UPDATE-ME` but we also don't need it to make much more complicated than it is and could defer to LW-8024, which is the desired proper solution either way.

Would that be ok? I reverted https://github.com/input-output-hk/cardano-js-sdk/pull/864/commits/25beff27b46b1b81ee01b0f1bc2a9fa9991cd8e8

gytis-ivaskevicius: Just delete the line? At least that is what I'd do

I am okay with some gibberish like `UPDATE-ME`  since that would also prevent it from failing silently

blaggacao: Implemented in https://github.com/input-output-hk/cardano-js-sdk/pull/864/commits/80e4eae962d7b5d7f01bfd29aff952a87cff8e76

.github/std-unexploded.yml
blaggacao: The most straight forward way was to work with the `aws` cli without complicated wrapping / manipulation:

```yaml
      - name: Configure K8S Cluster Access
        shell: bash
        run: |
          echo "Assuming role '$(aws sts get-caller-identity)' in cluster 'lace-prod-eu-central-1'."
          aws eks update-kubeconfig --name "lace-prod-eu-central-1"
```

I think this  sets a good precedent of simplicity, at least for this iteration.

It is also relevant in combination with #865 which requires us to gather more practical experience with how this integration ends up working and in which way we want to present and map environments.

nix/cardano-services/deployments.nix
blaggacao: Most certainly in this [force-push](https://github.com/input-output-hk/cardano-js-sdk/compare/80e4eae962d7b5d7f01bfd29aff952a87cff8e76..ad57dd41ec1b74dfea5df6acfe29630b9d49c24b)

nix/cardano-services/deployments/backend-deployment.yaml
gytis-ivaskevicius: Removed the lines which should not be here.  Instead, they should live together with overwrites since they are very mutable (just like OCI image)

rhyslbw: What's the reason for including them here @blaggacao ?

blaggacao: This is as green-to-green as it gets (and just designed to proof and establish the pipeline).

I captured this too for timely follow up in [LW-8029](https://input-output.atlassian.net/browse/LW-8029).

.github/std-unexploded.yml
blaggacao: I think this is can be resolved here?

Since this question will resurface as soon as we increase the coverage of environments one way or another, I'll abstain from a dedicated backlog, here.

nix/cardano-services/deployments.nix
blaggacao: What was the matter with blockfrost service? Did I still miss something?

gytis-ivaskevicius: yes, you missed a deployment

nix/cardano-services/deployments/backend-deployment.yaml
gytis-ivaskevicius: green-to-green does not mean that it should be in yaml, it just means that no additional changes should have been applied

nix/cardano-services/deployments/coingecko-proxy-deployment.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: apps/v1
```
gytis-ivaskevicius, line 1: As mentioned in my previous review this service should not exist here

nix/cardano-services/deployments/handle-provider-deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+            - name: POSTGRES_USER
+              valueFrom:
+                secretKeyRef:
+                  key: username
+                  name: handle-owner-user.dev-preview-dbsync-db.credentials.postgresql.acid.zalan.do
+            - name: SERVICE_NAMES
+              value: handle
+            - name: USE_KORA_LABS
+              value: "true"
+          image: FIXME-ADD-LATEST-IMAGE-REFERENCE
```
gytis-ivaskevicius, line 66: In slack you mentioned that this PR has already been tested - how was it tested since this should be crashing since there is no overwrite in `deployments.nix`

gytis-ivaskevicius (changes requested):
Please write down testing scenarios

nix/cardano-services/deployments/coingecko-proxy-deployment.yaml
rhyslbw: I agree with this. It's not a Cardano service, so best to remove and introduce it separately when the issue of it's home is determined. It's not being used by Lace yet so is not required to fulfil the needs of Lace developers.

rhyslbw (changes requested):
https://github.com/input-output-hk/cardano-js-sdk/pull/864/files/8f5804a7bff9639d0238ed98e7546e607025157e#r1298353403

nix/cardano-services/deployments/coingecko-proxy-deployment.yaml
rhyslbw: Is this a straight copy from wallet-world?

rhyslbw: Looks that way. I'm going to retract my change request in then interest of getting this initial setup complete. Any further refinements can be done separately.

blaggacao: > Is this a straight copy from wallet-world?

Yes it is. :+1:

gytis-ivaskevicius: yes it is, I think coingecko should be a separate deployment. it does not have much to do with `cardano-services` and it definitely does not belong in this repo

Also, this chart is  a little useless since it has image hardcoded thus no upgrade possible without manual intervention

blaggacao:
https://github.com/input-output-hk/cardano-js-sdk/pull/864#pullrequestreview-1580182086

The actionable parts have been satisfactorily addressed.

All requests for follow ups have been backlogged.

To close off the week with this.: merging on green CI.

All and any followups will be incoming early next week.

image

gytis-ivaskevicius:
I mentioned this bug somewhere in my reviews, and I asked you to test this before merging. We are not supposed to be doing whatever we feel like, please let's have big of quality assurance before pushing to master 
image

PR #867, opened 2023-08-15
Feat/lw 7732 implement  protol param updates serialization

AngelCastilloB:
# Context

# Proposed Solution

- Added protocol parameter update serialization classes.
- Added protocol parameter update  core types.


packages/core/src/Cardano/types/ProtocolParameters.ts
```diff
@@ -95,7 +101,19 @@ type NewProtocolParamsInBabbage = {
...
 type BabbageProtocolParameters = Omit<AlonzoProtocolParams, 'coinsPerUtxoWord' | 'extraEntropy'> &
   NewProtocolParamsInBabbage;
 
 export type ProtocolParameters = BabbageProtocolParameters;
+
+// Even tho extraEntropy was deprecated on babbage era, it is still present in the ProtocolParametersUpdate structure
+// since this structure is backward compatible with all eras.
+export type ProtocolParametersUpdate = Partial<ProtocolParameters & Pick<AlonzoProtocolParams, 'extraEntropy'>>;
+
+export type ProposedProtocolParameterUpdates = Map<Crypto.Hash28ByteBase16, ProtocolParametersUpdate>;
```
mkazlauskas, line 114: What is this hash of? :thinking:

packages/core/src/Serialization/Common/ExUnits.ts
```diff
@@ -0,0 +1,159 @@
...
+    const mem = this.#mem + other.#mem;
+    const steps = this.#steps + other.#steps;
+
+    return new ExUnits(mem, steps);
+  }
+
+  /**
+   * Used to create a dummy ExUnits that takes up the maximum size possible in cbor to provide an upper bound on tx size.
+   */
+  static dummy(): ExUnits {
```
mkazlauskas, line 156: Why do we need this?

packages/core/src/Cardano/types/ProtocolParameters.ts
mkazlauskas: There's only 1 genesis block though, so it would always be a map of size 1? If that's the case, then maybe it's better to model as `{hash, update}`?

packages/core/src/Serialization/Common/ExUnits.ts
mkazlauskas: Yeah in that case let's remove it for now, we can re-add it as test util or export from `util-dev` package if/when needed

mkazlauskas (dismissed):
Great work! :muscle:

mkazlauskas (approved):

mirceahasegan (approved):
Beautifully executed and great comments. Nice work @AngelCastilloB !

PR #861, opened 2023-08-11
fix(wallet): make PersonalWallet.submitTx idempotent

mkazlauskas:
# Context

Re-submitting transaction manually results in faulty state, as TransactionsTracker can't handle it.

LW-7873

# Proposed Solution

babe6a5cbb834adafe90fa23714cc74822601eca

PR #869, opened 2023-08-16
fix(wallet): do not prevent re-submission

mkazlauskas:
# Context

#861 made submitTx idempotent, but also broke internal logic of PersonalWallet preventing it to re-submit transactions

# Proposed Solution

ef64b30c7d02b64746344ea3042ae7b96db888f6

### DIFF
--- a/packages/wallet/test/PersonalWallet/methods.test.ts
+++ b/packages/wallet/test/PersonalWallet/methods.test.ts
@@ -442,14 +442,13 @@ describe('PersonalWallet methods', () => {
       });
 
       it('mightBeAlreadySubmitted option interprets ValueNotConservedError as success', async () => {
-        txSubmitProvider.submitTx.mockRejectedValueOnce(
-          new ProviderError(
-            ProviderFailure.BadRequest,
-            new CardanoNodeErrors.TxSubmissionErrors.ValueNotConservedError({
-              valueNotConserved: { consumed: 2, produced: 1 }
-            })
-          )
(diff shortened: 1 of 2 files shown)


mirceahasegan (dismissed):
Nice catch! 🕵️‍♂️

mirceahasegan (approved):

packages/wallet/test/PersonalWallet/methods.test.ts
```diff
@@ -462,6 +464,20 @@ describe('PersonalWallet methods', () => {
         expect(await txPending).toEqual(outgoingTx);
       });
 
+      it('attempts to resubmit the tx that is already in flight', async () => {
```
rhyslbw, line 467:
```suggestion
      it('only tracks the first submission as in-flight when there are multiple submission of the same tx', async () => {
```
Describe the expectation

rhyslbw (changes requested):

packages/wallet/test/PersonalWallet/methods.test.ts
mkazlauskas: Main assertion in this test is `expect(txSubmitProvider.submitTx).toBeCalledTimes(2);` though :thinking:

rhyslbw: Right I see. I'll retract my suggestion

rhyslbw (approved):

PR #870, opened 2023-08-16
ci: call std as release process only for specific release prefix

blaggacao:
# Context

The current setup triggers multiple STD pipelines, one for _each_ release tag.
But we only need the pipeline to run once for `cardano-sdk/cardano-services`.

# Proposed Solution

- Make it a call workflow
- Call it form a release workflow
  - Which is in turn able to filter on a broader spectrum of the available event data.

# Testing
- create a new release after merge (?)

ref: LW-8058

### DIFF
--- a/.github/std-unexploded.yml
+++ b/.github/std-unexploded.yml
@@ -3,6 +3,7 @@ name: STD
 
 on:
   workflow_dispatch:
+  workflow_call:
 
   pull_request:
     branches:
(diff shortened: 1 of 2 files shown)


rhyslbw (approved):
Nice!

blaggacao:
I can't really tell if it works until the next release.

However, in order to not generate undue and confusing diffs with concurrent improvements to the CD, I'm merging this now untested.

blaggacao:
Shoot I got distracted and missed a force push didn't succeed. Fixup: https://github.com/input-output-hk/cardano-js-sdk/pull/879

PR #871, opened 2023-08-17
fix(wallet): address key statuses are now updated properly when changes are detected

AngelCastilloB:
# Context

The reward accounts key statuses were not being updated consistently in the browser environment. The transaction history was being updated, and the parsing logic was detecting correctly the new key status, however, these status were not being emitted.

# Proposed Solution

addressKeyStatuses observable now uses deepEquals instead of shallowEquals


mirceahasegan (commented):
Great work @AngelCastilloB ! 🚀

mirceahasegan (approved):

PR #872, opened 2023-08-17
Feat: OpenAPI version in URL path

rhyslbw:
# Context
API-breaking changes will be a challenge to manage without introducing a version scheme in the URL. Since we're building the client in the same scope as the server, it's possible to abstract the versioning concerns within.

# Proposed Solution
- Include the OpenAPI version within the URL scheme 
- Embed the expected version within the client HTTP requests by reading the version from a file generated at build time
- The tests could use refactoring, as clearly there's some repetition evidenced by the number of changes this PR contains, but I've left that out of scope.

While completing this task I found the service level OpenAPI docs were missing the health check paths. I've fixed this in a preliminary commit.


iccicci (changes requested):
Great work!

Just some minor suggestions to remove some more code repetitions

packages/cardano-services/src/Http/schema.json
```diff
@@ -0,0 +1,64 @@
+{
```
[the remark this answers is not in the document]
iccicci: Thinking about this point a bit more: this can be helpful only if we plan to support multiple API versions with one `provider-server`; if, on the other hand, our target is to support only a single API version with one `provider-server` and handling versions at routing level, this proposal looses its value.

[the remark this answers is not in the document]
rhyslbw: Thanks for the prompt, and yes you're right, for now we'll just be supporting one version.

packages/cardano-services/src/Http/paths.json
```diff
@@ -0,0 +1,36 @@
+{
+  "Health": {
+    "post": {
```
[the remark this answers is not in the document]
iccicci: I had to revert this.

Even if this is accepted at api spec compile time, it seems it is not working correctly when validating responses...
Simply reverting this change fixed lots of tests.

packages/cardano-services/src/Http/HttpServer.ts
```diff
@@ -103,11 +109,11 @@ export class HttpServer extends RunnableModule {
       };
 
     this.app.use(
-      '/health',
+      path.join(versionPath, 'health'),
```
mkazlauskas, line 112: Maybe instead of duplicating `path.join(versionPath, ...)` everywhere we can do sth like `this.app = express().use(versionPath)` and then mount (use) all subpaths on that.

packages/cardano-services-client/scripts/createVersionSource.js
```diff
@@ -1,10 +1,27 @@
 const fs = require('fs');
 const path = require('path');
 
-const openApiSpec = require('../../cardano-services/src/Http/openApi.json');
+const cardanoServicesSrc = path.join('..', '..', 'cardano-services', 'src');
+
+const apiVersion = Object.fromEntries(
+  Object.entries({
+    assetInfo: 'Asset',
+    chainHistory: 'ChainHistory',
```
mkazlauskas, line 9: I'm wondering if we could walk through fs and find all `(.+)HttpService` alongside openApi.json in the same dir. Then we wouldn't need to maintain this file separately and rely on convention.

mkazlauskas (changes requested):
Looking good, just some suggestions

packages/cardano-services/src/Http/HttpService.ts
```diff
@@ -8,20 +8,25 @@ import {
...
 import express, { Router } from 'express';
 
 export abstract class HttpService extends RunnableModule {
   public router: express.Router;
   public slug: string;
   public provider: Provider;
+  public openApiPath: string;
 
-  constructor(slug: string, provider: Provider, router: express.Router, logger: Logger) {
+  constructor(slug: string, provider: Provider, router: express.Router, openApiPath: string, logger: Logger) {
```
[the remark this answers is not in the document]
mkazlauskas: Maybe this strategy would help: https://github.com/input-output-hk/cardano-js-sdk/pull/872#discussion_r1302993206

packages/cardano-services/test/util/openApi.test.ts
```diff
@@ -0,0 +1,15 @@
...
+import path from 'path';
+
+describe('openApi utils', () => {
+  describe('versionPathFromSpec', () => {
+    it('parses the OpenAPI version and returns as a path, prefixed with v', () => {
+      const apiSpecPath = path.join(__dirname, '..', '..', 'src', 'Http', 'openApi.json');
+      const {
+        info: { version }
+      } = require(apiSpecPath) as OpenAPIV3.Document;
+      expect(versionPathFromSpec(apiSpecPath)).toBe(`/v${version}`);
```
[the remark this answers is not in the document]
iccicci: ea396744a should solve this

packages/cardano-services-client/scripts/createVersionSource.js
mkazlauskas: Right we can't get rid of the file, the idea was to not have to update this file when we add/remove/rename providers. Anyway it's good as-is.

greatertomi (approved):

VanessaPC (approved):
It looks great to me, look forward to using it!

.github/workflows/continuous-integration-e2e.yaml
```diff
@@ -83,10 +83,10 @@ jobs:
         yarn workspace @cardano-sdk/e2e test:projection
         yarn workspace @cardano-sdk/e2e test:pg-boss
       env:
-        STAKE_POOL_PROVIDER_URL: 'http://localhost:4000/stake-pool'
+        STAKE_POOL_PROVIDER_URL: 'http://localhost:4000/'
 
     - name: 🔬 Test - e2e - wallet - typeorm stake pool provider
       run: |
-        yarn workspace @cardano-sdk/e2e test:providers
+        yarn workspace @cardano-sdk/e2e test:providers -t HandleProvider
```
mkazlauskas, line 90:
```suggestion
        yarn workspace @cardano-sdk/e2e test:providers
```

packages/e2e/.env.example
```diff
@@ -11,40 +11,38 @@ KEY_MANAGEMENT_PARAMS='{"bip32Ed25519": "CML", "accountIndex": 0, "chainId":{"ne
 
 # Providers setup - required by getWallet
 ASSET_PROVIDER=http
-ASSET_PROVIDER_PARAMS='{"baseUrl":"http://localhost:4000/asset"}'
+ASSET_PROVIDER_PARAMS='{"baseUrl":"http://localhost:4000/v1.0.0/asset"}'
```
mkazlauskas, line 14: Shouldn't this match what we have in [GH workflow definition](https://github.com/input-output-hk/cardano-js-sdk/pull/872/files#diff-33e5806dbe086222ebce154c323030b5a434ee370b7de1e0bf1c3d04584f28f7R10)?
```suggestion
ASSET_PROVIDER_PARAMS='{"baseUrl":"http://localhost:4000/"}'
```

mkazlauskas (changes requested):

.github/workflows/continuous-integration-e2e.yaml
iccicci: Currently `HandleProvider` is the only test under `test:providers`.

Since here we want to test the _typeorm stake pool provider_ we neet to run only this specific test with this specific `STAKE_POOL_PROVIDER_URL` setting.

If in the future we will add some other tests under `test:providers`, they will be executed about 5 lines before in the same file and they will be not required here.

packages/e2e/.env.example
iccicci: Yes, it should. Nice catch. Let me check better this.

iccicci: I confirm I changed only my `.env` file and I forgot to update `.env.example` accordingly.

083994d31 should solve this.

mkazlauskas: Looks good, thanks. Maybe we should create a follow-up task to do sth like
```
PROVIDER=http
PROVIDER_PARAMS={...}
```
to further simplify e2e .env. WDYT?

.github/workflows/continuous-integration-e2e.yaml
iccicci: 3af295122 should solve this

mkazlauskas (dismissed):
Great work @rhyslbw & @iccicci :muscle:

packages/e2e/.env.example
iccicci: Created LW-8181

iccicci (approved):

gytis-ivaskevicius:
Pushed deployment fixes, now it should be good and ready for merge. I want to test this out before doing that, @blaggacao could you help us with that?

blaggacao:
>  could you help us with that [assert the changed to the deployment manifests]?

The suggestive next steps are:
1. wait for an update to [the diff](https://github.com/input-output-hk/cardano-js-sdk/pull/872#issuecomment-1695493769) from CI
2. carefully audit the diff: in a CD scenario, that's what a final review on this PR signs off on, as well
3. merge and wait for deployment (side note: I've reasons to believe that this step turned out not working as intended, I'll need to revise this. I have this backlogged as LW-8219)

... iterate if the changes turned out not working as intended.

For an even shorter feedback loop to polish the deployment manifests, I can think of three options:
- `arion` / docker-compose playground
- kind playground
- some of the existing operational playgrounds

We may need to find and discuss our preferred option, still.

gytis-ivaskevicius:
I can not consider this 'testing', also alternative options arent satisfactory since we are dealing with k8s and AWS integration testing

PR #874, opened 2023-08-18
fix(wallet): remove tx from inFlight$ when loading by the time it's already on-chain

mkazlauskas:
# Context

Wallet can end up in a state where transaction is present in both:
- history store
- in-flight store

When that happens, wallet gets in a faulty state,
because TransactionsTracker looking only for new transactions
when waiting for a transaction to be on-chain.

# Proposed Solution

Check the entire transaction history when looking for on-chain confirmation


packages/wallet/src/services/TransactionsTracker.ts
```diff
@@ -250,7 +250,13 @@ export const createTransactionsTracker = (
   );
 
   const txOnChain$ = (evt: OutgoingTx): Observable<OutgoingOnChainTx> =>
-    onChainNewTxSuccess$.pipe(
+    merge(
+      historicalTransactions$.pipe(
+        take(1),
+        mergeMap((txs) => from(txs))
+      ),
+      onChainNewTxSuccess$
```
mirceahasegan, line 258: I think it's safe to remove onChainNewTxSuccess$. It is included in historicalTransactions$

mkazlauskas: It has a `take(1)` from `historicalTransactions$`, so that we don't have to check it against every historical transaction twice.

mirceahasegan: Missed that. good to go 😄

mirceahasegan (approved):
Nicely done @mkazlauskas ! 🚀

packages/wallet/test/PersonalWallet/methods.test.ts
```diff
@@ -464,6 +464,14 @@ describe('PersonalWallet methods', () => {
         expect(await txPending).toEqual(outgoingTx);
       });
 
+      it('does not re-serialize the transaction to compute transaction id', async () => {
```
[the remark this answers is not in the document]
rhyslbw: Leave it for now, but in future you could say as a comment "This transaction produces a different ID when round-tripping serialisation"

packages/wallet/src/PersonalWallet/PersonalWallet.ts
```diff
@@ -158,12 +158,15 @@ const isSignedTx = (input: Cardano.Tx | TxCBOR | OutgoingTx | SignedTx): input i
...
-      cbor: input,
-      // Do not re-serialize transaction body to compute transaction id
-      id: Cardano.TransactionId.fromTxBodyCbor(TxBodyCBOR.fromTxCBOR(input))
-    };
+    return usingAutoFree((scope) => {
+      const tx = scope.manage(Serialization.Transaction.fromCbor(input));
+      return {
+        body: tx.toCore().body,
+        cbor: input,
+        // Do not re-serialize transaction body to compute transaction id
```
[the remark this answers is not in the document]
rhyslbw: Looks that way 😅

rhyslbw (dismissed):
Great work @mkazlauskas 🚀

packages/core/src/Serialization/Transaction.ts
```diff
@@ -98,8 +99,8 @@ export class Transaction {
    * @param cbor The CBOR encoded transaction object.
    * @returns The new transaction instance.
    */
-  static fromCbor(cbor: HexBlob): Transaction {
-    const reader = new CborReader(cbor);
+  static fromCbor(cbor: TxCBOR): Transaction {
```
mirceahasegan, line 102: Isn't this a breaking change?

mirceahasegan (dismissed):

packages/wallet/test/PersonalWallet/methods.test.ts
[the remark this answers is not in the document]
mkazlauskas: Added the comment since I was updating the commits anyway

packages/core/src/Serialization/Transaction.ts
mkazlauskas: It is not at runtime, but could break builds :sweat_smile: nice catch, reworded the commit

mirceahasegan (approved):

rhyslbw (approved):

PR #883, opened 2023-08-22
Fix/greedy input selection only with active stake keys

mirceahasegan:
# Context

Currently, the greedy input selector is used whenever the user has more than 1 reward account and tries to delegate, even if delegating to a single stake pool.
Scenario:
1. User enables multi-delegation and delegates to 2 pools. 2 stake keys will be created.
2. User disables multi-delegation and delegates to a single pool.
3. User wants to change delegation to another pool. Unexpectedly, inputs would be selected by the Greedy algorithm instead of the round-robin one. This should not occur because the user has only one active stake key.

# Proposed Solution
Use greedy input selection only when:
1. Multi delegating: distribute the funds to multiple addresses.
5. Reducing the number of pools to 1, which implies we have multiple registered stake keys. This is needed to concentrate the funds back into a single address and delegate its stake key.


mkazlauskas (approved):
Nice catch! :detective:

blaggacao:
Hey Folks! In case you haven't noticed, this PR actually _does_ [break](https://github.com/input-output-hk/cardano-js-sdk/actions/runs/5941501375/job/16112517860) CI/CD. Let's scramble to see what we can do.

PR #885, opened 2023-08-23
Fix/ledger import

mchappell:
# Context

Using a hardware wallet with CIP-30 presents issues when signing transactions or data due to invoking a new key agent 

# Proposed Solution

Where possible, return an existing connection, only create a new one if none requested match the existing connections

### DIFF
--- a/packages/hardware-ledger/src/LedgerKeyAgent.ts
+++ b/packages/hardware-ledger/src/LedgerKeyAgent.ts
@@ -16,7 +16,8 @@ import { LedgerTransportType } from './types';
 import { ManagedFreeableScope } from '@cardano-sdk/util';
 import { str_to_path } from '@cardano-foundation/ledgerjs-hw-app-cardano/dist/utils/address';
 import { toLedgerTx } from './transformers';
-import LedgerConnection, {
+import TransportNodeHid from '@ledgerhq/hw-transport-node-hid-noevents';
+import _LedgerConnection, {
   Certificate,
   CertificateType,
   GetVersionResponse,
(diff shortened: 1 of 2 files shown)


packages/wallet/test/hardware/ledger/LedgerKeyAgent.test.ts
```diff
@@ -219,6 +219,53 @@ describe('LedgerKeyAgent', () => {
...
+    });
+
+    it('should return an existing connection if one exists', async () => {
+      LedgerKeyAgent.deviceConnections = [];
+      const connection1 = await LedgerKeyAgent.establishDeviceConnection(CommunicationType.Node);
+      const connection2 = await LedgerKeyAgent.establishDeviceConnection(CommunicationType.Node);
+
+      expect(transportSpy).toHaveBeenCalledTimes(1);
+
+      await connection1.transport.close();
```
mkazlauskas, line 243: It should be the same object so we don't need to close both, would be better to have an assertion like this instead
```suggestion
      expect(connection1).toBe(connection2);
```

mkazlauskas (commented):
Looks good! Just 1 suggestion for test

packages/hardware-ledger/src/LedgerKeyAgent.ts
```diff
@@ -120,16 +120,33 @@ const isMultiSig = (tx: Transaction): boolean => {
...
   constructor({ deviceConnection, ...serializableData }: LedgerKeyAgentProps, dependencies: KeyAgentDependencies) {
     super({ ...serializableData, __typename: KeyAgentType.Ledger }, dependencies);
     this.deviceConnection = deviceConnection;
     this.#communicationType = serializableData.communicationType;
   }
 
+  static findKeyAgentByCommunicationTypeAndDevicePath(
+    communicationType: CommunicationType,
+    devicePath?: string,
+    deviceConnections?: LedgerConnectionWithCommunicationTypeAndDevicePath[]
```
mirceahasegan, line 143: I do not understand the need to pass this as a param. Shouldn't it use the static one availble? `LedgerKeyAgent.deviceConnections`

mchappell: Good point, i'll update. Thanks for catching this

mchappell: Updated in [99891ec](https://github.com/input-output-hk/cardano-js-sdk/pull/885/commits/99891ec44752b799256cfd0a78d910b3f0589307)

mirceahasegan (approved):
Great work @mchappell ! 🚀

mkazlauskas (dismissed):
Nice work! :rocket:

flake.lock
```diff
@@ -311,11 +311,11 @@
         "yants": "yants"
       },
       "locked": {
-        "lastModified": 1692005050,
-        "narHash": "sha256-bFhmzmenx3Y/0eH1qAAZGrmFc3siz0BiAJRptdt2QIQ=",
+        "lastModified": 1692627026,
```
rhyslbw, line 314: Why are there changes to the `flake.lock` if there's no package changes in the scope of this commit?

mkazlauskas (changes requested):
Please fix the format the 2nd commit message (missing colon `:` after scope)

mkazlauskas (approved):

PR #887, opened 2023-08-23
feat: allow extensions for CIP-30 API

lgobbi-atix:
# Context

As a DApp developer, I want to enable the lace DApp connector with an array of extensions, so I can make use of the extensible DApp connector within my DApp.

# Proposed Solution
- Add CIP-95 as a supported extension
- Add `getActivePubStakeKeys` and `getPubDRepKey` to the `WalletApi` interface (without implementation)
- Add a parameter to the `Cip30Wallet.enable` method to be able to pass an array of extensions according to the CIP-30 specification
- Wrap the `Cip30Wallet.api` as a JS object with the allowed methods depending on the extensions array to make `enable` does not return a `Proxy`

### DIFF
--- a/packages/dapp-connector/src/WalletApi/Cip30Wallet.ts
+++ b/packages/dapp-connector/src/WalletApi/Cip30Wallet.ts
@@ -1,7 +1,63 @@
 import { APIErrorCode, ApiError } from '../errors';
+import { Bytes, Cbor, Paginate, WalletApi, WalletApiExtension, WalletMethod } from './types';
+import { Cardano } from '@cardano-sdk/core';
 import { Logger } from 'ts-log';
 import { RemoteAuthenticator } from '../AuthenticatorApi';
-import { WalletApi } from './types';
+import uniq from 'lodash/uniq';
+
+export const CipMethodsMapping: Record<number, WalletMethod[]> = {
+  30: [
+    'getNetworkId',
+    'getUtxos',
+    'getCollateral',
+    'getBalance',
+    'getUsedAddresses',
+    'getUnusedAddresses',
+    'getChangeAddress',
+    'getRewardAddresses',
+    'signTx',
+    'signData',
+    'submitTx'
+  ],
+  95: ['getActivePubStakeKeys', 'getPubDRepKey']
+};
+export const WalletApiMethodNames: WalletMethod[] = Object.values(CipMethodsMapping).flat();
+
+/**
+ * Wrap the proxy API object with a regular javascript object to avoid interop issues with some dApps.
+ *
+ * Only return the allowed API methods.
+ */
+const wrapAndEnableApi = (walletApi: WalletApi, allowedApiMethods: WalletMethod[]): WalletApi => {
+  const objectApi: WalletApi = {
+    getActivePubStakeKeys: () => walletApi.getActivePubStakeKeys(),
+    getBalance: () => walletApi.getBalance(),
+    getChangeAddress: () => walletApi.getChangeAddress(),
+    getCollateral: (params?: { amount?: Cbor }) => walletApi.getCollateral(params),
+    getNetworkId: () => walletApi.getNetworkId(),
+    getPubDRepKey: () => walletApi.getPubDRepKey(),
+    getRewardAddresses: () => walletApi.getRewardAddresses(),
+    getUnusedAddresses: () => walletApi.getUnusedAddresses(),
+    getUsedAddresses: (paginate?: Paginate) => walletApi.getUsedAddresses(paginate),
+    getUtxos: (amount?: Cbor, paginate?: Paginate) => walletApi.getUtxos(amount, paginate),
+    signData: (addr: Cardano.PaymentAddress | Bytes, payload: Bytes) => walletApi.signData(addr, payload),
+    signTx: (tx: Cbor, partialSign?: Boolean) => walletApi.signTx(tx, partialSign),
+    submitTx: (tx: Cbor) => walletApi.submitTx(tx)
+  };
+
+  const enabledApi = Object.fromEntries(
+    Object.entries(objectApi).filter(([methodName, _]) => allowedApiMethods.includes(methodName as WalletMethod))
+  );
+
+  // Add experimental.getCollateral to CIP-30 API
+  if (allowedApiMethods.includes('getCollateral')) {
+    enabledApi.experimental = {
+      getCollateral: (params?: { amount?: Cbor }) => walletApi.getCollateral(params)
+    };
+  }
+
+  return enabledApi as WalletApi;
+};
 
 /**
  * CIP30 API version
@@ -28,6 +84,8 @@ export type WalletDependencies = {
   api: WalletApi;
 };
 
+export type Cip30EnableOptions = { extensions: WalletApiExtension[] };
+
 /**
  * CIP-30 wallet that is injected to page
  */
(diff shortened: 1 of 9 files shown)


packages/dapp-connector/src/WalletApi/Cip30Wallet.ts
```diff
@@ -1,7 +1,71 @@
...
+    getUnusedAddresses: () => walletApi.getUnusedAddresses(),
+    getUsedAddresses: (paginate?: Paginate) => walletApi.getUsedAddresses(paginate),
+    getUtxos: (amount?: Cbor, paginate?: Paginate) => walletApi.getUtxos(amount, paginate),
+    signData: (addr: Cardano.PaymentAddress | Bytes, payload: Bytes) => walletApi.signData(addr, payload),
+    signTx: (tx: Cbor, partialSign?: Boolean) => walletApi.signTx(tx, partialSign),
+    submitTx: (tx: Cbor) => walletApi.submitTx(tx)
+  };
+
+  const enabledApi = allowedApiMethods.reduce(
+    (api, methodName) => ({ ...api, [methodName]: objectApi[methodName] }),
```
mchappell, line 55: Please convert this to something more like
```
const enableApi = Object.fromEntries(
  Object.entries(objectApi)
  .filter((methodName, _) => allowedMethods.includes(methodName));
```
to remove the reduce-spread pattern

mkazlauskas (changes requested):

mchappell (dismissed):

mkazlauskas (approved):
Great work! :rocket:

packages/dapp-connector/src/WalletApi/Cip30Wallet.ts
```diff
@@ -64,7 +57,7 @@ const wrapAndEnableApi = (walletApi: WalletApi, allowedApiMethods: WalletMethod[
     };
```
mirceahasegan, line 57: i don't think `enabledApi as any` cast is needed anymore:
```diff
-     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-     (enabledApi as any).experimental = {
+     enabledApi.experimental = {

File: packages/dapp-connector/test/WalletApi/Cip30Wallet.test.ts [hunk 3]
```diff
@@ -25,20 +32,50 @@ describe('Wallet', () => {
...
+      expect(await wallet.isEnabled()).toBe(false);
+      const api = await wallet.enable();
+      expect(typeof api).toBe('object');
+      const methods = new Set(Object.keys(api));
+      expect(methods).toEqual(new Set([...CipMethodsMapping[30], 'experimental']));
+      expect(await wallet.isEnabled()).toBe(true);
+    });
+
+    test('with cip95 extension', async () => {
+      expect(await wallet.isEnabled()).toBe(false);
```
mirceahasegan, line 54: The `not enabled` precondition is not needed in every test.
```diff
-       expect(await wallet.isEnabled()).toBe(false);
```

mirceahasegan (changes requested):

packages/dapp-connector/test/WalletApi/Cip30Wallet.test.ts
```diff
@@ -25,20 +32,50 @@ describe('Wallet', () => {
...
   });
 
-  test('enable', async () => {
-    expect(await wallet.isEnabled()).toBe(false);
-    const api = await wallet.enable();
-    expect(typeof api).toBe('object');
-    const methods = new Set(Object.keys(api));
-    expect(methods).toEqual(new Set(WalletApiMethodNames));
-    expect(await wallet.isEnabled()).toBe(true);
+  describe('enable', () => {
```
mirceahasegan, line 43: Please add a test for the `throws` scenario

mirceahasegan (changes requested):

packages/dapp-connector/test/WalletApi/Cip30Wallet.test.ts
```diff
@@ -25,20 +33,50 @@ describe('Wallet', () => {
...
```
mirceahasegan, line 43: Users might `enable()`, then `enable({extensions: [{cip: 95}]})`.
I think a test to see if the second call to enable, with the cip95 extension, will return the cip95 methods would be good to have.
```diff
+ await wallet.enable();
+ const api = await wallet.enable({ extensions: [{ cip: 95 }] });
```

lgobbi-atix: Done here with the rest of the requested changes: https://github.com/input-output-hk/cardano-js-sdk/pull/887/commits/d17eb7ba288aecd3015a61d1eb9824c001de94d9

packages/dapp-connector/src/WalletApi/Cip30Wallet.ts
```diff
@@ -50,6 +109,16 @@ export class Cip30Wallet {
     this.#authenticator = authenticator;
   }
 
+  /**
+   * Receives the array of extensions provided when `wallet.enable` was called and
+   * returns a list of methods names for the CIP-30 API and supported extensions.
+   */
+  #getAllowedApiMethods(extensions: WalletApiExtension[] = []): WalletMethod[] {
+    const enabledExtensions: WalletApiExtension[] =
```
mirceahasegan, line 117: we could have a try/catch block here to account for invalid extensions param format + a test for it.
```diff
+ try {
    const enabledExtensions: WalletApiExtension[] =
      extensions.filter((extension) => this.supportedExtensions.some(({ cip }) => cip === extension.cip)) ?? [];
+ } catch(error) {
+     throw new ApiError(APIErrorCode.InvalidRequest, `Couldn't parse extensions ${extensions}.`);
+ }
```

lgobbi-atix: Added here c1977cc153c7fed6618328c611c363f02c8053f4, but this still enables the API as it fails after a successful `requestAccess`. Is this ok? Or should we validate the extensions array format before calling `requestAccess`?

mirceahasegan: Nice catch 🕵️ ! I think it would be best to fail, so validate it before calling `requestAccess`.

mkazlauskas: Imho we should throw if extensions don't match the type. If we do this, then devs are going to try it without reading the cip, find that it works, then use it in an invalid way. Other wallets won't thank us for it :sweat_smile:

lgobbi-atix: Then I think this should do it 5a764cd6110f7571bc3b0a96ae7a0e997e585ad7

mirceahasegan (approved):
Fantastic work @lgobbi-atix 💪

mkazlauskas (approved):

packages/dapp-connector/src/WalletApi/Cip30Wallet.ts
```diff
@@ -1,7 +1,63 @@
...
+    'signTx',
+    'signData',
+    'submitTx'
+  ],
+  95: ['getActivePubStakeKeys', 'getPubDRepKey']
+};
+export const WalletApiMethodNames: WalletMethod[] = Object.values(CipMethodsMapping).flat();
+
+/**
+ * Wrap the proxy API object with a regular javascript object to avoid interop issues with some dApps.
```
rhyslbw, line 27: It would have been better to commit this as a separate change first as it's considered a fix in some contexts.

rhyslbw (approved):
Great work @lgobbi-atix 🚀

PR #888, opened 2023-08-24
fix: use chromedriver that matches the environment's chrome version

blaggacao:
# Context

PR #883 updated the pin of `chromedriver` dependency.

However, `nix/cardano-services/packages.nix` had been shipping with it's own pinned version as dynamic/lazy download of chromedriver binary is not possible in the nix sandbox. Since that version didn't match, anymore, download was attempted and failing in the sandbox, which failed CI/CD.

Furthermore, during local development, `chromedriver` should match the currently installed version of chrome.

# Proposed Solution

This PR removes the dependency and instructs the user to provide a global chromedriver version that maps their currently installed version of chrome.

Now, `DETECT_CHROMEDRIVER_VERSION=true yarn global add chromedriver` is part of the manual build setup.


README.md
```diff
@@ -67,6 +67,7 @@ git clone \
 #### Install and Build
 
 ```bash
+DETECT_CHROMEDRIVER_VERSION=true yarn global add chromedriver
```
mkazlauskas, line 70: This should go after `nvm use` or else you might install it to a different nodejs. Also a comment would be nice that version of global chromedriver installation has to be maintained to match the installed Google Chrome version

mkazlauskas (changes requested):

michalrus (commented):
Hmmm, adding another manual build step is a worse developer experience, isn’t it?

When building `cardano-js-sdk` for Windows, I do this to get the `chromedriver` version programmtically::

https://github.com/input-output-hk/lace/blob/274d316b263781c6afe612c8b983389a0daa7679/nix/lace-blockchain-services/internal/x86_64-windows.nix#L583-L586

blaggacao:
> Hmmm, adding another manual build step is a worse developer experience, isn’t it?

Well, yes, in principle. But as far as I had understood it in a quick chat from our colleagues, the version should _not_ be pinned but needs to match the Chrome that the user has installed, globally.

> https://github.com/input-output-hk/lace/blob/274d316b263781c6afe612c8b983389a0daa7679/nix/lace-blockchain-services/internal/x86_64-windows.nix#L583-L586

This is pretty nice, regardless.

@mkazlauskas does this change our (your) assessment?

michalrus:
> Follow up for @michalrus :
> 
> * Do you trace and update the hash manually on version updates?

Yes, but when the version changes in this repo, Nix re-downloads the archive, and then we get an error with a mismatched hash, and the whole build fails. So it’s hard to miss =)

mkazlauskas (approved):

mirceahasegan (approved):

michalrus (approved):

PR #893, opened 2023-08-29
tx builder minting

iadmytro:
# Context

Tx builder extended with minting functionality
LW-9394

### DIFF
--- a/packages/e2e/src/factories.ts
+++ b/packages/e2e/src/factories.ts
@@ -322,7 +322,8 @@ const patchInitializeTxToRespectEpochBoundary = <T extends ObservableWallet>(
  * @returns an object containing the wallet and providers passed to it
  */
 export const getWallet = async (props: GetWalletProps) => {
-  const { env, idx, logger, name, polling, stores, customKeyParams, keyAgent, witnesser } = props;
+  const { env, idx, logger, name, polling, stores, customKeyParams, witnesser } = props;
+  const keyAgent = props.keyAgent;
   const providers = {
     addressDiscovery: await addressDiscoveryFactory.create(
       env.ADDRESS_DISCOVERY,
@@ -364,6 +365,7 @@ export const getWallet = async (props: GetWalletProps) => {
   const keyManagementParams = { ...envKeyParams, ...(idx === undefined ? {} : { accountIndex: idx }) };
 
   const bip32Ed25519 = await bip32Ed25519Factory.create(env.KEY_MANAGEMENT_PARAMS.bip32Ed25519, null, logger);
+
   const asyncKeyAgent =
     keyAgent ||
     (await (
@@ -391,7 +393,12 @@ export const getWallet = async (props: GetWalletProps) => {
     polling?.maxInterval ||
     (polling?.interval && polling.interval * DEFAULT_POLLING_CONFIG.maxIntervalMultiplier) ||
     DEFAULT_POLLING_CONFIG.maxInterval;
-  return { bip32Account, providers, wallet: patchInitializeTxToRespectEpochBoundary(wallet, maxInterval) };
+  return {
+    asyncKeyAgent,
+    bip32Account,
+    providers,
+    wallet: patchInitializeTxToRespectEpochBoundary(wallet, maxInterval)
+  };
 };
 
 /**
--- a/packages/e2e/src/factories.ts
+++ b/packages/e2e/src/factories.ts
@@ -322,8 +322,7 @@ const patchInitializeTxToRespectEpochBoundary = <T extends ObservableWallet>(
  * @returns an object containing the wallet and providers passed to it
  */
 export const getWallet = async (props: GetWalletProps) => {
-  const { env, idx, logger, name, polling, stores, customKeyParams, witnesser } = props;
-  const keyAgent = props.keyAgent;
+  const { env, idx, logger, name, polling, stores, customKeyParams, keyAgent, witnesser } = props;
   const providers = {
     addressDiscovery: await addressDiscoveryFactory.create(
       env.ADDRESS_DISCOVERY,
--- a/packages/e2e/test/wallet/PersonalWallet/nft.test.ts
+++ b/packages/e2e/test/wallet/PersonalWallet/nft.test.ts
@@ -1,12 +1,8 @@
 /* eslint-disable sonarjs/no-duplicate-string */
-import { Asset, Cardano, metadatum, nativeScriptPolicyId } from '@cardano-sdk/core';
-import { Assets, BaseWallet, FinalizeTxProps } from '@cardano-sdk/wallet';
-import { InitializeTxProps } from '@cardano-sdk/tx-construction';
-import { KeyRole, TransactionSigner, util } from '@cardano-sdk/key-management';
+import { Asset, Cardano, metadatum } from '@cardano-sdk/core';
+import { Assets, BaseWallet } from '@cardano-sdk/wallet';
 import {
-  bip32Ed25519Factory,
   burnTokens,
-  createStandaloneKeyAgent,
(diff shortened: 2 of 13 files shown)


packages/tx-construction/src/tx-builder/PolicyBuilder.ts
```diff
@@ -0,0 +1,63 @@
...
+    return this.#derivationPath;
+  }
+
+  setPolicyScript(script: Cardano.NativeScript) {
+    this.#policyScript = script;
+    return this;
+  }
+
+  async getPolicyScript() {
+    if (!this.#policyScript) {
```
mirceahasegan, line 35: If `derivationPath` is updated, the policy script should also be recalculated as it depends on the `keyHash`.
It would be great to have these cases covered in unit tests.
```
- if (!this.#policyScript) {
+ if (!this.#policyScript || this.#derivationPathUpdated) {

Comment (iadmytro) [reply packages/tx-construction/src/tx-builder/PolicyBuilder.ts]: updated in [c7e44c7](https://github.com/input-output-hk/cardano-js-sdk/pull/893/commits/c7e44c75007d6bbb5447b0889030b0005219a75c)

File: packages/tx-construction/src/tx-builder/types.ts [hunk 2]
```diff
@@ -223,6 +223,44 @@ export interface TxBuilder {
...
+   *  ->          .assets(tokensOutput)
+   *              .build())
+   *        .build()
+   *        .sign();
+   *      await wallet.submitTx(tx);
+   */
+  addMint(mintToken: Cardano.TokenMap): TxBuilder;
+
+  /** @param script */
+  addNativeScript(script: Cardano.NativeScript): TxBuilder;
```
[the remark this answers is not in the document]
iadmytro: removed in [ffbc789](https://github.com/input-output-hk/cardano-js-sdk/pull/893/commits/ffbc7896c101d6dcf61758e8d680974796a4fa6b)

packages/tx-construction/src/tx-builder/TokenBuilder.ts
```diff
@@ -0,0 +1,26 @@
...
+      .map((s) => s.charCodeAt(0).toString(16))
+      .join('');
+
+  addAsset(tokenName: string, quantity: bigint): TxTokenBuilder {
+    this.#assets.set(Cardano.AssetId(`${this.#policyId}${this.#toHex(tokenName)}`), quantity);
+    return this;
+  }
+
+  build(): Cardano.TokenMap {
+    return this.#assets ? this.#assets : new Map();
```
mirceahasegan, line 24: this.#assets is initialised to `new Map()` in constructor.
Also, please return a copy of the assets. Otherwise, the returned assets map is just a reference to the internal `#assets` of the TxTokenBuilder and updating it from the TxTokenBuilder will also update the previously returned `#assets`.
```
-     return this.#assets ? this.#assets : new Map();
+    return new Map(this.#assets);;

File: packages/tx-construction/src/tx-builder/PolicyBuilder.ts
Comment (mirceahasegan) [inline packages/tx-construction/src/tx-builder/PolicyBuilder.ts:1]: Please add unit tests. `derivationPathUpdated` cases should be covered.

File: packages/e2e/test/wallet/PersonalWallet/nft.test.ts [hunk 4]
```diff
@@ -142,47 +90,41 @@ describe('PersonalWallet.assets/nft', () => {
...
+      Cardano.AssetId(`${policyId}${assetNames[TOKEN_METADATA_2_INDEX]}`),
+      Cardano.AssetId(`${policyId}${assetNames[TOKEN_BURN_INDEX]}`)
+    ];
 
-    const finalizeProps: FinalizeTxProps = {
-      auxiliaryData,
-      tx: unsignedTx,
-      witness: { extraSigners: [policySigner], scripts: [policyScript] }
-    };
+    const tokens = new Map([
```
mirceahasegan, line 101: Could you use the token builder here?

packages/tx-construction/src/tx-builder/types.ts
```diff
@@ -223,6 +223,44 @@ export interface TxBuilder {
    * @param portfolio the CIP17 delegation portfolio to apply. Using `null` will deregister all stake keys,
    *  reclaiming the deposits.
    */
+
+  /**
+   *
+   * @param mintToken
+   * list of tokens to mint/burn should be balanced with outputs
+   * to mint tokens use tokenBuilder
+   *      const policy = await txBuilder.buildPolicy();
```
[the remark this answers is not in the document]
mirceahasegan: Added

packages/web-extension/src/observableWallet/util.ts
```diff
@@ -38,7 +38,15 @@ export const outputBuilderProperties: RemoteApiProperties<OutputBuilder> = {
   }
 };
 
-export const txBuilderProperties: RemoteApiProperties<Omit<TxBuilder, 'customize'>> = {
+export const txBuilderProperties: RemoteApiProperties<Omit<TxBuilder, 'customize' | 'buildToken' | 'buildPolicy'>> = {
```
mirceahasegan, line 41: @mkazlauskas , I excluded the policy/token build utilities from the TxBuilder remote api interface to avoid cluttering.
I believe this might have been the intent in the original PR, but it was done by excluding these methods from the TxBuilder interface. I added them in the interface, but excluded them from the exposed properties.
Do you agree?

mirceahasegan:
There is also the `handle.test.ts` that could be migrated to the newer mint/burn api, but it currently uses well written utilities that I did not want to change.

packages/e2e/src/factories.ts
```diff
@@ -322,7 +322,8 @@ const patchInitializeTxToRespectEpochBoundary = <T extends ObservableWallet>(
  * @returns an object containing the wallet and providers passed to it
  */
 export const getWallet = async (props: GetWalletProps) => {
-  const { env, idx, logger, name, polling, stores, customKeyParams, keyAgent, witnesser } = props;
+  const { env, idx, logger, name, polling, stores, customKeyParams, witnesser } = props;
+  const keyAgent = props.keyAgent;
```
iccicci, line 326: Why this change?

And moreover, how could it be that the eslint rule `unicorn/consistent-destructuring` is not applied here?

iccicci (approved):
The benefits from this feature is clearly visible in the e2e tests! 💪

packages/e2e/src/factories.ts
mirceahasegan: Great catch!
Weird indeed. The change was due to me experimenting, and forgot to roll it back

iccicci: Just out of curiosity, mainly about the eslint rule...
I think no actions are required here

packages/web-extension/src/observableWallet/util.ts
mkazlauskas: Problem with this approach is that ObservableWallet.createTxBuilder returns a TxBuilder so those methods would appear as available but just fail. It would be better to make it actually compatible with messaging by following the pattern of OutputBuilder (`build` method async).

packages/tx-construction/src/tx-builder/TxBuilder.ts
```diff
@@ -155,14 +159,34 @@ export class GenericTxBuilder implements TxBuilder {
     });
   }
 
+  addMint(mintToken: Cardano.TokenMap): TxBuilder {
```
mkazlauskas, line 162: Missing tests for new methods in `TxBuilder.test.ts`

mkazlauskas (changes requested):
As agreed on Slack let's park this until we have a use case

PR #894, opened 2023-08-29
ci: print the current commit sha for auditing in all workflows

blaggacao:
# Context

It might be surprising to find out that the HEAD of a feature branch is not representing the same commit as is run during CI.

The reason is GitHubs Merge Queue feature which runs CI on a synthetic commit between a branch and the main branch.

# Proposed Solution

Print out the sha explicitly in every step to facilitate auditing and debugging and avoid false assumptions.

ref: LW-8217

### DIFF
--- a/.github/std-unexploded.yml
+++ b/.github/std-unexploded.yml
@@ -51,6 +51,12 @@ jobs:
       #   with:
       #     nixbuild_ssh_key: ${{ secrets.SSH_PRIVATE_KEY }}
       #     generate_summary_for: job
+
+      - name: Show commit
+        shell: bash
+        run: |
+          echo commit: ${{ github.sha }}
+
       - uses: divnix/std-action/discover@main
         with: {ffBuildInstructions: true}
         id: discovery
@@ -74,6 +80,10 @@ jobs:
           ssh_key: ${{ secrets.SSH_PRIVATE_KEY }}
           user_name: ${{ env.DISCOVERY_USER_NAME }}
           ssh_known_hosts_entry: ${{ env.DISCOVERY_KNOWN_HOSTS_ENTRY }}
+      - name: Show commit
+        shell: bash
+        run: |
+          echo commit: ${{ github.sha }}
       - uses: divnix/std-action/run@main
         with: {ffBuildInstructions: true, remoteStore: "ssh-ng://eu.nixbuild.net"}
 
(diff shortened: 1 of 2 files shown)


.github/std-unexploded.yml
```diff
@@ -51,6 +51,12 @@ jobs:
       #   with:
       #     nixbuild_ssh_key: ${{ secrets.SSH_PRIVATE_KEY }}
       #     generate_summary_for: job
+
+      - name: Show commit
```
rhyslbw, line 55: Could make this an `Action` to DRY

blaggacao: Good idea, addressed in [force-push](https://github.com/input-output-hk/cardano-js-sdk/compare/956e95b3e27cf401bb8530f6cd5d058b6824ff00..2f6a6a91806296061f82aa0295561b2f492981f5). Let's see if it works as expected :hourglass:

blaggacao: Didn't work, maybe [this](https://github.com/input-output-hk/cardano-js-sdk/compare/2f6a6a91806296061f82aa0295561b2f492981f5..a8fbae2e719dbd7d48171892aa770a2ba28dbc4c) does.

.github/std-unexploded.yml
```diff
@@ -51,6 +51,9 @@ jobs:
       #   with:
       #     nixbuild_ssh_key: ${{ secrets.SSH_PRIVATE_KEY }}
       #     generate_summary_for: job
+
+      - uses: ./.github/workflows/show-commit.yaml
```
rhyslbw, line 55:
```suggestion
      - uses: ./.github/actions/show-commit.yaml
```
The convention is to have a `actions` dir

.github/workflows/show-commit.yaml
```diff
@@ -0,0 +1,16 @@
+name: Show commit
```
rhyslbw, line 1: It's not the same structure as a workflow. See composite example here https://docs.github.com/en/actions/creating-actions/creating-a-composite-action#creating-an-action-metadata-file

rhyslbw (changes requested):

.github/workflows/show-commit.yaml
blaggacao: Oh thx! I've been taking this from a different example that was structured as a workflow. Let's see if it works after [this force-push](https://github.com/input-output-hk/cardano-js-sdk/compare/a8fbae2e719dbd7d48171892aa770a2ba28dbc4c..21e25874ccdd085f153549c6b353e985107098fa).

.github/std-unexploded.yml
blaggacao: Turn's out this is not going to work:
> Did you forget to run actions/checkout before running your local action?

I need to revert to the original version.

blaggacao: Well, in the original form, it at least works:
image
:smile:

rhyslbw (approved):

gytis-ivaskevicius (approved):
I feel disgusting approving this :D I assume we will revert it in few days?

blaggacao:
> I assume we will revert it in few days?

I think it may bite more people still until it's comon knowledge and the details around how the commit influences all nix hashes are better understood. :D

PR #897, opened 2023-08-30
ci: fix the deployment run on push to master

blaggacao:
ref: LW-8230

# Context

When someone directly pushes to master, then there's no branch, and hence no pull request that the gh CLI can infer for the purpose of posting a diff back to the PR as comment.

# Proposed Solution

- Run the diff posting job only when the event type is a PR
- Ensure that deploy runs on master regardless of whether it comes from a PR or not (e.g. a bot release directly pushed to master)

# Testing

To test this, a new direct push to master would need to be made.

### DIFF
--- a/.github/std-unexploded.yml
+++ b/.github/std-unexploded.yml
@@ -184,7 +184,10 @@ jobs:
       name: dev-preview
       url: https://backend.dev-preview.eks.lw.iog.io
     # Boolean input should be compared with string until https://github.com/actions/runner/issues/2238 resolved
-    if: fromJSON(needs.discover.outputs.hits).deployments.apply != '{}' && github.event.pull_request.merged == 'true'
+    if: >
+      fromJSON(needs.discover.outputs.hits).deployments.apply != '{}'
+      && github.event_name == 'push'
+      && github.ref == 'refs/heads/master'
     strategy:
       matrix:
         target: ${{ fromJSON(needs.discover.outputs.hits).deployments.apply }}
--- a/.github/std-unexploded.yml
+++ b/.github/std-unexploded.yml
@@ -131,7 +131,7 @@ jobs:
       id-token: write
       contents: read
       pull-requests: write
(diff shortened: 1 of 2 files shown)


.github/std-unexploded.yml
```diff
@@ -184,7 +184,7 @@ jobs:
       name: dev-preview
       url: https://backend.dev-preview.eks.lw.iog.io
     # Boolean input should be compared with string until https://github.com/actions/runner/issues/2238 resolved
-    if: fromJSON(needs.discover.outputs.hits).deployments.apply != '{}' && github.event.pull_request.merged == 'true'
+    if: fromJSON(needs.discover.outputs.hits).deployments.apply != '{}' && (github.event.pull_request.merged == 'true' || github.ref == 'refs/heads/master')
```
rhyslbw, line 187: Can this not be done exclusively with a [push event](https://docs.github.com/en/rest/overview/github-event-types?apiVersion=2022-11-28#pushevent)?

blaggacao: yes! 1 min

blaggacao: No, wait. I'm not sure if a merge (of a PR) counts as push event.

blaggacao: Addressed in [force-push](https://github.com/input-output-hk/cardano-js-sdk/compare/1347c1c444259e4dffe10b5a3e22907507108896..4f5614ebf37f1b5849c075b62b6b917ab944e3b2)

rhyslbw: Ah, no you're right: https://github.com/input-output-hk/cardano-js-sdk/blob/daa05d97a928ed1878bba16848ea40618979e50b/.github/workflows/continuous-integration-unit-tests.yaml#L7-L12

Will likely need to consider both events

rhyslbw: Hang on, that's actually triggering the workflow to run before merge. Let's just see if it runs after merging this PR

rhyslbw (approved):

PR #898, opened 2023-08-30
LW-7438 Refactor cli providers url

iccicci:
# Context

The stake pool rewards job requires the network info provider, so we need to add the network info provider url cli parameter and to propagate it to the job handler.

We had exactly the same problem with the stake pool provider when implementing the stake pool metrics job; in that occasion we introduced the `stakePoolProviderUrl` parameter.

I did this two refactoring rather than _repeating_ what we have for the `stakePoolProviderUrl` parameter to implement the `assetInfoProviderUrl` one; which would have opened the path to an even larger code repetition when adding new provider urls.

# Proposed Solution

- Refactored the `stakePoolProviderUrl` cli parameter into all the provider url parameters.
- Refactored how the configuration is propagated from cli to the job handlers to simply propagate it completely. 

The second commit also let us to move the check(s) about the presence of specific provider urls in a more appropriate place.


mkazlauskas (approved):
Nice refactor :1st_place_medal:

rhyslbw (approved):
Great work @iccicci 💪

PR #882, opened 2023-08-21
LW-7438 APY job

iccicci:
# Context

We need to re-enable the ROS (once called APY) in our stake pool metrics.

# Proposed Solution

- Added a _stake pool reward_ job scheduled on epoch rollover.

# Important Changes Introduced

_Renamed_ the properties for new ROS naming: actually new `ros` related properties was added while `apy` related properties was not removed to avoid a huge impact on `DbSyncStakePoolProvider`; better to remove these properties while removing the db-sync provider.

PR #899, opened 2023-08-31
LW-7438 Fix pg-boss and stake pool metrics errors

iccicci:
# Context

While working on #882 an error was flooding my logs slowing down my development; so I investigated it and I found two aspects of it which I think need to be addressed ASAP.

In same cases the `HttpStakePoolMetadataService.getStakePoolMetadata` method returns a `metric` with value `{ ext: undefined }` rather than `undefined`.

This makes the job to save the _empty_ (rather than `undefined`) `metadata`; which results in an `NOT NULL ticker VIOLATION`.

Not only! Since the first time the **pg-boss** worker gets this error, all subsequent queries performed by **pg-boss** itself (i.e. not by our job handlers) get an error caused by the raw PG error `invalid message format`; this makes **pg-boss** to enter a _locked_ state where it no longer launches jobs but stays alive.

# Proposed Solution

- Fixed `HttpStakePoolMetadataService.getStakePoolMetadata` to correctly return `undefined`.
- Handled the _mysterious_ PG error as a retriable error to let the already in place retry mechanism to recover it (restarting **pg-boss**).

Even if the first commit is enough to solve the problem of the error flooding my logs, I added the second commit as well to cover any other unexpected case which can make **pg-boss** to enter the _locked_ state.

### DIFF
--- a/packages/cardano-services/src/Program/services/pgboss.ts
+++ b/packages/cardano-services/src/Program/services/pgboss.ts
@@ -69,6 +69,15 @@ export interface PgBossServiceDependencies {
   logger: Logger;
 }
 
+const hasDriverError = (error: unknown): error is { driverError: unknown } =>
+  error instanceof Object && 'driverError' in error;
+
+const isRecoverableError = (error: unknown) =>
+  isRecoverableTypeormError(error) ||
+  (hasDriverError(error) &&
+    error.driverError instanceof Error &&
+    error.driverError.message === 'invalid message format');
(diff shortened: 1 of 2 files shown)


mkazlauskas (approved):
Nice catch! :detective: :muscle: 
Side note: those pgboss tests are a little hard to read :sweat_smile:

rhyslbw (approved):
Great work @iccicci 💪

PR #901, opened 2023-09-02
Feat/lw 7585 implement serialization of witness set

AngelCastilloB:
# Context

# Proposed Solution

- Added transaction witness set serialization classes.

### DIFF
--- a/packages/core/src/Serialization/TransactionWitnessSet/TransactionWitnessSet.ts
+++ b/packages/core/src/Serialization/TransactionWitnessSet/TransactionWitnessSet.ts
@@ -1,0 +1,483 @@
+/* eslint-disable sonarjs/cognitive-complexity, complexity, max-statements */
+
+import * as Cardano from '../../Cardano';
+import { BootstrapWitness } from './BootstrapWitness';
+import { CborReader, CborReaderState, CborWriter } from '../CBOR';
+import { HexBlob } from '@cardano-sdk/util';
+import { NativeScript, PlutusV1Script, PlutusV2Script } from '../Scripts';
+import { PlutusData } from '../PlutusData/PlutusData';
+import { Redeemer } from './Redeemer';
+import { SerializationError, SerializationFailure } from '../../errors';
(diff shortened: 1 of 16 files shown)


packages/core/src/Serialization/TransactionWitnessSet/TransactionWitnessSet.ts
```diff
@@ -0,0 +1,483 @@
...
+  native: Array<NativeScript> | undefined;
+  plutusV1: Array<PlutusV1Script> | undefined;
+  plutusV2: Array<PlutusV2Script> | undefined;
+};
+
+/**
+ * A witness is a piece of information that allows you to efficiently verify the
+ * authenticity of the transaction (also known as proof).
+ *
+ * In Cardano, transactions need multiple types of authentication proofs, these can range
```
rhyslbw, line 27:
```suggestion
 * In Cardano, transactions have multiple types of authentication proofs, these can range
```

rhyslbw (changes requested):
Awesome work @AngelCastilloB 🔥 

It looks like 0f7298aef7cea11973f973398979bff61e9e2940 should be squashed with the first commit though, if the tests were failing without it. It's great to see this implementation has a positive effect on transaction cost 😎

packages/tx-construction/test/fees/fees.test.ts
```diff
@@ -120,7 +120,7 @@ describe('fees', () => {
     it('calculate the correct min fee for transaction with scripts', () => {
       const prices = { memory: 0.0577, steps: 0.000_007_21 };
       const fee = minFee(babbageTx, prices, MinFeeConstant(155_381), MinFeeCoefficient(44));
-      expect(fee).toBe(218_807n);
+      expect(fee).toBe(218_763n);
```
mkazlauskas, line 123: Witness set is now smaller? :mage:

mkazlauskas (approved):
Great work!

rhyslbw (approved):

PR #902, opened 2023-09-04
Feat/Add separate withdrawals to implicit coin

xdzurman:
# Context

Currently there's no way of knowing whether the value in `implicitCoin`'s `input` comes from rewards, stake key de-registration, or a combination of them. With this change, this is solved.

# Proposed Solution

Add a `withdrawals` prop to the return object, which contains only the user's rewards. This is a non-breaking change, and by `diff`ing `input` and `withdrawals`, it's possible to also determine what the stake key deregistration return is.


mkazlauskas (approved):

rhyslbw (approved):
Nice work @xdzurman 👍

mkazlauskas (approved):

PR #903, opened 2023-09-04
Finalize hardware trezor package [LW-6803]

DominikGuzei:
# Context

This PR is about finalizing refactoring the legacy Trezor integration in the `key-management` package to its standalone `hardware-trezor` package. The goal is to finish all outstanding refactoring not associated with individual mapper task is completed (generalizing utils, code de-duplication, etc.)

# Proposed Solution

- Trezor / Ledger mappers are removed from `key-management` package
- shared utils for hardware wallet packages are refactored to a single location

- New shared key utils in `key-management` package
- Standalone `hardware-trezor` package is finalized
- Shared utils are refactored from `hardware-ledger` and `hardware-trezor`


DominikGuzei:
@tomislavhoracek @rhyslbw @marcin-mazurek any ideas where to put these potentially shared helpers for the hardware wallet packages? Or do you think it doesn't make sense to refactor them (as they contain some specific types)
CleanShot 2023-09-05 at 13 06 09@2x

tomislavhoracek:
> ou think it doesn't make sense to refactor the

@DominikGuzei thanks for thinking about unifying stuff :) 
It is not just assets mapper / util but we also have `keyPaths` transformer and maybe / probably something more. 
If I remember well we already asked @rhyslbw about that when we started working on Trezor Package and Rhys agreed we should unify that. 
Also, when we talked with @mchappell yesterday we touched that topic. 
From my perspective, if you agree, let's keep it simple and release it as is. Once we have both, Ledger & Trezort packages OUT, we will see what can be unified and we can do that with less stress. What do you think guys?
cc @marcin-mazurek @rhyslbw @mkazlauskas @mchappell

DominikGuzei:
> > ou think it doesn't make sense to refactor the
> 
> @DominikGuzei thanks for thinking about unifying stuff :) It is not just assets mapper / util but we also have `keyPaths` transformer and maybe / probably something more. If I remember well we already asked @rhyslbw about that when we started working on Trezor Package and Rhys agreed we should unify that. Also, when we talked with @mchappell yesterday we touched that topic. From my perspective, if you agree, let's keep it simple and release it as is. Once we have both, Ledger & Trezort packages OUT, we will see what can be unified and we can do that with less stress. What do you think guys? cc @marcin-mazurek @rhyslbw @mkazlauskas @mchappell

`keyPath` helpers are already moved into `key-management` 👍

DominikGuzei:
@marcin-mazurek @rhyslbw Tomo and I discussed that we see three options for the canonical ordering helpers:
- Leave them as-is because it's not a big win atm
- Refactor them to a higher level package (e.g: core or tx-construction) so that nobody needs to think about canonical sorting of data when constructing transactions.

There is also the possibility to defer this work to a separate ticket that can be tackled separately (in order to not to delay this PR unnecessarily)

rhyslbw:
Seems like deferring to a follow-up refactor is the best option. @AngelCastilloB will have some ideas about what that could look like too

packages/key-management/src/index.ts
tomislavhoracek, line 1: @DominikGuzei 💪

packages/key-management/src/TrezorKeyAgent.ts
tomislavhoracek, line 1: @DominikGuzei 💪

tomislavhoracek (approved):
@DominikGuzei Awesome 🎉

PR #904, opened 2023-09-05
Feat/lw 7586 implement serialization of aux metadata

AngelCastilloB:
# Context

# Proposed Solution

- Added transaction auxiliary data serialization classes.
- Implement hashing functions for transaction types.
- Remove CML as a dependency from the core package.


packages/core/src/Serialization/PlutusData/PlutusMap.ts
```diff
@@ -88,7 +88,13 @@ export class PlutusMap {
    * if there is no element with the given key.
    */
   get(key: PlutusData): PlutusData | undefined {
-    return this._map.get(key);
+    if (!this._map) return undefined;
+
+    const element = [...this._map.entries()].find((entry) => deepEquals(entry[0], key));
```
mkazlauskas, line 93: Just a nitpick suggestion, entries is fine too
```suggestion
    const element = [...this._map.keys()].find((mapKey) => deepEquals(mapKey, key));
```

packages/core/test/Serialization/PlutusData.test.ts
```diff
@@ -167,6 +167,14 @@ describe('PlutusData', () => {
 
       expect(data.toCbor()).toEqual('a10102');
     });
+
+    it('can find an element in a plutus map with key', () => {
+      const data = new Serialization.PlutusMap();
+
+      data.insert(Serialization.PlutusData.newInteger(1n), Serialization.PlutusData.newInteger(2n));
+
+      expect(data.get(Serialization.PlutusData.newInteger(1n))).toEqual(Serialization.PlutusData.newInteger(2n));
```
mkazlauskas, line 176: Would be great to add this assertion to make sure that 1n and 2n PlutusData are not deepEquals. Wouldn't be a bad idea to add `equals` method to all PlutusData classes to be sure it works as expected with all kinds.
```ts
expect(data.get(Serialization.PlutusData.newInteger(2n))).toBeUndefined();
```

mkazlauskas (changes requested):

packages/core/src/Serialization/PlutusData/PlutusData.ts
```diff
@@ -370,7 +370,43 @@ export class PlutusData {
...
+  }
+
+  /**
+   * Indicates whether some other PlutusData is "equal to" this one.
+   *
+   * @param other The other object to be compared.
+   * @returns true if objects are equals; otherwise false.
+   */
+  // eslint-disable-next-line complexity
+  equals(other: PlutusData): boolean {
```
mkazlauskas, line 383: Even though this is indirectly tested in `PlutusMap.get`, this is a public method that deserves a couple of direct tests

packages/core/src/Serialization/PlutusData/PlutusData.ts
```diff
@@ -370,7 +370,43 @@ export class PlutusData {
...
+    let result = false;
+    switch (this.#kind) {
+      case PlutusDataKind.Bytes:
+        return HexBlob.fromBytes(this.#bytes!) === HexBlob.fromBytes(other.#bytes!);
```
mkazlauskas, line 387: Possible optimization here (nitpick :sweat_smile:)
```suggestion
        return this.#bytes && this.#bytes.length === other.#bytes?.length && this.#bytes.every((value, index) => value === other.#bytes![index]);
```

packages/core/src/Serialization/PlutusData/PlutusData.ts
```diff
@@ -370,7 +370,43 @@ export class PlutusData {
...
+          return this.#list.equals(other.#list);
+        }
+        return false;
+      case PlutusDataKind.Map:
+        if (this.#map && other.#map) {
+          return this.#map.equals(other.#map);
+        }
+        return false;
+      default:
+        result = false;
```
mkazlauskas, line 406: One more nitpick
```suggestion
        return false;
```

mkazlauskas (changes requested):

mkazlauskas (approved):
Great work! :rocket: When squashing, please change the `equals` commit type to `feat` as it's really just adding a feature

rhyslbw (approved):
Fantastic work @AngelCastilloB :mage: :boom:

PR #907, opened 2023-09-06
feat: add support for signing data with a DRepID in CIP-95 API

lgobbi-atix:
# Context

Extend the capabilities of [CIP-30's .signData()](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0030/README.md#apisigndataaddr-address-payload-bytes-promisedatasignature) to allow for signatures using DRep keys.

Allow for `DRepID` to be passed in the `addr` field to sign using the associated DRep key.

To construct an address for DRep Key, the client application should construct a [type 6 address](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0019/CIP-0019-cardano-addresses.abnf#L7C8-L7C93) using an appropriate [Network Tag](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0019/CIP-0019-cardano-addresses.abnf#L13) and a hash of a public DRep Key.

# Proposed Solution

- Add new `DRepID` type with `DRepID.isValid()` function
- Add address type check in `PaymentAddress` in case it's a `DRepID`. Only allow type 6 addresses.


packages/key-management/src/cip8/cip30signData.ts
```diff
@@ -50,6 +50,11 @@ const getAddressBytes = (signWith: Cardano.PaymentAddress | Cardano.RewardAccoun
 };
 
 const getDerivationPath = async (signWith: Cardano.PaymentAddress | Cardano.RewardAccount, keyAgent: AsyncKeyAgent) => {
```
mirceahasegan, line 52: I think it would be better to add DrepID in the function signature:
```
- const getDerivationPath = async (signWith: Cardano.PaymentAddress | Cardano.RewardAccount, keyAgent: AsyncKeyAgent) => {
+ const getDerivationPath = async (
+  signWith: Cardano.PaymentAddress | Cardano.RewardAccount | Cardano.DRepID,
+  keyAgent: AsyncKeyAgent
+ ) => {

File: packages/key-management/src/cip8/cip30signData.ts [hunk 2]
```diff
@@ -50,6 +50,11 @@ const getAddressBytes = (signWith: Cardano.PaymentAddress | Cardano.RewardAccoun
...
+  const isDRepID = signWith.startsWith('drep_id');
+  if (isDRepID) {
```
mirceahasegan, line 54: Why not use DrepID.isValid?
```diff
-  const isDRepID = signWith.startsWith('drep_id');
-  if (isDRepID) {
+ if (DRepID.isValid(signWith)) {

File: packages/wallet/src/cip30.ts [hunk 3]
```diff
@@ -406,7 +406,10 @@ const baseCip30WalletApi = (
...
     }
   },
-  signData: async (addr: Cardano.PaymentAddress | Bytes, payload: Bytes): Promise<Cip30DataSignature> => {
+  signData: async (
+    addr: Cardano.PaymentAddress | Cardano.DRepID | Bytes,
+    payload: Bytes
+  ): Promise<Cip30DataSignature> => {
     logger.debug('signData');
     const hexBlobPayload = HexBlob(payload);
     const signWith = Cardano.PaymentAddress(addr);
```
mirceahasegan, line 415: I think it would be clearer if we use the DrepID type. It would also require updating `Cip30SignDataRequest` to accept `DrepID`.
```diff
- const signWith = Cardano.PaymentAddress(addr);
+ const signWith = DRepID.isValid(addr) ? DRepID(addr) : Cardano.PaymentAddress(addr);

File: packages/core/src/Cardano/Address/PaymentAddress.ts [hunk 4]
```diff
@@ -40,6 +41,9 @@ export const PaymentAddress = (value: string): PaymentAddress => {
     if (isRewardAccount(value)) {
       throw new InvalidStringError(value, 'Address type can only be used for payment addresses');
     }
+    if (DRepID.isValid(value) && Address.fromBech32(value).getType() !== AddressType.EnterpriseKey) {
```
mirceahasegan, line 44: Why validate DRepID as payment address?

packages/wallet/src/cip30.ts
lgobbi-atix: Should I also add a `DRepID.canSign()` function to check if the DRepID is the correct type?

According to CIP-95 only type

lgobbi-atix: Otherwise, it lets me sign with types 6, 7, 14 and 15. The rest fail because of `Invalid address` in `getAddressBytes() (cip30signData.ts)`

lgobbi-atix: Added this and addressed the rest of the comments in https://github.com/input-output-hk/cardano-js-sdk/pull/907/commits/c74bed59c4b80616b1cc71167cb67ac7656f2921

mirceahasegan (dismissed):
Awesome work, @lgobbi-atix 👏

rhyslbw (commented):
Point to note here that we have a package called `governance` already, so we should probably consider migrating it's existing code into other packages in a later refactor considering this code is not being implemented there.

66122d7b9d33fa97637a51462fc0e7270919200a can also be squashed here as you're introducing this functionality in the commit prior, and it's naturally part of that change.

rhyslbw (dismissed):
Well done @lgobbi-atix 🦾

rhyslbw:
@AngelCastilloB please review

lgobbi-atix:
@rhyslbw @AngelCastilloB @mirceahasegan pushed a new commit cc92e8387945b5f10584b956533cd621a3f2c902 as the prefix for DRep IDs changed to just `drep` from `drep_id`. Could you review again?

mirceahasegan (approved):

rhyslbw (approved):

PR #909, opened 2023-09-07
Adds live endpoint and fixes DbSyncProvider ready endpoint

rhyslbw:
# Context
Services need to report when they are live and also when they are ready, which can be used to interpret the service health. Orchestration logic can be tied to the live result, to determine if a restart is required, or the ready result for starting up dependent services.

# Proposed Solution
- Adds `/live` endpoint that simply returns a `200` if the server is reachable.
- `DbSyncProvider` now only as to access both the node and database to be considered ready.

`/ready` endpoint no longer returns the health check response, and it's been included in the `HttpServer` OpenApi document

### DIFF
--- a/nix/cardano-services/deployments/handle-provider-deployment.yaml
+++ b/nix/cardano-services/deployments/handle-provider-deployment.yaml
@@ -66,7 +66,7 @@ spec:
           image: FIXME-ADD-LATEST-IMAGE-REFERENCE
           livenessProbe:
             httpGet:
-              path: /v1.0.0/health
+              path: /v1.0.0/live
               port: 3000
             timeoutSeconds: 5
           name: handle-provider
--- a/packages/cardano-services/src/Http/HttpServer.ts
+++ b/packages/cardano-services/src/Http/HttpServer.ts
@@ -137,6 +137,10 @@ export class HttpServer extends RunnableModule {
 
     const handlers = {
       health: healthCheckHandler(() => 200),
+      live: async (req: express.Request, res: express.Response) => {
+        this.logger.debug('/live', { ip: req.ip });
+        return res.sendStatus(200);
+      },
       meta: serverMetadataHandler,
       ready: healthCheckHandler(({ ok }) => (ok ? 200 : 503))
     };
--- a/packages/cardano-services/src/Http/HttpServer.ts
+++ b/packages/cardano-services/src/Http/HttpServer.ts
@@ -97,20 +97,6 @@ export class HttpServer extends RunnableModule {
       this.logger.info(`Using ${serviceBaseUrl}`);
     }
 
-    const healthCheckHandler =
-      (statusCodeFromHealth: (health: ServicesHealthCheckResponse) => number) =>
-      async (req: express.Request, res: express.Response) => {
-        this.logger.debug('/ready', { ip: req.ip });
-        let health: ServicesHealthCheckResponse | Error['message'];
-        try {
-          health = await this.#getServicesHealth();
-        } catch (error) {
-          this.logger.error(error);
-          return HttpServer.sendJSON(res, new ProviderError(ProviderFailure.Unhealthy, error), 500);
-        }
-        return HttpServer.sendJSON(res, health, statusCodeFromHealth(health));
-      };
-
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     this.app.use((err: any, _req: express.Request, res: express.Response, _n: express.NextFunction) => {
       if (err instanceof ProviderError) {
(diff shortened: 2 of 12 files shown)


mkazlauskas (approved):
LGTM :rocket:

blaggacao:
> or the ready result for starting up dependent services

or the ready result for knowing when to hook up a service into the load balancer

blaggacao (approved):
Nice! This is still `version`-spaced in the path, right? I understand we need to keep this for a while, still, until we have a more capable ingress class.

rhyslbw:
> This is still version-spaced in the path, right? I understand we need to keep this for a while, still, until we have a more capable ingress class.

Yes, I mentioned this could be hoisted, but then I realised it's still an API that still deserves to be specified, and we have versioning in place already. Chances are this version won't change BTW

rhyslbw:
The test failure was due to the `/ready` endpoint returning the health response in the body, which is a violation of the API spec. I added https://github.com/input-output-hk/cardano-js-sdk/pull/909/commits/6a7097f2cf65d522d6f4d2accd629e28057a84e5 to address this, and will rename the commit when squashing. Stylistic change in https://github.com/input-output-hk/cardano-js-sdk/pull/909/commits/a1043169b54e08b9127f63cc259d7a58a81213bd

blaggacao (dismissed):

iccicci (dismissed):
LGTM!

rhyslbw:
@blaggacao do you want to update the service config in this PR to use the new `/live` endpoint instead?

rhyslbw:
@blaggacao Actually, it was fairly easy for me to update it in https://github.com/input-output-hk/cardano-js-sdk/commit/2ba08acebaaa402a7322b341f6bfd5f541fb6b8f

nix/cardano-services/deployments/handle-provider-deployment.yaml
```diff
@@ -66,7 +66,7 @@ spec:
           image: FIXME-ADD-LATEST-IMAGE-REFERENCE
           livenessProbe:
             httpGet:
-              path: /v1.0.0/health
+              path: /v1.0.0/live
```
iccicci, line 69: Could it be this change should be reflected on health checks of our `docker-compose` files?
I see CI is passing... I would have expected all containers to be unhealthy...

Could it be we are adding the `/live` endpoint **and** we are keeping the `/health`?

Thank you

iccicci (changes requested):

nix/cardano-services/deployments/handle-provider-deployment.yaml
rhyslbw: Yes, great catch! Yes, health is staying since it includes information about the service state. `live` and `ready` are binary responses

iccicci (approved):

nix/cardano-services/deployments/handle-provider-deployment.yaml
iccicci: Much better! We don't need to change `docker-compose` cfg as it's heath checks are not impacted.

rhyslbw:
I had a look at the `/health` usage in compose files, and identified just [this](https://github.com/input-output-hk/cardano-js-sdk/pull/909/commits/95573b74254a50470f7d00546099ec01322e8151) one as a candidate for replacement with `/live`, since the others interpret the response.

rhyslbw:
@iccicci I had a look at the `/health` usage in compose files, and identified just [this](https://github.com/input-output-hk/cardano-js-sdk/pull/909/commits/95573b74254a50470f7d00546099ec01322e8151) one as a candidate for replacement with `/live`, since the others interpret the response.

iccicci (approved):

packages/cardano-services/src/Http/HttpServer.ts
```diff
@@ -136,9 +122,33 @@ export class HttpServer extends RunnableModule {
...
+        } catch (error) {
+          this.logger.error(error);
+          return HttpServer.sendJSON(res, new ProviderError(ProviderFailure.Unhealthy, error), 500);
+        }
+      },
+      live: async (req: express.Request, res: express.Response) => {
+        this.logger.debug('/live', { ip: req.ip });
+        return res.sendStatus(200);
+      },
       meta: serverMetadataHandler,
```
gytis-ivaskevicius, line 140: Wait a second, this does not do much, in definitions we should not replace liveliness probe because this acts more like startup probe

gytis-ivaskevicius (changes requested):

packages/cardano-services/src/Http/HttpServer.ts
rhyslbw: We have `/live` and `/ready`, with the former being somewhat of a startup test. Is that not what the liveliness probe needs?

gytis-ivaskevicius: Short summary is:
liveliness probe - assesses if an application is healthy
startup probe - assesses if an application has started
readiness probe - assesses if an application is ready to accept connections

PR #912, opened 2023-09-08
fix: service name url mismatch

blaggacao:
# Context

The url link from pg-boss-worker needs to point to the correct service name.

# Solution

Without really understanding how they are rendered and if the actual service name would be rendered as expceted, ensure that they are both rendered the same way (either right or wrong).

# Testing

This is currently a root cause for a failure in `dev-preview` deployment.

- [x] As per: https://github.com/input-output-hk/wallet-world/pull/2854

### DIFF
--- a/nix/cardano-services/deployments/pgboss-deployment.yaml
+++ b/nix/cardano-services/deployments/pgboss-deployment.yaml
@@ -74,7 +74,7 @@ spec:
             - name: QUEUES
               value: pool-metadata,pool-metrics
             - name: STAKE_POOL_PROVIDER_URL
-              value: https://dev-preview-cardanojs-backend.dev-preview.svc.cluster.local/stake-pool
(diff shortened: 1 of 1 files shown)


gytis-ivaskevicius (commented):
I see the Testing section but I don't see any test scenarios 🤔
Why wasn't this picked up by /health???

iccicci (approved):

gytis-ivaskevicius (approved):

iccicci (approved):

blaggacao:
The CI failure is due to an old package version of Nix in `discovery` runners. The image tag, however, does not change via this change set. Test have been successfully executed according to the protocol in https://github.com/input-output-hk/wallet-world/pull/2854.

Deploying now manually,

rhyslbw (approved):

PR #914, opened 2023-09-12
ci: remove gpg passphrase

blaggacao:
# Context

The new Lace Bot GPG key has no passphase.

# Proposed Solution

Remove the passphrase config variable.

# Testing

- [x] run the ci


mkazlauskas:
Please test the workflow change by running `release` with workflow_dispatch against this branch workflow.

blaggacao:
> Please test the workflow change by running release with workflow_dispatch against this branch workflow.

&rarr; https://github.com/input-output-hk/cardano-js-sdk/actions/runs/6158389380 :pray:

mkazlauskas (approved):
Nice work!

.github/workflows/release.yaml
```diff
@@ -7,48 +7,47 @@ jobs:
...
+        with:
+          node-version: 18.12.0
+          registry-url: 'https://registry.npmjs.org'
 
-    - name: Import bot's GPG key for signing commits
-      id: import-gpg
-      uses: crazy-max/ghaction-import-gpg@v5
-      with:
-        gpg_private_key: ${{ secrets.BOT_GPG_KEY }}
-        passphrase: ${{ secrets.BOT_GPG_PASSPHRASE }}
```
gytis-ivaskevicius, line 32: I guess this is the only relevant change? It feels weird to have gpg key without a passphrase but I guess it does not make much difference

gytis-ivaskevicius (approved):

.github/workflows/release.yaml
blaggacao: Both secrets are held by the GH action and the same GH secret store, so a passphrase only adds complexity at no gain in security.

PR #915, opened 2023-09-12
fix: pgboss url configuration

blaggacao:
# Context

The application adds the path suffix all on its own, furthermore no https (port 443) is configured _within_ the cluster network.

# Proposed Solutioni
- remove the path suffix
- change https to http

# Testing
- [x] `dev-preview` is still not working and this gets us one step closer, see [comment](https://github.com/input-output-hk/cardano-js-sdk/pull/915#pullrequestreview-1622325439)


gytis-ivaskevicius (approved):
Why wasn't this picked up in testing?

iccicci (approved):
LGTM!

rhyslbw (approved):
Nice work @blaggacao. We're still using the `dev-preview` env as a test bed so testing can be completed after merge, with any changes required done immediately.

PR #916, opened 2023-09-14
ci: fix handle provider OCI image reference

gytis-ivaskevicius:
# Context
For the past a while we have handle provider failing due to invalid OCI image name. This is a fix to reduce the amount of alerts

### DIFF
--- a/nix/cardano-services/deployments.nix
+++ b/nix/cardano-services/deployments.nix
@@ -24,6 +24,14 @@ in {
         }
       ];
     };
+    handle-provider-deployment = {
+      spec.template.spec.containers = dmerge.updateOn "name" [
+        {
+          name = "handle-provider";
+          image = cell.oci-images.cardano-services.image.name;
+        }
+      ];
+    };
     handle-projector-deployment = {
       spec.template.spec.containers = dmerge.updateOn "name" [
         {


nix/cardano-services/deployments.nix
```diff
@@ -24,6 +24,14 @@ in {
         }
       ];
     };
+    handle-provider-deployment = {
```
rhyslbw, line 27: What's the default this overrides?

nix/cardano-services/deployments.nix
```diff
@@ -24,6 +24,14 @@ in {
...
+      spec.template.spec.containers = dmerge.updateOn "name" [
+        {
+          name = "handle-provider";
+          image = cell.oci-images.cardano-services.image.name;
```
rhyslbw, line 31: What's the default this overrides?

gytis-ivaskevicius: While diff is not very clear but here is the difference, default is invalid image name to avoid it failing silently

Diff:
```diff
-        image: FIXME-ADD-LATEST-IMAGE-REFERENCE
+        image: 926093910549.dkr.ecr.us-east-1.amazonaws.com/cardano-services:vh1wcqh4y1i85bvjdfic9a965dy1d848
```

bernokl (approved):

PR #924, opened 2023-09-21
Sanchonet support

michalrus:
So far only `x86_64-linux` works, I’ll fix an `aarch64-linux` version shortly (for easy running on macOS on Apple ~M1~ Silicon)

# Context

We need a Dockerfile where we can specify exact `cardano-node` and `ogmios` commits.

# Proposed Solution

A Dockerfile that’s using haskell.nix inside (you don’t have to have Nix installed!).

Ogmios stopped having Nix definitions in their tree, but with slight tweaking, we can build it with haskell.nix taken from the specified `cardano-node` commit.

As a developer choosing those commits, you have to make sure they're at least somewhat compatible for Ogmios to work with cardano-node, and that kind of guarantees, that haskell.nix and Cardano Haskell Packages from this `cardano-node` will be able to build the selected commit of `ogmios`.

# Important Changes Introduced

n/a

PR #925, opened 2023-09-22
fix(ogmios): fix our Dockerfile for Ogmios 5.x and cardano-node 1.35.x

michalrus:
# Context

It turns out that the Dockerfile from the previous PR #924 doesn’t work for older Ogmios 5.6.x and cardano-node 1.35.5

# Proposed Solution

Now they do 

n/a


michalrus:
Closing as already cherry-picked onto PR #924

PR #929, opened 2023-09-25
Feat/lw 8188 upgrade serialization to conway era

AngelCastilloB:
# Context

We need to upgrade our serialization classes in order to support the Conway era transactions.

# Proposed Solution

Add the necessary serialization classes to support Conway era.

New serialization classes were added


packages/core/src/Cardano/types/Certificate.ts
```diff
@@ -40,7 +40,7 @@ export interface NewStakeAddressCertificate {
 export interface VoteDelegationCertificate {
   __typename: CertificateType.VoteDelegation;
   stakeKeyHash: Crypto.Ed25519KeyHashHex;
-  drep: DelegateRepresentative;
+  dRep: DelegateRepresentative;
```
mkazlauskas, line 43: This and some other type changes in `Governance.ts` are technically breaking, please reflect it in commit message

mkazlauskas (changes requested):

packages/core/src/Cardano/types/Certificate.ts
mkazlauskas: If we intend to squash this commit to the one that created Governance.ts then yeah

packages/core/src/Serialization/TransactionBody/ProposalProcedure/InfoAction.ts
```diff
@@ -0,0 +1,69 @@
...
+   */
+  static fromCbor(cbor: HexBlob): InfoAction {
+    const reader = new CborReader(cbor);
+
+    const kind = Number(reader.readUInt());
+
+    if (kind !== GovernanceActionKind.Info)
+      throw new InvalidArgumentError(
+        'cbor',
+        `Expected action kind, expected ${GovernanceActionKind.HardForkInitiation}  but got ${kind}`
```
mchappell, line 43:
```suggestion
        `Expected action kind, expected ${GovernanceActionKind.Info}  but got ${kind}`
```

packages/core/src/Serialization/TransactionBody/ProposalProcedure/NewConstitution.ts
```diff
@@ -0,0 +1,134 @@
...
+        'cbor',
+        `Expected an array of ${EMBEDDED_GROUP_SIZE} elements, but got an array of ${length} elements`
+      );
+
+    const kind = Number(reader.readUInt());
+
+    if (kind !== GovernanceActionKind.NewConstitution)
+      throw new InvalidArgumentError(
+        'cbor',
+        `Expected action kind, expected ${GovernanceActionKind.HardForkInitiation}  but got ${kind}`
```
mchappell, line 72:
```suggestion
        `Expected action kind, expected ${GovernanceActionKind.NewConstitution}  but got ${kind}`
```

mchappell (changes requested):

rhyslbw (approved):
This is incredible work @AngelCastilloB :mage: 

I'm happy to see this be merged since this era is pre-production, and the priority for us is early integration. We can perform further reviews once there's test vectors available for gaining confidence in the serialisation.

packages/core/src/Cardano/types/Certificate.ts
rhyslbw: Could also just cherry-pick this commit onto the other branch then squash it when ready to merge. I agree it's part of the same atomic change

PR #931, opened 2023-09-25
Fix/ogmios dockerfile

iccicci:
# Context

#924 is a long story PR.

# Proposed Solution

Some minor fixes.

### DIFF
--- a/compose/ogmios/Dockerfile
+++ b/compose/ogmios/Dockerfile
@@ -81,4 +81,5 @@ EXPOSE 1337/tcp 3000/tcp 12788/tcp 12798/tcp
 HEALTHCHECK --interval=10s --timeout=5s --retries=1 CMD /bin/ogmios health-check
 STOPSIGNAL SIGINT
 COPY --from=haskell-builder /app/ogmios-src/scripts/cardano-node-ogmios.sh cardano-node-ogmios.sh
+RUN ln -s /sbin/tini /tini
 ENTRYPOINT ["tini", "-g", "--", "/root/cardano-node-ogmios.sh" ]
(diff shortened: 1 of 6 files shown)


compose/ogmios/Dockerfile
```diff
@@ -81,4 +81,5 @@ EXPOSE 1337/tcp 3000/tcp 12788/tcp 12798/tcp
 HEALTHCHECK --interval=10s --timeout=5s --retries=1 CMD /bin/ogmios health-check
 STOPSIGNAL SIGINT
 COPY --from=haskell-builder /app/ogmios-src/scripts/cardano-node-ogmios.sh cardano-node-ogmios.sh
+RUN ln -s /sbin/tini /tini
```
rhyslbw, line 84: @michalrus FYI

michalrus: I don’t know if this is needed if later, in `ENTRYPOINT`, we call `tiny` (from `PATH`) instead of `/tiny` (absolute/relative), but it doesn’t hurt =)

The original Dockerfile calls `/tiny`

iccicci: Yep, later we have [this override](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/e2e/docker-compose.yml#L43).

Since
> The original Dockerfile calls `/tiny`

I preferred to make our custom image as more compliant to the original as possible rather than applying a workaround the the override to handle the slight difference fixed by this change.

michalrus: Ah, you’re right! Sorry, I missed it 🫣

rhyslbw:
Superseded by https://github.com/input-output-hk/cardano-js-sdk/pull/935

PR #932, opened 2023-09-25
Fix/LW-8192 fix for local network running on mac

VanessaPC:
# Context

Local-network wasn't running on Mac, we couldn't access the full functionality of the SDK. 
The issue was that we couldn't do inline replacement because of permission errors in Mac, this couldn't be solved by giving it root permission. 

# Proposed Solution
Use a temporary file to make the changes and copy them back to where we want the change to be. 

### DIFF
--- a/packages/e2e/local-network/scripts/make-babbage.sh
+++ b/packages/e2e/local-network/scripts/make-babbage.sh
@@ -273,14 +273,24 @@ mkdir -p ./config/network/genesis/
 cp ./templates/babbage/db-sync-config.json ./config/network/cardano-db-sync/config.json
 cp ./templates/babbage/node-config.json ./config/network/cardano-node/config.json
 
-$SED -i -E "s/\"ByronGenesisHash\": \".*\"/\"ByronGenesisHash\": \"${byronGenesisHash}\"/" ./config/network/cardano-node/config.json
-$SED -i -E "s/\"ShelleyGenesisHash\": \".*\"/\"ShelleyGenesisHash\": \"${shelleyGenesisHash}\"/" ./config/network/cardano-node/config.json
-$SED -i -E "s/\"AlonzoGenesisHash\": \".*\"/\"AlonzoGenesisHash\": \"${alonzoGenesisHash}\"/" ./config/network/cardano-node/config.json
+###############################
+# NOTE: We can't use -i to do inline replacement because this breaks due to permission errors on MacOS when creating the temp output file in this case, giving Root permissions to Docker is also not 
+# a valid working option for MacOS.
+# https://forums.docker.com/t/sed-couldnt-open-temporary-file-xyz-permission-denied-when-using-virtiofs/125473
+###############################
+
+$SED -E "s/\"ByronGenesisHash\": \".*\"/\"ByronGenesisHash\": \"${byronGenesisHash}\"/" ./config/network/cardano-node/config.json > ./config/network/cardano-node/config.json.tmp
+mv ./config/network/cardano-node/config.json.tmp ./config/network/cardano-node/config.json
+$SED -E "s/\"ShelleyGenesisHash\": \".*\"/\"ShelleyGenesisHash\": \"${shelleyGenesisHash}\"/" ./config/network/cardano-node/config.json > ./config/network/cardano-node/config.json.tmp
+mv ./config/network/cardano-node/config.json.tmp ./config/network/cardano-node/config.json
+$SED -E "s/\"AlonzoGenesisHash\": \".*\"/\"AlonzoGenesisHash\": \"${alonzoGenesisHash}\"/" ./config/network/cardano-node/config.json > ./config/network/cardano-node/config.json.tmp
+mv ./config/network/cardano-node/config.json.tmp ./config/network/cardano-node/config.json
 
 cp ./templates/babbage/topology.json ./config/network/cardano-node/topology.json
 # docker hostname in topology.json isn't working, so need to specify ip of local network
 CONTAINER_IP=$(hostname -I | xargs)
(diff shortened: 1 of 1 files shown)


iccicci (dismissed):
LGTM!

packages/e2e/local-network/scripts/make-babbage.sh
```diff
@@ -273,14 +275,24 @@ mkdir -p ./config/network/genesis/
 cp ./templates/babbage/db-sync-config.json ./config/network/cardano-db-sync/config.json
 cp ./templates/babbage/node-config.json ./config/network/cardano-node/config.json
 
-$SED -i -E "s/\"ByronGenesisHash\": \".*\"/\"ByronGenesisHash\": \"${byronGenesisHash}\"/" ./config/network/cardano-node/config.json
-$SED -i -E "s/\"ShelleyGenesisHash\": \".*\"/\"ShelleyGenesisHash\": \"${shelleyGenesisHash}\"/" ./config/network/cardano-node/config.json
-$SED -i -E "s/\"AlonzoGenesisHash\": \".*\"/\"AlonzoGenesisHash\": \"${alonzoGenesisHash}\"/" ./config/network/cardano-node/config.json
+###############################
+# NOTE: We can't use -i to do inline replacement because this breaks due to permission errors on MacOS when creating the temp output file in this case, giving Root permissions to Docker is also not
```
pczeglik-iohk, line 279: This comment is a great note for the person who is doing a code review, additional context to reason about the change. 

I would prefer to see it as a comment in Github UI though, as you put focus on what cannot be done / be used. Not on what you have actually decided to do.

I like the content you have put in PR's **Proposed solution**.

`Use a temporary file to make the changes and copy them back to where we want the change to be.`

Clean and straight.

If you like to keep this link to the forum or leave more context you can extend it ex 

```
###############################
# Use a temporary file to make the changes and copy them back to where we want
# the change to be, as we can't use -i to do inline replacement because this breaks 
# due to permission errors on MacOS
#
# https://forums.docker.com/t/sed-couldnt-open-temporary-file-xyz-permission-denied-when-using-virtiofs/125473
###############################
```

What do you think about it @VanessaPC?

btw. We can slightly improve this comment by limiting lines' length to about 80 chars.

packages/e2e/local-network/scripts/make-babbage.sh
```diff
@@ -273,14 +275,24 @@ mkdir -p ./config/network/genesis/
...
-$SED -i -E "s/\"ShelleyGenesisHash\": \".*\"/\"ShelleyGenesisHash\": \"${shelleyGenesisHash}\"/" ./config/network/cardano-node/config.json
-$SED -i -E "s/\"AlonzoGenesisHash\": \".*\"/\"AlonzoGenesisHash\": \"${alonzoGenesisHash}\"/" ./config/network/cardano-node/config.json
+###############################
+# NOTE: We can't use -i to do inline replacement because this breaks due to permission errors on MacOS when creating the temp output file in this case, giving Root permissions to Docker is also not 
+# a valid working option for MacOS.
+# https://forums.docker.com/t/sed-couldnt-open-temporary-file-xyz-permission-denied-when-using-virtiofs/125473
+###############################
+
+$SED -E "s/\"ByronGenesisHash\": \".*\"/\"ByronGenesisHash\": \"${byronGenesisHash}\"/" ./config/network/cardano-node/config.json > ./config/network/cardano-node/config.json.tmp
+mv ./config/network/cardano-node/config.json.tmp ./config/network/cardano-node/config.json
```
pczeglik-iohk, line 285: I would recommend inserting blank lines to group related commands, as it improves code's readability :)

```
$SED -E "s ...
mv ./config/network ...

$SED -E "s ...
mv ./config/network ...

$SED -E "s ...
mv ./config/network ...
```

pczeglik-iohk (commented):
Hey @VanessaPC I left two comments for you to consider. 

I think it may improve other devs experience when they will be reading the code in the future and try to follow what is happening here and why :)

Please let me know what do you think.

packages/e2e/local-network/scripts/make-babbage.sh
VanessaPC: Yeah, that sounds a good idea, I'll do the changes, thank you

packages/e2e/local-network/scripts/make-babbage.sh
```diff
@@ -11,6 +11,8 @@ root_path="$(cd "$here/.." && pwd)"
 cd "$root_path"
 
 export PATH=$PWD/bin:$PATH
+export CARDANO_NODE_CHAINDB_LOG_LEVEL="Warning"
```
iccicci, line 14: This makes the setting _hard coded_ in the container image, it can't be overridden through environment variables.

```suggestion
export CARDANO_NODE_CHAINDB_LOG_LEVEL=${CARDANO_NODE_CHAINDB_LOG_LEVEL:-Warning}
```

iccicci (changes requested):

packages/e2e/local-network/scripts/make-babbage.sh
michalrus: Agreed – it was a `warning` before my change, which broke the node, I just changed the case to `Warning` to make it work

michalrus: In that case, the error is not very informative, no stack trace, no anything, just a failed `read`:
```
cardano-node: Prelude.read: no parse
```
Reported [here](https://input-output-rnd.slack.com/archives/CR599HMFX/p1697107953967799).

iccicci: Sorry for my initial comment not specifying this, but the next line is affected by the same problem

PR #934, opened 2023-09-26
build/LW-7438 run lint on build only if required

iccicci:
# Context

Both `yarn build` and `yarn test` run `yarn build:version` which runs `yarn lint:fix` which takes about a couple of minutes.
Actually `yarn lint:fix` needs to run in this context only when the change reflects in a API version change.

# Proposed Solution

Changed the build script to check if is there a change which requires `yarn lint:fix` to run in order to actually run it.

Applied this change in my local branch a couple of hours ago and it incredibly speed up my job; so I thought this needs to be shared with the rest of the team ASAP.

### DIFF
--- a/packages/cardano-services-client/package.json
+++ b/packages/cardano-services-client/package.json
@@ -22,7 +22,7 @@
   "scripts": {
     "build:esm": "tsc -p src/tsconfig.json --outDir ./dist/esm --module es2020",
     "build:cjs": "tsc --build src",
-    "build:version": "node ./scripts/createVersionSource.js && yarn lint:fix && shx cp ./package.json ./dist/package.json",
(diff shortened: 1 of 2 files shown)


rhyslbw:
Test @ginnun

rhyslbw (commented):
Thanks for looking at this optimisation @iccicci. Did you consider manipulating the format of the file in the write step instead, so it doesn't need linting?

iccicci:
> Thanks for looking at this optimisation @iccicci. Did you consider manipulating the format of the file in the write step instead, so it doesn't need linting?

Nope @rhyslbw , I just had this idea, it taken a blink of an eye to be implemented and I opened the PR.

mkazlauskas (commented):
Thanks @iccicci, that will help a lot. Can we use `shx` instead of bash for the script so that it works on windows?

rhyslbw (approved):

PR #936, opened 2023-09-28
Generate drepPubKey when creating keyAgent

mchappell:
# Context

Currently the drepPubKey is only created at the point of having a running keyAgent, change this so it's created immediately accessible after creation.

# Proposed Solution

Generate the drepPubKey immediately on keyAgent creation, so that it may be accessed by DApps without the need for an active keyAgent


packages/wallet/src/PersonalWallet/PersonalWallet.ts
```diff
@@ -225,6 +225,7 @@ export class PersonalWallet implements ObservableWallet {
   readonly handleProvider: HandleProvider;
   readonly changeAddressResolver: ChangeAddressResolver;
   readonly publicStakeKeys$: TrackerSubject<PubStakeKeyAndStatus[]>;
+  drepPubKey: Ed25519PublicKeyHex;
```
mirceahasegan, line 228: let's make this private to avoid having it modified by external means.

packages/wallet/src/PersonalWallet/PersonalWallet.ts
```diff
@@ -707,7 +712,7 @@ export class PersonalWallet implements ObservableWallet {
     );
   }
 
-  async getPubDRepKey(): Promise<Ed25519PublicKeyHex> {
-    return this.keyAgent.derivePublicKey(keyManagementUtil.DREP_KEY_DERIVATION_PATH);
+  getPubDRepKey(): Ed25519PublicKeyHex {
```
mirceahasegan, line 715: This should be `async`, otherwise it won't work with remoteApi

packages/wallet/src/PersonalWallet/PersonalWallet.ts
```diff
@@ -518,6 +518,11 @@ export class PersonalWallet implements ObservableWallet {
       utxo: this.utxo
     });
 
+    this.keyAgent
+      .derivePublicKey(keyManagementUtil.DREP_KEY_DERIVATION_PATH)
+      .then((ed25519PublicKeyHex) => (this.drepPubKey = ed25519PublicKeyHex))
```
mirceahasegan, line 523: If the derive call fails it's pretty bad because there's no retry mechanism in place.
We could retry when user calls `getPubDrepKey` so we can return the drep key or an error.
```diff
-  this.keyAgent.derivePublicKey
+ this.getPubDrepKey()
// ....
+ async getPubDRepKey(): Ed25519PublicKeyHex {
+   if (!this.drepPubKey) {
+     this.drepPubKey = await this.keyAgent.derivePublicKey(...)
+   }
+  return this.drepPubKey;
+ }
```

packages/wallet/src/PersonalWallet/PersonalWallet.ts
mchappell: done in [274e020](https://github.com/input-output-hk/cardano-js-sdk/pull/936/commits/274e0201bc9165a19d955ff31a8c41e6bebc0f59)

packages/wallet/src/PersonalWallet/PersonalWallet.ts
mchappell: done in [274e020](https://github.com/input-output-hk/cardano-js-sdk/pull/936/commits/274e0201bc9165a19d955ff31a8c41e6bebc0f59)

packages/wallet/src/PersonalWallet/PersonalWallet.ts
mchappell: added in [2aedd12](https://github.com/input-output-hk/cardano-js-sdk/pull/936/commits/2aedd125d87e57aa01d82c17366f19ea9e910929)

rhyslbw: Please review @mirceahasegan, then @mchappell please squash and rebase, prior to my review

mirceahasegan (approved):

mirceahasegan (approved):

rhyslbw (approved):

PR #945, opened 2023-10-05
LW-8113 add Asset projection

VanessaPC:
# Context
We want to create a new `Asset` projector from the TypeOrmProjector. 
We want the projector to only store UTXO's that has assets in them. 
We need to set up the projector in the docker infrastructure.

# Proposed Solution
We are adding the set-up to filter the UTXO's in the files that affect the `projection` package and `projection typeorm`. Setting up the docker compose for the Asset projector and provider. 
Making sure the `asset` db exists. 
 
DB showing asset and asset test db. 
Screenshot 2023-10-05 at 09 12 42

Projection and provider running in `yarn preprod:up asset-projector asset-provider-server`  (the screenshot shows starting, because I had restarted but it starts like the other ones and its healthy.
Screenshot 2023-10-04 at 15 28 10


compose/common.yml
```diff
@@ -262,6 +287,20 @@ services:
     ports:
       - ${STAKE_POOL_API_PORT:-4010}:3000
 
+  asset-provider-server:
```
iccicci, line 290: The `handle-provider-server` in _local-testnet_ (`packages/e2e/docker-compose.yml`) has following configuration:
```
    TOKEN_METADATA_SERVER_URL: stub://
```

I'm not sure about it, but I guess `asset-provider-server` requires the same setting.
Could you please confirm @mkazlauskas ?

iccicci (changes requested):
Apart from a doubt, very good.

compose/common.yml
VanessaPC: Yea.. I was looking at that this afternoon, I think the `asset-projector` also needs configuration in that file.

mkazlauskas: Asset **projector** shouldn't need it, but **provider** does

packages/projection/test/operators/Mappers/withUtxo.test.ts
```diff
@@ -57,7 +58,16 @@ export const validTxSource$ = of({
               address:
                 'addr_test1qzrf8t56qhzcp2chrtn7deqhep0dttr3eemhnut6lth3gulj7cuplfarmnq5fyumgl0lklddvau9dhamaexykljzvpyswqt56p',
               value: {
-                assets: new Map(),
+                assets: new Map([
```
mkazlauskas, line 61: It would be better to keep this output as-is and add a new one instead. This way we can test that we actually filter the outputs. Right now, not applying `filterProducedUtxoByAssetsPresence()` would probably still pass the test.

compose/common.yml
VanessaPC: It should be addressed here: 91f41e5fffc8a29cddbb82bf555f97492aa0e51b

packages/projection-typeorm/test/operators/storeHandles/util.ts
```diff
@@ -100,13 +100,18 @@ const applyMappers = (evt$: Observable<ProjectionEvent<{}>>) =>
     Mappers.withUtxo(),
     Mappers.withMint(),
     Mappers.withAddresses(),
+    Mappers.filterProducedUtxoByAssetsPresence(),
```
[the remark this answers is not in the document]
VanessaPC: addressed here: 69d8eadfc8d48433777d2ceae53f1a4bdd624fac

packages/projection/test/operators/Mappers/withUtxo.test.ts
VanessaPC: addressed here: 69d8eadfc8d48433777d2ceae53f1a4bdd624fac

packages/projection/test/operators/Mappers/withUtxo.test.ts
```diff
@@ -207,6 +217,22 @@ describe('withUtxo', () => {
     });
   });
 
+  describe('filterProducedUtxoByAssetsPresence', () => {
+    it('keeps only utxo produced for supplied addresses that contain assets', async () => {
```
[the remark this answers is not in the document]
VanessaPC: addressed here: b4bde72f5e320777663da62af761becb84345dbe

mkazlauskas (approved):
Code changes looks good! :rocket:

Some issues with commits:
- Order seems to be wrong, [feat: add helper to especifically store UTXO's with assets](https://github.com/input-output-hk/cardano-js-sdk/pull/945/commits/100b5a1715e600c85847a780153f098483106eff) should come first as other commits depend on it.
- We use package name as commit scope, `compose` is not a valid commit scope in this repo. If it can't be scoped to 1 package, then omit the scope.

VanessaPC:
@mkazlauskas  The order of commits and docker changes scope are now addressed.

rhyslbw (approved):
Great work @VanessaPC 💪

PR #946, opened 2023-10-06
fix(wallet): delegation tracker now searches for portfolio updates on…

AngelCastilloB:
# Context

The current strategy for finding a delegation portfolio update is by looking at all delegation transactions (transactions with some delegation related certificate) and then checking whether the CIP-17 metadata object is present, however this is not entirely correct, if the user only alters the weights of the portfolio, this doesn't trigger a delegation certificate, because the delegation state of the wallet remains unchanged. So the only way we can detect all portfolio updates is by checking every transaction auxiliary data and looking for the portfolio there.

# Proposed Solution

Delegation tracker now searches all transactions in the transaction history.

### DIFF
--- a/packages/wallet/test/services/ChangeAddress/DynamicChangeAddressResolver.test.ts
+++ b/packages/wallet/test/services/ChangeAddress/DynamicChangeAddressResolver.test.ts
@@ -256,6 +256,90 @@ describe('DynamicChangeAddressResolver', () => {
     ]);
   });
 
+  it('doesnt throw if there are entries with 0% in the portfolio, ', async () => {
+    const changeAddressResolver = new DynamicChangeAddressResolver(
+      knownAddresses$,
+      createMockDelegateTracker(
+        new Map<Cardano.PoolId, DelegatedStake>([
+          [
+            poolId1,
+            {
+              percentage: Percent(0),
+              pool: pool1,
+              rewardAccounts: [rewardAccount_1],
+              stake: 0n
+            }
+          ],
+          [
+            poolId2,
+            {
+              percentage: Percent(0),
+              pool: pool2,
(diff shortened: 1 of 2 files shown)


packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
```diff
@@ -71,19 +71,50 @@ export const certificateTransactionsWithEpochs = (
...
+    }
+
+    return hasCert;
+  });
+};
+
+export const createDelegationPortfolioTracker = (transactions: Observable<Cardano.HydratedTx[]>) =>
+  transactions.pipe(
+    map((hydratedTxs) => {
+      const sortedTransactions = [...hydratedTxs].sort((lhs, rhs) => rhs.blockHeader.slot - lhs.blockHeader.slot);
```
mkazlauskas, line 97: It's already sorted, so here's a possible optimization. Or even just a `for` going in reverse
```suggestion
      const sortedTransactions = [...hydratedTxs].reverse();
```

packages/wallet/test/services/ChangeAddress/DynamicChangeAddressResolver.test.ts
```diff
@@ -256,6 +256,90 @@ describe('DynamicChangeAddressResolver', () => {
     ]);
   });
 
+  it('doesnt throw if there are entries with 0% in the portfolio, ', async () => {
```
mkazlauskas, line 259: Nice catch!

mkazlauskas (dismissed):
Great work! :rocket:

rhyslbw (approved):

mkazlauskas (approved):

PR #947, opened 2023-10-13
feat: no stability window buffer til volatile

mkazlauskas:
# Context

Projector is inserting block data into the database even when it's far away from the tip and there is no chance of a fork/rollback happening.

LW-6953

# Proposed Solution

Projector no longer inserts block data until it reaches `tip-securityParameter` block height. This is implemented by extracting Tip tracking capabilities out of StabilityWindowBuffer interface.

Preliminary performance testing of utxo projection on my computer resulted in
- ~3x projector speed on empty blocks
- ~2x projector speed on non-empty blocks

See commit log

### DIFF
--- a/packages/cardano-services/src/Program/services/pgboss.ts
+++ b/packages/cardano-services/src/Program/services/pgboss.ts
@@ -6,6 +6,7 @@ import {
   PoolRegistrationEntity,
   PoolRetirementEntity,
   StakePoolEntity,
+  createDataSource,
   createPgBoss,
   isRecoverableTypeormError
 } from '@cardano-sdk/projection-typeorm';
@@ -32,14 +33,13 @@ import { Pool } from 'pg';
 import { Router } from 'express';
 import { StakePoolMetadataProgramOptions } from '../options/stakePoolMetadata';
 import { contextLogger } from '@cardano-sdk/util';
-import { createObservableDataSource } from '../../Projection/createTypeormProjection';
 import { retryBackoff } from 'backoff-rxjs';
 import PgBoss from 'pg-boss';
 
 /**
  * The entities required by the job handlers
  */
-export const pgBossEntities = [
+export const pgBossEntities: Function[] = [
   CurrentPoolMetricsEntity,
   BlockEntity,
   PoolMetadataEntity,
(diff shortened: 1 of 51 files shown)


iccicci (approved):
LGTM!

packages/projection-typeorm/src/createDataSource.ts
```diff
@@ -204,14 +204,14 @@ const releaseConnection =
       try {
         await connection.queryRunner.rollbackTransaction();
       } catch (error) {
-        logger.error('Failed to rollback transaction', error);
+        logger.warn('Failed to rollback transaction', error);
```
rhyslbw, line 207: Why is this now a warning?

packages/cardano-services/src/Program/services/pgboss.ts
```diff
@@ -49,13 +49,28 @@ export const pgBossEntities = [
...
 
 export const createPgBossDataSource = (connectionConfig$: Observable<PgConnectionConfig>, logger: Logger) =>
-  createObservableDataSource({
-    connectionConfig$,
-    entities: pgBossEntities,
-    extensions: {},
-    logger,
-    migrationsRun: false
-  });
+  // TODO: use createObservableDataSource from projection-typeorm package.
```
rhyslbw, line 52: Is this captured as a separate task?

packages/cardano-services/src/Program/services/pgboss.ts
```diff
@@ -49,13 +49,28 @@ export const pgBossEntities = [
...
+  // A challenge in doing that is that we call subscriber.error on retryable errors in order to reconnect.
+  // Doing that with createObservableDataSource will 'destroy' the data source that's currently used,
+  // so pg-boss is then unable to update job status and it stays 'active', not available for the newly
+  // recreated worker to be picked up.
+  // TODO: this raises another question - what happens when database connection drops while working on a job?
```
rhyslbw, line 57: Do we need to do some testing to better understand?

rhyslbw (commented):
Have you done a performance test @mkazlauskas ? I'm curios to know the gains we're getting here

packages/projection-typeorm/src/createDataSource.ts
mkazlauskas: This is called every time the observable is unsubscribed or errors. In some cases (e.g. legit stopping the projector), it might not have an active transaction. so this is expected to throw.

packages/cardano-services/src/Program/services/pgboss.ts
mkazlauskas: Captured it in LW-8844

packages/cardano-services/src/Program/services/pgboss.ts
mkazlauskas: Yes, captured it in LW-8844

rhyslbw (approved):
Awesome work @mkazlauskas 🧙

PR #948, opened 2023-10-13
LW-7438 refactor Dockerfile to take advantage from cache

iccicci:
# Context

Due to a too early `COPY packages packages` without any exclusion, each `yarn *:up --build` execution requires to run `yarn workspaces focus --all --production`, `yarn --immutable --inline-builds` and `yarn build` even if neither a single file was changed from previous execution.

# Proposed Solution

- Excluded a lot of files through `.dockerignore` both to reduce the size of images and to decrease the probability a full rebuild is required.
- Changed the order of some `RUN` and `COPY` commands to skip not required rebuilds (taking advantage from docker build cache).

- Aggregated some consecutive `RUN` commands into a single one to reduce the number of intermediate images.
- Upgraded the commands to install **Node.js** accordingly to its latest install documentation in order to remove a script deprecation warning (usually not seen thanks to docker build cache).


mkazlauskas (dismissed):
Nice initiative :bow:

Dockerfile
```diff
@@ -3,48 +3,66 @@ ARG UBUNTU_VERSION=20.04
...
+  NODE_OPTIONS=--max_old_space_size=10240 yarn build
 
 FROM nodejs-builder as cardano-services-production-deps
 RUN yarn workspaces focus --all --production
 
 FROM ubuntu-nodejs as cardano-services
 COPY --from=cardano-services-production-deps /app/node_modules /app/node_modules
 COPY --from=cardano-services-production-deps /app/packages/cardano-services/node_modules /app/packages/cardano-services/node_modules
-COPY --from=cardano-services-production-deps /app/packages/projection-typeorm/node_modules /app/packages/projection-typeorm/node_modules
+COPY --from=cardano-services-production-deps /app/packages/core/node_modules /app/packages/core/node_modules
```
mkazlauskas, line 65: Maybe we should copy node_modules from all packages?

mkazlauskas (approved):

mirceahasegan (approved):
great work @iccicci ! 💪

ginnun (approved):
lgtm

Dockerfile
iccicci: Not all packages have production dependencies.

To double check which one of them requires its own `node_modules` directory I've added following command (temporary, only to investigate this).

```
#39 [cardano-services-production-deps 2/2] RUN find . -name node_modules | grep -ve "^\.\/node"
#39 sha256:741d6c81dc0e63565f48b91e07b16a7ac09ef678a787a0f0360f3222af7216a6
#39 0.283 ./packages/cardano-services/node_modules
#39 0.283 ./packages/core/node_modules
#39 0.283 ./packages/e2e/node_modules
#39 0.283 ./packages/web-extension/node_modules
#39 0.283 ./packages/input-selection/node_modules
#39 0.283 ./packages/wallet/node_modules
#39 0.283 ./packages/golden-test-generator/node_modules
#39 DONE 0.3s
```

Here we need to copy only those directories actually required by the backend.

The only one I'm not sure is `input-selection`: could you please confirm it is used only client side?

Dockerfile
```diff
@@ -3,48 +3,66 @@ ARG UBUNTU_VERSION=20.04
...
   echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list &&\
-  apt-get update && apt-get install pkg-config libusb-1.0 libudev-dev gcc g++ make gnupg2 yarn -y
-RUN yarn global add node-gyp@9.0.0
-RUN mkdir -p /app/packages
+  apt-get update && apt-get install pkg-config libusb-1.0 libudev-dev gcc g++ make gnupg2 yarn -y &&\
+  yarn global add node-gyp@9.0.0
 WORKDIR /app
 COPY build build
-COPY packages packages
+COPY packages/cardano-services/package.json packages/cardano-services/package.json
```
pczeglik-iohk, line 27: Are we interested to copy all packages, or almost all, but some are excluded purposefully?

I'm asking as it may be worth checking if we can replace all those `COPY`s with one line command `COPY packages/*/package.json packages/`

This command will care for every single package, and in return we will have less maintenance of Dockerfile in case of new packages etc.

Dockerfile
mkazlauskas: Yes input-selection is only used client-side. The reason I suggested copying `node_modules` of all packages (or all probably packages that are used in the backend) is brittle - a change in dependencies can result in a broken build.

pczeglik-iohk (approved):

Dockerfile
pczeglik-iohk: Discussed off-line. This is a bit more complicated than it looks. Agreed to merge this as it is.

PR #949, opened 2023-10-16
fix/LW-8833 return active stake from previous epoch insted

iccicci:
# Context

Due to our wrong interpretation of `epoch_stake` population time, the active stake of our `DbSynNetworkInfoProvider.stake` call is wrong.

# Proposed Temporary Solution

Return active stake from previous epoch.

Minor query optimization.


mkazlauskas (approved):

ginnun (approved):
LGTM

PR #950, opened 2023-10-16
Feat/lw 8398 handle delisted pools

ginnun:
# Context

SMASH server is used to de-list stake pools. The list of the delisted pools will change over time. A stake pool can be de-listed or un-delisted at any time. This implies that we must regularly update the complete list of delisted stake pools.

With this PR, we can schedule a worker job to regularly fetch delisted pools from SMASH and record them in our database. Delisted stake pools are discarded in our stake pool responses. 

# Proposed Solution

Two commits are made. 
-  Created a new database entity and it's migrations
-  Implemented scheduling capability using pgboss
- Schedules can be configured via a provided json file


packages/cardano-services/src/Projection/migrations/1695899010515-pool-delisted-table.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`CREATE TABLE "pool_delisted"
+                             (
+
+                                 "stake_pool_id" character(56)           NOT NULL,
+                                 CONSTRAINT "PK_pool_delisted_stake_pool_id" PRIMARY KEY ("stake_pool_id")
+                             )`);
+
+    await queryRunner.query(`ALTER TABLE "pool_delisted"
```
mkazlauskas, line 15: Was this generated or did you add it manually? I noticed PoolDelistedEntity doesn't have `@OneToOne` so I thought it wouldn't generate the foreign key

packages/cardano-services/src/Projection/prepareTypeormProjection.ts
```diff
@@ -146,7 +148,7 @@ const storeEntities: Partial<Record<StoreName, EntityName[]>> = {
   storePoolMetricsUpdateJob: ['stakePool', 'currentPoolMetrics', 'poolMetadata'],
   storeStakeKeyRegistrations: ['block', 'stakeKeyRegistration'],
   storeStakePoolMetadataJob: ['stakePool', 'currentPoolMetrics', 'poolMetadata'],
-  storeStakePools: ['stakePool', 'currentPoolMetrics', 'poolMetadata'],
+  storeStakePools: ['stakePool', 'currentPoolMetrics', 'poolMetadata', 'poolDelisted'],
```
mkazlauskas, line 151: This will work fine, but it's not semantically correct: `storeStakePools` can work fine without `poolDelisted` entity. There's no other way to set it up right now, so this is fine, but maybe let's leave a comment.

```suggestion
  // 'stake-pool' projection requires it, but `storeStakePools` does not.
  // at the time of writing there was no way to specify a direct projection->entity dependency.
  storeStakePools: ['stakePool', 'currentPoolMetrics', 'poolMetadata', 'poolDelisted'],
```

packages/cardano-services/test/StakePool/TypeormStakePoolProvider/TypeormStakePoolProvider.test.ts
```diff
@@ -264,6 +264,27 @@ describe('TypeormStakePoolProvider', () => {
           expect(response.totalResultCount).toEqual(2);
         });
 
+        it('response is an array of stake pools excluding delisted pool', async () => {
```
mkazlauskas, line 267: How does this test work? It seems like it's querying 2 pools by id, then expects 2 pools in the response. Doesn't seem to be asserting exclusion of delisted pools :thinking:

compose/common.yml
```diff
@@ -164,7 +164,7 @@ services:
     environment:
       <<: *sdk-environment
       API_URL: http://0.0.0.0:3003
-      QUEUES: ${QUEUES:-pool-metadata,pool-metrics}
+      QUEUES: ${QUEUES:-pool-metadata,pool-metrics,,pool-delist-schedule}
```
mkazlauskas, line 167: Is this a typo?
```suggestion
      QUEUES: ${QUEUES:-pool-metadata,pool-metrics,pool-delist-schedule}
```

packages/cardano-services/environments/.schedule.preview.json
```diff
@@ -0,0 +1,10 @@
+{
+  "schedules": {
+    "list": [
+      {
+        "queue": "pool-delist-schedule",
+        "cron": "0 * * * *"
```
mkazlauskas, line 6: Is there a reason to have a schedule config per network? They all seem to be the same

packages/cardano-services/src/Program/services/pgboss.ts
```diff
@@ -203,9 +215,27 @@ export class PgBossHttpService extends HttpService {
...
+    _dataSource: DataSource,
+    pgBoss: PgBoss,
+    schedule: ScheduleConfig,
+    _db: Pool
+  ): Observable<unknown> {
+    const { queue, cron, data, scheduleOptions } = schedule;
+    const logger = contextLogger(this.logger, queue);
+
+    return new Observable((subscriber) => {
+      pgBoss.schedule(queue, cron, data, scheduleOptions).catch((error: unknown) => {
```
mkazlauskas, line 234: Will this update the schedule if `cron` changes?

packages/cardano-services/test/PgBoss/stakePoolBatchDelistHandler.test.ts
```diff
@@ -0,0 +1,53 @@
...
+describe('stakePoolBatchDelistHandler', () => {
+  let dataSource: DataSource;
+
+  beforeAll(async () => {
+    const testData = await initHandlerTest();
+    dataSource = testData.dataSource;
+  });
+  it('contain latest delisted pools only', async () => {
+    const expected = [
+      'pool1                                                   ',
```
mkazlauskas, line 20: Took a while to understand, probably worth a comment :sweat_smile: 
```suggestion
      // Padded by spaces because pool id length is fixed
      'pool1                                                   ',
```

packages/cardano-services/test/cli.test.ts
```diff
@@ -2671,6 +2671,37 @@ describe('CLI', () => {
...
+            '--queues',
+            'pool-metadata',
+            '--postgres-connection-string-db-sync',
+            postgresConnectionString,
+            '--postgres-connection-string-stake-pool',
+            postgresConnectionStringStakePool,
+            '--stake-pool-provider-url',
+            'http://localhost:4000/stake-pool',
+            '--schedules',
+            'does_not_exist'
```
mkazlauskas, line 2691: I'm not convinced on using file to specify schedules, I think it makes it more difficult to use, as all of the other options are just direct arguments. Let's consider `--pool-delist-schedule '0 * * * *'`. Thoughts @ginnun @iccicci ?

mkazlauskas (changes requested):

packages/cardano-services/src/Projection/migrations/1695899010515-pool-delisted-table.ts
ginnun: It was generated, but later on, I edited the entity file, and not generated again. I will look into it.

mkazlauskas: I think we should keep the foreign key. Just saw this as a mismatch. We probably just need to re-add a decorator for the relation.

packages/cardano-services/environments/.schedule.preview.json
ginnun: For the given task of delisting, they are the same. But they might become different in time, and there will be for sure more schedules that we can't anticipate right now.
What I can do is, I can step down to a single file and once we need more, we can have more.

mkazlauskas: Let's consider https://github.com/input-output-hk/cardano-js-sdk/pull/950#discussion_r1360698371 first, if we decide to keep the files then there's no need to downgrade

packages/cardano-services/test/PgBoss/stakePoolBatchDelistHandler.test.ts
ginnun: Haha, it took a while for me to find out why my tests were failing!

packages/cardano-services/src/Projection/migrations/1695899010515-pool-delisted-table.ts
ginnun: Unfortunately, typeorm doesn't provide an elegant way to do what we want here  (see https://github.com/typeorm/typeorm/issues/4569). Since we don't benefit from having a foreign key, I am gonna remove it.

mkazlauskas: Sure, it's fine since we're re-populating the entire table every time (no use of cascade etc.).

Out of curiosity, why didn't you want to use a typeorm relation?

ginnun: Having the relation is not needed and doesn't add any value. It only adds an extra layer/complexity in my opinion.

packages/cardano-services/src/cli.ts
```diff
@@ -67,7 +67,9 @@ import { EPOCH_POLL_INTERVAL_DEFAULT } from './util';
 import { HttpServer } from './Http';
 import { PgBossQueue, isValidQueue } from './PgBoss';
 import { ProjectionName } from './Projection';
+import { URL } from 'url';
```
iccicci, line 70: As far as I remember the SDK often (always?) uses the built-in `URL` type and in some PRs I had to remove the imported `URL` type as it was causing some problems (sorry I can't remember about details).

I would remove this line; everything should work as it does with this line.

packages/cardano-services/src/cli.ts
```diff
@@ -456,7 +458,17 @@ addOptions(
       for (const queue of queuesArray) if (!isValidQueue(queue)) throw new Error(`Unknown queue name: '${queue}'`);
 
       return queuesArray;
-    }).makeOptionMandatory()
+    }).makeOptionMandatory(),
+    newOption(
+      '--schedules <schedules>',
+      'temp',
```
iccicci, line 464: `'temp'` 😄 
It happens... that's why we perform code review 🤝

packages/cardano-services/src/util/schedule.ts
```diff
@@ -0,0 +1,29 @@
...
+};
+type RawSchedulesConfig = {
+  schedules: {
+    sharedScheduleOptions: {};
+    list: Array<ScheduleConfig>;
+  };
+};
+export const readScheduleConfig = (filePath: string): Array<ScheduleConfig> => {
+  try {
+    const { schedules } = JSON.parse(fs.readFileSync(filePath, 'utf-8')) as RawSchedulesConfig;
```
iccicci, line 17: As per my experience, the use synchronous `fs` functions is fine in scripts (build scripts and so on..) while in the services I would use asynchronous version of the same functions.

Since applying the change I'm suggesting implies a `readScheduleConfig` signature change with consequential changes where it is called, this may require a considerable amount of time. Please raise this in next team meetings to check together if applying this change actually worth.

packages/cardano-services/test/cli.test.ts
iccicci: Personally I don't like so much combining configurations through files and environment variables, but I understand and agree your point @ginnun .

I would vote to keep this as is, once we have some other idea we can improve this.

packages/cardano-services/test/jest-setup/snapshots/stake_pool.sql
```diff
@@ -2655,6 +2667,15 @@ pool1m7y2gakwewqyaz05tvmqqwvl2h4jj2qxtknkzew0dhtuwuupreq	activating	191820000000
 \.
 
 
+--
+-- Data for Name: pool_delisted; Type: TABLE DATA; Schema: public; Owner: postgres
+--
+
+COPY public.pool_delisted (stake_pool_id) FROM stdin;
+pool1vj30jr7wn83dzn928qk6fx34h3d3f3cesr47j5ymeumf65wdw9x
```
iccicci, line 2675: How was this record inserted before taking the snapshot?

If this was inserted with some _actions by hands_, next time we generate new fixtures, we will loose this.

iccicci (changes requested):
Good job!

I have some minor comment, the main one is the one about fixtures.

P.S.
Which is the meaning of the newly added file `/sil.txt`?

packages/cardano-services/test/jest-setup/snapshots/stake_pool.sql
ginnun: I manually insert it. Don't the fixtures run on top of what is already in the database?

packages/cardano-services/test/cli.test.ts
ginnun: In my opinion for the rest of the configurations, we should switch to simple key/value program config files, which could be overridden by an environment or CLI param.

packages/cardano-services/src/util/schedule.ts
ginnun: I am not sure if I understood the suggestion's value. Let's talk at a team meeting and update back here.

ginnun:
> P.S. Which is the meaning of the newly added file `/sil.txt`?
sil->delete . It's a part of my hidden agenda of advertising Turkish language.

packages/cardano-services/src/PgBoss/stakePoolBatchDelistHandler.ts
```diff
@@ -0,0 +1,42 @@
...
+  const { logger, smashUrl, dataSource } = options;
+
+  const service = createService(smashUrl);
+  const repo = dataSource.getRepository(PoolDelistedEntity);
+
+  return async (_data: unknown) => {
+    logger.info(`Getting list of delisted pools from ${smashUrl}`);
+    const delists = await service.getDelistedStakePoolIds();
+
+    if (delists instanceof CustomError) {
```
pczeglik-iohk, line 22: I like this pattern. Returning error instead of throwing it, sounds more granular, explicit.

Recently I red this [article](https://medium.com/javascript-in-plain-english/stop-using-try-catch-and-use-this-instead-2742b51cf97b) on Medium

packages/cardano-services/src/PgBoss/stakePoolBatchDelistHandler.ts
```diff
@@ -0,0 +1,42 @@
...
+  return createSmashStakePoolDelistedService(smashUrl);
+};
+export const stakePoolBatchDelistHandlerFactory: WorkerHandlerFactory = (options) => {
+  const { logger, smashUrl, dataSource } = options;
+
+  const service = createService(smashUrl);
+  const repo = dataSource.getRepository(PoolDelistedEntity);
+
+  return async (_data: unknown) => {
+    logger.info(`Getting list of delisted pools from ${smashUrl}`);
```
pczeglik-iohk, line 19: In this factory I see 6 `logger` calls. Each contains `stake`, but not this one. 
This one contains only `pools`. It is really low issue, but we can improve it before merging. 

In the end we will have more cohesive logs :)

packages/cardano-services/src/cli.ts
ginnun: Thanks! Fixed in c91247ca2

packages/cardano-services/test/jest-setup/snapshots/stake_pool.sql
ginnun: Removed in c91247ca2

packages/cardano-services/src/PgBoss/stakePoolBatchDelistHandler.ts
ginnun: Nooo! I can't let it ruin our logs! I solved it in c4f1525

packages/cardano-services/src/PgBoss/stakePoolBatchDelistHandler.ts
mkazlauskas: I have mixed feelings about this: on one hand errors are very common in JS, you have to build abstraction layers for everything to not have to worry that something would throw. So it's difficult to get away from, might as well embrace it :sweat_smile: 

On the other hand, when using TypeScript, returning errors has a clear advantage that it forces documenting the errors (as types) as well as handling them.

A couple other simple ways to do that:
```ts
// tuples are simple to implement, but you can technically return `[undefined, undefined]`
async getDelistedStakePoolIds(): Promise<[CustomError | undefined, Array<string> | undefined]>

// a little more preparation but more a more accurate representation of what we're trying to do
type Ok<R> = { ok: true; result: R };
type Err<E> = { ok: false; error: E };
type OkOrErr<T, E> = Ok<R> | Err<E>;
async getDelistedStakePoolIds(): Promise<OkOrErr<Array<string>, CustomError>>
```

packages/cardano-services/src/util/schedule.ts
iccicci: Being this called by argument parser (which requires synchronous functions) I'm fine keeping this as is

packages/cardano-services/src/cli.ts
mkazlauskas: Agree, it is available as global in all target environments https://developer.mozilla.org/en-US/docs/Web/API/URL

ginnun: I had no problems importing it. Removed here 9b2db5b

iccicci (approved):

mkazlauskas (approved):
Great work! :rocket:

PR #951, opened 2023-10-17
fix(hardware-trezor): Trezor connect web import LW-6514

DominikGuzei:
# Context

This PR fixes an issue related to importing the `@trezor/connect-web` package. The previous version (seemingly correct) causes an import issue in Lace.

### DIFF
--- a/packages/hardware-trezor/src/TrezorKeyAgent.ts
+++ b/packages/hardware-trezor/src/TrezorKeyAgent.ts
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import * as Crypto from '@cardano-sdk/crypto';
 import * as Trezor from '@trezor/connect';
+import * as TrezorWeb from '@trezor/connect-web';
 import { Cardano, NotImplementedError } from '@cardano-sdk/core';
 import {
   CardanoKeyConst,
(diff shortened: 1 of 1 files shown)


packages/hardware-trezor/src/TrezorKeyAgent.ts
```diff
@@ -14,9 +15,9 @@ import {
   errors
 } from '@cardano-sdk/key-management';
 import { txToTrezor } from './transformers/tx';
-import TrezorConnectWeb from '@trezor/connect-web';
 
 const TrezorConnectNode = Trezor.default;
```
marcin-mazurek, line 19: Do you know why is this weird import style needed? Is it worth leaving a comment?

DominikGuzei: @marcin-mazurek no, i think nobody really knows - @mchappell helped me with this because they had the same weird issue with ledger

marcin-mazurek (approved):

mchappell (approved):
LGTM 👍

packages/hardware-trezor/src/TrezorKeyAgent.ts
mkazlauskas: It should really be equivalent. It indicates that webpack is bundling `cjs` build rather than `esm`. I don't mind merging this, but someone should audit and cleanup lace builds.

mkazlauskas (approved):

DominikGuzei:
@marcin-mazurek @mkazlauskas @mchappell I just realized that this PR (and a new release) is not necessary - the code is fine as it was before and only the import changes in Lace have been relevant to fix the issues we had.

PR #952, opened 2023-10-18
docs: projection

mkazlauskas:
# Context

Projection packages can be difficult to understand.

# Proposed Solution

Add some documentation in READMEs

Delete obsolete `docs/` directory. `cardano-services` package is a better example.


packages/projection/README.md
```diff
@@ -1,20 +1,100 @@
 # Cardano JS SDK | projection
 
-Chain Sync event projection utilities.
+This package is a library of generic projection types and utilities.
```
ginnun, line 3: Wouldn't it be nice if we explained what 'projecting' or 'projection' means in our context? (in a relevant README if not here)

packages/projection/README.md
```diff
@@ -1,20 +1,100 @@
...
 ```
+
+### Other Utils
+
+#### InMemoryStabilityWindowBuffer
+
+If your application doesn't need persistence, you can use in-memory implementation of [StabilityWindowBuffer](#stabilitywindowbuffer), which can also be used as local tip tracker for Bootstrap (provides `Observable<TipOrOrigin>`).
+
+--- 
+*<sup>1</sup>Currently our projection source observable is not a pure 'Producer', as events have a `requestNext` method that is used to control the source. We'll probably be changing how this works (LW-7812).*
```
ginnun, line 98: Is LW-7812 accessible to the public? If not, we might want to say something like:
Maybe we can say:
> This is subject to change in the future.

packages/projection/README.md
```diff
@@ -1,20 +1,100 @@
...
+2. Events include some additional properties: `{eraSumaries, genesisParameters, epochNo, crossEpochBoundary}`.
+3. `RollBackward` events include block data (instead of just specifying the rollback point), and are emitted once for **each** rolled back block.
+
+#### ExtraProps (Generic Parameter)
+
+Source observable can be piped through a series of RxJS operators, which may add some properties to the event. In a nutshell, `ProjectionEvent<T> = ProjectionEvent<{}> & T`. For more detail see [Mappers](#mappers).
+
+### StabilityWindowBuffer
+
+Since `ProjectionEvent{eventType=RollBackward}` includes block data, we must store some number (at least `k`, which is the "security parameter") of latest blocks so that if a fork/rollback happens, we can call `StabilityWindowBuffer.getBlock(hash)` in order to build a valid `ProjectionEvent`.
```
mirceahasegan, line 31: I would rephrase it to emphasize that StabilityWindowBuffer is needed for including block data in ProjectionEvent. As it is now, I understand that ProjectionEvent includes block data, which causes storing some number of blocks.
> "In order to include block data in `ProjectionEvent{eventType=RollBackward}`, we must..."
or
> "Previous blocks data must be stored in order to support including block data in `ProjectionEvent{eventType=RollBackward}`. Therefore, we must store...."

packages/projection/README.md
```diff
@@ -1,20 +1,100 @@
...
+2. Doing something with it (e.g. store into the database). 
+
+Here's an example:
+
+```typescript
+// source$: Observable<ProjectionEvent<{}>>
+source$.pipe(
+  // policyId is a parameter that should be provided
+  // by handle issuer, e.g. ADA Handle
+  Mappers.withHandles({ policyId }),
```
mirceahasegan, line 82: I am tempted to think that "withHandles" also does filtering. But I understand that it does not. If there are no handles for `policyId`, it will still emit the ProjectionEvent, with emtpy `event.handles`.
I my understanding correct?

mirceahasegan (dismissed):
Great work, @mkazlauskas!
It is very helpful to have this documentation. It creates context and direction for anyone wanting to create a new projection 💪 .

packages/projection/README.md
mkazlauskas: Yes, it would still emit an event with empty handles. Filtering is a good idea and it would improve performance quite a lot, but it should be implemented on the application level (cardano-services), because otherwise we'd lose ability to project multiple things

packages/projection/README.md
rhyslbw: Just link this @mkazlauskas https://event-driven.io/en/projections_and_read_models_in_event_driven_architecture/

packages/projection/README.md
mkazlauskas: Done in https://github.com/input-output-hk/cardano-js-sdk/pull/952/commits/e8d9e3ac77b682ce213d9fc817c72f56893acf14

packages/projection/README.md
mkazlauskas: Done in https://github.com/input-output-hk/cardano-js-sdk/pull/952/commits/e8d9e3ac77b682ce213d9fc817c72f56893acf14

ginnun (approved):
lgtm!

rhyslbw (changes requested):
Screenshot 2024-07-26 at 10 33 01

Looks good, just a spelling mistake in the commit message 😬

rhyslbw (approved):

PR #956, opened 2023-10-19
fix(hardware-trezor): trezor tx signing LW-6522

DominikGuzei:
# Context

This PR fixes the Trezor transaction signing which was failing due to an invalid fee calculation.

# Proposed Solution

Filter the Trezor signatures to those with the expected public keys (based on `ownSignatureKeyPaths`), so that the fees calculation (which is based on the number of signatures) is correct.

none


mkazlauskas (approved):

iccicci (approved):
LGTM

PR #958, opened 2023-10-20
feat: add support for only updating outdated pool metrics at a given …

ginnun:
Add support for only updating outdated pool metrics at a given interval.

# Context

In case of an error during the stake pool metrics job, metrics for a given stake pool might get outdated. This change allows us to run in another interval so that only outdated metrics are updated.

# Proposed Solution
We have modified the existing metrics projector and job to handle outdated metrics as well.

None

### DIFF
--- a/packages/cardano-services/src/PgBoss/stakePoolMetricsHandler.ts
+++ b/packages/cardano-services/src/PgBoss/stakePoolMetricsHandler.ts
@@ -5,7 +5,7 @@ import {
   StakePoolEntity,
   StakePoolMetricsUpdateJob
 } from '@cardano-sdk/projection-typeorm';
-import { DataSource } from 'typeorm';
+import { DataSource, LessThan } from 'typeorm';
 import { Logger } from 'ts-log';
 import { ServiceNames } from '../Program/programs/types';
 import { WorkerHandlerFactory } from './types';
@@ -72,6 +72,14 @@ export const refreshPoolMetrics = async (options: RefreshPoolMetricsOptions) => 
   }
 };
 
+export const getPoolIdsToUpdate = async (dataSource: DataSource, outdatedSlot?: Cardano.Slot) =>
+  outdatedSlot
+    ? await dataSource.getRepository(StakePoolEntity).find({
+        select: { id: true },
+        where: [{ metrics: { slot: LessThan(outdatedSlot) } }, { metrics: undefined }]
+      })
+    : await dataSource.getRepository(StakePoolEntity).find({ select: { id: true } });
+
 export const stakePoolMetricsHandlerFactory: WorkerHandlerFactory = (options) => {
   const { dataSource, logger, stakePoolProviderUrl } = options;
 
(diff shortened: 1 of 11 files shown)


packages/cardano-services/src/PgBoss/stakePoolMetricsHandler.ts
```diff
@@ -72,6 +72,14 @@ export const refreshPoolMetrics = async (options: RefreshPoolMetricsOptions) =>
   }
 };
 
+export const getPoolIdsToUpdate = async (dataSource: DataSource, outdatedSlot?: Cardano.Slot) =>
```
pczeglik-iohk, line 75: Should this function provide TSDoc, as it is public (exported from the module)?

Do not know what are recommendations here or good practices.

packages/cardano-services/src/PgBoss/stakePoolMetricsHandler.ts
```diff
@@ -72,6 +72,14 @@ export const refreshPoolMetrics = async (options: RefreshPoolMetricsOptions) =>
...
+  outdatedSlot
```
pczeglik-iohk, line 76: What do you think about such change 👇  I think it reduces cognitive load.

```
const options = {
    select: { id: true },
};

if (outdatedSlot) {
    options['where]: [{ metrics: { slot: LessThan(outdatedSlot) } }, { metrics: undefined }]
}

await dataSource.getRepository(StakePoolEntity).find(options);
```

pczeglik-iohk: instead of `options` we may want to use `query`, `selector` or whatever you think works best :)

ginnun: Nice perspective. Unfortunately (or fortunately?), our type-safety constraints don't allow us to assign `options['where]` since it will have any type. The only way I could do it is to define a type and then use it, which is an overkill.

In addition, I think the provided suggestion is not favorable for me since it operates on an already initialized object in a side-effect kind of way.

Another doable implementation would be the below one, but it's not nicer:
```
export const getPoolIdsToUpdate = async (dataSource: DataSource, outdatedSlot?: Cardano.Slot) => {
  const options = outdatedSlot
    ? {
        select: { id: true },
        where: [{ metrics: { slot: LessThan(outdatedSlot) } }, { metrics: undefined }]
      }
    : { select: { id: true } };
  return await dataSource.getRepository(StakePoolEntity).find(options);
};
```

packages/cardano-services/src/PgBoss/stakePoolMetricsHandler.ts
ginnun: This function is only exposed for testing purposes. In our project, we are using index.ts files as an importing point. So no one should be directly 'importing' and using `getPoolIdsToUpdate`. Besides, I think the function name is explanatory enough for such a simple internal implementation. I could still and comments to explain it if you desire.

I noticed that we have these methods available in the index.ts. Removed them with f47a09fc4f

packages/cardano-services/src/PgBoss/stakePoolMetricsHandler.ts
iccicci: My 2 cents...

```
export const getPoolIdsToUpdate = (dataSource: DataSource, outdatedSlot?: Cardano.Slot) =>
  dataSource.getRepository(StakePoolEntity).find({
    select: { id: true },
    ...(outdatedSlot ? { where: [{ metrics: { slot: LessThan(outdatedSlot) } }, { metrics: undefined }] } : undefined)
  });
```

iccicci (approved):
LGTM!

packages/cardano-services/src/PgBoss/stakePoolMetricsHandler.ts
ginnun: @iccicci This one-liner is also nice, but not worth the change. Also, it's a little bit harder to read.

pczeglik-iohk (approved):

PR #959, opened 2023-10-20
chore: LW-7438 enforce single line jsdoc when possible

iccicci:
# Context

We have lots of multiline jsdoc comments which actually can fit a single line. Changing them in single line will make our code a bit more compact (when possible) increasing a bit its readability.

# Proposed Solution

Added  `"jsdoc/multiline-blocks": ["error", { "minimumLengthForMultiline": 120, "noMultilineBlocks": true }]` eslint rule and fixed all the files.


mkazlauskas (approved):

pczeglik-iohk:
Thank you @iccicci for preparing this PR.

I appreciate consistency in code, and this change enhances it, particularly since you've adjusted the ESLint config as well.

I would like to hear what other engineers think about this change, as it is about code style, and maybe there are some people who favors multiline.

mirceahasegan (approved):
I agree. 🚀

pczeglik-iohk (approved):

PR #966, opened 2023-10-27
[LW-8695] Update Nft metadata and asset provider tests

VanessaPC:
# Context
Update Asset provider and NFT metadata test to use Asset db. 

# Proposed Solution


pczeglik-iohk (approved):
It looks good to me. Thank you

packages/cardano-services/test/cli.test.ts
```diff
@@ -1562,6 +1564,8 @@ describe('CLI', () => {
               callCliAndAssertExit(
                 {
                   args: [
+                    '--postgres-connection-string-asset',
+                    postgresConnectionStringAsset,
```
mkazlauskas, line 1568: `tx-submit` provider shouldn't depend on asset database, so this is probably redundant (same with the tx-submit test that uses env variables). It does depend on `handle` database because it's validating handle resolutions before submitting to the node

iccicci (approved):
Great job!

mkazlauskas (changes requested):
Some notes for commit messages:
- for consistency in changelogs, please use imperative/present tense in commit messages (e.g. update~~s~~ NFT metadata).
- 2655c894fc1cbadb38620653e4e896b5b49631c9 should probably be a `chore:`, because it's not relevant for changelogs
- 354c1e2f5491d6fc3ca50d64e3bd6e5fbcff45d7 should probably be the 1st commit as other commits seem to depend on it

packages/cardano-services/test/cli.test.ts
VanessaPC: addressed in this commit now: 80eb76d12e79aca67a1624cffdcc60efbf5ad819

VanessaPC:
@mkazlauskas  I have addressed the changelog as well.

packages/cardano-services/test/cli.test.ts
```diff
@@ -1590,6 +1592,7 @@ describe('CLI', () => {
                     HANDLE_POLICY_IDS,
                     HANDLE_PROVIDER_SERVER_URL,
                     OGMIOS_SRV_SERVICE_NAME: ogmiosSrvServiceName,
+                    POSTGRES_CONNECTION_STRING_ASSET: postgresConnectionStringAsset,
```
mkazlauskas, line 1595:
```suggestion
```

packages/cardano-services/test/cli.test.ts
```diff
@@ -1640,6 +1643,8 @@ describe('CLI', () => {
               callCliAndAssertExit(
                 {
                   args: [
+                    '--postgres-connection-string-asset',
+                    postgresConnectionStringAsset,
```
mkazlauskas, line 1647:
```suggestion
```

mkazlauskas (changes requested):

packages/cardano-services/test/cli.test.ts
VanessaPC: addressed here: 190929af68712ff574403f8ce843b37a1594c9c9

packages/cardano-services/test/cli.test.ts
VanessaPC: addressed here: 190929af68712ff574403f8ce843b37a1594c9c9

mkazlauskas (approved):
Looks good! Tests seem to be failing though :thinking: re-running...

iccicci (approved):

pczeglik-iohk (approved):

PR #970, opened 2023-10-30
Fix/handle bigint inner error

mirceahasegan:
# Context

Ogmios 6.0 errors can have bigint data in error data.

# Proposed Solution
Parse it accordingly


rhyslbw (approved):

mchappell (approved):
Great work @mirceahasegan!

PR #971, opened 2023-10-31
fix: resilient ogmios v6 tx submission

mkazlauskas:
# Context

OgmiosTxSubmitProvider is bricked whenever websocket connection drops.

# Proposed Solution

Replace `OgmiosTxSubmitProvider` with `ObservableCardanoNode.submitTx` and a new `NodeTxSubmitProvider`. See commit messages for more details.

## Note

I didn't run e2e tests, but they seem to be broken in general in `conway-era` branch. This kind of change would be great to test e2e, but we'll probably have to defer it until later, after this is merged to `conway-era`.

### Follow-up work

Consider removing non-observable variant of `CardanoNode`


packages/cardano-services/test/Program/services/ogmios.test.ts
```diff
@@ -119,9 +118,9 @@ describe.skip('Service dependency abstractions', () => {
           port = await getPort();
           apiUrlBase = `http://localhost:${port}${services.txSubmit.versionPath}/${services.txSubmit.name}`;
           config = { listen: { port } };
-          txSubmitProvider = await getOgmiosTxSubmitProvider(dnsResolver, logger, {
-            ogmiosSrvServiceName: process.env.OGMIOS_SRV_SERVICE_NAME
-          });
+          // txSubmitProvider = await getOgmiosTxSubmitProvider(dnsResolver, logger, {
```
rhyslbw, line 121: Should this be removed or is it here until the tx submit worker has been migrated?

rhyslbw (approved):
Nice work @mkazlauskas, that's a solid improvement

packages/cardano-services/test/Program/services/ogmios.test.ts
mkazlauskas: These tests are all `describe.skip`ed in `conway-era`. There is a separate ticket to rework/re-enable those tests, and I just did some minimal changes here so that it compiles

PR #972, opened 2023-11-02
Cardano node 8.6.0-pre upgrade

mirceahasegan:
# Context
1. Upgraded sanchonet cardano-node version to 8.6.0-pre
2. Upgraded config files subrepo to match the new node version.
3. Re-enable running all services (projector services work with sanchonet versions now).
4. Use sanchonet default versions for all networks: 
  - Running sanchonet and non-sanchonet SDK packages is not possible due to the breaking nature of the changes done in the SDK `cardano-services/ogmios/core` packages.

So, it does not make sense to maintain pre-sanchonet versions.

# Proposed Solution

Use sanchonet version defaults. Once node/ogmios/db-sync/SDK are conway-era stable they will be integrated into master, and used as default versions for all networks.


mchappell (approved):
LGTM

PR #973, opened 2023-11-02
[LW-8984] Accept handles with char matching @

VanessaPC:
# Context
As a phase 1 of the ADA handle support, we didn't support @ symbols, that were used in sub handles. Currently we will be supporting the character in preparation for further support. 

# Proposed Solution
Allow the projection to add/return this symbols 

### DIFF
--- a/packages/core/src/Asset/util/isValidHandle.ts
+++ b/packages/core/src/Asset/util/isValidHandle.ts
@@ -1,4 +1,4 @@
 export const isValidHandle = (handle: string) => {
-  const pattern = /^[\w.-]+$/;
+  const pattern = /^[\w.@\-]+$/;
   return pattern.test(handle);
 };
--- a/packages/core/test/Asset/util/isValidHandle.test.ts
+++ b/packages/core/test/Asset/util/isValidHandle.test.ts
@@ -8,10 +8,12 @@ const validHandles = [
   'test.handle',
   'test-handle-123',
   'test-handle-123.456',
-  'handle_name'
+  'handle_name',
+  '@alice',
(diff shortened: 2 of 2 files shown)


packages/core/src/Asset/util/isValidHandle.ts
```diff
@@ -1,4 +1,4 @@
 export const isValidHandle = (handle: string) => {
-  const pattern = /^[\w.-]+$/;
+  const pattern = /^[\w.-\s@]+$/;
```
iccicci, line 2: Added `\s` as well. Doesn't this means accepting tabs and new lines as well?
I would increase the strength of the tests adding a handle with a space in `invalidHandles` (I guess)

iccicci (changes requested):

gytis-ivaskevicius (dismissed):
Do we have a spec or some definition which lists all supported characters? Maybe there are more edge cases like this

gytis-ivaskevicius:
Actually, isn't it being validated on blockchain already? If that's the case maybe we can remove validation from our side?

packages/core/src/Asset/util/isValidHandle.ts
VanessaPC: you are right, it does support tabs and new lines. Will re do it, and will add to the tests.

VanessaPC:
@gytis-ivaskevicius  we are taking the rules from [here](https://docs.adahandle.com/best-practices). 
@iccicci  we are using the handle for the validation of inputs in Lace.

iccicci (approved):
LGTM!

PR #976, opened 2023-11-06
fix(tx-construction): txBuilder now properly redistributes all outputs to first address after de-registration of all stake keys

AngelCastilloB:
# Context

Currently the transaction builder is not redistributing all outputs back to the first address when de-registering all stake keys.

# Proposed Solution

delegatePortfolio should properly move all funds back to the first address when all stake keys are deregistered.

delegatePortfolio now uses the greedy input selector if the wallet is currently multi delegating and then decidedes to unregister all stake keys.

### DIFF
--- a/packages/tx-construction/test/tx-builder/TxBuilderDelegatePortfolio.test.ts
+++ b/packages/tx-construction/test/tx-builder/TxBuilderDelegatePortfolio.test.ts
@@ -281,6 +281,8 @@ describe('TxBuilder/delegatePortfolio', () => {
 
   describe('pre-existing multi-delegation on all stake keys', () => {
     let txBuilderProviders: jest.Mocked<TxBuilderProviders>;
+    let nonDelegatingWalletTxBuilder: GenericTxBuilder;
+
     beforeEach(async () => {
       const txBuilderFactory = await createTxBuilder({
         keyAgent,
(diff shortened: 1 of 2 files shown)


mkazlauskas (dismissed):
LGTM :rocket:

pczeglik-iohk (approved):
Code looks good. Thanks @AngelCastilloB

packages/tx-construction/test/tx-builder/TxBuilderDelegatePortfolio.test.ts
```diff
@@ -292,9 +294,23 @@ describe('TxBuilder/delegatePortfolio', () => {
       groupedAddresses = txBuilderFactory.groupedAddresses;
       txBuilder = txBuilderFactory.txBuilder;
       txBuilderProviders = txBuilderFactory.txBuilderProviders;
+
+      nonDelegatingWalletTxBuilder = (
+        await createTxBuilder({
+          keyAgent,
+          stakeKeyDelegations: []
```
mirceahasegan, line 301: Could you also include 2 unregistered stake keys to the scenario, so we test that it ignores them and uses random improve?

mirceahasegan (approved):
Great work, @AngelCastilloB ! 🚀

pczeglik-iohk (approved):

PR #977, opened 2023-11-07
feat(wallet): re-fetch cip68 nft metadata

mkazlauskas:
# Context

Currently, PersonalWallet only fetches asset metadata once when it is started. Metadata of some of the CIP-68 NFTs is updated in a transaction made by the owner of the user NFT (e.g. adahandle personalization).

LW-8332

# Proposed Solution

PersonalWallet will re-fetch CIP-68 NFT metadata when it sees an on-chain transaction with a corresponding reference NFT


mirceahasegan (approved):
great job! 🥇

rhyslbw (approved):
Great work!

PR #978, opened 2023-11-07
 LW-8967 db sync chain history conway era certificates

iccicci:
# Context

Conway era introduced new certificates.

# Proposed Solution

Implemented the new certificates in `ChainHistoryBuilder`.

Skipped the two tests files testing `ChainHistoryBuilder`.

### DIFF
--- a/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/mappers.ts
+++ b/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/mappers.ts
@@ -21,13 +21,25 @@ import {
   WithdrawalModel
 } from './types';
 import { Cardano } from '@cardano-sdk/core';
-import { Hash32ByteBase16 } from '@cardano-sdk/crypto';
+import { Hash28ByteBase16, Hash32ByteBase16 } from '@cardano-sdk/crypto';
 import {
+  isAuthorizeCommitteeHotCertModel,
+  // isAuthorizeCommitteeHotCertModel,
   isDelegationCertModel,
+  isDrepRegistrationCertModel,
+  isDrepUnregistrationCertModel,
   isMirCertModel,
   isPoolRegisterCertModel,
   isPoolRetireCertModel,
-  isStakeCertModel
+  isResignCommitteeColdCertModel,
+  // isResignCommitteeColdCertModel,
+  isStakeCertModel,
+  isStakeRegistrationDelegationCertModel,
+  isStakeVoteDelegationCertModel,
+  isStakeVoteRegistrationDelegationCertModel,
+  isUpdateDrepCertModel,
+  isVoteDelegationCertModel,
+  isVoteRegistrationDelegationCertModel
 } from './util';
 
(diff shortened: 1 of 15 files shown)


packages/cardano-services/test/ChainHistory/ChainHistoryHttpService.test.ts
```diff
@@ -33,7 +33,7 @@ const APPLICATION_CBOR = 'application/cbor';
 const APPLICATION_JSON = 'application/json';
 const PAGINATION_PAGE_SIZE_LIMIT = 5;
 
-describe('ChainHistoryHttpService', () => {
+describe.skip('ChainHistoryHttpService', () => {
```
mirceahasegan, line 36: Why are these tests skipped?

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/mappers.ts
```diff
@@ -148,16 +170,17 @@ export const mapCertificate = (
   if (isStakeCertModel(certModel))
     return {
       __typename: certModel.registration
-        ? Cardano.CertificateType.StakeRegistration
-        : Cardano.CertificateType.StakeDeregistration,
+        ? Cardano.CertificateType.Registration
+        : Cardano.CertificateType.Unregistration,
       cert_index: certModel.cert_index,
+      deposit: BigInt(certModel.deposit),
```
mkazlauskas, line 176: Is this available in the db? It's not part of the old certificates

mkazlauskas (approved):
Looks good! Just one question below

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/mappers.ts
iccicci: Yes it is, not directly in the `stake_registration` table (some `JOIN`s with `epoch_parameter` table is required) and this is common both to old and new certificates.
What's missing in the **db-sync** schema is the ability to distinguish a new certificate from an old certificate.
Anyway the old certificate is a subset of the new certificate and they are equivalent (at logical level there is no changes regardless from the certificate has or not an explicit `deposit` attribute in the tx body) so we map both of them to the same Core type.

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/mappers.ts
```diff
@@ -172,6 +195,128 @@ export const mapCertificate = (
...
+      stakeCredential: {
+        hash: Cardano.RewardAccount.toHash(
+          Cardano.RewardAccount(certModel.address)
+        ) as unknown as Crypto.Hash28ByteBase16,
+        type: Cardano.CredentialType.KeyHash
+      }
+    };
+
+  /**
+   * TODO: Can these be returned from db-sync?
```
mchappell, line 315: Can we link to the issue raised against `db-sync` so that we can track it?

iccicci: a6d11ce74f should solve this

mchappell (approved):
Great work @iccicci 🚀

PR #979, opened 2023-11-08
LW-9018 fix unit tests 

mirceahasegan:
# Context

# Proposed Solution

### DIFF
--- a/packages/util/src/serializableObject.ts
+++ b/packages/util/src/serializableObject.ts
@@ -23,6 +23,12 @@ const defaultGetErrorPrototype: GetErrorPrototype = () => Error.prototype;
 const defaultTransformKey: TransformKey = (key) => key;
 const defaultTransformationTypeKey = '__type';
 
+const hasInnerError = (error: unknown): error is { innerError: unknown } =>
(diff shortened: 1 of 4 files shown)


iccicci (approved):
Simply awesome!
Thank you very much!

packages/util/src/serializableObject.ts
```diff
@@ -41,6 +47,11 @@ export const toSerializableObject = (obj: unknown, options: ToSerializableObject
       const value = Buffer.from(arr).toString('hex');
       return { [transformationTypeKey]: 'Buffer', value };
     }
+
+    if (hasInnerError(obj) && innerErrorHasData(obj.innerError)) {
+      obj.innerError.data = toSerializableObject(obj.innerError.data);
```
mkazlauskas, line 52: I think this results in serializing innerError as plain object, which means that it won't get `[transformationTypeKey]: 'Error'` and won't run through `getErrorPrototype`. Maybe we could wrap `serializeError` in a new function that does this recursively.

mkazlauskas (changes requested):

mkazlauskas (approved):
Nice work! :rocket:

PR #984, opened 2023-11-13
Feat/lw 9027 cardano submit api

iccicci:
# Context

We need to implement tx submission through `cardano-submit-api`.

# Proposed Solution

- Implemented `TxSubmitApiProvider` in `cardano-services-client`.
- Implemented tx submission through `TxSubmitApiProvider` in `provider-server`.

# TODO

- A complete error handling.
- `TxSubmitApiProvider` unit tests.


compose/common.yml
```diff
@@ -64,8 +64,10 @@ x-provider-server-environment: &provider-server-environment
   ENABLE_METRICS: ${ENABLE_METRICS:-false}
   EPOCH_POLL_INTERVAL: ${EPOCH_POLL_INTERVAL:-500}
   SERVICE_NAMES: ${SERVICE_NAMES:-asset,chain-history,handle,network-info,rewards,stake-pool,tx-submit,utxo}
+  SUBMIT_API_URL: ${SUBMIT_API_URL:-http://cardano-submit-api:8090/api/submit/tx}
```
rhyslbw, line 67:
```suggestion
  SUBMIT_API_URL: ${SUBMIT_API_URL:-http://cardano-submit-api:8090}
```
No need to expose the internal path of the service, just the root URL

packages/cardano-services-client/src/TxSubmitProvider/TxSubmitApiProvider.ts
```diff
@@ -0,0 +1,47 @@
...
+
+      // TODO improve (implement?) error handling
+      throw axios.isAxiosError(error)
+        ? new ProviderError(ProviderFailure.BadRequest, null, error.response?.data as string)
+        : new ProviderError(ProviderFailure.Unknown, error, 'submitting tx');
+    }
+  }
+
+  healthCheck() {
+    return Promise.resolve({ ok: true });
```
rhyslbw, line 45: I was going to suggest calling a path we know not to exist, and assert a 404, but it's more important to ship something that works ASAP. Maybe we should rethink if this method is even needed 🤔

rhyslbw (changes requested):

packages/cardano-services-client/src/TxSubmitProvider/TxSubmitApiProvider.ts
iccicci: > I was going to suggest calling a path we know not to exist, and assert a 404
I thought the same, but I wasn't able to find a way to perform an ok dry call to get the `ok` status.

compose/common.yml
iccicci: This is the `provider-server` configuration which contains the docker compose internal URL, nothing new is exposed.

rhyslbw: Right, but it's better to just configure the root URL, then add `/api/submit/tx` inside the client

packages/cardano-services-client/src/TxSubmitProvider/TxSubmitApiProvider.ts
mkazlauskas: An improvement over what we currently have is mapping status of last submission: `2xx` or `4xx` is ok, everything else is not ok

mkazlauskas: I don't think it's meaningless: afaik client is not checking health before submitting, but health is used to track backend services

mkazlauskas (changes requested):
Nice work! Just a minor change request below, and also one of the commits scope is `2e2` (should be `e2e`)

packages/cardano-services-client/src/TxSubmitProvider/TxSubmitApiProvider.ts
```diff
@@ -0,0 +1,47 @@
...
+  #path: string;
+
+  constructor(submitApiUrl: URL, logger: Logger) {
+    this.#axios = axios.create({ baseURL: submitApiUrl.origin });
+    this.#logger = logger;
+    this.#path = submitApiUrl.pathname;
+  }
+
+  async submitTx({ signedTransaction }: SubmitTxArgs) {
+    let txId = '' as Cardano.TransactionId;
```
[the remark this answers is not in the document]
iccicci: 10c97ca4c should solve this

packages/cardano-services-client/src/TxSubmitProvider/TxSubmitApiProvider.ts
iccicci: b6dc40411 should solve this

compose/common.yml
iccicci: b6dc40411 and 81d6db18f should solve this

mkazlauskas (approved):
Great work!

mirceahasegan (approved):
Fantastic work, @iccicci 🧙‍♂️ 🚀

rhyslbw (approved):
Great work @iccicci 🚀

PR #985, opened 2023-11-13
chore(hardware-trezor): add trezor tx inline datum and datum hash to outputs mapper

tomislavhoracek:
This PR introduces additional outputs mapper data where we added `inlineDatum` and `datumHash`.

# Context

Add trezor tx inline datum and datum hash to outputs mapper


VanessaPC (approved):
LGTM!

PR #987, opened 2023-11-14
Fix/lw 8951 reject tx submit errors

mirceahasegan:
# Context

cardano-submit-api tx provider does not emit innerError with code and data like ogmios does.
When innerError is missing or is not parsable in to a known txSubmitError, the error is reported as ProviderFailure.Unknown.
Furthermore, SmartTxSubmitProvider is retrying Unknown errors, but shouldn't.

# Proposed Solution

TxSubmit errors without innerError will be mapped to a ProviderFailure based on the response reason.
Update SmartTxSubmitProvider to retry only Unhealthy or ConnectionFailure unreachable errors


iccicci (approved):
LGTM

PR #995, opened 2023-11-20
chore(hardware-trezor): add trezor tx reference inputs and script mappers

tomislavhoracek:
This PR introduces additional data mappers where we added `referenceInputs` and `referenceScript`.
[JIRA Ticket](https://input-output.atlassian.net/browse/LW-9015)

# Context

SDK testing standards were followed when developing this ticket
It will not be tested by SDETs in isolation! (Described in JIRA Ticket)

# Proposed Solution

Add Trezor TX reference inputs and script mappers

New mappers were added. No breaking changes

### DIFF
--- a/packages/hardware-trezor/src/transformers/tx.ts
+++ b/packages/hardware-trezor/src/transformers/tx.ts
@@ -28,6 +28,7 @@ const trezorTxTransformer = async (
     networkId: context.chainId.networkId,
     outputs: mapTxOuts(body.outputs, context),
     protocolMagic: context.chainId.networkMagic,
+    referenceInputs: body.referenceInputs ? await mapTxIns(body.referenceInputs, context) : undefined,
     ttl: body.validityInterval?.invalidHereafter?.toString(),
     validityIntervalStart: body.validityInterval?.invalidBefore?.toString(),
     withdrawals: mapWithdrawals(body.withdrawals ?? [], context)
--- a/packages/hardware-trezor/src/transformers/txOut.ts
+++ b/packages/hardware-trezor/src/transformers/txOut.ts
@@ -1,7 +1,7 @@
 import * as Trezor from '@trezor/connect';
 import { Cardano, Serialization } from '@cardano-sdk/core';
 import { GroupedAddress, util } from '@cardano-sdk/key-management';
-import { InvalidArgumentError, Transform } from '@cardano-sdk/util';
+import { HexBlob, InvalidArgumentError, Transform } from '@cardano-sdk/util';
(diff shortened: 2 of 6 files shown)


packages/hardware-trezor/src/transformers/tx.ts
```diff
@@ -28,6 +28,7 @@ const trezorTxTransformer = async (
     networkId: context.chainId.networkId,
     outputs: mapTxOuts(body.outputs, context),
     protocolMagic: context.chainId.networkMagic,
+    referenceInputs: body.referenceInputs ? mapReferenceInputs(body.referenceInputs) : undefined,
```
mkazlauskas, line 31: Reference inputs seem to have the same format as regular inputs. Is there a reason to duplicate the mapper?
```suggestion
    referenceInputs: body.referenceInputs ? mapTxIns(body.referenceInputs) : undefined,
```

mkazlauskas (changes requested):

packages/hardware-trezor/src/transformers/tx.ts
tomislavhoracek: Thanks @mkazlauskas 👍 
I was also thinking about that. But, then I wanted to isolate it to have a specific output type declaration `CardanoReferenceInput` since `connect/trezor` package is changing a lot and we often face some breaking changes. Also, it is isolated in `hardware-ledger`  (https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/hardware-ledger/src/transformers/referenceInputs.ts) so I thought it is maybe a more correct way to do it. 

I'm happy to change it if you think it is better to use existing inputs mapper...?
cc @AngelCastilloB

mkazlauskas: In CDDL it is the same type, so chance of it diverging is probably low
```
0 : set<transaction_input>    ; inputs
? 13 : set<transaction_input> ; collateral inputs
? 18 : set<transaction_input> ; reference inputs; New
```

tomislavhoracek: 👌 then I will change it. Thanks a lot @mkazlauskas

tomislavhoracek: Changed in: [05495d6](https://github.com/input-output-hk/cardano-js-sdk/pull/995/commits/05495d6d7ed2e19d65a137a92dae0fd409c43d58)

mkazlauskas (approved):
LGTM :rocket:

AngelCastilloB (approved):
LGTM 🚀

PR #996, opened 2023-11-24
feat: wallet repository

mkazlauskas:
# Context

In order to implement multi-wallet/multi-account feature, we need to have a persistent repository of multiple wallets/accounts, so that they are restored when application is reloaded.

# Proposed Solution

See commit log


mirceahasegan (approved):
Great work, @mkazlauskas ! 🚀

PR #997, opened 2023-11-25
Feat/lw 9089 add address manager and witnesser to personal wallet

AngelCastilloB:
# Context

In order to add multi-wallet/account support, we need to re-design wallet manager. We know that we’ll have to implement Shared Wallets after this epic, and the concept of KeyAgent doesn’t work with this type of wallets. Replacing PersonalWallet dependency on KeyAgent with new abstractions now will result in less work overall. 

# Proposed Solution

Replacing PersonalWallet dependency on KeyAgent with new abstractions (AddressManager and Witnesser). 

PersonalWallet now takes an AddressManager and a Witnesser instead of a key agent.

### DIFF
--- a/packages/key-management/src/types.ts
+++ b/packages/key-management/src/types.ts
@@ -201,3 +201,34 @@ export interface TransactionSigner {
    */
   sign(tx: Cardano.TxBodyWithHash): Promise<TransactionSignerResult>;
 }
+
+export type WitnessOptions = SignTransactionOptions;
+
+/** Interface for an entity capable of generating witness data for a transaction. */
+export interface Witnesser {
+  /**
+   * Generates the witness data for a given transaction.
+   *
+   * @param txInternals The transaction body along with its hash for which the witness data is to be generated.
+   * @param options Optional additional parameters that may influence how the witness data is generated.
+   * @returns A promise that resolves to the generated witness data for the transaction.
+   */
+  witness(txInternals: Cardano.TxBodyWithHash, options?: WitnessOptions): Promise<Cardano.Witness>;
+
+  /**
+   * @throws AuthenticationError
+   */
+  signBlob(derivationPath: AccountKeyDerivationPath, blob: HexBlob): Promise<SignBlobResult>;
+}
+
+/** Interface for managing blockchain addresses. */
+export interface AddressManager {
+  knownAddresses$: Observable<GroupedAddress[]>;
+
+  /**
+   * Sets or updates the list of known addresses managed by this instance.
+   *
+   * @param addresses An array of grouped addresses to be managed.
+   */
+  setKnownAddresses(addresses: GroupedAddress[]): Promise<void>;
+}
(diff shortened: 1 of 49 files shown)


packages/key-management/src/types.ts
```diff
@@ -201,3 +201,34 @@ export interface TransactionSigner {
...
+/** Interface for managing blockchain addresses. */
+export interface AddressManager {
+  knownAddresses$: Observable<GroupedAddress[]>;
+
+  /**
+   * Sets or updates the list of known addresses managed by this instance.
+   *
+   * @param addresses An array of grouped addresses to be managed.
+   */
+  setKnownAddresses(addresses: GroupedAddress[]): void;
```
mkazlauskas, line 233: `Bip32Ed25519AddressManager` method is async
```suggestion
  setKnownAddresses(addresses: GroupedAddress[]): Promise<void>;
```

packages/wallet/src/PersonalWallet/PersonalWallet.ts
```diff
@@ -622,7 +626,14 @@ export class PersonalWallet implements ObservableWallet {
   }
 
   signData(props: SignDataProps): Promise<Cip30DataSignature> {
-    return cip8.cip30signData({ keyAgent: this.keyAgent, ...props });
+    return cip8.cip30signData({
+      // TODO: signData probably needs to be refactored out of the wallet and supported as a stand alone util
```
mkazlauskas, line 630: Please capture in a ticket

mkazlauskas (changes requested):

packages/key-management/src/types.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/997/commits/c6958599638a7fd4e0de8fac8dc2048e3f5e8908

packages/wallet/src/PersonalWallet/PersonalWallet.ts
AngelCastilloB: done!

mkazlauskas (approved):
Great work! :rocket:

rhyslbw (approved):
Nice work @AngelCastilloB 🚀

packages/key-management/src/util/createAddressManager.ts
```diff
@@ -0,0 +1,53 @@
...
+  AccountAddressDerivationPath,
+  AccountKeyDerivationPath,
+  AddressManager,
+  AsyncKeyAgent,
+  GroupedAddress
+} from '../types';
+import { Observable } from 'rxjs';
+
+/** An address manager that uses a {@link AsyncKeyAgent} to derive addresses. */
+export class Bip32Ed25519AddressManager implements AddressManager {
```
DominikGuzei, line 12: `AdressManager` looks like a duplicate of `AsyncKeyAgent` or at least it's just using the same methods of the passed key agent?

PR #1004, opened 2023-11-29
refactor(crypto): replaced BN library with custom implementation of big number arithmetic operations

AngelCastilloB:
# Context

During our recent internal security audit of our javascript/typescrip implementation of Bip32 key derivation and Ed25519 cryptography operations over libsodium, several issues were found:

***
**1. Invalid base for all number conversions**

BN initial function is defined as `function init (number, base, endian)` But all current usage in the crypto code set base=16, this is surprising to the reader as parsing from bytes imply base=256.
***

**2. Bug in padding**

The padding used is invalid, resulting in padding never added (unless the number is less than half the size, which should happen only extremely rarely) This code snippet transforms a “32 bytes number” into the hexadecimal string, which should be 64 characters long, In a case of a short number, the padEnd is passed the value 32 characters to pad to, when it should be 64 characters .

```
if (r.length !== 32) {
    r = Buffer.from(r.toString('hex').padEnd(32, '0'), 'hex');
}
```
***

**3. Timing attack doing arithmetic operations**

BN usage to do addition and multiplication is not subject to constant time, unless due care has been applied. So for certain classes of number, this creates a potential timing attack oracle that could result in leaking cryptographic key material.
***

As recommended by the auditor, the BN library was removed and replaced by implementing the operations directly over the little endian uint8array.

A reference implementation for these primitives were provided in Rust:

https://github.com/typed-io/rust-ed25519-bip32/blob/5a14675ecf9c2940054d984943d71a23e604fcfd/src/derivation/v2.rs

# Proposed Solution

- Remove BN library.
- Port the required primitives from the provided Rust implementation to Typescript.
- Update the derivation code to use the new custom primitive rather than the BN library.


vincenthz (approved):
lgtm

PR #1008, opened 2023-12-05
LW-9237 typeorm stake pool provider backward compatibility

iccicci:
# Context

When I introduced `ros` I broken back compatibility; we need to restore it. 

# Proposed Solution

- copy the property `ros` into `apy` one; 
- when sort by `apy` is requested, act as sort by `ros` was requested;

- minor change to `yarn cleanup` to make it faster
- increased a delay in test to make the test more strong

### DIFF
--- a/packages/cardano-services-client/package.json
+++ b/packages/cardano-services-client/package.json
@@ -29,7 +29,7 @@
     "cleanup:dist": "rm -rf dist",
     "circular-deps:check": "madge --circular dist/cjs",
     "cleanup:nm": "rm -rf node_modules",
-    "cleanup": "run-s cleanup:dist cleanup:nm",
+    "cleanup": "rm -rf dist node_modules",
     "lint": "eslint -c ../../complete.eslintrc.js \"src/**/*.ts\" \"test/**/*.ts\"",
     "lint:fix": "yarn lint --fix",
     "test": "yarn build:version && jest -c ./jest.config.js",
--- a/packages/cardano-services/src/StakePool/TypeormStakePoolProvider/util.ts
+++ b/packages/cardano-services/src/StakePool/TypeormStakePoolProvider/util.ts
@@ -3,8 +3,6 @@
 import {
   Cardano,
   FilterCondition,
-  ProviderError,
-  ProviderFailure,
   QueryStakePoolsArgs,
   SortField,
   SortOrder,
@@ -51,7 +49,7 @@ export const stakePoolSearchSelection = [
 ];
 
 export const sortSelectionMap: { [key in SortField]: string } = {
-  apy: 'metrics_apy',
+  apy: 'metrics_ros',
   cost: 'params.cost',
   lastRos: 'metrics_last_ros',
   name: 'metadata.name',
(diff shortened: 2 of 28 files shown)


packages/cardano-services/src/StakePool/TypeormStakePoolProvider/util.ts
```diff
@@ -63,32 +61,14 @@ export const nullsInSort = 'NULLS LAST';
...
-    sort = { field: defaultField, order: defaultOrder };
-  }
-  if (sort.field === 'name') {
-    return {
-      field: `lower(${sortSelectionMap[sort.field as SortField]})`,
-      order: sort.order.toUpperCase() as Uppercase<SortOrder>
-    };
-  }
+export const getSortOptions = (sort?: StakePoolSortOptions) => {
+  if (!sort) sort = { field: 'name', order: 'asc' };
```
pczeglik-iohk, line 65: Do we have any recommendation / convention to keep curly brackets `{...}` even for cases like this ☝️ It's a good practice to always put brackets.

mkazlauskas: We're using eslint for stuff like this. It's currently valid since there's no rule enforcing brackets. If we want to change it, then a rule should be added and changes applied in the entire codebase.

pczeglik-iohk: In line 67 there is `toUpperCase()` for order. Maybe it's good to put here `{ field: 'name', order: 'ASC'}` if we expect this to be uppercase?

packages/cardano-services-client/package.json
```diff
@@ -29,7 +29,7 @@
     "cleanup:dist": "rm -rf dist",
     "circular-deps:check": "madge --circular dist/cjs",
     "cleanup:nm": "rm -rf node_modules",
-    "cleanup": "run-s cleanup:dist cleanup:nm",
+    "cleanup": "rm -rf dist node_modules",
```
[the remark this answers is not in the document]
iccicci: Yes, both of them are used by other scripts

mkazlauskas (commented):
Looks good! I think bumping API version should be treated as a breaking change, so please add a footer in commit message like
```
BREAKING CHANGE: bump stake pool provider api version to 1.1.0
```

packages/cardano-services/src/StakePool/TypeormStakePoolProvider/util.ts
iccicci: 09a2b9584 should partially solve this.

About the case: in input `order` is of type
```typescript
export type SortOrder = 'asc' | 'desc';
```
while the output `order` property will be used in **TypeORM** `orderBy` method which accepts type `"ASC" | "DESC"`; that's why the function needs to make uppercase and why the default value for input should be expressed in lowercase.

mkazlauskas: Disable eslint rule suggests that you should use a const/variable instead of that object literal :)

iccicci: f5e262f89 should fix this

VanessaPC (approved):
LGTM

pczeglik-iohk (approved):

PR #1011, opened 2023-12-06
Fix signed/unsigned bug in Value.fromCBOR

itsfarseen:
It looks like the author meant to do:

    if(reader.peek() === UNSIGNED_INT) { reader.readUInt() }

instead of

    if(reader.peek() === UNSIGNED_INT) { reader.readInt() }

### DIFF
--- a/packages/core/src/Serialization/TransactionBody/Value.ts
+++ b/packages/core/src/Serialization/TransactionBody/Value.ts
@@ -81,7 +81,7 @@ export class Value {
     const reader = new CborReader(cbor);
 
     if (reader.peekState() === CborReaderState.UnsignedInteger) {
-      const coins = reader.readInt();
+      const coins = reader.readUInt();
       return new Value(coins);
     }
 


AngelCastilloB:
Hi @itsfarseen thanks for the feedback, in our typescript CBOR implementation both `readInt` and `readUInt` map to a bigint (because we dont have signed types in javascript and bigint is large enough in both directions), and particularly `readInt` is 'smart' enough to figure out if the number is negative or positive and do the right thing when deserializing, so this wont cause any issues here. 

That being said, the `readUInt` version will throw if it finds a signed number which is probably semantically more correct, so I am happy to accept the change, but please fix the commit message to use conventional commits, I.E:

`fix(core): deserialization of Value now checks that both coin and multiasset amount are unsigned`

> Not sure if the same bug is present in the following lines as well. Coins and quantity can't be negative right?

Correct, you can apply the same change here.

mkazlauskas (changes requested):
Thanks for contribution! A fix should come with a test (or some change in an existing test)

rhyslbw:
@itsfarseen thanks for the PR but we're unable to merge given it's incomplete. Will you be able to address @mkazlauskas [feedback](https://github.com/input-output-hk/cardano-js-sdk/pull/1011#pullrequestreview-1771597716)?

AngelCastilloB:
I will continue this work on https://github.com/input-output-hk/cardano-js-sdk/pull/1482 since this commit is in a fork

PR #1012, opened 2023-12-07
Feat/lw 9092 rework wallet manager

AngelCastilloB:
# Context

Existing WalletManager api supports only a single wallet. It has to be re-designed for multi-wallet support.
This task connects most of the SDK pieces together.

# Proposed Solution

Re-designed WalletManager api to supports multi-wallet.


packages/e2e/test/web-extension/extension/background/walletManager.ts
```diff
@@ -3,22 +3,25 @@ import {
...
-  walletManagerProperties
+  walletManagerProperties,
+  walletRepositoryProperties
 } from '@cardano-sdk/web-extension';
 
-import { AsyncKeyAgent } from '@cardano-sdk/key-management';
 import { storage as WebExtensionStorage, runtime } from 'webextension-polyfill';
-import { env, logger } from '../util';
+import { Witnesser } from '@cardano-sdk/key-management';
+import { env, logger, Metadata } from "../util";
```
mkazlauskas, line 16: Some lint errors here

packages/e2e/test/web-extension/extension/ui.ts
```diff
@@ -198,7 +211,48 @@ const cleanupMultidelegationInfo = (multiDelegationDiv: Element) => {
...
+
+const passphraseByteArray = Uint8Array.from(
+  env.KEY_MANAGEMENT_PARAMS.passphrase.split('').map((letter) => letter.charCodeAt(0))
+);
+merge(signerManager.signDataRequest$, signerManager.transactionWitnessRequest$).subscribe((req) => {
+  logger.info('Sign request', req);
+  if (req.walletType === WalletType.InMemory) {
+    req.sign(new Uint8Array(passphraseByteArray));
+  } else {
+    req.sign();
```
mkazlauskas, line 240: Lint error
```suggestion
    void req.sign(new Uint8Array(passphraseByteArray));
  } else {
    void req.sign();
```

packages/e2e/test/web-extension/extension/ui.ts
```diff
@@ -198,7 +211,48 @@ const cleanupMultidelegationInfo = (multiDelegationDiv: Element) => {
...
+
+// Expose signer manager.
+exposeSignerManagerApi(
+  {
+    signerManager
+  },
+  { logger, runtime }
+);
+
+// TODO: toSerializableObj does not support serializing empty obj {}
```
mkazlauskas, line 253:
```suggestion
```

packages/e2e/test/web-extension/extension/ui.ts
```diff
@@ -264,6 +363,7 @@ document.querySelector(selectors.btnSignAndBuildTx)!.addEventListener('click', a
   const { body } = await builtTx.inspect();
   logger.info('Built tx', body.outputs.length);
   const { tx: signedTx } = await builtTx.sign();
+  logger.error(signedTx);
```
mkazlauskas, line 366:
```suggestion
```

packages/key-management/src/types.ts
```diff
@@ -222,12 +222,13 @@ export interface Witnesser {
...
    *
-   * @param txInternals The transaction body along with its hash for which the witness data is to be generated.
+   * @param transaction The transaction along with its hash for which the witness data is to be generated.
+   * @param context The witness sign transaction context
    * @param options Optional additional parameters that may influence how the witness data is generated.
    * @returns A promise that resolves to the generated witness data for the transaction.
    */
   witness(
-    txInternals: Cardano.TxBodyWithHash,
+    transaction: Serialization.Transaction,
```
mkazlauskas, line 231: Would've been nice to do this refactor in a separate commit, please see if you can do this without too much effort. Otherwise feel free to resolve this change request

packages/web-extension/src/walletManager/walletManagerUi.ts
```diff
@@ -15,11 +14,9 @@ import { observableWalletProperties } from '../observableWallet';
  * requires a unique `walletId`, which is calculated by {@link WalletManagerUi} before passing the message to
  * {@link WalletManagerWorker}. It is not meant to be calculated or maintained by the user.
  */
-export class WalletManagerUi implements Shutdown {
+export class WalletManagerUi<Metadata extends {}> implements Shutdown {
```
mkazlauskas, line 17: Let's delete this as you suggested in DM. Then we can rename WalletManagerWorker to just WalletManager

packages/web-extension/src/walletManager/walletManagerWorker.ts
```diff
@@ -11,48 +21,85 @@ import {
...
+      )
+    );
+
     this.#walletFactory = walletFactory;
     this.#managerStorageKey = `${walletName}-activate`;
     this.#managerStorage = managerStorage;
     this.#storesFactory = storesFactory;
     this.#logger = logger;
     this.#runtime = runtime;
+    this.#signerManagerApi = consumeSignerManagerApi({ logger: this.#logger, runtime: this.#runtime });
```
mkazlauskas, line 101: For testability and separation of concerns / boundary, it's better to just take `{signerManager: SignerManagerSignApi}` as a dependency in constructor. Application is already exposing and consuming stuff, so it can do this as well when wiring things up.

packages/web-extension/src/walletManager/walletManagerWorker.ts
```diff
@@ -11,48 +21,85 @@ import {
...
+
     this.#hostSubscription = exposeApi(
```
mkazlauskas, line 103: Same as above, application can expose it. Maybe we can just update the interface to `activeWallet$: Observable<ObservableWallet | null>;`
```suggestion
```

packages/web-extension/src/walletManager/walletManager.types.ts
```diff
@@ -8,11 +11,15 @@ export interface WalletManagerProps {
 export interface WalletManagerActivateProps<P extends string | number = string, O = unknown> {
   /** User given name for the observableWallet being activated */
   observableWalletName: string;
```
mkazlauskas, line 13: This should be replaced with name in metadata, that is stored in WalletRepository. We can add a constraint to metadata,e.g. `WalletManager<Metadata extends {name: string}>`
```suggestion
```

packages/web-extension/src/walletManager/walletManager.types.ts
```diff
@@ -8,11 +11,15 @@ export interface WalletManagerProps {
...
-   */
-  walletId: string;
+
+  /** The walletId of the wallet to activate */
+  walletId: WalletId;
+
+  accountIndex?: number;
+
+  /** The chainId of the network to activate the wallet in */
+  chainId: Cardano.ChainId;
```
mkazlauskas, line 21: WalletManager already knows chainId. Is it supposed to switch network when you activate? Maybe it's better to keep activation just for wallet id, and switch network separately?
```suggestion
```

packages/web-extension/src/walletManager/walletManager.types.ts
```diff
@@ -46,11 +53,28 @@ export interface WalletManagerActivateProps<P extends string | number = string,
...
    * Reuses the store if the wallet was previously deactivated but not destroyed.
    */
   activate(props: WalletManagerActivateProps): Promise<void>;
+
+  /**
+   * Returns true if the wallet is active.
+   *
+   * @param props The properties of the wallet to check.
+   */
+  isActive(props: WalletManagerActivateProps): Promise<boolean>;
```
mkazlauskas, line 69: This suggestion is only valid if we keep activation functionality to set walletId only
```suggestion
  isActive(walletId: WalletId): Promise<boolean>;
```

packages/web-extension/src/walletManager/walletManager.types.ts
```diff
@@ -46,11 +53,28 @@ export interface WalletManagerActivateProps<P extends string | number = string,
...
+
+  /**
+   * Switches the network of the active wallet.
+   *
+   * @param id The chain id of the network to switch to.
+   */
+  switchNetwork(id: Cardano.ChainId): Promise<void>;
+
```
mkazlauskas, line 77: It wouldn't let me add a comment for `destroy()` as there were no changes. Let's rename it to `destroyData(walletId: WalletId)`. It shouldn't deactivate, user/app can call it separately. It can throw if you try to destroy wallet that is currently active.

packages/web-extension/src/walletManager/walletManager.types.ts
```diff
@@ -46,11 +53,28 @@ export interface WalletManagerActivateProps<P extends string | number = string,
 }
 
 export interface WalletManagerApi {
+  chainId$: Observable<Cardano.ChainId>;
```
mkazlauskas, line 56: I think there's currently no way to know which wallet is active
```suggestion
  chainId$: Observable<Cardano.ChainId>;
  activeWalletId$: Observable<WalletId>;
```

packages/web-extension/src/walletManager/walletManagerWorker.ts
```diff
@@ -11,48 +21,85 @@ import {
...
+      switchMap((wallet) =>
+        wallet.genesisParameters$.pipe(
+          map(
+            (params): Cardano.ChainId => ({
+              networkId: params.networkId,
+              networkMagic: params.networkMagic
+            })
+          )
+        )
+      )
```
mkazlauskas, line 92: It would be a better/more optimal behavior, because currently it will re-emit same chain id when you activate different wallet on same network
```suggestion
      ),
      distinctUntilChanged(deepEquals)
```

packages/e2e/test/web-extension/extension/background/walletManager.ts
```diff
@@ -28,30 +31,40 @@ export interface WalletFactoryDependencies {
...
-        stores
+        stores,
+        witnesser
       })
     ).wallet
 };
 
 const storesFactory: StoresFactory = {
-  create: ({ walletId }) => storage.createPouchDbWalletStores(walletId, { logger })
+  create: ({ name }) => storage.createPouchDbWalletStores(name, { logger })
```
mkazlauskas, line 47: Please confirm that this is compatible with existing store name (won't re-create the stores for existing wallets)

packages/web-extension/test/walletManager/walletManagerWorker.test.ts
```diff
@@ -41,48 +54,74 @@ const expectWalletChannelClosed = () => {
 
 describe('WalletManagerWorker', () => {
```
mkazlauskas, line 55: We should probably add a couple new tests for new functionality

mkazlauskas (changes requested):

package.json
```diff
@@ -88,5 +88,8 @@
     "typedoc-plugin-missing-exports": "^1.0.0",
     "typescript": "^4.7.4"
   },
-  "packageManager": "yarn@3.2.1"
+  "packageManager": "yarn@3.2.1",
+  "dependencies": {
+    "chromedriver": "^120.0.0"
```
mkazlauskas, line 93: Please move this dependency to e2e package

packages/e2e/test/web-extension/extension/background/walletManager.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/a048f61d16b6c6969bdd63ae367f853c9a269e2f

packages/e2e/test/web-extension/extension/ui.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/a048f61d16b6c6969bdd63ae367f853c9a269e2f

packages/e2e/test/web-extension/extension/ui.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/a048f61d16b6c6969bdd63ae367f853c9a269e2f

packages/e2e/test/web-extension/extension/ui.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/a048f61d16b6c6969bdd63ae367f853c9a269e2f

package.json
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/a048f61d16b6c6969bdd63ae367f853c9a269e2f

packages/web-extension/src/walletManager/walletManagerUi.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/813029a3123dac04ffb0f783c6e239a3c19ae8e9

packages/web-extension/src/walletManager/walletManagerWorker.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/813029a3123dac04ffb0f783c6e239a3c19ae8e9

packages/web-extension/src/walletManager/walletManagerWorker.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/813029a3123dac04ffb0f783c6e239a3c19ae8e9

packages/web-extension/src/walletManager/walletManager.types.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/813029a3123dac04ffb0f783c6e239a3c19ae8e9

packages/web-extension/src/walletManager/walletManager.types.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/813029a3123dac04ffb0f783c6e239a3c19ae8e9

packages/web-extension/src/walletManager/walletManager.types.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/813029a3123dac04ffb0f783c6e239a3c19ae8e9

packages/web-extension/src/walletManager/walletManagerWorker.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/813029a3123dac04ffb0f783c6e239a3c19ae8e9

packages/e2e/test/web-extension/extension/background/walletManager.ts
AngelCastilloB: It should be fine, the name of the argument/props object is not used for the db name but rather its content:

```
  const baseDbName = walletName.replace(/[^\da-z]/gi, '');
  const docsDbName = `${baseDbName}Docs`;
  ````

File: packages/web-extension/src/walletManager/walletManager.types.ts (hunk 10 above)
Comment (AngelCastilloB) [reply packages/web-extension/src/walletManager/walletManager.types.ts:18]: So in order to activate a wallet we need to retrieve its state from the storage and build the providers, for this we need all three arguments, walletId, chainId and account index. This was how it worked before too btw, is just that the client code would build the store id (pubkeyhash + networkId). We are now passing the pieces separated and the wallet manager is now in charge of building the storeage ID.

I also think this makes sense, the same credentials is still a different wallet if used in a different network. If we just take the wallet ID here, we would need to guess which network id and account index the client means to activate, which we could guess by looking at the currently activa wallet and reuse the same network id and account index, but we could run into problems, lets say the user has two wallets, in one wallet he has 3 accounts and in the second wallet has 1 account, what should we do here if he is currently at account 3 on the first wallet and switches to the first wallet? we could see the accounts are different and default to account 0 (or always do this), but ultimately I think Lace should be explicit about what wallet he means to activate (which credentials, which account and in which network), and also Lace knows this, because the user will trigger this action via the GUI so Lace knows exactly what wallet the user wants (credentials, network, account). I added the utility method "switchNetwork" which does this guess work and reuses the current active wallet properties and just tries to activate with a different network.

But let me know what you think, if you dont agree with this approach I can add the logic to make it work like you suggest

File: packages/web-extension/test/walletManager/walletManagerWorker.test.ts (hunk 16 above)
Comment (AngelCastilloB) [reply packages/web-extension/test/walletManager/walletManagerWorker.test.ts]: Once we decide the actual interface functionality of the wallet manager, I will then proceed to complete the unit tests as the last step

File: packages/web-extension/src/walletManager/walletManager.types.ts (hunk 13 above)
Comment (mkazlauskas) [reply packages/web-extension/src/walletManager/walletManager.types.ts:53]: Let's combine `activeWalletId$` and `activeWalletAccount$` into 1 `activeWallet$: Observable<{walletId: WalletId; accountIndex?: number}>`, because they are always set together in 1 operation. Depending on decision https://github.com/input-output-hk/cardano-js-sdk/pull/1012#discussion_r1427616805 we could also add `chainId: ChainId` to it.

File: packages/web-extension/src/walletManager/walletManager.types.ts (hunk 11 above)
Comment (mkazlauskas) [reply packages/web-extension/src/walletManager/walletManager.types.ts:68]: Maybe we should delete this method? Doesn't seem very useful since we add the observable for active wallet props

Comment (AngelCastilloB) [reply packages/web-extension/src/walletManager/walletManager.types.ts:68]: agree will remove

File: packages/e2e/test/web-extension/extension/background/walletManager.ts (hunk 15 above)
Comment (mkazlauskas) [reply packages/e2e/test/web-extension/extension/background/walletManager.ts:50]: The 'walletName' here is confusing as we're using xpub (hash?). This is how it looks now: 
image

File: package.json (hunk 17 above)
Comment (mkazlauskas) [reply package.json]: It should probably be in `devDependencies` as it's only used for running the tests, and not for anything under `src/`

File: packages/web-extension/src/walletManager/walletManager.types.ts (hunk 11 above)
Comment (AngelCastilloB) [reply packages/web-extension/src/walletManager/walletManager.types.ts]: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/cd24430f838afcaaa5d8271ec804d2f53a167ad4

File: packages/web-extension/src/walletManager/walletManager.types.ts (hunk 13 above)
Comment (AngelCastilloB) [reply packages/web-extension/src/walletManager/walletManager.types.ts]: changes as discussed in slack, done here: https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/cd24430f838afcaaa5d8271ec804d2f53a167ad4

File: package.json (hunk 17 above)
Comment (AngelCastilloB) [reply package.json]: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/bacced6cb9d380b38fb7fba77c5489f75518655e

File: packages/web-extension/test/walletManager/walletManagerWorker.test.ts (hunk 16 above)
Comment (AngelCastilloB) [reply packages/web-extension/test/walletManager/walletManagerWorker.test.ts]: added some more tests to cover the extra cases here: https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/ee775430185f2444c530835a5ce0306e00a4986c

**Review by mkazlauskas — APPROVED** [review]
Great work! :muscle: I would reword last commit when squashing to something more generic, like `feat!: rework WalletManager` and then detail it in commit body, because it's doing a lot more than just changing type of the argument

File: packages/web-extension/src/walletManager/types.ts [hunk 18]
```diff
@@ -10,7 +10,7 @@ export enum WalletType {
 }
 
 /** For BIP-32 wallets: hash of extended account public key. For script wallets: script hash */
```
mirceahasegan, line 12: Is the comment still relevant?

mkazlauskas: I think yes, but it can be a different kind of hash. If wanted to make this an OpaqueString then I think it should be a WalletId rather than hash type.

packages/e2e/package.json
```diff
@@ -149,6 +149,7 @@
     "babel-loader": "^8.2.5",
     "blake2b-no-wasm": "2.1.4",
     "buffer": "^6.0.3",
+    "chromedriver": "^120.0.0",
```
mirceahasegan, line 152: This dependency was removed in the past because the actual version has to be in sync with the developer environment chrome version, which can vary.
Ideally devs would have the desired chromedriver installed globally in their setup, or use an alternative solution.
If we choose to keep the version here, [this README comment](https://github.com/input-output-hk/cardano-js-sdk/blob/master/README.md#L77) should be updated to inform the need to set this version based on the developer's chrome browser version.
Sidenote: the instructions in the [README](https://github.com/input-output-hk/cardano-js-sdk/blob/master/README.md#L77) do not work 😞

mkazlauskas: It was re-added because global installation approach no longer works. I think we should remove the instructions and try to come up with a better way to manage this in the future

mirceahasegan: Agree. Let's replace the instructions with "Update chromedrive dependency to the local chrome version", or something in the lines of that :D

packages/web-extension/src/walletManager/walletManager.ts
```diff
@@ -0,0 +1,388 @@
...
+    const { walletId, chainId, accountIndex } = props;
+
+    const wallets = await firstValueFrom(this.#walletRepository.wallets$);
+    const activeWallet = wallets.find((wallet) => wallet.walletId === walletId);
+
+    if (!activeWallet) {
+      throw new InvalidArgumentError('walletId', `Wallet ${walletId} not found`);
+    }
+
+    this.#deactivateWallet();
```
mirceahasegan, line 151: There doesn't seem to be any guard against activating the same wallet twice. It would deactivate (which shuts it down), then build the witnesser and reactivate

AngelCastilloB: The methods first line has this guard which should prevent double activation:

```
    if (this.#isActive(props)) {
      return;
    }
```

packages/e2e/package.json
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/aba05830935a2d5a441802fab53ccbcd5aea4a32

packages/web-extension/src/walletManager/types.ts
AngelCastilloB: Yes the comment explains what the contents of wallet id could be, in the end we decided to just make it a string, should we change this?

mirceahasegan: No need to change it. Thank you

mirceahasegan (approved):
Great job, @AngelCastilloB ! 🚀

packages/web-extension/src/walletManager/walletManager.ts
```diff
@@ -0,0 +1,388 @@
...
+    // {
+    //   walletId: WalletId; <-- I.E 'da7b4795b11a79116eb5232c83d2c862', only the 16 byte length pub key hash.
+    //   accountIndex?: number; <-- optional account number
+    //   chainId: Cardano.ChainId; <-- I.E { networkId: Cardano.NetworkId.Testnet, networkMagic: 1097911063 }
+    //   provider?: { type: P; options: O };
+    // }
+    //
+    // If we find a legacy wallet id, we need to map it to the new format and save it to storage.
+    if (this.#isLegacyProps(lastActivateProps.walletId)) {
+      lastActivateProps = this.#mapLegacyProps(lastActivateProps);
```
mkazlauskas, line 122: This would make it easier to migrate, because Lace stores wallet name in UI context: it will have to create the wallet in repository from the UI.
```suggestion
      lastActivateProps = this.#mapLegacyProps(lastActivateProps);
      if (!activeWalletExistsInRepository(lastActivateProps)) return;
```

mkazlauskas (changes requested):

packages/web-extension/src/walletManager/walletManager.ts
AngelCastilloB: done

mkazlauskas (dismissed):
Great work! :rocket:

mirceahasegan (dismissed):

packages/web-extension/src/walletManager/walletManager.ts
```diff
@@ -0,0 +1,337 @@
...
+   * Destroys all stores for the given wallet id.
+   *
+   * @param walletId The wallet id to destroy.
+   * @private
+   */
+  async #destroyWalletStores(walletId: WalletId): Promise<void> {
+    if (this.#activeWalletProps?.walletId === walletId)
+      throw new InvalidArgumentError('walletId', 'Cannot destroy active wallet');
+
+    const stores = [...this.#walletStores.keys()].filter((key) => key.includes(walletId));
```
mkazlauskas, line 203: This won't destroy the store data if it's not loaded in memory (if wallet wasn't activated since starting)

packages/web-extension/src/walletManager/walletManager.types.ts
```diff
@@ -59,17 +73,19 @@ export interface WalletManagerApi {
...
    */
-  destroy(): Promise<void>;
+  destroyData(walletId: WalletId): Promise<void>;
 }
 
 export interface WalletFactory {
   create: (
-    props: WalletManagerActivateProps,
-    dependencies: { keyAgent: AsyncKeyAgent; stores: storage.WalletStores }
+    name: string,
```
mkazlauskas, line 84: Since wallet factory is provided by Lace, it needs to know the provider and chain id. I suggest to pass through all the activate props.
```suggestion
    props: WalletManagerActivateProps & { name: string },
```

mkazlauskas (changes requested):

packages/web-extension/src/walletManager/walletManager.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/17d86e545285d041a6fd47c4d48cc5b1d8897c70

packages/web-extension/src/walletManager/walletManager.types.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1012/commits/39951537f0ec93792820e128dfc09540d2270a1c

mkazlauskas (approved):
Nice work!

mirceahasegan (approved):
🥇

PR #1009, opened 2023-12-05
refactor: remove indirect KeyAgent dependency on ObservableWallet

mkazlauskas:
# Context

1. `KeyAgent` depends on `InputResolver`, which in is implemented by a util that depends on `ObservableWallet`. Since `ObservableWallet` depends on `KeyAgent`, it makes it a circular dependency. It was worked around by having LazyWalletUtil and `setupWallet` utils.
2.  Addresses are stored by both `KeyAgent` and `PersonalWallet`.

# Proposed Solution

1. Replaces the `KeyAgent`->`InputResolver` dependency with an extra `TxContext` argument for `signTransaction` method, as well as removes the obsolete utils.
2. Consolidate address storage in `PersonalWallet`, making `KeyAgent` stateless and establishing single source of truth.

# Important Changes Introduced

- remove createAsyncKeyAgent tests instead of refactoring them, because `AsyncKeyAgent` is no longer required and should be removed.
- extract address discovery and storage from `PersonalWallet` into a new `AddressTracker` component
- remove `StubKeyAgent` as it was only used in tests where specific addresses are required
- refactor `KeyAgentBase` class to utilize `Bip32Account` for key and address derivations. This can be refactored further by removing those methods from KeyAgent and relying Bip32Account everywhere it's used instead.

**BREAKING CHANGES**:
- remove `KeyAgent.knownAddresses`
- remove `SerializableKeyAgentData.knownAddreses`
- remove `AsyncKeyAgent.knownAddresses$`
- remove `LazyWalletUtil` and `setupWallet` utils
- replace `KeyAgent` dependency on `InputResolver` with props passed to sign method
- re-purpose `AddressManager` into `Bip32Account`: addresses are now stored only by the wallet

## Checklist

- [x] `yarn build && yarn test:build:verify && yarn lint && yarn test`
- [ ] hw integration tests pass (not tested yet)
- [ ] e2e tests pass (not tested yet)
- [ ] e2e web-extension tests pass (not tested yet)

PR #1014, opened 2023-12-08
feat: add ObservableWallet.discoverAddreses

mkazlauskas:
# Context

we consolidated address storage in the wallet in #1009 and there was no way to re-discover addresses,
since storage is managed by the wallet internally

# Proposed Solution

add ObservableWallet.discoverAddreses


mchappell (approved):
LGTM @mkazlauskas 💪

rhyslbw (approved):

PR #1023, opened 2023-12-13
feat(wallet): forward MessageSender to GetCollateralCallback

renanvalentin:
# Context

When `cip30.getCollateral` is called, Lace will present the dapp information. Thus, we need a reliable way to get the dapp tab. See [cip30.ts](https://github.com/input-output-hk/lace/blob/f7c4fd1972184bf5d84b196894a6e0700236ca0b/apps/browser-extension-wallet/src/lib/scripts/background/cip30.ts#L70-L71)

### DIFF
--- a/packages/wallet/src/cip30.ts
+++ b/packages/wallet/src/cip30.ts
@@ -59,6 +59,7 @@ export type SubmitTxCallbackParams = {
 
 // Optional callback
 export type GetCollateralCallbackParams = {
+  sender: MessageSender;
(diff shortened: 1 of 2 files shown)


mkazlauskas (commented):
Looks good! I think `feat` is a more appropriate commit type for this, as it's adding functionality

mkazlauskas (approved):
Nice work!

iccicci (approved):
LGTM

PR #1029, opened 2023-12-19
fix(core): take care of skipped eras in era summaries

iccicci:
# Context

Some of our `slotCalc` utilities are not taking care of skipped eras in `EraSummary[]`.

In the real world **this is not impacting _mainnet_ nor _preprod_** while this impacts on _preview_ since it has this use case.

# Proposed Solution

Fixed those utilities affected by the bug: `epochSlotsCalc` and `createSlotTimeCalc`.

Not added a specific test for `createSlotEpochInfoCalc` since it relies on `createSlotEpochCalcImpl` (not affected by the bug) and `createSlotTimeCalc` (fixed and tested in this PR).

### DIFF
--- a/packages/core/test/util/slotCalc.test.ts
+++ b/packages/core/test/util/slotCalc.test.ts
@@ -14,6 +14,7 @@ import {
   createSlotTimeCalc,
   epochSlotsCalc
 } from '../../src';
+import { fromSerializableObject } from '@cardano-sdk/util';
 
 import merge from 'lodash/merge';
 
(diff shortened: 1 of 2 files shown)


packages/core/test/util/slotCalc.test.ts
```diff
@@ -63,9 +64,56 @@ const preprodEraSummaries: EraSummary[] = [
...
 const SomeByronSlot = Cardano.Slot(1_209_592);
 
 describe('slotCalc utils', () => {
+  describe('preview', () => {
+    // following test data is taken from last block of epoch 418
+    // https://preview.cexplorer.io/block/b25f4238b9d9d70a2942d277baf2312cc0562990e38b2ce5888861454e05e6c4
+    const testEpoch = Cardano.EpochNo(418);
+    const testSlot = Cardano.Slot(36_201_583);
+    const testDate = new Date('2023-12-17T23:59:43Z');
+    it('epochStartCalc correctly computes last slot of epoch 418', () => {
```
mkazlauskas, line 104:
```suggestion
    it('epochSlotsCalc correctly computes last slot of epoch 418', () => {
```

packages/core/test/util/slotCalc.test.ts
```diff
@@ -63,9 +64,56 @@ const preprodEraSummaries: EraSummary[] = [
...
+      const range = epochSlotsCalc(testEpoch, previewEraSummaries);
+      expect(range.lastSlot).toBeGreaterThanOrEqual(testSlot);
+    });
```
mkazlauskas, line 107: This would be a useful test too
```suggestion
    });
    it('SlotEpochInfoCalc correctly computes info for last slot of epoch 418', () =>
```

mkazlauskas (changes requested):

packages/core/test/util/slotCalc.test.ts
iccicci: 6fefbb1d6c (already squashed to speed up) should fix this

packages/core/test/util/slotCalc.test.ts
iccicci: 6fefbb1d6c (already squashed to speed up) should fix this

mkazlauskas (approved):
Nice work!

mirceahasegan (approved):
LGTM!

PR #1030, opened 2023-12-20
Feat/extend observable wallet addresses to support script addresses

AngelCastilloB:
# Context

It should be an Observable<WalletAddress>, where type WalletAddress = GroupedAddress | ScriptAddress

LW-9424

# Proposed Solution

Introduce a new type WalletAddress

Observable wallet now emits addresses$ as an array of WalletAddress[] rather than GroupedAddress[]

### DIFF
--- a/packages/wallet/src/types.ts
+++ b/packages/wallet/src/types.ts
@@ -51,6 +51,22 @@ export type FinalizeTxProps = Omit<TxContext, 'signingContext'> & {
 
 export type HandleInfo = HandleResolution & Asset.AssetInfo;
 
+export interface ScriptAddress {
+  networkId: Cardano.NetworkId;
+  address: Cardano.PaymentAddress;
+  rewardAccount: Cardano.RewardAccount;
+  scripts: {
+    payment: Cardano.NativeScript;
+    stake: Cardano.NativeScript;
+  };
+}
+
+export type WalletAddress = GroupedAddress | ScriptAddress;
+
+export const isScriptAddress = (address: WalletAddress): address is ScriptAddress => 'scripts' in address;
+
+export const isKeyHashAddress = (address: WalletAddress): address is GroupedAddress => !isScriptAddress(address);
+
 export interface ObservableWallet {
   readonly balance: BalanceTracker;
   readonly delegation: DelegationTracker;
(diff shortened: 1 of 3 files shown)


packages/wallet/src/types.ts
```diff
@@ -51,6 +51,25 @@ export type FinalizeTxProps = Omit<TxContext, 'signingContext'> & {
...
+}
+
+export type WalletAddress = GroupedAddress | ScriptAddress;
+
+export const isKeyHashAddress = (address: WalletAddress): address is GroupedAddress =>
+  Cardano.Address.fromBech32(address.address).asBase()?.getPaymentCredential()?.type === Cardano.CredentialType.KeyHash;
+
+export const isScriptAddress = (address: WalletAddress): address is ScriptAddress =>
+  Cardano.Address.fromBech32(address.address).asBase()?.getPaymentCredential()?.type ===
+  Cardano.CredentialType.ScriptHash;
```
mkazlauskas, line 71: That's a lighter implemenatation, could do similar approach for `isKeyHashAddress`
```suggestion
  'scripts' in address
```

packages/wallet/test/services/WalletUtil.test.ts
```diff
@@ -607,5 +613,344 @@ describe('WalletUtil', () => {
...
+          },
+          ...tx.body.inputs
+        ];
+
+        expect(tx.body.inputs.length).toBeGreaterThanOrEqual(2);
+        expect(tx.body.certificates!.length).toBe(1);
+        expect(await requiresForeignSignatures(tx, scriptWallet)).toBeTruthy();
+      });
+
+      it('Voting procedures are always foreign', async () => {
```
mkazlauskas, line 922: Maybe we could implement the proper check that compares script hash for script credentials on this level. It won't be possible to use via cip30, but that would be a more correct implementation, right? Same with DRepCredential and probably some voting certificate types

mkazlauskas (changes requested):

packages/wallet/src/types.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1030/commits/7134e634d3d1c5437c45f029383bd1213d7756db

packages/wallet/test/services/WalletUtil.test.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1030/commits/508d7e4f778e8ac6f98c0905a8e7dca178e49b56

mkazlauskas (approved):
Nice work! :rocket:

pczeglik-iohk (approved):
Code looks good to me. Thank you @AngelCastilloB

PR #1035, opened 2024-01-03
ci: take the exact version from lib/chromedriver.js

michalrus:
# Context

**LW-9464**

ChromeDriver version was hardcoded. This would break in a non-obvious way on the next ChromeDriver update.

# Proposed Solution

Read the version from `node_modules/chromedriver/lib/chromedriver.js`. The hash will still need to be updated, but at least the error message will be clear, providing both hashes (you can `grep` for the old one and replace it).

It costs us around 5 seconds, I thought it's the least awkward option for the end user (who updates this dependency).

—

### DIFF
--- a/nix/cardano-services/packages.nix
+++ b/nix/cardano-services/packages.nix
@@ -43,8 +43,21 @@ in {
       '';
     });
 
-    chromedriverVersion = "120.0.6099.71";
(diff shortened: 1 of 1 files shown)


mkazlauskas (approved):
Thanks, looks good!

Ideally we would find a way to omit chromedriver from cardano-services build entirely, because it is only used in `e2e` package tests

rhyslbw (approved):

PR #1036, opened 2024-01-04
Fix/wallet manager active wallet id null

AngelCastilloB:
# Context

The wallet manager was not emitting null for the activeWalletId$ observable when there was no active wallet. Also the activate method was not comparing the active wallet properties correctly.

LW-9094

# Proposed Solution

Fix both bugs on the wallet manager


mkazlauskas (approved):
Looks good! :rocket:

pczeglik-iohk (approved):
Code looks good. Thank you @AngelCastilloB

PR #1043, opened 2024-01-10
test(wallet): hardware wallets multi-delegation

tomislavhoracek:
This PR introduces additional tests for signing multi-delegation transaction with Ledger and Trezor devices
[JIRA Ticket](https://input-output.atlassian.net/browse/LW-9390)

# Context

SDK testing standards were followed when developing this ticket
It will not be tested by SDETs in isolation!

# Proposed Solution

- Add support for signing Ledger multi-delegation transactions
- Add support for signing Trezor multi-delegation transactions


mkazlauskas (commented):
Changes looks good! It appears that tests from 1st commit are testing fix from 2nd commit. If that's the case then the order is incorrect and I suggest to squash into a single `fix` commit

packages/core/src/Serialization/TransactionBody/TransactionBody.ts
```diff
@@ -132,19 +132,24 @@ export class TransactionBody {
...
-
         if (!rewardAddress) {
           throw new SerializationError(SerializationFailure.InvalidAddress, `Invalid withdrawal address: ${key}`);
         }
+        const rewardAddressBytes = rewardAddress.toAddress().toBytes();
+        withdrawalsWithAddressBytes.set(rewardAddressBytes, value);
+      }
+
+      // Sort withdrawals by address bytes, canonically
+      const sortedCanonically = new Map([...withdrawalsWithAddressBytes].sort((a, b) => (a > b ? 1 : -1)));
```
mkazlauskas, line 147: A map here looks odd since we're looking for ordered collection.
```suggestion
      const sortedCanonically = [...withdrawalsWithAddressBytes].sort((a, b) => (a > b ? 1 : -1));
```

tomislavhoracek: this will create an array and we need a Map. I reused the existing logic but transformed addresses to address bytes
https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/core/src/Serialization/TransactionBody/TransactionBody.ts#L136

mkazlauskas: Why do we need a map? for-of where it's used should work with array just as well :thinking:

tomislavhoracek: aaaah, you are right, please ignore my last comment. Sorry

VanessaPC (approved):
Nice work!

mkazlauskas (approved):
Nice work! :rocket: 

The commit could be scoped to `(core)` because it's the only package that has `src/` changes

PR #1045, opened 2024-01-11
feat!: signerManager and Witnesser now propagate signData confirmation address

AngelCastilloB:
# Context

SignerManager signData requests don’t have ‘address’ property. Instead, it only has derivation path to sign with.

In order to get the address, we need to update Witnesser and SignerManager interfaces to preserve address as additional context of the signing operation.

# Proposed Solution

- SignDataContext now has an additional optional address field.
- Witnesser signBlob now takes a context as one of its arguments (of type SignDataContext, same as SignerManager signData).


rhyslbw (approved):

pczeglik-iohk (approved):
Code looks good. Thank you @AngelCastilloB

PR #1048, opened 2024-01-15
LW-9535 configurable metadata job retry delay

iccicci:
# Context

Latest applied changes in the SDK introduced some delays in fetching stake pool metadata. Since the default retry delay for the job which fetches the metadata is tailored for `live-mainnet` no retries are performed during the rebuild fixtures procedure.

The sum of this two facts makes the newly generated fixtures missing the metadata.

# Proposed Solution

- made the retry delay for the job configurable
- configured said delay to 60" in `local-network`


mkazlauskas (changes requested):
Looks good! Just please squash the 1st two commits and mark it as breaking change due to rename of operator exported from `projection-typeorm`

rhyslbw (dismissed):
Nice work @iccicci 🦾

rhyslbw (approved):

mkazlauskas (approved):
Nice work! :rocket: 

Just a note (no action needed): when commit message header doesn't describe the breaking change, it's better to use `BREAKING CHANGE:` footer to describe it separately instead of `!`

PR #1050, opened 2024-01-16
feat(web-extension): handle wallet metadata in repository [LW-9503]

DominikGuzei:
# Context

This PR refactors the `WalletRepository` and related code to handle wallet metadata as well as account metadata.

# Proposed Solution

Introduces a separate `AccountMetadata` type that can be different from the wallet metadata and also splits the `WalletRepository::updateMetadata` into two separate methods `updateWalletMetadata` and `updateAccountMetadata` to avoid confusion and complicated checks.

Both types need to be passed explicitly in most cases as something like `<WalletMetadata extends {}, AccountMetadata extends {} = WalletMetadata>` might be convenient in the first place but obscures the fact that the metadata for wallets and accounts can be different eventually.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -22,6 +22,7 @@ module.exports = {
     '@typescript-eslint/no-unused-vars': ['warn', { argsIgnorePattern: '^_', varsIgnorePattern: '^_' }],
     // needed for inference from type guards
     '@typescript-eslint/no-var-requires': 0,
+    'brace-style': 0, // collides with prettier formatting
     // typescript checks return types
     camelcase: 0,
     'consistent-return': 0,
--- a/packages/e2e/test/web-extension/extension/background/walletManager.ts
+++ b/packages/e2e/test/web-extension/extension/background/walletManager.ts
@@ -38,7 +38,7 @@ export interface WalletFactoryDependencies {
  * @param accountIndex The account index to get the name from.
  * @private
  */
-const getWalletName = (wallet: AnyWallet<Metadata>, accountIndex?: number): string => {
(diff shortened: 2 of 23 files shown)


.eslintrc.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
     '@typescript-eslint/no-unused-vars': ['warn', { argsIgnorePattern: '^_', varsIgnorePattern: '^_' }],
     // needed for inference from type guards
     '@typescript-eslint/no-var-requires': 0,
+    'brace-style': 0, // collides with prettier formatting
```
mkazlauskas, line 25: Would be great to extract this into a separate commit

Also please squash the rest into a single `feat` commit that has a footer formatted like that

```
BREAKING CHANGE: add metadata to bip32 wallets
- split WalletRepository.updateMetadata into 2 methods
```

it will then bump the package version correctly as well as add notes to CHANGELOG.md

mkazlauskas (changes requested):

.eslintrc.js
DominikGuzei: done ✅

mkazlauskas (approved):
Great work! :rocket:

.eslintrc.js
mkazlauskas: The chore commit should be unscoped, as it applies to more than 1 (all in this case) package. Like this: `chore: disable brace style eslint rule for prettier co...`. Also it should be the 1st commit ,because the 2nd commit depends on it (you would get lint errors if you checked out that commit.

mkazlauskas (dismissed):

AngelCastilloB (dismissed):
LGTM @DominikGuzei 🚀

mkazlauskas (approved):

PR #1053, opened 2024-01-17
LW-9518 Use `pgboss.archive` table to check computed epoch rewards

iccicci:
# Context

The `pool-rewards` job uses `pgboss.job` table as a registry to check which is the last epoch for which it run.

This is not enough as **pg-boss** archives completed jobs in `pgboss.archive` table:

# Proposed Solution

Added `COUNT(*) FROM pgboss.archive` to the query which performs given check.


mkazlauskas (approved):
Nice work! :detective:

mirceahasegan (approved):
:rocket:

PR #1056, opened 2024-01-18
Fix smash settings in preview and preprod env files

iccicci:
# Context

Our `preview` and `preprod` **docker compose** environments are configured to use **smash**, while they should use the `direct` method.

# Proposed Solution

Fixed the cfg


rhyslbw (commented):
I've just done some digging and it looks like there's deployments at:
`https://preprod-smash.world.dev.cardano.org/api/v1`
`https://preview-smash.world.dev.cardano.org/api/v1`

rhyslbw (approved):

mkazlauskas (approved):

mirceahasegan (approved):
Great work, @iccicci 🚀

PR #1057, opened 2024-01-18
Feat/lw 9145 implement utilities to compute asset diff

AngelCastilloB:
# Context

To implement new UI components Summary in Lace, From Address and To Address sections, we’ll need to compute the data from the transaction body and resolved inputs.

# Proposed Solution

Implement two new transaction inspectors to produce a renderable object with the summary result.

- TxInspectors are now async
- Added new TxSummaryInspector
- Added new TokenTransferInspector

### DIFF
--- a/packages/core/src/util/transactionSummaryInspector.ts
+++ b/packages/core/src/util/transactionSummaryInspector.ts
@@ -1,0 +1,153 @@
+import * as Cardano from '../Cardano';
+import { AssetId } from '../Cardano';
+import {
+  AssetsMintedInspection,
+  Inspector,
+  ResolutionResult,
+  assetsBurnedInspector,
+  assetsMintedInspector,
+  resolveInputs,
+  totalAddressInputsValueInspector,
+  totalAddressOutputsValueInspector
+} from './txInspector';
+import { BigIntMath } from '@cardano-sdk/util';
+import { coalesceTokenMaps, subtractTokenMaps } from '../Asset/util';
+import { coalesceValueQuantities } from './coalesceValueQuantities';
+import { computeImplicitCoin } from '../Cardano/util';
+import { subtractValueQuantities } from './subtractValueQuantities';
+
+interface TransactionSummaryInspectorArgs {
+  addresses: Cardano.PaymentAddress[];
+  rewardAccounts: Cardano.RewardAccount[];
+  inputResolver: Cardano.InputResolver;
+  protocolParameters: Cardano.ProtocolParameters;
+}
+
+export type TransactionSummaryInspection = {
+  assets: Cardano.TokenMap;
+  coins: Cardano.Lovelace;
+  collateral: Cardano.Lovelace;
+  deposit: Cardano.Lovelace;
+  returnedDeposit: Cardano.Lovelace;
+  fee: Cardano.Lovelace;
+  unresolved: {
+    inputs: Cardano.TxIn[];
+    value: Cardano.Value;
+  };
+};
+
+export type TransactionSummaryInspector = (
+  args: TransactionSummaryInspectorArgs
+) => Inspector<TransactionSummaryInspection>;
+
+/**
+ * Gets the collateral specified for this transaction.
+ *
+ * @param tx transaction to inspect.
+ * @param inputResolver input resolver.
+ * @param addresses addresses to inspect.
+ */
+export const getCollateral = async (
+  tx: Cardano.Tx,
+  inputResolver: Cardano.InputResolver,
+  addresses: Cardano.PaymentAddress[]
+): Promise<Cardano.Lovelace> => {
+  if (!tx.body.collaterals || tx.body.collaterals.length === 0) return 0n;
+
+  const resolvedCollateralInputs = (await resolveInputs(tx.body.collaterals, inputResolver)).resolvedInputs.filter(
+    (input) => addresses.includes(input.address)
+  );
+
+  const totalOwnedValueAtRisk = BigIntMath.sum(resolvedCollateralInputs.map(({ value }) => value.coins));
+
+  // If collateral return is specified, it means that not all the balance in the collateral inputs would be spent given a validation failure
+  if (tx.body.collateralReturn) {
+    // In the case of CIP-40 we need to sum all the collateral inputs we own, and check if the collateral return is coming to one of our addresses
+    // If it is, we can simply subtract the sum of all our collateral inputs with the amount returned in the collateral returned, if the collateral return
+    // is going somewhere else, it means we will lose whatever is on those outputs given a failure.
+    if (!addresses.includes(tx.body.collateralReturn.address)) return totalOwnedValueAtRisk;
+
+    // Should never be negative in a correctly balanced transaction, but let's do a sanity check anyway.
+    return BigIntMath.max([totalOwnedValueAtRisk - tx.body.collateralReturn.value.coins, 0n]) ?? 0n;
+  }
+
+  // Legacy collaterals, we simply return the addition of the collateral inputs we own.
+  return totalOwnedValueAtRisk;
+};
+
+const totalInputsValue = (resolvedInputs: ResolutionResult) => {
+  const receivedInputsValues = resolvedInputs.resolvedInputs.map((input) => input.value);
+  return coalesceValueQuantities(receivedInputsValues);
+};
+
+const totalOutputsValue = (outputs: Cardano.TxOut[]) => coalesceValueQuantities(outputs.map((output) => output.value));
+
+const mintInspectionToTokenMap = (mintedAssets: AssetsMintedInspection) =>
+  new Map<Cardano.AssetId, bigint>(
+    mintedAssets.map((asset) => [AssetId.fromParts(asset.policyId, asset.assetName), asset.quantity])
+  );
+
+const getImplicitAssets = async (tx: Cardano.Tx) => {
+  const mintedAssets = mintInspectionToTokenMap(await assetsMintedInspector(tx));
+  const burnedAssets = mintInspectionToTokenMap(await assetsBurnedInspector(tx));
+
+  return coalesceTokenMaps([mintedAssets, burnedAssets]);
+};
+
+const getUnaccountedFunds = async (
+  tx: Cardano.Tx,
+  resolvedInputs: ResolutionResult,
+  implicitCoin: Cardano.Lovelace,
+  implicitAssets: Cardano.TokenMap = new Map()
+): Promise<Cardano.Value> => {
+  const totalInputs = totalInputsValue(resolvedInputs);
+  const totalOutputs = totalOutputsValue(tx.body.outputs);
+
+  totalInputs.assets = coalesceTokenMaps([totalInputs.assets, implicitAssets]);
+  totalInputs.coins += implicitCoin;
+
+  return subtractValueQuantities([totalOutputs, totalInputs]);
+};
+
+/**
+ * Inspects a transaction and produces a summary.
+ *
+ * @param {TransactionSummaryInspectorArgs} args The arguments for the inspector.
+ */
+export const transactionSummaryInspector: TransactionSummaryInspector =
+  (args: TransactionSummaryInspectorArgs) => async (tx) => {
+    const { inputResolver, addresses, rewardAccounts, protocolParameters } = args;
+    const resolvedInputs = await resolveInputs(tx.body.inputs, inputResolver);
+    const fee = tx.body.fee;
+
+    const implicit = computeImplicitCoin(
+      protocolParameters,
+      { certificates: tx.body.certificates, withdrawals: tx.body.withdrawals },
+      rewardAccounts || []
+    );
+
+    const collateral = await getCollateral(tx, inputResolver, addresses);
+
+    const totalOutputValue = await totalAddressOutputsValueInspector(addresses)(tx);
+    const totalInputValue = await totalAddressInputsValueInspector(addresses, inputResolver)(tx);
+    const implicitCoin = (implicit.withdrawals || 0n) + (implicit.reclaimDeposit || 0n) - (implicit.deposit || 0n);
+    const implicitAssets = await getImplicitAssets(tx);
+
+    const diff = {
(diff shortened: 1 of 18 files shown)


packages/core/src/util/txInspector.ts
```diff
@@ -81,20 +84,64 @@ export type MetadataInspector = Inspector<MetadataInspection>;
 export type PoolRegistrationInspector = Inspector<PoolRegistrationInspection>;
 export type PoolRetirementInspector = Inspector<PoolRetirementInspection>;
 
+type ResolvedInput = {
```
mkazlauskas, line 87: Any reason not to use TxOut type here (to also resolve datum etc.)?
```suggestion
type ResolvedInput = TxIn & TxOut;
```

packages/core/test/Cardano/util/subtractValueQuantities.test.ts
```diff
@@ -36,7 +36,7 @@ describe('Cardano.util.subtractValueQuantities', () => {
       ]),
       coins: 200n
     };
-    expect(subtractValueQuantities([q1, q1])).toEqual({ assets: undefined, coins: 0n });
+    expect(subtractValueQuantities([q1, q1])).toEqual({ assets: new Map(), coins: 0n });
```
mkazlauskas, line 39: Does this mean it no longer works when assets are `undefined`?

packages/core/src/Asset/util/addTokenMaps.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  return result;
+};
+
+/**
+ * Given a list of TokenMaps, compute a TokenMap with the addition of all TokenMaps.
+ *
+ * @param assets The list of TokenMaps to add.
+ */
+export const addTokenMaps = (assets: (Cardano.TokenMap | undefined)[]): Cardano.TokenMap | undefined => {
```
mkazlauskas, line 57: This looks similar to https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/core/src/Asset/util/coalesceTokenMaps.ts#L5 . If the purpose is the same, then we should probably consolidate the 2

packages/core/src/util/transactionSummaryInspector.ts
```diff
@@ -0,0 +1,242 @@
...
+  depositParams: DepositProtocolParams,
+  certificates: Cardano.Certificate[],
+  rewardAccounts: Cardano.RewardAccount[],
+  poolIds: Set<Cardano.PoolId>,
+  networkId: Cardano.NetworkId
+): { deposit: Cardano.Lovelace; returnedDeposit: Cardano.Lovelace } => {
+  let deposit = 0n;
+  let returnedDeposit = 0n;
+
+  // TODO: For the case of deregistration (StakeDeregistration and PoolRetirement) the code here is not entirely correct
```
mkazlauskas, line 54: Let's capture this in a ticket. Maybe @mirceahasegan can advice an appropriate parent ticket

packages/core/src/util/transactionSummaryInspector.ts
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  return {
+    deposit,
+    returnedDeposit
+  };
+};
+
+/** Inspects a transaction for its deposits and returned deposits. */
+export const getDeposits = (
```
mkazlauskas, line 117: Consider consolidating this with `computeImplicitCoin`. Perhaps adding an optional `rewardAccounts` argument to it rather than reimplementing the computation entirely would be better?

packages/core/src/util/transactionSummaryInspector.ts
```diff
@@ -0,0 +1,242 @@
...
+  protocolParameters: Cardano.ProtocolParameters;
+}
+
+export type TransactionSummaryInspection = {
+  tokens: Cardano.TokenMap;
+  coins: Cardano.Lovelace;
+  collateral: Cardano.Lovelace;
+  deposit: Cardano.Lovelace;
+  returnDeposit: Cardano.Lovelace;
+  totalFees: Cardano.Lovelace;
```
mkazlauskas, line 31: Name implies that there is more than 1 type of fee accounted for here, which is not true. If we ever need to find other types of fees (e.g. dex fees), then those would probably be listed in a new property or separate inspectors to have more detail.
```suggestion
  fee: Cardano.Lovelace;
```

packages/core/src/util/transactionSummaryInspector.ts
```diff
@@ -0,0 +1,242 @@
...
+  };
+};
+
+/**
+ * Gets the collateral specified for this transaction.
+ *
+ * @param tx transaction to inspect.
+ * @param inputResolver input resolver.
+ */
+export const getCollateral = async (
```
mkazlauskas, line 152: @renanvalentin please review this function to confirm it matches requirements for your current task

packages/core/src/util/transactionSummaryInspector.ts
```diff
@@ -0,0 +1,242 @@
...
+  totalFees: Cardano.Lovelace;
+  unresolved: {
+    inputs: Cardano.TxIn[];
+    value: Cardano.Value;
+  };
+};
+
+export type TransactionSummaryInspector = (
+  args: TransactionSummaryInspectorArgs
+) => Inspector<TransactionSummaryInspection>;
```
mkazlauskas, line 40: Inspector currently takes in a `HydratedTx` which is a problem, because we want to use these new inspectors for tx confirmation. Can we change it to `Tx`? I think it's also assignable to `HydratedTx` so shouldn't break anything, but also some other inspectors might need `InputResolver` argument added.

mkazlauskas (changes requested):

packages/core/src/util/transactionSummaryInspector.ts
renanvalentin: @mkazlauskas I used @AngelCastilloB  as a base to also consider own vs foreign address. See https://github.com/input-output-hk/lace/pull/844/files#diff-3841c99e0a0dd2e29e4b14c462861db6300ba1fdd9889b82b79a901aae056096R4

packages/core/src/util/transactionSummaryInspector.ts
AngelCastilloB: this PR already makes this change @mkazlauskas , you can see it here in this commit:

https://github.com/input-output-hk/cardano-js-sdk/pull/1057/commits/cff6ef1dbca1c57c07ac102aba9a429fc1ffac2f

But I see I forgot to add it as a breaking change to the commit, message will update it

packages/util-dev/src/stringifyReplacers.ts
```diff
@@ -0,0 +1,28 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const stringifyReplacer = (_key: any, value: any) => {
```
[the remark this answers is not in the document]
AngelCastilloB: This is not currently used, but I added to the ulti-dev packkage, because is a pice of code I find myself often rewriting when debugin transaction objects, this is a replacer funciton top be use with the JSON.stringify:

`console.log(JSON.stringify(x, stringifyReplacer))`

If the serializable object can be used in the same way I will remove this.

packages/core/src/util/transactionSummaryInspector.ts
AngelCastilloB: I will use @renanvalentin version of the function then. Thanks @renanvalentin

packages/util-dev/src/stringifyReplacers.ts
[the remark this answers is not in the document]
mkazlauskas: Yep it can be used the same way. You can `JSON.stringify(toSerializableObj(x))`

packages/core/test/Cardano/util/subtractValueQuantities.test.ts
AngelCastilloB: I does work with undefined, the change in behavior here is that it will return an empty map when the subtraction of all assets is 0 (before it would return undefined here), I can make it return undefined in this case if we want to maintain this interface

mkazlauskas: Ah got it, it's fine

packages/util-dev/src/stringifyReplacers.ts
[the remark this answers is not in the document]
AngelCastilloB: commit that introduced this change was removed

packages/core/src/util/txInspector.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1057/commits/08dafef4829ec29541ec7ff4c3cbee7a3f3d6494

packages/core/src/Asset/util/addTokenMaps.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1057/commits/85564707d99c7b9eccd034de4c0c841263fce521

packages/core/src/util/transactionSummaryInspector.ts
AngelCastilloB: done: LW-9604

packages/core/src/util/transactionSummaryInspector.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1057/commits/289d353cfccce8f9b1f5d6871367e9234a559bce (and ifxed a few bugs with the type as well)

packages/core/src/util/transactionSummaryInspector.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1057/commits/0f21727f0f32928b0363582ca7592d38f850cc2e

packages/core/src/util/transactionSummaryInspector.ts
AngelCastilloB: I thought about this, but in the end I re-implemented the code because it is a bit different, computeImplicitCoin assumes all certificates belong to the wallet (I guess is because its only ever internally used by the PersonalWallet during input selection for transaction its building) and it consolidates the return deposits with the withdrawals (although it can be easily reversed with a subtraction). 

I will still need to have separate logic for deposits calculation if we make rewardAddresses optional. I can move the getDeposits function into the util and call that if the rewardAddresses are provided, and otherwise, call the original logic (but now that I look at it closely it probably also needs to be updated to work with conway certificates). 

Alternatively we can make the util take a mandatory rewardAddresses array rather than an optional one and we keep the logic the same for both cases. Let me know what you think (adding an extra field "returnedDeposits" would be also convenient)

mkazlauskas: Returning `returnedDeposits` (we call it 'deposit reclaim' elsewhere in code) from `computeImplicitCoin` would be good. However Making `rewardAddresses` mandatory wouldn't work for input selection, so you'd have to do some conditional logic there with optional addresses. I don't like having 2 utilities that do (pretty much) the same thing here. I think we should either:
- hoist `computeImplicitCoin` to `input-selection` package so that it's more explicitly scoped for input selection usage
- consolidate the 2 in core package and do conditional logic based on presence of rewardAddresses param. Even if implementation is different based on parameters, there would be some reuse, and more importantly there would be a single function to maintain

AngelCastilloB: done here @mkazlauskas https://github.com/input-output-hk/cardano-js-sdk/pull/1057/commits/293cb9a7df5fb1c52ada864c5ab4e19c606c3ce1

packages/core/src/util/transactionSummaryInspector.ts
```diff
@@ -0,0 +1,154 @@
...
+      protocolParameters,
+      { certificates: tx.body.certificates, withdrawals: tx.body.withdrawals },
+      rewardAccounts || []
+    );
+
+    const collateral = await getCollateral(tx, inputResolver, addresses);
+
+    const totalOutputValue = await totalAddressOutputsValueInspector(addresses)(tx);
+    const totalInputValue = await totalAddressInputsValueInspector(addresses, inputResolver)(tx);
+    const implicitCoin = implicit.withdrawals! + implicit.reclaimDeposit! - implicit.deposit!;
```
mkazlauskas, line 134: When would those be null/undefined? Shouldn't we use 0 as default then?
```suggestion
    const implicitCoin = implicit.withdrawals + implicit.reclaimDeposit - implicit.deposit;
```

mkazlauskas (approved):
Great work! Just a question below

packages/core/src/util/transactionSummaryInspector.ts
AngelCastilloB: I didn't understand why they are optional either, they should be 0, this is the code where it is defined:

```
/** Implicit coin quantities used in the transaction */
export interface ImplicitCoin {
  /** Reward withdrawals */
  withdrawals?: Lovelace;
  /** Reward withdrawals + deposit reclaims (total return) */
  input?: Lovelace;
  /** Delegation registration deposit */
  deposit?: Lovelace;
  /** Deposits returned */
  reclaimDeposit?: Lovelace;
}

```
I removed the ? from the fields and broke the property testing and it was not obvious how to fix them for me 😅 , so I left them optional

mkazlauskas: Got it, so I think the code in suggested change should do `|| 0n` for each of those

AngelCastilloB: done here: https://github.com/input-output-hk/cardano-js-sdk/pull/1057/commits/76d140d23288335920f0074dd659f7bd5f148109

mkazlauskas (dismissed):

mkazlauskas (approved):
Great work! :rocket:

rhyslbw (approved):
Great work @AngelCastilloB

PR #1060, opened 2024-01-19
fix(projection-typeorm): ignore orphan subhandles

mkazlauskas:
# Context

ADA Handle rules require all subhandles to have a parent handle. Therefore, we must validate that subhandle has a valid parent handle.

# Proposed Solution

Query for existence of parent handle before inserting a subhandle


iccicci (approved):
Good job!
Nice test refactoring as well

mirceahasegan (approved):
LGTM :muscle:

rhyslbw (approved):

PR #1061, opened 2024-01-20
fix(core): fromMetadatum now uses asset_name as utf8 when name field is missing

AngelCastilloB:
# Context

[CIP25](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0025) defines the name field as required, but some established projects (Cardano Crocs Club, Radioactive Crocs Club) do not follow the standard, relying just on the Asset name.

# Proposed Solution

Modify SDK to handle NFTs (v1) without the name field, using asset_name when name not available

- fromMetadatum now uses asset_name as utf8 when name field is missing for CIP-25 V1


mkazlauskas (approved):
Looks good! :rocket: 

I think `feat` would be a more appropriate commit type for this, but `fix` is fine too

VanessaPC (approved):
LGTM!

PR #1062, opened 2024-01-23
LW-9518 Use SIGINT to stop cardano-submit-api

iccicci:
# Context

Shutting down a docker compose infrastructure take a long time due to `cardano-subit-api` hangs.

# Proposed Solution

Changed its stop signal into `SIGINT`.

### DIFF
--- a/compose/common.yml
+++ b/compose/common.yml
@@ -177,6 +177,7 @@ services:
     ports:
       - 8090:8090
     restart: on-failure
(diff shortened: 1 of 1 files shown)


rhyslbw (approved):
Nice catch @iccicci

mirceahasegan (approved):
Nice catch! :male_detective:

PR #1068, opened 2024-01-25
fix: ignore SigningCoordinator responses from unintended targets

mkazlauskas:
# Context

Wallet can have more than 1 UI open at the same time (e.g. multiple tabs).
Each of them exposes a SigningCoordinator and receives all sign requests.
However, only one of them is the intended target (the one that has
subscriptions at the time). Other coordinators can safely ignore the sign request.

LW-9094

# Proposed Solution

Define a new error type in web-extension messaging and use it to ignore the responses.

Also fix large collection storage error LW-9628


rhyslbw (approved):
Nice work @mkazlauskas

mirceahasegan (approved):
Great work @mkazlauskas ! 🥇

mirceahasegan (approved):

PR #1069, opened 2024-01-25
Refactor/lw 9414 unify deposits calculation

mirceahasegan:
# Context

Take drep registration/retirement in account when calculating deposit and reclaims.
Unify own account and any account computation functions.

### DIFF
--- a/packages/core/src/Cardano/util/computeImplicitCoin.ts
+++ b/packages/core/src/Cardano/util/computeImplicitCoin.ts
@@ -1,5 +1,6 @@
 /* eslint-disable sonarjs/cognitive-complexity */
 /* eslint-disable complexity */
+import * as Crypto from '@cardano-sdk/crypto';
 import { BigIntMath } from '@cardano-sdk/util';
 import { Cardano } from '../..';
 import { HydratedTxBody, Lovelace } from '../types';
@@ -75,7 +76,8 @@ const computeShellyDeposits = (
 
 const computeConwayDeposits = (
   certificates: Cardano.Certificate[],
-  rewardAccounts: Cardano.RewardAccount[]
+  rewardAccounts: Cardano.RewardAccount[],
+  dRepKeyHash?: Crypto.Ed25519KeyHashHex
 ): { deposit: Cardano.Lovelace; reclaimDeposit: Cardano.Lovelace } => {
   let deposit = 0n;
   let reclaimDeposit = 0n;
@@ -91,6 +93,18 @@ const computeConwayDeposits = (
       case Cardano.CertificateType.Unregistration:
         if (stakeCredentialInRewardAccounts(cert.stakeCredential, rewardAccounts)) reclaimDeposit += cert.deposit;
         break;
+      case Cardano.CertificateType.RegisterDelegateRepresentative:
+      case Cardano.CertificateType.UnregisterDelegateRepresentative:
+        if (
+          !dRepKeyHash ||
+          (cert.dRepCredential.type === Cardano.CredentialType.KeyHash &&
+            cert.dRepCredential.hash === Crypto.Hash28ByteBase16.fromEd25519KeyHashHex(dRepKeyHash))
+        ) {
+          cert.__typename === Cardano.CertificateType.RegisterDelegateRepresentative
+            ? (deposit += cert.deposit)
+            : (reclaimDeposit += cert.deposit);
+        }
+        break;
     }
   }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/Cardano/util/computeImplicitCoin.ts
```diff
@@ -93,6 +93,15 @@ const computeConwayDeposits = (
       case Cardano.CertificateType.Unregistration:
         if (stakeCredentialInRewardAccounts(cert.stakeCredential, rewardAccounts)) reclaimDeposit += cert.deposit;
         break;
+      case Cardano.CertificateType.RegisterDelegateRepresentative:
+        // REVIEW: assuming a single drep scenario. If this should work in a multiple account scenario,
```
mirceahasegan, line 97: @AngelCastilloB , @mkazlauskas, is the context of multi-sig, should I include an "ownDrepId" parameter to the function?

packages/core/src/Cardano/util/computeImplicitCoin.ts
```diff
@@ -16,15 +18,30 @@ export interface ImplicitCoin {
...
+  rewardAccounts: Cardano.RewardAccount[]
 ): { deposit: Cardano.Lovelace; reclaimDeposit: Cardano.Lovelace } => {
   let deposit = 0n;
   let reclaimDeposit = 0n;
+  const anyRewardAccount = rewardAccounts.length === 0;
+
+  const poolIds = new Set(
+    rewardAccounts
+      .map((account) => Cardano.RewardAccount.toHash(account))
+      .map((hash) => Cardano.PoolId.fromKeyHash(hash))
```
iccicci, line 43:
```suggestion
    rewardAccounts.map((account) => Cardano.RewardAccount.toHash(account).Cardano.PoolId.fromKeyHash(hash))
```

Could this work as well? It would use une less `map`.

mirceahasegan: done here [cbc1744](https://github.com/input-output-hk/cardano-js-sdk/pull/1069/commits/cbc174488c54ff0d55d91d5b1b9559569aba5cbc)

packages/core/src/Cardano/util/computeImplicitCoin.ts
mkazlauskas: I think we need to include it even for single wallets. This is now also used for producing transaction summary and reclaiming a deposit is basically spending user's owned funds. Similarly for registering/depositing: it is relevant to the user how much of the funds in the transaction he remains in control of

packages/core/src/Cardano/util/computeImplicitCoin.ts
```diff
@@ -89,67 +103,27 @@ const computeConwayDeposits = (
...
 
 /** Inspects a transaction for its deposits and returned deposits. */
-const getTxOwnDeposits = (
+const getTxDeposits = (
   { stakeKeyDeposit, poolDeposit }: Pick<Cardano.ProtocolParameters, 'stakeKeyDeposit' | 'poolDeposit'>,
   certificates: Cardano.Certificate[],
-  rewardAccounts: Cardano.RewardAccount[]
+  rewardAccounts: Cardano.RewardAccount[] = []
 ): { deposit: Cardano.Lovelace; reclaimDeposit: Cardano.Lovelace } => {
-  if (rewardAccounts.length === 0 || certificates.length === 0) return { deposit: 0n, reclaimDeposit: 0n };
```
mirceahasegan, line 97: @AngelCastilloB, what is the intent of `rewardAccounts.length === 0`?
Is it to avoid collecting/reclaiming deposits when the user has no reward accounts?

AngelCastilloB: This function (computeImplicitCoin) is used by the input selector and our util to compute transaction summary/display. The input selector doesn't have reward accounts to provide, but because we are building the transaction internally we know all the certificates are ours, so in this case the input selector pass an empy/undefined array to this function. However the transaction summary/display is receiving a transaction from a dApp, and can have mixed certificates (foreign and ours), so we need the list of reward accounts to be able to distinguish which deposits are going to our rewardaccounts from the ones that could potentially go to a different reward accounts that we dont control (same with reclaims)

mkazlauskas: Nice summary, maybe worth putting it as a comment in code?

mirceahasegan: done here [3a4da4f](https://github.com/input-output-hk/cardano-js-sdk/pull/1069/commits/3a4da4fa07dacc071d0f3cd0e567333f491c261a)

packages/core/src/Cardano/util/computeImplicitCoin.ts
mirceahasegan: done here [3a4da4f](https://github.com/input-output-hk/cardano-js-sdk/pull/1069/commits/3a4da4fa07dacc071d0f3cd0e567333f491c261a)

mkazlauskas (approved):
Nice work! :rocket:

iccicci (approved):
Great work! 🤝

PR #1074, opened 2024-01-31
Fix/lw 9414 deposit at unregister stake key on master

mirceahasegan:
# Context

Certificates that register stake key include the deposit value, as configured in the protocol parameters at the time of submitting the transaction

Currently, the deposit value is taken from the protocol parameters, and used as [implicit coin](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/core/src/Cardano/util/computeImplicitCoin.ts#L16) when deregistering the stake key

In the conway era, the protocol parameters could change often as a  result of a governance action, leading to a discrepancy in the deposit value at stake key registration and the deposit value at stake key deregistration.

# Proposed Solution
For this reason, the stake key deposit value at the time of registering the stake key should be stored in RewardAccountInfo from ObservableWallet.delegation.rewardAccounts$.

It could for example be undefined when the stake key is not registered, and the real value when the stake key is registered.

This way, on deregistering a stake key, SDK can use this value as deposit value.

### DIFF
--- a/packages/core/src/Cardano/types/Certificate.ts
+++ b/packages/core/src/Cardano/types/Certificate.ts
@@ -5,6 +5,7 @@ import { EpochNo } from './Block';
 import { Hash28ByteBase16 } from '@cardano-sdk/crypto';
 import { Lovelace } from './Value';
 import { PoolId, PoolParameters } from './StakePool';
+import { isNotNil } from '@cardano-sdk/util';
 
 export enum CertificateType {
   StakeRegistration = 'StakeRegistrationCertificate',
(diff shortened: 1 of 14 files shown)


packages/core/src/util/txInspector.ts
```diff
@@ -167,16 +169,18 @@ export const getCertificatesByType = (
...
 
   return certificates.filter((certificate) => {
-    if ('stakeCredential' in certificate && certificate.stakeCredential) {
+    if (isCertType(certificate, StakeCredentialCertificateTypes)) {
       const credHash = Crypto.Ed25519KeyHashHex(certificate.stakeCredential.hash);
       return rewardAccounts.some((account) => RewardAccount.toHash(account) === credHash);
     }
 
-    if ('poolParameters' in certificate) return rewardAccounts.includes(certificate.poolParameters.rewardAccount);
+    if (isCertType(certificate, [CertificateType.PoolRegistration]) && 'poolParameters' in certificate)
```
mirceahasegan: I missed this one. Added a fixup to address it
```diff
-    if (isCertType(certificate, [CertificateType.PoolRegistration]) && 'poolParameters' in certificate)
+    if (isCertType(certificate, [CertificateType.PoolRegistration]))

File: packages/core/src/Cardano/types/Certificate.ts [hunk 2]
```diff
@@ -215,6 +207,40 @@ export const RegAndDeregCertificateTypes = [
...
+  [CertificateType.StakeVoteRegistrationDelegation]: StakeVoteRegistrationDelegationCertificate;
+  [CertificateType.UnregisterDelegateRepresentative]: UnRegisterDelegateRepresentativeCertificate;
+  [CertificateType.Unregistration]: NewStakeAddressCertificate;
+  [CertificateType.UpdateDelegateRepresentative]: UpdateDelegateRepresentativeCertificate;
+  [CertificateType.VoteDelegation]: VoteDelegationCertificate;
+  [CertificateType.VoteRegistrationDelegation]: VoteRegistrationDelegationCertificate;
+};
+
+/** Checks if {@link certificate} is one {@link certificateTypes}, and narrows down its type */
+export const isCertType = <K extends keyof CertificateTypeMap>(
```
iccicci, line 239: Very nice!

packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
```diff
@@ -73,10 +73,9 @@ export const certificateTransactionsWithEpochs = (
 
 const hasDelegationCert = (certificates: Array<Cardano.Certificate> | undefined): boolean =>
   !!certificates &&
-  certificates.some((cert) =>
-    [...Cardano.RegAndDeregCertificateTypes, ...Cardano.StakeDelegationCertificateTypes].includes(
-      cert.__typename as Cardano.RegAndDeregCertificateTypes | Cardano.StakeDelegationCertificateTypes
-    )
+  certificates.some(
+    (cert) =>
+      !!Cardano.isCertType(cert, [...Cardano.RegAndDeregCertificateTypes, ...Cardano.StakeDelegationCertificateTypes])
```
iccicci, line 78: `!!` operator seems to be redundant: type checkers return `boolean`s.

iccicci (changes requested):
Great job! Well done!

Just a minor minor minor note

packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
mirceahasegan: Done here: [5138707](https://github.com/input-output-hk/cardano-js-sdk/pull/1074/commits/5138707eaf274991bdc1231ba64e14b79c2ba4db)

packages/core/src/util/txInspector.ts
mirceahasegan: Done here: [efd05a8](https://github.com/input-output-hk/cardano-js-sdk/pull/1074/commits/efd05a8d06766935d33d9a2c6e5ea726c9c5b29c)

iccicci (dismissed):

mkazlauskas (changes requested):
Code looks good, but would be great to add basic some tests for new utils that are exported from index file(s)

mirceahasegan:
> Code looks good, but would be great to add basic some tests for new utils that are exported from index file(s)

Done here: [65880a5](https://github.com/input-output-hk/cardano-js-sdk/pull/1074/commits/65880a50ef18288890bbfbc3d83fdaacdbb0ed4f) and [cc0c3c7](https://github.com/input-output-hk/cardano-js-sdk/pull/1074/commits/cc0c3c7c1040b07273451f0fcb50e1f254c83ad1)

mkazlauskas (approved):
Nice work! :rocket:

AngelCastilloB (approved):
LGTM @mirceahasegan

PR #1075, opened 2024-01-31
Feat/lw 9673 create new asset provider that uses personalwallet cache

AngelCastilloB:
# Context

In order to properly display the token metadata in the transaction summary, we must fetch the AssetInfo of all tokens in the transaction, this involves fetching from the backend, but we can also laverage the ObservableWallet assets$ cache, and we must also consider tokens that are not present on chain but are currently being minted, which involves resolving the asset info from the Tx mint and auxData fields.

# Proposed Solution

Implement new AssetProvider that looks up (in this order)

1. ObservableWallet.assetInfo$
3. Datum (cip68)
4. Metadata (CIP25)
5. cardano-services AssetProvider

### DIFF
--- a/packages/wallet/src/services/WalletAssetProvider.ts
+++ b/packages/wallet/src/services/WalletAssetProvider.ts
@@ -1,0 +1,233 @@
+import { Asset, AssetProvider, Cardano, GetAssetArgs, GetAssetsArgs, HealthCheckResponse } from '@cardano-sdk/core';
+import { Assets } from '../types';
+import { Logger } from 'ts-log';
+import { Observable, firstValueFrom } from 'rxjs';
+import { isNotNil } from '@cardano-sdk/util';
+
+export interface AssetProviderContext {
+  assetProvider: AssetProvider;
+  assetInfo$: Observable<Assets>;
+  tx?: Cardano.Tx;
+  logger: Logger;
+}
+
+const tryCip68NftMetadata = (
+  policyId: Cardano.PolicyId,
+  name: Cardano.AssetName,
+  tx: Cardano.Tx,
+  logger: Logger
+): Asset.NftMetadata | null => {
+  const decoded = Asset.AssetNameLabel.decode(name);
+
+  if (decoded?.label === Asset.AssetNameLabelNum.UserNFT) {
+    const referenceAssetId = Cardano.AssetId.fromParts(
+      policyId,
+      Asset.AssetNameLabel.encode(decoded.content, Asset.AssetNameLabelNum.ReferenceNFT)
+    );
+
+    // TODO: It is possible that the reference NFT is not in one of the outputs of the transaction and was previously minted. We
+    // need a way to find the reference NFT TxOut from the current active UTXO set on the network.
+    for (const output of tx.body.outputs) {
+      if (output.value.assets?.get(referenceAssetId)) {
+        return Asset.NftMetadata.fromPlutusData(output.datum, logger);
+      }
+    }
+  }
+
+  return null;
+};
+
+const getNftMetadata = (
+  name: Cardano.AssetName,
+  policyId: Cardano.PolicyId,
+  tx: Cardano.Tx,
+  logger: Logger
+): Asset.NftMetadata | null => {
+  // First, try CIP-68
+  let metadata = tryCip68NftMetadata(policyId, name, tx, logger);
+
+  // If metadata is not found, try CIP-25
+  if (!metadata) {
+    metadata = tx.auxiliaryData?.blob
+      ? Asset.NftMetadata.fromMetadatum({ name, policyId }, tx.auxiliaryData.blob, logger)
+      : null;
+  }
+
+  return metadata;
+};
+
+const createAssetInfo = (assetId: Cardano.AssetId, amount: bigint, tx: Cardano.Tx, logger: Logger): Asset.AssetInfo => {
+  const name = Cardano.AssetId.getAssetName(assetId);
+  const policyId = Cardano.AssetId.getPolicyId(assetId);
+  const assetInfo: Asset.AssetInfo = {
+    assetId,
+    fingerprint: Cardano.AssetFingerprint.fromParts(policyId, name),
+    name,
+    policyId,
+    quantity: amount,
+    supply: amount
+  };
+
+  assetInfo.nftMetadata = getNftMetadata(name, policyId, tx, logger);
+
+  return assetInfo;
+};
+
+const getMintedAssetInfosFromTx = async (tx: Cardano.Tx, logger: Logger): Promise<Asset.AssetInfo[] | null> => {
+  const mints = tx.body.mint;
+
+  if (!mints) return null;
+
+  return [...mints.entries()]
+    .filter(([_, amount]) => amount > 0)
+    .map(([assetId, amount]) => createAssetInfo(assetId, amount, tx, logger));
+};
+
+const fetchAssetsFromProvider = async (
+  provider: AssetProvider,
+  assetIds: Cardano.AssetId[],
+  logger: Logger
+): Promise<Asset.AssetInfo[]> => {
+  const assetsFromProvider: Asset.AssetInfo[] = [];
+
+  // We need to fetch assets one by one because the provider will throw if any of the assets requests to the getAssets endpoint is not found.
+  // We want to fetch the ones we can and return a simplified AssetInfo for the ones we can't.
+  for (const assetId of assetIds) {
+    try {
+      const fetchedAsset = await provider.getAsset({
+        assetId,
+        extraData: { nftMetadata: true, tokenMetadata: true }
+      });
+      assetsFromProvider.push(fetchedAsset);
+    } catch (error) {
+      logger.error(error);
+    }
+  }
+
+  return assetsFromProvider;
+};
+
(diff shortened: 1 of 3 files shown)


packages/wallet/src/services/WalletAssetProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+): Asset.NftMetadata | null => {
+  const decoded = Asset.AssetNameLabel.decode(name);
+  if (decoded?.label === Asset.AssetNameLabelNum.UserNFT) {
+    const referenceAssetId = Cardano.AssetId.fromParts(
+      policyId,
+      Asset.AssetNameLabel.encode(decoded.content, Asset.AssetNameLabelNum.ReferenceNFT)
+    );
+
+    // TODO: It is possible that the reference NFT is not in one of the outputs of the transaction and was previously minted. We
+    // need a way to find the reference NFT TxOut from the current active UTXO set on the network.
```
mkazlauskas, line 28: Yes, we're currently projecting reference nft metadata even when user token does not exist yet, but I think there's no way to query it. From what I've seen, reference NFTs are being minted in the same tx, but it doesn't have to be. Let's consider this an edge case and address this in a separate task:
- add `AssetProvider.getNftMetadata`
- use it here when not found

packages/wallet/src/services/WalletAssetProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+import { isNotNil } from '@cardano-sdk/util';
+
+export interface AssetProviderContext {
+  assetProvider: AssetProvider;
+  assetInfo$: Observable<Assets>;
+  tx: Cardano.Tx;
+  logger: Logger;
+}
+
+const tryCip68Metadata = (
```
mkazlauskas, line 14: There's also Cip68 FT metadata (333 label) which we're not handling. In fact, we don't even have utils to parse it yet, and it's not being projected. Let's be more explicit here.
```suggestion
const tryCip68NftMetadata = (
```

packages/wallet/src/services/WalletAssetProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+  return null;
+};
+
+const getNftMetadata = (
+  name: Cardano.AssetName,
+  policyId: Cardano.PolicyId,
+  tx: Cardano.Tx,
+  logger: Logger
+): Asset.NftMetadata | null => {
+  // First, try CIP-25
```
mkazlauskas, line 45: Our backend logic looks for CIP-68 metadata first, let's reorder those here too for consistency

packages/wallet/src/services/WalletAssetProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+
+const fetchAssetsFromProvider = async (
+  provider: AssetProvider,
+  assetIds: Cardano.AssetId[],
+  logger: Logger
+): Promise<Asset.AssetInfo[]> => {
+  const assetsFromProvider: Asset.AssetInfo[] = [];
+
+  // TODO: We need to fetch assets one by one because the provider will throw if any of the assets requests to the getAssets endpoint is not found.
+  // We want to fetch the ones we can and return a simplified AssetInfo for the ones we can't.
```
mkazlauskas, line 93: Makes sense. Let's remove the 'TODO:' prefix of this comment

packages/wallet/src/services/WalletAssetProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+const createAssetInfo = (assetId: Cardano.AssetId, amount: bigint, tx: Cardano.Tx, logger: Logger): Asset.AssetInfo => {
+  const name = Cardano.AssetId.getAssetName(assetId);
+  const policyId = Cardano.AssetId.getPolicyId(assetId);
+  const assetInfo: Asset.AssetInfo = {
+    assetId,
+    fingerprint: Cardano.AssetFingerprint.fromParts(policyId, name),
+    name,
+    policyId,
+    quantity: 0n,
+    supply: amount
```
mkazlauskas, line 67: `quantity` was renamed to `supply` and is the same value, but is a deprecated field
```suggestion
    quantity: amount,
    supply: amount
```

packages/wallet/src/services/WalletAssetProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+    if (!asset && !mintedAsset) {
+      return createFallbackAsset(assetId);
+    }
+
+    if (!asset && mintedAsset) {
+      return mintedAsset;
+    }
+
+    if (asset && mintedAsset) {
+      asset.supply += mintedAsset.supply;
```
mkazlauskas, line 141:
```suggestion
      asset.supply += mintedAsset.supply;
      asset.quantity = asset.supply;
```

packages/wallet/src/services/WalletAssetProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+    if (!asset && !mintedAsset) {
+      const name = Cardano.AssetId.getAssetName(assetId);
+      const policyId = Cardano.AssetId.getPolicyId(assetId);
+      return {
+        assetId,
+        fingerprint: Cardano.AssetFingerprint.fromParts(policyId, name),
+        name,
+        policyId,
+        quantity: 0n,
+        supply: 0n
```
mkazlauskas, line 183: How about this? Same suggestion for createFallbackAsset. Either way, I think it's important to add a doc comment explaining this behavior when asset info couldn't be resolved, because without looking at the code I would expect it to throw instead.
```suggestion
        quantity: Number.NaN,
        supply: Number.NaN
```

mkazlauskas (changes requested):

packages/wallet/src/services/WalletAssetProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+      };
+    }
+
+    if (!asset) return mintedAsset!;
+
+    if (mintedAsset) {
+      asset.supply += mintedAsset.supply;
+
+      // We give preference to the metadata in the transaction if preset as this would be the most up to date.
```
mkazlauskas, line 192: Actually for cip-68 metadata, tokens don't have to be minted to update metadata. We might have a logic error here - we should probably always search for reference tokens in outputs.
```suggestion
    if (mintedAsset) {
      asset.supply += mintedAsset.supply;

      // We give preference to the metadata in the transaction if preset as this would be the most up to date.
```

AngelCastilloB: So we need to do the search for all tokens, see if they are CIP-68 and then check again vs the outputs. I will add a final step to do this

packages/wallet/src/services/WalletAssetProvider.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1075/commits/80c61e293ca1968f366c665694957293a8d41bfe

packages/wallet/src/services/WalletAssetProvider.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1075/commits/bfff8ef2a42b4b6951f6885ecf47e961de148fca

packages/wallet/src/services/WalletAssetProvider.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1075/commits/5a6bfa60663fa702b84ed715740047fde129cc6f

packages/wallet/src/services/WalletAssetProvider.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1075/commits/c39bcf81694122eb4d49fef57ab00e41b3480634

packages/wallet/src/services/WalletAssetProvider.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1075/commits/c39bcf81694122eb4d49fef57ab00e41b3480634

packages/wallet/src/services/WalletAssetProvider.ts
AngelCastilloB: I cant assing Number.NaN since those are bigints, but I think 0 supply could be also good? since 0 supply is not valid (tokens must have at least 1 to exists on chain). I did add the documentation as suggested tho:

https://github.com/input-output-hk/cardano-js-sdk/pull/1075/commits/c39bcf81694122eb4d49fef57ab00e41b3480634

packages/wallet/src/services/WalletAssetProvider.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1075/commits/eb8eb75ad82bcefdee30286759d3114cfd6f5580

packages/wallet/src/services/WalletAssetProvider.ts
AngelCastilloB: separate task was logged

mkazlauskas (approved):
Great work! :rocket:

VanessaPC (approved):
Nice work!

PR #1076, opened 2024-01-31
Feat/lw 9670 tx summary inspector and token transfer inspector must return assetinfo

AngelCastilloB:
# Context

Current implementation of the inspectors returns the raw asset ids that are present on the transaction. Lace must then fetch the AssetInfo and combine this information with the mint and metadata details of the transaction to properly display the tokens or NFTs.

This logic is probably better encapsulated in the inspector itself by simply adding an extra parameter (AssetProvider).

# Proposed Solution

The inspectors now take an extra argument, asset provider

### DIFF
--- a/packages/core/src/util/tokenTransferInspector.ts
+++ b/packages/core/src/util/tokenTransferInspector.ts
@@ -1,16 +1,33 @@
 import * as Cardano from '../Cardano';
+import { AssetInfo } from '../Asset';
+import { AssetProvider } from '../Provider';
 import { Inspector, resolveInputs } from './txInspector';
 import { coalesceValueQuantities } from './coalesceValueQuantities';
 import { subtractValueQuantities } from './subtractValueQuantities';
 import uniq from 'lodash/uniq';
 
-export type TokenTransferInspection = {
-  fromAddress: Map<Cardano.PaymentAddress, Cardano.Value>;
-  toAddress: Map<Cardano.PaymentAddress, Cardano.Value>;
+export type AssetInfoWithAmount = { amount: Cardano.Lovelace; assetInfo: AssetInfo };
+
+export type TokenTransferValue = {
+  assets: Map<Cardano.AssetId, AssetInfoWithAmount>;
+  coins: Cardano.Lovelace;
 };
 
+export type TokenTransferInspection = {
+  fromAddress: Map<Cardano.PaymentAddress, TokenTransferValue>;
+  toAddress: Map<Cardano.PaymentAddress, TokenTransferValue>;
+};
+
+/** Arguments for the token transfer inspector. */
 export interface TokenTransferInspectorArgs {
+  /** The input resolver. */
   inputResolver: Cardano.InputResolver;
+
+  /** The asset provider to resolve AssetInfo for assets in the fromAddress field. */
+  fromAddressAssetProvider: AssetProvider;
+
+  /** The asset provider to resolve AssetInfo for assets in the toAddress field. */
+  toAddressAssetProvider: AssetProvider;
 }
 
 export type TokenTransferInspector = (args: TokenTransferInspectorArgs) => Inspector<TokenTransferInspection>;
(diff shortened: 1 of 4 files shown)


packages/core/package.json
```diff
@@ -37,6 +37,7 @@
     "prepack": "yarn build"
   },
   "devDependencies": {
+    "@cardano-sdk/util-dev": "workspace:~",
```
mkazlauskas, line 40: This creates a circular dependency. I suggest following the existing pattern to use relative import instead (see `packages/core/test/Cardano/Address/Address.test.ts`). There must be a better way to do it, but for now that's what we have.

packages/core/src/util/transactionSummaryInspector.ts
```diff
@@ -111,14 +115,38 @@ const getUnaccountedFunds = async (
   return subtractValueQuantities([totalOutputs, totalInputs]);
 };
 
+const toAssetInfoWithAmount = async (
+  assetProvider: AssetProvider,
+  tokenMap?: TokenMap
+): Promise<Map<Cardano.AssetId, AssetInfoWithAmount>> => {
+  if (!tokenMap) return new Map();
+
+  const assetIds = uniq(tokenMap && tokenMap.size > 0 ? [...tokenMap.keys()] : []);
```
mkazlauskas, line 124: A `Map` should already have uniq keys :thinking: 
```suggestion
  const assetIds =tokenMap && tokenMap.size > 0 ? [...tokenMap.keys()] : [];
```

mkazlauskas (changes requested):

packages/core/src/util/tokenTransferInspector.ts
```diff
@@ -118,5 +160,8 @@ export const tokenTransferInspector: TokenTransferInspector =
     removeZeroBalanceEntries(fromAddress);
     removeZeroBalanceEntries(toAddress);
 
-    return { fromAddress, toAddress };
+    return {
+      fromAddress: await toTokenTransferValue(assetProvider, fromAddress),
```
mkazlauskas, line 164: It would be great to **not** look up cip68 metadata in `fromAddress`: there will be dapps that makes transactions to 'mutate' metadata, so ideally wallet would show old metadata in 'fromAddress' and new metadata in 'toAddress'

DominikGuzei (dismissed):
great work 🎉 except for @mkazlauskas feedback LGTM

packages/core/package.json
AngelCastilloB: Ah, I didn't check if util-dev had a dependency on core, my bad (I was expecting it not to)

packages/core/src/util/tokenTransferInspector.ts
AngelCastilloB: Ok so basically, in the fromAddress we do not want to check the transaction for metadata, and only rely on cache and backend correct?

mkazlauskas: Correct

packages/core/package.json
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1076/commits/d747a9da096e4db7f5c85e34d9afbb7b01f697b4

packages/core/src/util/transactionSummaryInspector.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1076/commits/2ee48609f1598d9c98ec1a235d796302a0c5b4e6

packages/core/src/util/tokenTransferInspector.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1076/commits/f16e095a0f51587495ca35b81ff2538b4baab671

mkazlauskas (approved):
Great work! Let's not forget to add an option that enables/disables looking up tx metadata in #1075

AngelCastilloB:
> Great work! Let's not forget to add an option that enables/disables looking up tx metadata in #1075

Done here https://github.com/input-output-hk/cardano-js-sdk/pull/1075/commits/ec38d9a25c0b07d16b553e71710b74c0494ac118 @mkazlauskas

mkazlauskas (approved):

VanessaPC (approved):
great work!

PR #1077, opened 2024-02-01
feat(cardano-services): resolve collateralReturn property

renanvalentin:
# Context

The confirmed tx body was missing the `collateralReturn` property, which is required for displaying collaterals in Lace's activity detail view.

### DIFF
--- a/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
+++ b/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
@@ -51,7 +51,7 @@ export class ChainHistoryBuilder {
   public async queryTransactionInputsByIds(ids: string[], collateral = false): Promise<TxInput[]> {
     this.#logger.debug(`About to find inputs (collateral: ${collateral}) for transactions with ids:`, ids);
     const result: QueryResult<TxInputModel> = await this.#db.query({
-      name: `tx_${collateral ? 'collateral' : 'inputs'}_by_tx_ids`,
+      name: `tx_${collateral ? 'collateral_' : ''}inputs_by_tx_ids`,
       text: collateral ? Queries.findTxCollateralsByIds : Queries.findTxInputsByIds,
       values: [ids]
     });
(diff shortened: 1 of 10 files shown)


packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
```diff
@@ -82,6 +82,20 @@ export class ChainHistoryBuilder {
     return result.rows.map((txOut) => mapTxOutModel(txOut, multiAssets.get(txOut.id)));
   }
 
+  public async queryCollateralOutputsByTxIds(ids: string[]): Promise<TxOutput[]> {
```
iccicci, line 85: Why this new method rather than adding a new parameter (`collateral = false`) to `queryTransactionOutputsByIds` as `queryTransactionInputsByIds` already has?

I see the required mechanics are already implemented in the `queries.ts` file.

iccicci (changes requested):

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
renanvalentin: I intended to make it explicit and avoid changing any existing code. Should I follow the same pattern as `queryTransactionInputsByIds`?

iccicci: Oh, thank you for intention, but yes, I think it's better to follow the same pattern as `queryTransactionInputsByIds`, that will reduce code repetition.

Thank you

renanvalentin: @iccicci see https://github.com/input-output-hk/cardano-js-sdk/pull/1077/files#diff-89b29702e05c8f20b81c0e9180f9da3d1d1edaf833180de1c61f3d0a5888ab65R71

iccicci: Very nice, but you made me notice another problem (originated not by this change, but already present):

Now, both `queryTransactionInputsByIds` and `queryTransactionOutputsByIds` produce the same statement name when they have `collateral = true`.

Better to change
```
      name: `tx_${collateral ? 'collateral' : 'outputs'}_by_tx_ids`,
```
into
```
      name: `tx_${collateral ? 'collateral_' : ''}outputs_by_tx_ids`,
```

(and same change in `queryTransactionInputsByIds`, thank you)

renanvalentin: Good catch! https://github.com/input-output-hk/cardano-js-sdk/pull/1077/commits/983495ed1552b02554ed0c34a463d478d9c5b21d

iccicci (dismissed):
Very nice!

mkazlauskas (dismissed):
Great work! :rocket:

iccicci (approved):

PR #1082, opened 2024-02-02
feat(wallet): added a new input resolver decorator that also fetches from the backend

AngelCastilloB:
# Context

In order to inspect the transaction that’s build by the dapp, input resolver that we have in WalletUtil is not going to work because it only resolves own addresses.

Create a new input resolver that utilizes the existing one and queries chain history provider for inputs that are not resolved as own

# Proposed Solution

Add a decorator to the inputResolver that additional resolve inputs from the backend if the inputs are not found in our UTXO set.

### DIFF
--- a/packages/wallet/src/services/WalletUtil.ts
+++ b/packages/wallet/src/services/WalletUtil.ts
@@ -23,11 +23,20 @@ export interface WalletOutputValidatorContext {
 export type WalletUtilContext = WalletOutputValidatorContext & InputResolverContext;
 
 export const createInputResolver = ({ utxo }: InputResolverContext): Cardano.InputResolver => ({
-  async resolveInput(input: Cardano.TxIn) {
+  async resolveInput(input: Cardano.TxIn, options?: Cardano.ResolveOptions) {
     const utxoAvailable = await firstValueFrom(utxo.available$);
     const availableUtxo = utxoAvailable?.find(([txIn]) => txInEquals(txIn, input));
-    if (!availableUtxo) return null;
-    return availableUtxo[1];
+
+    if (availableUtxo) return availableUtxo[1];
+
+    if (options?.hints) {
+      const tx = options?.hints.find((hint) => hint.id === input.txId);
+
+      if (tx && tx.body.outputs.length > input.index) {
+        return tx.body.outputs[input.index];
+      }
+    }
+    return null;
   }
 });
 
@@ -59,7 +68,14 @@ export const createBackendInputResolver = (provider: ChainHistoryProvider): Card
   };
 
   return {
-    async resolveInput(input: Cardano.TxIn) {
+    async resolveInput(input: Cardano.TxIn, options?: Cardano.ResolveOptions) {
+      // Add hints to the cache
+      if (options?.hints) {
+        for (const hint of options.hints) {
+          txCache.set(hint.id, hint);
+        }
+      }
+
       const tx = await fetchAndCacheTransaction(input.txId);
       if (!tx) return null;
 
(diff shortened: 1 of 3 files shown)


AngelCastilloB:
@mkazlauskas I also added an optional parameters that allow us to pass an array of local transaction to the input resolver, if given these will be used before fetching from the backend. I am thinking this could be useful for transaction chaining in the future.

packages/wallet/src/services/WalletUtil.ts
```diff
@@ -31,6 +31,61 @@ export const createInputResolver = ({ utxo }: InputResolverContext): Cardano.Inp
...
+ * @param inputResolver The original input resolver to be decorated.
+ * @param provider The backend provider used to fetch transactions by their hashes if
+ * they are not found by the inputResolver.
+ * @param localTxs An optional array of transactions to search before querying the backend.
+ * @returns A decorated input resolver with enhanced functionality to fetch unresolved inputs from the backend.
+ */
+export const createBackendInputResolver = (
+  inputResolver: Cardano.InputResolver,
+  provider: ChainHistoryProvider,
+  localTxs: Cardano.Tx[] = []
```
mkazlauskas, line 53: How about this instead? 
```ts
const resolver = combineInputResolvers(walletUtil, createBackendInputResolver)
```

This way we could have a resolver that lives longer (not recreated every time) and has a more effective cache. We already have a local InputResolver in WalletUtil so we could just pass that here.

> I am thinking this could be useful for transaction chaining in the future (or to have the resolver look into any transaction not yet on chain)

That makes sense, we could then add an optional parameter to `resolveInput` function. There's been some discussion about adding this parameter to cip30 signTx.

DominikGuzei (approved):

packages/wallet/src/services/WalletUtil.ts
AngelCastilloB: done here: https://github.com/input-output-hk/cardano-js-sdk/pull/1082/commits/d0dc28827485d2744c7fd9aba045a3066c2e4417 and here: https://github.com/input-output-hk/cardano-js-sdk/pull/1082/commits/c90750bed3c30425a8237f053577d980229dc926

packages/wallet/src/services/WalletUtil.ts
```diff
@@ -23,11 +23,79 @@ export interface WalletOutputValidatorContext {
...
+  const txCache = new Map<Cardano.TransactionId, Cardano.Tx>();
+
+  const fetchAndCacheTransaction = async (txId: Cardano.TransactionId): Promise<Cardano.Tx | null> => {
+    if (txCache.has(txId)) {
+      return txCache.get(txId)!;
+    }
+
+    const txs = await provider.transactionsByHashes({ ids: [txId] });
+    if (txs.length > 0) {
+      txCache.set(txId, txs[0]);
```
mirceahasegan, line 63: Could this be caching transactions that will later be rolled back?
We could cache only if it's outside of the stability window.

mirceahasegan (changes requested):

pczeglik-iohk (approved):

mkazlauskas (approved):
Nice work! :rocket:

packages/wallet/src/services/WalletUtil.ts
AngelCastilloB: I also think it is harmless to cache transactions that re rolled back, because the point is to retrieve the information to properly display what the transaction is doing, if the user is referencing inputs from such transactions and there are more than one input from within the transaction we still save some fetching by caching it, because the input resolver only resolves one input at a time.

Does the input selector gets destroyed when the extension closes?, or whats the lifecycle of the input selector? if it is very long lived, I agree we need to limit the size of the cache just in case, altho very unlikely this cache will ever be too high since it only fetches foreign TXs and most inputs will be from the wallet most of the time. Anyways, let me know how strong you guys feel about this and I can add this extra change. We could use something like https://www.npmjs.com/package/lru-cache

mkazlauskas: > Does the input selector gets destroyed when the extension closes?, or whats the lifecycle of the input selector?

It does when the service worker dies, so after like 10 or 15 minutes of inactivity.

> Anyways, let me know how strong you guys feel about this and I can add this extra change.

I don't think it's worth managing this cache as it's not likely that it would take up lots of memory

PR #1083, opened 2024-02-04
fix(web-extension): delete active wallet from storage and emit null

mkazlauskas:
# Context

When wallet is deactivated, it should emit null to indicate there's no active wallet

When wallet manager is reinitialized after deactivation, it shouldn't load any wallet

LW-9094

# Proposed Solution


VanessaPC (approved):
nicely done!

PR #1086, opened 2024-02-06
chore(cardano-services): un-skip previously skipped unit tests

iccicci:
# Context

Fixtures generated with current cardano stack versions no longer causes problems.

# Proposed Solution

Un-skipped a couple of tests skipped in the past as a quick workaround.


mirceahasegan (approved):
LGTM

mkazlauskas (approved):

PR #1088, opened 2024-02-07
fix(core): transactionSummaryInspector now correctly accounts for transaction fees

AngelCastilloB:
# Context

The TransactionSummaryInspector doesn't properly account for transaction fees and always displays them as unresolved balance.

# Proposed Solution

Fix TransactionSummaryInspector to properly account for transaction fees


mkazlauskas (approved):
Looks good! :detective:

VanessaPC (approved):

PR #1089, opened 2024-02-08
fix: handle projector crash on invalid name

mkazlauskas:
# Context

`InvalidStringError: Invalid string: "Invalid handle 0|0"` when projecting handles on mainnet

# Proposed Solution

- Update handle validation regular expression to be the same as used by ada handle
- Allow '0|0' handle as an exception as it exists on mainnet
- Ignore invalid handles instead of crashing the projector

- Update handles minted in e2e to be lowercase, because on-chain handle asset names are always lowercase


packages/projection/test/operators/Mappers/withHandleMetadata.test.ts
```diff
@@ -29,43 +30,45 @@ const project = (tx: Cardano.OnChainTx) =>
     )
   );
 
+const createCip25HandleMetadata = (handle: Handle) => ({
```
gytis-ivaskevicius, line 33: Thats one weird data structure :sweat_smile:

packages/projection/src/operators/Mappers/util.ts
```diff
@@ -1,11 +1,14 @@
...
 export const computeCompactTxId = (blockHeight: number, txIndex: number) => blockHeight * 100_000 + txIndex;
 
-export const assetNameToUTF8Handle = (assetName: Cardano.AssetName): Handle => {
+export const assetNameToUTF8Handle = (assetName: Cardano.AssetName, logger: Logger): Handle | null => {
   const handle = Cardano.AssetName.toUTF8(assetName);
-  if (!Asset.util.isValidHandle(handle)) throw new InvalidStringError(`Invalid handle ${handle}`);
+  if (!Asset.util.isValidHandle(handle)) {
+    logger.warn(`Invalid handle: '${handle}' / '${assetName}'`);
+    return null;
+  }
```
gytis-ivaskevicius, line 12: :+1: 

Just wondering if we wish to do something with these invalid handles. I kinda wish that we had some queue in this project, we could just submit it to it and at some point review whatever invalid handles we find

gytis-ivaskevicius (dismissed):

packages/core/src/Asset/util/isValidHandle.ts
```diff
@@ -1,4 +1,11 @@
+/**
+ * Alphanumeric: [a-z][A-Z][0-9]
+ * Dash: -
+ * Underscore: _
+ * Period: .
+ * Pipe: | (in general it is not valid, but it exists on mainnet)
```
VanessaPC, line 6: maybe it's worth wording like 'an exception exists in... '

packages/core/test/Asset/util/isValidHandle.test.ts
```diff
@@ -23,6 +27,12 @@ const invalidHandles = [
   'ada%1_test',
   'lace ',
   'wallet ',
+  'comma,',
+  'comma,@sub',
+  'sub@comma,',
```
VanessaPC, line 32: 👍🏼

VanessaPC (approved):
Looks good!

packages/projection/src/operators/Mappers/util.ts
mkazlauskas: Yes, that would be nice, but it's a bigger task. This will at least unblock deployments for now.

packages/core/src/Asset/util/isValidHandle.ts
mkazlauskas: https://github.com/input-output-hk/cardano-js-sdk/pull/1089/commits/57126533d688cdbbcd243eca11a12cc0b8869a6b (already squashed)

iccicci (dismissed):
Excellent!

VanessaPC (dismissed):
Great!

iccicci (approved):

VanessaPC (approved):

packages/core/src/Asset/util/isValidHandle.ts
```diff
@@ -1,4 +1,26 @@
...
-  return pattern.test(handle);
+  // 'g' modifier makes it stateful
+  REGEX_SUB_HANDLE.lastIndex = 0;
+  return (
+    REGEX_HANDLE.test(handle) ||
+    REGEX_SUB_HANDLE.test(handle) ||
+    // Pipe | in general is not valid, but an exception exists in mainnet
+    handle === '0|0'
+  );
 };
```
gytis-ivaskevicius, line 26: I just wanted to mention that there were some cases in my career where I shot myself in the foot with schema validation, as in colleagues push data to database without appropriate checks. To be fair this is a simple schema with one entrypoint

Anyway, even tho it is considered a bad practice I like to enforce it in the database with constraint/check syntax or with triggers

In this case it would be trigger on insert which would set `is_valid` column to true/false depending on the handle itself


Just a suggestion, and in this case I don't think that its necessary due to simplicity of the schema. I just would like you guys to consider this in the future

gytis-ivaskevicius (approved):
Looks good :))

PR #1091, opened 2024-02-12
feat: add explicit handle validation arg

mkazlauskas:
# Context

Handle validation on submit is nice to have, but it should not prevent submitting transactions when it's not possible to do the validation (e.g. due to a bug, handle provider outage, or just not having a handle provider available).

# Proposed Solution

Make handle validation an opt-in feature, so it is now possible to start a tx-submit provider without it.


gytis-ivaskevicius (approved):

VanessaPC (approved):
looks good to me!

rhyslbw (approved):

PR #1093, opened 2024-02-12
ci: make CI run on all release types

gytis-ivaskevicius:
# Context

CI does not run on some releases for some reason, this should resolve issue: https://github.com/input-output-hk/lace-ops/issues/29
Context: https://input-output-rnd.slack.com/archives/C032DECBR6G/p1707739178455329?thread_ts=1707396396.997519&cid=C032DECBR6G

Info on the trigger: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

### DIFF
--- a/.github/workflows/std-release.yaml
+++ b/.github/workflows/std-release.yaml
@@ -2,7 +2,6 @@ name: STD-on-release
 
 on:
   release:
-    types: [published]
 
 jobs:
   call-std:


michalrus (approved):

mkazlauskas (approved):

gytis-ivaskevicius:
image
It should work as it is, maybe commit message is overwriting it anyways? Really not sure if we should merge this

.github/workflows/std-release.yaml
```diff
@@ -2,7 +2,6 @@ name: STD-on-release
 
 on:
```
iccicci, line 3: It seems three more lines could be removed as well..

PR #1094, opened 2024-02-12
test(cardano-services): test ogmios utils

mirceahasegan:
# Context

These tests previously relied on a mocked OgmiosServer, making them tightly coupled with the ogmios implementation.
Since our ogmios implementation was abstracted behind the OgmiosCardanoNode/OgmiosObservableCardanoNode, we can mock them and test the factory utils.

# Proposed Solution
Complete rewrite to not require ogmios server.
Test only the ogmios.ts utils.


iccicci (approved):
Excellent!

mkazlauskas (approved):
Looks good! :rocket:

PR #1097, opened 2024-02-14
Chore/lw 8952 update comments regarding compact genesis

mirceahasegan:
# Context

Pre ogmios6: ogmios would return a Compact genesis.
Currently: ogmios can return genesis parameters per era.

SDK uses genesis parameters that are currently available in the Shelley era.

If in the future we require Conway-era parameters, we can do 2 queries and merge the result in our unified CompactGenesis type.

### DIFF
--- a/packages/ogmios/src/CardanoNode/queries.ts
+++ b/packages/ogmios/src/CardanoNode/queries.ts
@@ -94,8 +94,6 @@ export const queryEraSummaries = (client: LedgerStateQuery.LedgerStateQueryClien
 export const queryGenesisParameters = (client: LedgerStateQuery.LedgerStateQueryClient, logger: Logger) =>
   withCoreCardanoNodeError(async () => {
     logger.info('Querying genesis parameters');
-    // REVIEW: The queryNetwork/genesis local-state-query now expects one era as argument (either 'byron', 'shelley' or 'alonzo')
-    // to retrieve the corresponding genesis configuration.
-    // 'shelley' genesis maps best to the compact genesis we're using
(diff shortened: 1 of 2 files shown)


iccicci (approved):
LGTM

packages/ogmios/src/CardanoNode/queries.ts
```diff
@@ -94,8 +94,6 @@ export const queryEraSummaries = (client: LedgerStateQuery.LedgerStateQueryClien
 export const queryGenesisParameters = (client: LedgerStateQuery.LedgerStateQueryClient, logger: Logger) =>
   withCoreCardanoNodeError(async () => {
     logger.info('Querying genesis parameters');
-    // REVIEW: The queryNetwork/genesis local-state-query now expects one era as argument (either 'byron', 'shelley' or 'alonzo')
-    // to retrieve the corresponding genesis configuration.
-    // 'shelley' genesis maps best to the compact genesis we're using
+    // Update this to query multiple eras if the CompactGenesis type will have to include params from other eras
```
rhyslbw, line 97: Isn't this a `todo:`?

rhyslbw (approved):

packages/ogmios/src/CardanoNode/queries.ts
mirceahasegan: No, but it does look like it. It's a hint regarding what to do in case we need genesis params from other eras.
But if you think it brings more confusion than clarification, I can remove it altogether.

rhyslbw: Ah I see, no it's fine.

mkazlauskas (approved):

PR #1098, opened 2024-02-15
LW-8949 rework cli argument parsing and relative tests

iccicci:
# Context

CLI tests was using a mock ogmios server. The introduction of ogmios 6 made our mock server no longer compatible so CLI tests was disabled.

# Proposed Solution

Reworked the CLI tests to just dump the result of argument parsing (except from some few cases) so they no longer require the ogmios mock server.

- refactored/added some validators to increase output verbosity;
- moved `healthCheckCacheTtl` from `withCommonOptions` to `start-provider-server` being dedicated to the latter;
- fixed the description for `tokenMetadataRequestTimeout`;
- removed a couple of checks at runtime as they was checking for required options;
- excluded `cli.test.ts` from `yarn test`;
- added `yarn test:cli` to run `cli.test.ts` (added to CI as well);


packages/cardano-services/src/Program/options/common.ts
```diff
@@ -41,27 +42,23 @@ export type CommonProgramOptions = {
 
 export const withCommonOptions = (command: Command, apiUrl: URL) => {
   addOptions(command, [
-    newOption('--api-url <apiUrl>', Descriptions.ApiUrl, 'API_URL', (url) => new URL(url), apiUrl),
+    newOption('--api-url <apiUrl>', Descriptions.ApiUrl, 'API_URL', urlValidator(Descriptions.ApiUrl), apiUrl),
     newOption('--build-info <buildInfo>', Descriptions.BuildInfo, 'BUILD_INFO', buildInfoValidator),
+    newOption('--dump-only <true/false>', Descriptions.DumpOnly, 'DUMP_ONLY', (dumpOnly) =>
```
mirceahasegan, line 47: I commonly saw this kind of option named `--dry-run`. We could use the same if you agree that users would intuitively use `--dry-run` for this kind of functionality. 
It is just a suggestion.

iccicci: That option name is already used by the projector for a different purpose

mirceahasegan (approved):
Fantastic work , @iccicci ! 🚀

packages/cardano-services/src/Program/options/common.ts
mkazlauskas: To me it looks kinda similar. Perhaps we could consolidate the 2 options? :thinking: 
```
  DryRun = 'Initialize the projection, but do not start it',
```

packages/cardano-services/package.json
```diff
@@ -47,8 +47,9 @@
     "preview:dev": "FILES='-f ../../compose/dev.yml -f dev.yml' yarn preview:up",
     "preview:up": "NETWORK=preview SUBMIT_API_ARGS='--testnet-magic 2' yarn compose:up",
     "preview:down": "NETWORK=preview yarn compose:down",
-    "test": "jest --runInBand -c ./jest.config.js",
+    "test": "jest --runInBand -c ./jest.config.js --selectProjects unit",
     "test:build:verify": "tsc --build ./test",
+    "test:cli": "jest --runInBand -c ./jest.config.js --selectProjects cli",
```
mkazlauskas, line 52: Why do we need a separate test script for cli? It should be faster than before now :thinking:

packages/cardano-services/src/Program/options/common.ts
iccicci: It's not so easy.
The newly added `--dump-only` option, dumps the options and exits before actually starting the program while the `--dry-run` option, for the projector, let it to perform the initialization, and for the blockfrost worker actually starts everything but don't run the refresh job.
We use the `--dry-run` options in some tests to check the initialization: we should also rework these tests.

packages/cardano-services/package.json
iccicci: It's not faster because before this change lots of option was not tested at all.
We spoken about adding a script which only performs arguments checking, but I abandoned that path because it was not giving the expected improvement: argument parsing (to import options descriptions and default values) already imports more or less the entire SDK: this path was producing code duplication without providing any advantage.

packages/cardano-services/src/Program/options/common.ts
mkazlauskas: I understand it's not so easy, but now we have 2 slightly different options for what it seems to be the same purpose. Perhaps we should create a separate ticket to consolidate the options?

mkazlauskas (approved):
Great work! :rocket:

PR #1101, opened 2024-02-20
LW-9604 deposit values

iccicci:
# Context

Stake key de-registration certificates currently has a wrong deposit value (the value from relative epoch parameters rather than the value from epoch parameters relative to when stake key was registered).
Stake pool registration certificates doesn't show any deposit.

# Proposed Solution

- Fixed the deposit value for stake key de-registration certificates.
- Added the deposit for stake pool registration certificates.

### DIFF
--- a/packages/core/src/Serialization/Certificates/PoolRegistration.ts
+++ b/packages/core/src/Serialization/Certificates/PoolRegistration.ts
@@ -170,7 +170,7 @@ export class PoolRegistration {
    * @param cert core PoolRegistrationCertificate object.
    */
   static fromCore(cert: Cardano.PoolRegistrationCertificate) {
-    return new PoolRegistration(PoolParams.fromCore(cert.poolParameters)); // TODO: Core type does not support script hash as credential, fix?
+    return new PoolRegistration(PoolParams.fromCore(cert.poolParameters));
   }
 
   /**
--- a/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/mappers.ts
+++ b/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/mappers.ts
@@ -155,7 +155,7 @@ export const mapAnchor = (anchorUrl: string, anchorDataHash: string): Cardano.An
 export const mapCertificate = (
   certModel: WithCertType<CertificateModel>
   // eslint-disable-next-line sonarjs/cognitive-complexity
-): WithCertIndex<Cardano.Certificate> | null => {
+): WithCertIndex<Cardano.HydratedCertificate> | null => {
   if (isPoolRetireCertModel(certModel))
     return {
       __typename: Cardano.CertificateType.PoolRetirement,
@@ -168,8 +168,9 @@ export const mapCertificate = (
     return {
       __typename: Cardano.CertificateType.PoolRegistration,
       cert_index: certModel.cert_index,
+      deposit: BigInt(certModel.deposit),
       poolParameters: null as unknown as Cardano.PoolParameters
-    } as WithCertIndex<Cardano.PoolRegistrationCertificate>;
+    } as WithCertIndex<Cardano.HydratedPoolRegistrationCertificate>;
 
   if (isMirCertModel(certModel)) {
     const credential = Cardano.Address.fromString(certModel.address)?.asReward()?.getPaymentCredential();
--- a/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/queries.ts
+++ b/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/queries.ts
@@ -238,13 +238,32 @@ export const findPoolRetireCertsTxIds = `
 	ORDER BY tx.id ASC`;
 
 export const findPoolRegisterCertsByTxIds = `
-	SELECT	
+	SELECT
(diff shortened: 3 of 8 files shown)


packages/core/src/Serialization/Certificates/PoolRegistration.ts
```diff
@@ -13,6 +13,7 @@ const EMBEDDED_GROUP_SIZE = 10;
  * about the pool such as the pledge, costs, margin, reward account, and the pool's owners and relays.
  */
 export class PoolRegistration {
+  #deposit: Cardano.Lovelace;
```
AngelCastilloB, line 16: I am not sure we should add this here 🤔 . This class should be a representation of the CBOR pool registration certificate and it does not have this field. This means that if you serialize and deserialize to and from CBOR this information will be lost, which is unexpected. IMHO this class should be kept as is without the field and default to 0 when converting to the core type.

This field is only relevant in the core type when we get the information from the backend, but not when we read from CBOR as the information is not there, another option would be to make the field optional in the core type?

iccicci: Crystal clear why this property shouldn't be here: I'll remove it.
While I didn't understood why you are proposing to make the field optional in the core type; could you please elaborate?
Thank you

packages/core/src/Serialization/Certificates/PoolRegistration.ts
```diff
@@ -170,7 +174,7 @@ export class PoolRegistration {
    * @param cert core PoolRegistrationCertificate object.
    */
   static fromCore(cert: Cardano.PoolRegistrationCertificate) {
-    return new PoolRegistration(PoolParams.fromCore(cert.poolParameters)); // TODO: Core type does not support script hash as credential, fix?
+    return new PoolRegistration(PoolParams.fromCore(cert.poolParameters), cert.deposit); // TODO: Core type does not support script hash as credential, fix?
```
AngelCastilloB, line 177:
```suggestion
    return new PoolRegistration(PoolParams.fromCore(cert.poolParameters), cert.deposit); 
```
This was not introduced by this PR, but I noticed this TODO makes no sense here and is probably a result of a bad copy/paste 😅

AngelCastilloB (changes requested):

packages/core/src/Serialization/Certificates/PoolRegistration.ts
AngelCastilloB: This class maps to its equivalent core type (toCore), but doesn't have this information (the deposit), so we need to either default the value to 0, or make the field optional so the mapping can occur

iccicci: Much more clear; thank you.
I'll dig this aspect a bit more.

packages/core/src/Serialization/Certificates/PoolRegistration.ts
iccicci: Indeed I had some difficulties understanding its meaning...
If there are no objections, I'll remove it

iccicci: 8a790b0aef should solve this

packages/core/src/Serialization/Certificates/PoolRegistration.ts
iccicci: f2c51e9e8e should solve this

AngelCastilloB (approved):
Great work @iccicci ❤️

mirceahasegan (approved):
Great work, @iccicci ! 💪

packages/core/src/Cardano/types/Certificate.ts
```diff
@@ -115,6 +115,7 @@ export interface StakeAddressCertificate {
 
 export interface PoolRegistrationCertificate {
   __typename: CertificateType.PoolRegistration;
+  deposit?: Lovelace;
```
mkazlauskas, line 118: I would prefer to have a separate type for HydratedTx, but this will do for now. Let's add a comment :)
```suggestion
  /**
   * This type is used for both `Tx` and `HydratedTx` certificates.
   * It will always be `undefined` for `Tx` when building/deserializing a transaction.
   */
  deposit?: Lovelace;
```

mkazlauskas (changes requested):

packages/core/src/Cardano/types/Certificate.ts
iccicci: Sorry @mkazlauskas , does 34347c1495 address this?

mkazlauskas: Almost! Just need to update https://github.com/input-output-hk/cardano-js-sdk/commit/34347c1495fd032a46d76474a956939b4fcde455#diff-a4136fb38ea6bd360a39db1a56bf7429bc431c6be1c8937bcd98f58e2dc52462R89 to exclude `certificates` from base type, and use the regular `Certificate[]` instead. The inheritance is a little bit backwards there :sweat_smile:

iccicci: 77461a742f should address this as well

mkazlauskas (approved):
Great work! :rocket:

PR #1105, opened 2024-02-22
fix(core): make the projector not exit on asset name error

iccicci:
# Context

The `asset-projector` exit with error while projecting.

# Proposed Solution

Wrapped the line throwing in a `try / catch` block.

### DIFF
--- a/packages/core/src/Asset/NftMetadata/fromMetadatum.ts
+++ b/packages/core/src/Asset/NftMetadata/fromMetadatum.ts
@@ -121,7 +121,11 @@ export const fromMetadatum = (
   const assetId = Cardano.AssetId.fromParts(asset.policyId, asset.name);
 
   if (version === '1.0' && !name) {
-    name = AssetName.toUTF8(asset.name);
+    try {
+      name = AssetName.toUTF8(asset.name);
+    } catch (error) {
+      logger.warn(error);
+    }
   }
 
   if (!name || !image) {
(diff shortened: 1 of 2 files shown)


VanessaPC (approved):
I think this is a good solution in that we catch the asset name in the logs and let docker continue. 
Nice solution

packages/core/src/Asset/NftMetadata/fromMetadatum.ts
```diff
@@ -121,7 +121,11 @@ export const fromMetadatum = (
   const assetId = Cardano.AssetId.fromParts(asset.policyId, asset.name);
 
   if (version === '1.0' && !name) {
-    name = AssetName.toUTF8(asset.name);
+    try {
+      name = AssetName.toUTF8(asset.name);
+    } catch (error) {
+      logger.warn(error);
```
mkazlauskas, line 127: We shouldn't continue in this case, as metadata is invalid

```suggestion
      logger.warn(error);
      return null;
```

mkazlauskas (changes requested):
What is the problematic asset id? I suspect it may be for cip67-encoded asset :thinking:

packages/core/src/Asset/NftMetadata/fromMetadatum.ts
iccicci: The `return null` is already in the next line

mkazlauskas (approved):

PR #1107, opened 2024-02-23
fix: trezor bundler compatibility

mkazlauskas:
# Context

Webpack exports ledger and trezor libaries under `{default: {...}}`.

# Proposed Solution

We're [already using a hack for this for Ledger](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/hardware-ledger/src/LedgerKeyAgent.ts#L35-L40). This PR applies the same for Trezor.

Also fixed Ledger error mapping, when device is already open

### DIFF
--- a/packages/hardware-ledger/src/LedgerKeyAgent.ts
+++ b/packages/hardware-ledger/src/LedgerKeyAgent.ts
@@ -40,6 +40,16 @@ const LedgerConnection = (_LedgerConnection as any).default
   : _LedgerConnection;
 type LedgerConnection = _LedgerConnection;
 
+const isDeviceAlreadyOpenError = (error: unknown) => {
+  if (typeof error !== 'object') return false;
+  const innerError = (error as any).innerError;
+  if (typeof innerError !== 'object') return false;
+  return (
+    innerError.code === 11 ||
+    (typeof innerError.message === 'string' && innerError.message.includes('cannot open device with path'))
+  );
+};
+
 export interface LedgerKeyAgentProps extends Omit<SerializableLedgerKeyAgentData, '__typename'> {
   deviceConnection?: LedgerConnection;
 }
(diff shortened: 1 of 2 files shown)


packages/hardware-ledger/src/LedgerKeyAgent.ts
```diff
@@ -40,6 +40,16 @@ const LedgerConnection = (_LedgerConnection as any).default
   : _LedgerConnection;
 type LedgerConnection = _LedgerConnection;
 
+const isDeviceAlreadyOpenError = (error: unknown) => {
+  if (typeof error !== 'object') return false;
```
mirceahasegan, line 44: `null` is an object. I think might cause issues on the following line. Can you add some unit tests for this function?

mirceahasegan (approved):
Nice work, @mkazlauskas 💪

packages/hardware-trezor/src/TrezorKeyAgent.ts
```diff
@@ -16,9 +16,12 @@ import {
   util
 } from '@cardano-sdk/key-management';
 import { txToTrezor } from './transformers/tx';
-import TrezorConnectWeb from '@trezor/connect-web';
+import _TrezorConnectWeb from '@trezor/connect-web';
 
 const TrezorConnectNode = Trezor.default;
+const TrezorConnectWeb = (_TrezorConnectWeb as any).default
+  ? ((_TrezorConnectWeb as any).default as typeof _TrezorConnectWeb)
+  : _TrezorConnectWeb;
```
tomislavhoracek, line 24: @mkazlauskas This won't work in the Lace :( 
I made the same change when I was debugging it in the scope of https://input-output.atlassian.net/wiki/spaces/CWT/pages/4129554601/External+Dapps+testing+Trezor ticket.
It looks good when you are testing it in SDK but you will still have `.default` when the method is called from Lace

tomislavhoracek: Update: 
I checked this comment https://input-output-rnd.slack.com/archives/C01S2DD5B0D/p1708699319597529?thread_ts=1708695474.577899&cid=C01S2DD5B0D

Maybe that wasn't working because I tested it by using / setting `dist` from SDK directly. Let me check it.

tomislavhoracek: @mkazlauskas Checked, problem was in the Lace webpack. 
Ignore previous comments!

PR #1109, opened 2024-02-26
feat(core): cbor ser/des handle 128 bytes anchors

mirceahasegan:
# Context

For cardano-node-8.8 the length of urls will be increased from 64 -> 128 bytes - see in CDDL here.
This will allow much more flexibility in URLs which can be used for governance metadata anchors and pool metadata anchors.
For examples, IPFS links will be able to be used to store off-chain metadata. (edited).

Master branch (pre-conway) does not support the 128bytes anchors. Even so, usage of >64bytes anchors will result in a failed anchor, until `master` branch also uses Cardano Node 8.8+


iccicci (approved):
LGTM!

AngelCastilloB (approved):
LGTM @mirceahasegan

PR #1110, opened 2024-02-26
feat(core): conway-era cddl changes new policy_hash field

mirceahasegan:
# Context

parameter_change_action and treasury_withdrawals_action got an extra field policy_hash / null - [here](https://github.com/IntersectMBO/cardano-ledger/blob/74dead1f4a48b876ea787d1c42a693798ba84e6e/eras/conway/impl/cddl-files/conway.cddl#L106) and [here](https://github.com/IntersectMBO/cardano-ledger/blob/74dead1f4a48b876ea787d1c42a693798ba84e6e/eras/conway/impl/cddl-files/conway.cddl#L110)

# Proposed Solution

Update CBOR ser/des


packages/core/src/Serialization/TransactionBody/ProposalProcedure/ParameterChangeAction.ts
```diff
@@ -79,7 +88,16 @@ export class ParameterChangeAction {
...
 
-    const action = new ParameterChangeAction(parameterUpdate, govActionId);
+    let policyHash: Hash28ByteBase16 | undefined;
+    if (reader.peekState() === CborReaderState.Null) {
+      reader.readNull();
+    } else {
+      policyHash = HexBlob.fromBytes(reader.readByteString()) as unknown as Hash28ByteBase16;
+    }
+
+    reader.readEndArray();
```
iccicci, line 98: This was added, but it seems to be unrelated from the CDDL change.

Could I ask which is its purpose?

Another thing I do not understand is why the same change about `TreasuryWithdrawalsAction` doesn't require this statement.

Thank you

iccicci (commented):
I feel not strong enough to do a valuable review of this PR.
I have just a question to better understand.

packages/core/src/Serialization/TransactionBody/ProposalProcedure/ParameterChangeAction.ts
AngelCastilloB: I implemented the CBOR reader as a stack-based virtual machine, so every time you read a start array/map/object, you are pushed one level into the stack, you need to read the end of the array/map/object to go out one level on the stack. If you read the contents of an array for example, you need to readStartArray -> read each element of the array -> and finally readEndArray -> then continue reading objects at the same level of the array in the object. 

This is only required if you read more items after the array ends or if there are more elements at the same level of the array (which is not the case here). This readEndArray can probably be skipped, but I think is good to add those because this has the extra benefit of making sure that indefinite arrays are terminated correctly in the CBOR stream.

packages/core/test/Serialization/TransactionBody/ProposalProcedure/TreasuryWithdrawalsAction.test.ts
```diff
@@ -1,13 +1,17 @@
 /* eslint-disable sonarjs/no-duplicate-string */
 import * as Cardano from '../../../../src/Cardano';
+import { Hash28ByteBase16 } from '@cardano-sdk/crypto';
 import { HexBlob } from '@cardano-sdk/util';
 import { TreasuryWithdrawalsAction } from '../../../../src/Serialization';
 
 // Test data used in the following tests was generated with the cardano-serialization-lib
-const cbor = HexBlob('8202a1581de1cb0ec2692497b458e46812c8a5bfa2931d1a2d965a99893828ec810f01');
+const cbor = HexBlob(
+  '8302a1581de1cb0ec2692497b458e46812c8a5bfa2931d1a2d965a99893828ec810f01581c8293d319ef5b3ac72366dd28006bd315b715f7e7cfcbd3004129b80d'
```
AngelCastilloB, line 9: @mirceahasegan same as comment above, just confirm this has been verified with at least one more sources (CSL or ledger)

packages/core/test/Serialization/TransactionBody/ProposalProcedure/ParameterChangeAction.test.ts
```diff
@@ -1,12 +1,13 @@
...
...9d818641959d81864186418641959d81864194c5118201a0002acfa182019b551041a000363151901ff00011a00015c3518201a000797751936f404021a0002ff941a0006ea7818dc0001011903e8196ff604021a0003bd081a00034ec5183e011a00102e0f19312a011a00032e801901a5011a0002da781903e819cf06011a00013a34182019a8f118201903e818201a00013aac0119e143041903e80a1a00030219189c011a00030219189c011a0003207c1901d9011a000330001901ff0119ccf3182019fd40182019ffd5182019581e18201940b318201a00012adf18201a0002ff941a0006ea7818dc0001011a00010f92192da7000119eabb18201a0002ff941a0006ea7818dc0001011a0002ff941a0006ea7818dc0001011a0011b22c1a0005fdde00021a000c504e197712041a001d6af61a0001425b041a00040c660004001a00014fab18201a0003236119032c010119a0de18201a00033d7618201979f41820197fb8182019a95d1820197df718201995aa18201a0223accc0a1a0374f693194a1f0a1a02515e841980b30a1382d81e820102d81e82010214821b00000001000000001b000000010000000015821b00000001000000001b0000000100000000161903ba1719035418181864181984d81e820000d81e820101d81e820202d81e820303181a8ad81e820000d81e820101d81e820202d81e820303d81e820404d81e820505d81e820606d81e820707d81e820808d81e820909181b1864181c18c8181d19012c181e1903e8181f1907d01820191388581c8293d319ef5b3ac72366dd28006bd315b715f7e7cfcbd3004129b80d'
```
AngelCastilloB, line 10: @mirceahasegan did you generate this CBOR or are you referencing a specific source? We usually cross reference the CBOR test vectors of the tests with the CSL/CML or the ledger tests, would be good to do this cross verification (maybe you have already done it)

AngelCastilloB (approved):
Looks good @mirceahasegan ⭐ , just left a few comments regarding test vectors

packages/core/src/Serialization/TransactionBody/ProposalProcedure/ParameterChangeAction.ts
iccicci: So, if I understood correctly, we could consider a _pseudo problem_ the absence of this statement before this PR; am I right?

And following the same logic, it should be a good idea also adding it at the end of `TreasuryWithdrawalsAction.fromCBOR`: it starts with a `reader.readStartArray()` but I can't find a relative `reader.readEndArray()`; am I missing something else?

Thank you

mirceahasegan: > we could consider a pseudo problem the absence 
Yes, for the sake of setting a good example, we should call `readEndArray()`.

packages/core/test/Serialization/TransactionBody/ProposalProcedure/TreasuryWithdrawalsAction.test.ts
mirceahasegan: Verified with @emurgo/cardano-serialization-lib-nodejs@12.0.0-alpha.16

packages/core/test/Serialization/TransactionBody/ProposalProcedure/ParameterChangeAction.test.ts
mirceahasegan: Verified with @emurgo/cardano-serialization-lib-nodejs@12.0.0-alpha.16 by decoding the CBOR with CSL and checking the policy hash field, and a few other fields at random.

iccicci (approved):
Good job @mirceahasegan ...
but moreover thank you very much for involving me in this review: it was an exceptional learning occasion for me! (maybe at cost of your time... that's one more reason to say "thank you!")

PR #1117, opened 2024-02-29
fix(core): transaction body can now serializes bodies with no inputs correctly

AngelCastilloB:
# Context

There is a bug in the transaction body serialization logic which causes incorrect CBOR to be produced if there are no inputs in the transaction.

# Proposed Solution

Fix the bug

### DIFF
--- a/packages/core/src/Serialization/TransactionBody/TransactionBody.ts
+++ b/packages/core/src/Serialization/TransactionBody/TransactionBody.ts
@@ -990,7 +990,7 @@ export class TransactionBody {
     let mapSize = 0;
 
     if (this.#inputs !== undefined && this.#inputs.length > 0) ++mapSize;
-    if (this.#outputs !== undefined && this.#inputs.length > 0) ++mapSize;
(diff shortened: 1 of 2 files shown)


mkazlauskas (approved):
Nice work! :detective:

iccicci (approved):
The change is coherent with the code in the same context, but seems to be not so much coherent with subject and description of the PR; I read:
`bodies with no inputs`
but reading at the code I would have expected
`bodies with no outputs`.
Maybe this is just a minor refuse...

AngelCastilloB:
> The change is coherent with the code in the same context, but seems to be not so much coherent with subject and description of the PR; I read: `bodies with no inputs` but reading at the code I would have expected `bodies with no outputs`. Maybe this is just a minor refuse...

The error was caused when the transaction body had no inputs. Because we were using the inputs presence to account for both inputs and outputs during the computation of the map size. When the inputs were missing, the size of the map was reduced by 2 (rather than 1) as if when there were no inputs, there weren't also any outputs (even when outputs were present)

PR #1119, opened 2024-02-29
[LW-9879] Move deployment code from lace-ops to here

michalrus:
# Context

https://input-output.atlassian.net/browse/LW-9879

# Proposed Solution

n/a

n/a


bernokl (dismissed):

michalrus:
Removing stale reviews, I need to debug [this GHA error](https://github.com/input-output-hk/cardano-js-sdk/actions/runs/8098454555?pr=1119):

```
Error when evaluating 'strategy' for job 'images'. .github/workflows/std.yml (Line: 75, Col: 17): Unexpected value ''
```

.github/workflows/std.yml
```diff
@@ -95,100 +110,131 @@ jobs:
...
+            aws configure list
+
+            aws sts get-caller-identity
+
+            kubectl cluster-info -n dev-preview
+
+            echo ==================================================
+
+            ### FIXME: change to dev-preview!!!
+
```
gytis-ivaskevicius: I dont think any of this is necessary

.github/workflows/std.yml
```diff
@@ -95,100 +110,131 @@ jobs:
...
-    needs: [discover, images]
-    name: ${{ matrix.target.jobName }} (us-east-1)
-    env:
-      AWS_REGION: us-east-1
-    permissions:
-      id-token: write
-      contents: read
-      deployments: write
+  deploy-to-us:
+    if: github.event_name == 'push' && github.ref_name == 'dev-preview'
```
gytis-ivaskevicius, line 167: we should autodeploy from master

.github/workflows/std.yml
```diff
@@ -95,100 +110,131 @@ jobs:
...
-      url: https://backend.dev-preview.eks.lw.iog.io
-    # Boolean input should be compared with string until https://github.com/actions/runner/issues/2238 resolved
-    if: >
-      fromJSON(needs.discover.outputs.hits).deployments.apply != '{}' && ((github.event_name == 'push' && github.ref == 'refs/heads/master') || inputs.deploy)
-
-    strategy:
-      matrix:
-        target: ${{ fromJSON(needs.discover.outputs.hits).deployments.apply }}
+      name: ${{ github.ref_name }}
+      url: https://${{ github.ref_name }}.dapps.lw.iog.io/
```
gytis-ivaskevicius, line 177: dapps?

.github/workflows/std.yml
```diff
@@ -95,100 +110,131 @@ jobs:
...
           generate_summary_for: job
+      # Further steps assume AWS_PROFILE=lw, while the official action has no way to specify that profile:
+      - name: Set up AWS credentials
+        run: |
+          mkdir -p ~/.aws
+
+          cat <<EOF >~/.aws/credentials
+          [lw]
+          aws_access_key_id = ${{ secrets.AWS_ACCESS_KEY}}
+          aws_secret_access_key = ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
gytis-ivaskevicius, line 193: Arent env variables enough?

.github/workflows/std.yml
```diff
@@ -1,33 +1,23 @@
 # yq -y < std-unexploded.yml > workflows/std.yml
 name: STD
 on:
-  workflow_dispatch:
-    inputs:
-      deploy:
-        description: Deploy to dev-preview
-        type: boolean
-        required: true
-        default: false
```
gytis-ivaskevicius, line 10: I think we should preserve this functionallity

.github/workflows/std.yml
```diff
@@ -67,25 +63,36 @@ jobs:
...
+    permissions:
+      id-token: write
+      contents: read
     strategy:
       matrix:
-        target: ${{ fromJSON(needs.discover.outputs.hits).oci-images.publish }}
+        target: ${{ contains(fromJSON(needs.discover.outputs.hits), 'oci-images') && fromJSON(needs.discover.outputs.hits).oci-images.publish || fromJSON('["dummy-target"]') }}
     steps:
       - name: Configure AWS Credentials
+        if: matrix.target != 'dummy-target'
```
gytis-ivaskevicius, line 78: Can we make do without dummy target?

nix/local/envs.nix
```diff
@@ -33,16 +34,33 @@ in {
     name = "Cardano JS SDK Local Env";
     imports = [formattingModule];
 
+    packages = with inputs.nixpkgs; [
+      awscli2
+      python310
```
gytis-ivaskevicius, line 39: python?

gytis-ivaskevicius (changes requested):

.github/workflows/std.yml
michalrus: I tried a few times, but failed. We need this `images` matrix job to succeed, whether it builds anything or not. GitHub doesn’t consider empty matrix jobs as successful (it doesn’t even allow that when evaluating the YAML). If you add an `if:` to the job, it won’t run if the matrix was to be empty, and then our diff and deployment won't run either. So this ugly thing is the only thing I found working… WDYT?

nix/local/envs.nix
michalrus: It was present in dapps, a left over from trying to align those, I’ll remove it

.github/workflows/std.yml
michalrus: I’ll use them, okay

.github/workflows/std.yml
michalrus: Right :facepalm:

Which URL would you suggest? https://backend.dev-preview.eks.lw.iog.io ?

michalrus: Yup, that’s what the current deployments use

.github/workflows/std.yml
michalrus: Okay. So no `dev-preview` branch, and `dev-preview` always goes from `master`?

Will there be a branch for `dev-preprod`, or all `dev-*` are deployed from `master`?

.github/workflows/std.yml
michalrus: It complicates triggers, but okay, let me think.

Preserve it _only_ for `dev-preview`?

michalrus: Maybe this is something to discuss duringe the call starting in 1 hr

.github/workflows/std.yml
michalrus: Maybe this is something to discuss during the call starting in 1 hr

gytis-ivaskevicius: For the time being only dev-preview deployed from master. in the future we probably will have more envs deployed from master

.github/workflows/std.yml
gytis-ivaskevicius: Please use `https://dev-preview.lw.iog.io/`

.github/workflows/std.yml
gytis-ivaskevicius: We should have it for all enviornments but I am okay with keeping this only for dev-preview until future iterations

michalrus: I like the button, too, but I'm afraid we won’t be allowed to do this. :pensive:

Because even in `dapps` (which is of lower significance and not production code yet), Rhys wanted it removed, and do everything through PRs to deployment branches…

.github/workflows/std.yml
gytis-ivaskevicius: Cant we have an if statement like this for whole job, as in just don't run the job if there are no tagets

michalrus: Yes, we can, but then, as I wrote, we won’t run the deployment:

> if the [build] matrix was to be empty, […] then our diff and deployment won't run either

.github/workflows/std.yml
michalrus: The we need to continue posting the diff comments to all PRs, hmmm…

Okay, you’re right, let’s make this PR work the same as the current code, and then we can discuss and change things

.github/workflows/std.yml
michalrus: Fixed in d1c671b5b0ae79783fec28ad11d3d76ef63d75da

nix/local/envs.nix
michalrus: Fixed in f8ffbac1bc0a35b3aa17df904f501f43ebd9632e

.github/workflows/std.yml
michalrus: Fixed in d3b5fdc8c479a58e0bfc041b474bae06dcbc0ab9

.github/workflows/std.yml
michalrus: I think [env. variables are not enough](https://github.com/input-output-hk/cardano-js-sdk/actions/runs/8139418959/job/22242557797?pr=1119) (same error was happening in Dapps). 

```
Error: Failed to get release dev-preprod-cardanojs in namespace dev-preprod: exit status 1: 
The config profile (lw) could not be found
Error: Kubernetes cluster unreachable: Get "https://eb65a880b90df0a9b8455038ef27a45e.gr7.us-east-1.eks.amazonaws.com/version": getting credentials: exec: executable aws failed with exit code 255
```

The `lw` part in the INI file seems crucial :thinking:

michalrus: Let me revert d1c671b5b0ae79783fec28ad11d3d76ef63d75da…

michalrus: And now it works again :green_circle:

rhyslbw (changes requested):
There's a lot of commits here, do they all represent atomic changes? It would be easier for me to review once squashed into a smaller set of logical changes, if possible,

michalrus:
@rhyslbw I’ve just pushed a cleaned up history

gytis-ivaskevicius (dismissed):

bernokl (approved):

gytis-ivaskevicius (approved):

PR #1120, opened 2024-02-29
fix(cardano-services): phase-2 failure tx mapper

renanvalentin:
# Context

# Proposed Solution

- db-sync chain-history provider maps consumed collateral amount to fee property. It should be setting totalCollateral instead, and (if the original fee property of transaction body is unavailable), set fee: 0n

- similarly, we seem to be mapping the consumed collateral inputs into body.inputs. Test this, and update the mapping to map to body.collaterals (leaving inputs as an empty array, like with the fee)

- lastly, same for outputs and collateralOutput

### DIFF
--- a/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/mappers.ts
+++ b/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/mappers.ts
@@ -215,13 +215,25 @@ export const mapTxAlonzo = (
     slot: Cardano.Slot(Number(txModel.block_slot_no))
   },
   body: {
+    ...(inputSource === Cardano.InputSource.collaterals
+      ? {
+          collateralReturn: outputs.length > 0 ? outputs[0] : undefined,
+          collaterals: inputs,
+          fee: BigInt(0),
+          inputs: [],
+          outputs: [],
+          totalCollateral: BigInt(txModel.fee)
+        }
+      : {
+          collateralReturn: collateralOutputs.length > 0 ? collateralOutputs[0] : undefined,
+          collaterals,
+          fee: BigInt(txModel.fee),
+          inputs,
+          outputs
+        }),
     certificates,
-    collateralReturn: collateralOutputs.length > 0 ? collateralOutputs[0] : undefined,
-    collaterals,
-    fee: BigInt(txModel.fee),
-    inputs,
     mint,
-    outputs,
+    totalCollateral: inputSource === Cardano.InputSource.collaterals ? BigInt(txModel.fee) : undefined,
     validityInterval: {
       invalidBefore: Cardano.Slot(Number(txModel.invalid_before)) || undefined,
       invalidHereafter: Cardano.Slot(Number(txModel.invalid_hereafter)) || undefined
--- a/packages/cardano-services-client/src/version.ts
+++ b/packages/cardano-services-client/src/version.ts
@@ -1,7 +1,7 @@
 // auto-generated using ../scripts/createVersionSource.js
 export const apiVersion = {
   assetInfo: '1.0.0',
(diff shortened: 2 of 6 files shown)


AngelCastilloB:
The e2e test `test/wallet/PersonalWallet/phase2validation.test.ts` needs to be adjusted

`expect(txFoundInHistory.body.fee).toEqual(collateralCoinValue);`

=>

`expect(txFoundInHistory.body.totalCollateral).toEqual(collateralCoinValue);`

mkazlauskas (approved):
Nice work! Please squash the 2 commits into 1. It can still be scoped to `(cardano-services)`, since there are no code changes in `e2e` package (only test adjustment).

AngelCastilloB (dismissed):
LGTM 🚀

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/mappers.ts
```diff
@@ -215,13 +215,25 @@ export const mapTxAlonzo = (
...
   },
   body: {
+    ...(inputSource === Cardano.InputSource.collaterals
+      ? {
+          collateralReturn: outputs.length > 0 ? outputs[0] : undefined,
+          collaterals: inputs,
+          fee: BigInt(0),
+          inputs: [],
+          outputs: [],
+          totalCollateral: BigInt(txModel.fee)
```
iccicci, line 225: Maybe this is why I'm not still confident with these fields, but I would have said:

```suggestion
          collaterals: inputs,
          fee: BigInt(txModel.fee),
          inputs: [],
          outputs: [],
          totalCollateral: /* pseudo code: sum(inputs.coins) - outputs[0]?.coins */
```

Could you please explain me?
Thank you

mkazlauskas (changes requested):
I've just realized that we need to bump http api version to follow semantic versioning, because it's changing the behavior slightly. Please
- bump `info.version` to `3.0.1` in `packages/cardano-services/src/ChainHistory/openApi.json`
- build `cardano-services-client` to re-generate `version.ts`

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/mappers.ts
renanvalentin: When the input source is collateral, db sync stores the collateral consumed as a fee, so we just need to map `fee` to `totalCollateral`

iccicci: AFAIK the target of this task is to make SDK compliant to Ledger rather than to db-sync.
IDK how a tx which fails phase 2 validation is represented at the lowest level:

- if it has distinct fee and collateral, IMHO we should make the distinction as well
- if everything is collected under the fee field, I'm fine with the change as is

mirceahasegan (approved):
Great work, @renanvalentin 🚀

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/mappers.ts
mkazlauskas: > AFAIK the target of this task is to make SDK compliant to Ledger rather than to db-sync.

That's correct

> if it has distinct fee and collateral, IMHO we should make the distinction as well
> if everything is collected under the fee field, I'm fine with the change as is

Some information is lost in db-sync, since the `fee` field is used for collateral amount. We are setting it to 0n as that is best we can do in this circumstance.

mkazlauskas (approved):
:rocket:

iccicci (approved):
Good job!

mirceahasegan (approved):

mkazlauskas (approved):
Nice work! :rocket:

PR #1126, opened 2024-03-05
chore: pass fake chromedriver to package build to avoid IFD

gytis-ivaskevicius:
# Context

During OCI image build there is an IFD which is super annoying during deployments, this PR attempts to fix it

# Proposed Solution
Pass empty file instead of chromedriver since it is unused during build anyways

### DIFF
--- a/nix/cardano-services/packages.nix
+++ b/nix/cardano-services/packages.nix
@@ -42,25 +42,6 @@ in {
         runHook postInstall
       '';
     });
-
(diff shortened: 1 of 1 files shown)


michalrus (dismissed):
That’s great =)

gytis-ivaskevicius:
Deployed it to ops-preview-1. Works as expected

michalrus (approved):

bernokl (approved):

PR #1135, opened 2024-03-07
LW-9418 hard fork proposal fixes

iccicci:
# Context

While working on LW-9418 I discovered a couple of problems.

# Proposed Solution

Please check the commits

### DIFF
--- a/packages/core/src/Serialization/TransactionBody/ProposalProcedure/HardForkInitiationAction.ts
+++ b/packages/core/src/Serialization/TransactionBody/ProposalProcedure/HardForkInitiationAction.ts
@@ -41,7 +41,6 @@ export class HardForkInitiationAction {
     writer.writeInt(GovernanceActionKind.HardForkInitiation);
     this.#govActionId ? writer.writeEncodedValue(hexToBytes(this.#govActionId.toCbor())) : writer.writeNull();
 
-    writer.writeStartArray(1);
     writer.writeEncodedValue(hexToBytes(this.#protocolVersion.toCbor()));
 
     return writer.encodeAsHex();
@@ -79,10 +78,7 @@ export class HardForkInitiationAction {
       govActionId = GovernanceActionId.fromCbor(HexBlob.fromBytes(reader.readEncodedValue()));
     }
 
-    reader.readStartArray();
     const protocolVersion = ProtocolVersion.fromCbor(HexBlob.fromBytes(reader.readEncodedValue()));
-    reader.readEndArray();
-
     const action = new HardForkInitiationAction(protocolVersion, govActionId);
     action.#originalBytes = cbor;
 
--- a/packages/core/src/Cardano/types/Governance.ts
+++ b/packages/core/src/Cardano/types/Governance.ts
@@ -50,7 +50,7 @@ export type ParameterChangeAction = {
 export type HardForkInitiationAction = {
   __typename: GovernanceActionType.hard_fork_initiation_action;
   governanceActionId: GovernanceActionId | null;
-  protocolVersion: ProtocolVersion;
+  protocolVersion: Pick<ProtocolVersion, 'major' | 'minor'>;
 };
 
 export type TreasuryWithdrawalsAction = {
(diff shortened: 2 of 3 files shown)


packages/core/src/Serialization/TransactionBody/ProposalProcedure/HardForkInitiationAction.ts
```diff
@@ -41,7 +41,6 @@ export class HardForkInitiationAction {
     writer.writeInt(GovernanceActionKind.HardForkInitiation);
     this.#govActionId ? writer.writeEncodedValue(hexToBytes(this.#govActionId.toCbor())) : writer.writeNull();
 
-    writer.writeStartArray(1);
```
AngelCastilloB, line 44: I think we are making the change in the wrong place 🤔 . The ProtocolVersion is the object that is not serializing correctly, this array here is correctly specified in the CDDL. This will have the right overall effect, with the bug with ProtocolVersion serialization is still present. Would be better to fix it there and leave this as it was.

AngelCastilloB (changes requested):

packages/core/src/Serialization/TransactionBody/ProposalProcedure/HardForkInitiationAction.ts
iccicci: This writes an array of 1 element, while the one you are speaking about writes an array of 2 element.
To go in the direction you are proposing we should also change the size of the array here (importing the size from the other file? defining in this file of the size of the array written in the other file?).
More than this, going that way we should also hoist array read/write [here](https://github.com/input-output-hk/cardano-js-sdk/blob/5afd6d872085dff39b79ff866601d3e8083dafe7/packages/core/src/Serialization/Update/ProtocolParamUpdate.ts#L168-L171) and [here](https://github.com/input-output-hk/cardano-js-sdk/blob/5afd6d872085dff39b79ff866601d3e8083dafe7/packages/core/src/Serialization/Update/ProtocolParamUpdate.ts#L322-L324) actually producing some code repetition. 

In conclusion I think the real problem here is a code replication in CDDL file: `protocol_version` is always referred as `[ protocol_version ]` and defined as `protocol_version = (major_protocol_version, uint)`.

I'm strongly considering to open a PR in Ledger repos to change how `protocol_version` is referred and defined.

iccicci: Looking at previous eras CDDLs I'm even more convinced:
the definition of `protocol_version` as _elements of an array_ rather than as _an array_, seems to be a refuse from previous eras where the elements was expanded in `header_body` while in Conway era they are (IMHO correctly) grouped into the array we are discussing in this thread.

AngelCastilloB: Correct, we can export from the ProtocolVersion file a constant with the right size of the array, and we could also add the missing arrays in the other places. 

I think there is value in having the serialization code strictly follow the CDDL, if we deviate from this we should at least document it (such as this case, we can document that the ProtocolVersion class always serializes the object inside an array. 

But I see your point. I think the main take away here is to avoid this sort of bugs from happening again. If we leave it as is, we need to remember to remove the array from the top level object every time it involves/contains a ProtocolVersion (which is why we got this bug here in the first place)

iccicci: > I think there is value in having the serialization code strictly follow the CDDL

100% agree.

One more reason to open the PR in Ledger repos I'm proposing.

AngelCastilloB: So the ProtocolVersion has been always defined as a [group](https://www.rfc-editor.org/rfc/rfc8610#section-2.1) in the CDDL. The advantage of defining it as such is that it can be composed with other types (such as the case of the header_body), when you add this type to an object its sort of like an spread operator, now in some cases you only need it by itself, and in those cases its inside an array. There are a few other types like this btw

AngelCastilloB: Marking this a resolved since Ledger team is on board with @iccicci suggestion to change the CDDL: https://github.com/IntersectMBO/cardano-ledger/pull/4178

AngelCastilloB (approved):
LGTM

PR #1141, opened 2024-03-12
chore: build smash server from the source code for Docker usage

ginnun:
# Context

The current Smash server's Dockerfile downloads and uses a binary. It creates an external dependency on that hosted binary and causes CI failures when the hosted file is removed.

# Proposed Solution
We are going to compile the smash server from the source code. This requires a compilation of multiple other libraries from the sources. So, I prefer using nix to compile the binary.

It is expected to take longer to build the docker image now.


compose/smash/Dockerfile
```diff
@@ -1,19 +1,29 @@
-ARG UBUNTU_VERSION=20.04
+FROM nixos/nix AS nix-configured
```
michalrus, line 1: If you want a minimal image (say, based on Alpine), you could build a _static_ binary of the smash server, and copy only it (no dependencies needed) from the build stage to a runtime stage. This would only replace the part downloading the binary.

If needed, I can find a Nix attribute to build the static self-contained binary

compose/smash/Dockerfile
```diff
@@ -1,19 +1,29 @@
...
 
-ARG DBSYNC_BINARY_URL="https://update-cardano-mainnet.iohk.io/cardano-db-sync/cardano-db-sync-${CARDANO_DB_SYNC_VERSION:-13.1.0.0}-linux.tar.gz"
 ENV NETWORK=${NETWORK:-"mainnet"}
 
-RUN apt-get update && apt-get install curl -y
-RUN curl ${DBSYNC_BINARY_URL} --output /tmp/dbsync-binary.tar.gz &&\
-  tar --extract --file /tmp/dbsync-binary.tar.gz --directory /bin ./cardano-smash-server &&\
-  rm /tmp/dbsync-binary.tar.gz
+# checkout the source
+RUN git clone --depth 1 --branch ${CARDANO_DB_SYNC_VERSION:-13.1.0.0} \
```
michalrus, line 13: Instead of:

```
git clone https://github.com/some-org/some-repo \
  && cd some-repo \
  && nix build .#something
```

you can do just:

```
nix build github:some-org/some-repo#something
```

compose/smash/nix/nix.conf
```diff
@@ -0,0 +1,10 @@
+#this file attached in volume to docker because we need substituters to be there before nix starts
+
+build-users-group = nixbld
+
+experimental-features = nix-command flakes
```
michalrus, line 5: You could (maybe) simplify the setup ceremony with just:

```
FROM alpine:…

RUN apk add --no-cache curl && curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix/tag/v0.11.0 | sh -s -- install linux \
  --extra-conf "sandbox = false" \
  --extra-conf "substituters = https://cache.nixos.org https://cache.iog.io" \
  --extra-conf "trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=" \
  --init none --no-confirm
```

This should allow you to run `nix build github:…#…smash-server`

michalrus (approved):
LGTM, a few (optional) comments:

compose/smash/Dockerfile
ginnun: @michalrus That's a great idea! I couldn't find how to 'package' a stand-alone binary. Could you help me with it?

compose/smash/nix/nix.conf
ginnun: I am going to try this; never thought of it.

michalrus: Yeah, I think the official NixOS image is kind of heavy, this is bare bones

compose/smash/Dockerfile
michalrus: Yes, I don't know off the top of my head, but let's try :thinking:

michalrus: I think the attribute is `static/cardano-smash-server`, so it will be something like:

```
RUN nix build -L github:input-output-hk/cardano-db-sync/${CARDANO_DB_SYNC_VERSION:-13.1.0.0}#static/cardano-smash-server

FROM alpine:… as runtime

COPY --from=buildtime /where/you/built/result/cardano-smash-server /bin/cardano-smash-server

RUN cardano-smash-server --version
```

compose/smash/nix/nix.conf
ginnun: Works as expected, thanks!

compose/smash/Dockerfile
ginnun: Hmm, that won't work before because there are linked libraries as well. (we are compiling some libraries from the source while )

michalrus: Yeah, you'd need to use this statically-linked binary instead: `#static/cardano-smash-server` – it is self-contained and has all the (previously shared) libraries inside

ginnun: @michalrus done in a908181e4f6b6d438a66cb2f8da5844425f3f8dc

compose/smash/Dockerfile
ginnun: @michalrus done in a908181e4f6b6d438a66cb2f8da5844425f3f8dc

compose/smash/nix/nix.conf
ginnun: @michalrus done in a908181e4f6b6d438a66cb2f8da5844425f3f8dc

mkazlauskas (commented):
Changes looks good but CI e2e is failing with [error: writing to file: No space left on device](https://github.com/input-output-hk/cardano-js-sdk/actions/runs/8279833317/job/22655053980?pr=1141#step:5:6370)

ginnun:
I am gonna close this PR, since we are going to use a Smash docker image prepared [here](https://github.com/IntersectMBO/cardano-db-sync/pull/1668).

PR #1143, opened 2024-03-12
ci: fix calling convention of std.yml from std-release.yml

michalrus:
# Context

Deployment on release failed – https://github.com/input-output-hk/cardano-js-sdk/actions/runs/8249290630.

# Proposed Solution

You don’t use `steps:`, see <https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow>

~~I don’t know how this ever worked… (again)~~ — it [never has](https://github.com/input-output-hk/cardano-js-sdk/actions/workflows/std-release.yaml)

n/a

### DIFF
--- a/.github/workflows/std-release.yaml
+++ b/.github/workflows/std-release.yaml
@@ -7,8 +7,7 @@ on:
 jobs:
   call-std:
     if: startsWith(github.event.release.name, '@cardano-sdk/cardano-services@')
-    steps:
-      - uses: ./.github/workflows/std.yml
-        with:
-          deploy-dev-preprod: true
-          deploy-dev-mainnet: true
+    uses: ./.github/workflows/std.yml
+    with:
+      deploy-dev-preprod: true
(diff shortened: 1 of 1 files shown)


gytis-ivaskevicius (approved):
Are you sure that this is correct syntax? I would think previous one is the correct one

michalrus:
@gytis-ivaskevicius not 100% sure until we run it, but for some evidence:

1. The [original workflow failed](https://github.com/input-output-hk/cardano-js-sdk/actions/runs/8249290630) with missing `runs-on:` which is required for a `step`, not for a `job`.
2. See the syntax in the [3rd bullet in the official docs](https://github.com/input-output-hk/cardano-js-sdk/actions/runs/8249290630):

  image

bernokl (approved):

mkazlauskas (approved):

PR #1144, opened 2024-03-12
chore: get sanchonet services ready for migration

bernokl:
# Context

Migrating dev-sanchonet from old to new clusters

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ source
 
 # direnv
 /.direnv
+/.access.local
 
 # prj-spec dirs
 # <https://github.com/numtide/prj-spec>
--- a/.gitignore
+++ b/.gitignore
@@ -27,7 +27,6 @@ source
 
 # direnv
 /.direnv
-/.access.local
 
 # prj-spec dirs
 # <https://github.com/numtide/prj-spec>
--- a/nix/cardano-services/deployments/default.nix
+++ b/nix/cardano-services/deployments/default.nix
@@ -200,8 +200,9 @@ in
         };
       };
 
-      "dev-sanchonet@us-east-1@v1" = final: {
-        name = "${final.namespace}-cardanojs-v1";
+      "dev-sanchonet@us-east-1@v1" = final: let
+        oci = inputs.cardanojs-dev-sanchonet.x86_64-linux.cardano-services.oci-images.cardano-services;
+   		in {
         namespace = "dev-sanchonet";
 
         providers = {
@@ -222,6 +223,7 @@ in
           network = "sanchonet-1";
           region = "us-east-1";
 
+          name = "${final.namespace}-cardanojs-v1";
           blockfrost-worker.enabled = false;
           pg-boss-worker.enabled = true;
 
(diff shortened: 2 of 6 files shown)


nix/cardano-services/deployments/default.nix
```diff
@@ -200,9 +200,11 @@ in
         };
       };
 
-      "dev-sanchonet@us-east-1@v1" = final: {
-        name = "${final.namespace}-cardanojs-v1";
+      "dev-sanchonet@us-east-1@v1" = final: let
+        oci = inputs.cardanojs-dev-sanchonet.x86_64-linux.cardano-services.oci-images.cardano-services;
+   		in {
```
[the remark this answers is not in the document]
bernokl: Thanks for catching that, I am used to it being part of commit pre-hook in wallet-world

nix/cardano-services/deployments/default.nix
```diff
@@ -219,7 +220,7 @@ in
         };
 
         values = {
-          network = "sanchonet-1";
+          network = "sanchonet";
```
michalrus, line 223: I think this is the only change?

michalrus (approved):
LGTM, but `treefmt` is a bit confusing, I don't know if devs run it regularly, because our run in this PR changed some TypeScript files 🤔

iccicci:
> LGTM, but `treefmt` is a bit confusing, I don't know if devs run it regularly, because our run in this PR changed some TypeScript files 🤔

Nope, we don't use it, we _lint_ the code through `eslint`.

.gitignore
```diff
@@ -27,6 +27,7 @@ source
 
 # direnv
 /.direnv
+/.access.local
```
gytis-ivaskevicius, line 30: everything should be in .envrc.local?

gytis-ivaskevicius (dismissed):

.gitignore
bernokl: Fair point, removed the entry

michalrus:
> > LGTM, but `treefmt` is a bit confusing, I don't know if devs run it regularly, because our run in this PR changed some TypeScript files 🤔
> 
> Nope, we don't use it, we _lint_ the code through `eslint`.

@iccicci but you do use `prettier` for code formatting? This is what `treefmt` is [configured to use](https://github.com/input-output-hk/cardano-js-sdk/blob/fbd851a3be28652356777d303f2025e9a56d7159/treefmt.toml#L8-L9).

michalrus (approved):

nix/cardano-services/deployments/default.nix
bernokl: correct

iccicci:
> you do use `prettier` for code formatting? This is what `treefmt` is [configured to use](https://github.com/input-output-hk/cardano-js-sdk/blob/fbd851a3be28652356777d303f2025e9a56d7159/treefmt.toml#L8-L9).

I'm not so prepared to answer correctly to this question, but as far as I remember we use `eslint` with `prettier` rules, so I'm not surprised if formatting the code through `prettier` or `eslint` gives two different results.

A possible solution could be to configure `treefmt` to format only nix files and keep TypeScript formatting as is (in a new distinct activity anyway).

michalrus:
@iccicci thank you :bow: 

@bernokl so considering ↑, I think we should exclude the whitespace changes in TypeScript files in this PR, as they would be reverted soon with `eslint`

gytis-ivaskevicius (approved):

bernokl:
Merging as only remaining changes are in .nix and readme files

PR #1151, opened 2024-03-14
LW-9685 stake pools text search

iccicci:
# Context

We need to add a text search filter to stake pools search to perform a fuzzy search on metadata.

# Proposed Solution

- Added a **fuse.js** index on stake pools metadata.
- Cached it with `InMemoryCache`.
- Inject result of fuzzy search into a temporary table to take advantage from:
  - using it with an inner join works as a where condition as well
  - the DB can perform the sort through score to respect the paginated output

- Added `ticker` to search stake pool sort options.
- For development tests I added the option to override the fuzzy options configuration provided through command line or environment variables with a new _on the fly_ value provided through the API call itself; once done I left this option for test purposes.
- Added an option to disable the test feature in `live-*` environments.


packages/cardano-services/src/StakePool/TypeormStakePoolProvider/TypeormStakePoolProvider.ts
```diff
@@ -50,9 +74,45 @@ export class TypeormStakePoolProvider extends TypeormProvider implements StakePo
...
+    const {
+      threshold,
+      weights: { description, homepage, name, poolId, ticker }
+    } = this.#overrideFuzzyOptions ? { ...this.#fuzzyOptions, ...fuzzyOptions } : this.#fuzzyOptions;
+
+    const cacheKey = this.#overrideFuzzyOptions
+      ? `all-metadata-${JSON.stringify([description, homepage, name, threshold, ticker])}`
+      : 'all-metadata';
+
+    return this.#cache.get(cacheKey, async () => {
```
mkazlauskas, line 87: We need to have a strategy for updating cache so that we don't get stale values in case server is running for a really long time

mkazlauskas (changes requested):
Interesting solution :bulb: I think it'll work, just need to refine some details :)

packages/cardano-services/src/StakePool/TypeormStakePoolProvider/TypeormStakePoolProvider.ts
```diff
@@ -72,28 +132,62 @@ export class TypeormStakePoolProvider extends TypeormProvider implements StakePo
...
+          textFilter = true;
+
+          const values = (await this.getFuse(dataSource, fuzzyOptions))
+            .search(filters.text)
+            .map(({ item: { pool_id }, score }) => `('${pool_id}',${score})`)
+            .join(',');
+
+          await queryRunner.query(
+            [
+              'DROP TABLE IF EXISTS tmp_fuzzy',
```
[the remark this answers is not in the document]
iccicci: Temporary tables are per session: many connections can create many temporary tables with the same name without interfering with each other

packages/cardano-services/src/StakePool/TypeormStakePoolProvider/TypeormStakePoolProvider.ts
iccicci: It uses the standard one created in `providerServer.ts`, I thought it was enough to behave accordingly with `DB_CACHE_TTL` value

mkazlauskas: Got it, makes sense. Existing cache strategy will still result in some slow requests once it expires. For this use case, I think it would be better if it would always have it in cache and update it periodically (instead of when it no longer exists in cache). Not a requirement for this PR, just some thoughts for future improvement.

packages/cardano-services/src/StakePool/TypeormStakePoolProvider/TypeormStakePoolProvider.ts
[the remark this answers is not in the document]
iccicci: Sorry, with session I meant a DB session, each new db connection (in **TypeORM**, a `queryRunner`) is a new session.

Each session

1. drops the table... just in case
2. creates the temporary table and populates it
3. uses it one shot
4. even if in case of errors, it tries to drop the table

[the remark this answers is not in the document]
mkazlauskas: But it's probably using a connection pool, does it re-create the session when it picks up an existing connection from the pool? :thinking:

packages/cardano-services/src/StakePool/TypeormStakePoolProvider/TypeormStakePoolProvider.ts
```diff
@@ -72,28 +132,62 @@ export class TypeormStakePoolProvider extends TypeormProvider implements StakePo
...
+      let rawResult: PoolModel[];
+      let sortByScore = false;
+      let textFilter = false;
+
+      try {
+        if (withTextFilter(filters)) {
+          sortByScore = sort === undefined;
+          textFilter = true;
+
+          const values = (await this.getFuse(dataSource, fuzzyOptions))
```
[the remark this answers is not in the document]
iccicci: e92cd83fc0 should solve this

packages/cardano-services/src/StakePool/TypeormStakePoolProvider/TypeormStakePoolProvider.ts
iccicci: LW-10063 should solve this

packages/core/src/Provider/StakePoolProvider/types/StakePoolProvider.ts
```diff
@@ -40,8 +49,14 @@ export interface QueryStakePoolsArgs {
 
     /** If provided, returns all the pools in any of the given status. */
     status?: Cardano.StakePoolStatus[];
+
+    /** Fuzzy search on text metadata fields. Ignored if shorter than 3 characters. If provided and valid, `identifier` is ignored. */
+    text?: string;
   };
 
+  /** Overrides default fuzzy options. Ignored in _live_ environments. */
+  fuzzyOptions?: FuzzyOptions;
```
mkazlauskas, line 58: This might be a potential DoS vulnerability: making a lot of requests with slightly different fuzzy options weights or threshold would make it miss cache. Is this needed with current product requirements? If so, maybe we should separately cache [the result of all stake pool metadata query](https://github.com/input-output-hk/cardano-js-sdk/pull/1151/files#diff-a3d6cb4bc9a7a5e71b9eb05a20d21dde6e10155515c1a8769c2aad1ac5c66a2cR94) :thinking: 
```suggestion
  fuzzyOptions?: FuzzyOptions;
```

packages/cardano-services/src/StakePool/TypeormStakePoolProvider/TypeormStakePoolProvider.ts
```diff
@@ -50,9 +74,51 @@ export class TypeormStakePoolProvider extends TypeormProvider implements StakePo
...
+    } = this.#overrideFuzzyOptions ? { ...this.#fuzzyOptions, ...fuzzyOptions } : this.#fuzzyOptions;
+
+    const cacheKey = this.#overrideFuzzyOptions
+      ? `all-metadata-${JSON.stringify([description, homepage, name, threshold, ticker])}`
+      : 'all-metadata';
+
+    return this.#cache.get(cacheKey, async () => {
+      const metadata = await dataSource
+        .createQueryBuilder()
+        .from(PoolMetadataEntity, 'm')
```
mkazlauskas, line 96: I think this will return multiple metadata per pool (if exists). Maybe we should query `StakePool->lastRegistration->metadata` instead?
```suggestion
        .from(StakePool, 'm')
```

packages/cardano-services/src/StakePool/TypeormStakePoolProvider/TypeormStakePoolProvider.ts
```diff
@@ -50,9 +74,51 @@ export class TypeormStakePoolProvider extends TypeormProvider implements StakePo
     this.#lastRosEpochs = lastRosEpochs;
   }
 
+  async initializeImpl() {
+    await super.initializeImpl();
+
+    void this.withDataSource((dataSource) => this.getFuse(dataSource));
```
mkazlauskas, line 80:
```suggestion
    await this.withDataSource((dataSource) => this.getFuse(dataSource));
```

mkazlauskas (changes requested):

packages/core/src/Provider/StakePoolProvider/types/StakePoolProvider.ts
iccicci: > This might be a potential DoS vulnerability

Yes, indeed I added the option to disable this in `live-*` envs.

This is not a product requirement, I added the ability to override the options through API call as I found useful during development.

Next rather than removing it (as I thought somebody else may find it useful as I did) I tried to address the abuse attempts with the option.

Once said that, yes, caching separately query result and fuzzy index could be an interesting improvement quite cheap to be achieved: I'll do this.

packages/cardano-services/src/StakePool/TypeormStakePoolProvider/TypeormStakePoolProvider.ts
iccicci: Nice catch! Thank you

packages/core/src/Provider/StakePoolProvider/types/StakePoolProvider.ts
iccicci: 6deffdf484 should solve this

packages/cardano-services/src/StakePool/TypeormStakePoolProvider/TypeormStakePoolProvider.ts
iccicci: 6deffdf484 should solve this

packages/cardano-services/src/StakePool/TypeormStakePoolProvider/TypeormStakePoolProvider.ts
iccicci: This makes the code to hand, the `withDataSource` method must be called once `initialize` procedure completed.

mkazlauskas: Why does it have to be called after initialize is completed?

iccicci: I don't know... I say this empirically: `await`ing for that call makes the code to hang.

mkazlauskas: I would like to understand this before approving the workaround :)

iccicci: It seems because [this](https://github.com/input-output-hk/cardano-js-sdk/blob/51de9c56f451c15bed5b3fdbb136e8b024c53206/packages/cardano-services/src/util/TypeormService/TypeormService.ts#L58) is called in `startImpl` which runs later than `initializeImpl`.

Calling `getFuse` during the second event rather than during the first one, should fix.
4d200a9928

mkazlauskas: So the query doesn't work until some other service is `initialize`d?

mkazlauskas: Oh ok I understand it now. I agree moving it to `start` is a good solution :handshake:

iccicci: I would say until the service itself is not `start`ed; `withDataSource` is implemented in `TypeormService` which is in the hierarchy chain of `TypeormStakePoolProvider`.

mirceahasegan (approved):
Fantastic work, @iccicci ! 💪 🥇

PR #1152, opened 2024-03-14
Feat/lw 9886 cddl conway era changes

mirceahasegan:
# Context

- Ability to array or set (backwards compatibility reading). Conway era CDDL introduces a new tagged type: [set](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/cddl-files/extra.cddl#L5) (tag 258). Previously, these tags were encoded as arrays (4).
- Ability to write as `set`

- redeemers can also be a [Map](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/cddl-files/conway.cddl#L480), not just a flat array, so our decoder must be able to detect and read accordingly.
- ability to write Map redeemers

# Draft
Still a draft until:
- [x] Fix redeemer serialization. Submitting a tx with Map redemeers is currently rejected by the cardano-submit-api with deserialization error
- [x] Once the redeemers is fixed, retarget to `master`. Currently it's targetting `conway-era` because it was easier to test submitting against the local-network

# Tests
- Ran e2e tests with the new serialization enabled. All tests except for the `phase2Validation.test.ts`, which uses redeemers, passed

  | unit test reading cbor | CSL decode SDK encoded CBOR.(manual test)
-- | -- | --
txBody inputs | ✅ | ✅
txBody certs | ✅ | ✅
txBody collaterals | ✅ | ✅
txBody required signers | ✅ | ✅
txBody ref inputs | ✅ | ✅
txBody proposal procedures | ✅ | ✅
update committee | ✅ | ✅
witness vkeywitness | ✅ | ✅
witness native_scripts | ✅ | ✅
witness bootstrap_witness | ✅ | ✅
witness plutus_v1_scripts | ✅ | ✅
witness plutus_data | ✅ | ❌ does not work for pre-existing cbor before sets:'to_json: Serialization error: JsValue("Integer -18446744073709551616 too big for our JSON support")’
witness plutus_v2_script | ✅ | ✅
witness plutus_v3_script | ✅ | -
pool_params | ✅ | ✅
redeemers map |   | no. CSL does not support it


packages/core/src/Serialization/CBOR/CborWriter.ts
```diff
@@ -28,6 +28,15 @@ const BUF_INF_POS = Buffer.from('7ff0000000000000', 'hex');
...
 /** A simple writer for Concise Binary Object Representation (CBOR) encoded data. */
 export class CborWriter {
+  /**
+   * Enable encoding `set` type as 6.258, according to the Conway era CDDL.
+   * https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/cddl-files/extra.cddl#L5
+   *
+   * Enable encoding `redeemers` as `Map`:
+   * https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/cddl-files/conway.cddl#L480
+   */
+  static useConwaySerialization = false;
```
AngelCastilloB, line 38: another nitpick, This "feature flag" seems a bit out of place here, the CborWriter is general purpose and should be agnostic to Cardano domain concepts. Also the writer doesn't really do anything different based on this flag, maybe would be better to hoist this to a different module?. But again, just a small nitpick since this is temporary

packages/core/src/Serialization/TransactionBody/Utils.ts
```diff
@@ -59,3 +60,10 @@ export const multiAssetsToTokenMap = (
 
   return tokenMap;
 };
+
+/** Peeks for a tag indicating a `set` 6.258, and performs read if present */
+export const tryReadSet = (reader: CborReader): void => {
```
AngelCastilloB, line 65: just an idea, what if we create a new serializable class 'Set.ts', which takes in the constructor an array, and knows how to serialize/deserialize sets, and we can put this in the set itself, it can deserialize both arrays with and without the tag, and use the tag to serialize only if the feature flag is enabled (and we can place the feature flag within the set module)

packages/core/src/Serialization/TransactionWitnessSet/TransactionWitnessSet.ts
```diff
@@ -206,18 +247,44 @@ export class TransactionWitnessSet {
...
-          reader.readStartArray();
-
-          while (reader.peekState() !== CborReaderState.EndArray) {
-            witness.redeemers()!.push(Redeemer.fromCbor(HexBlob.fromBytes(reader.readEncodedValue())));
+          if (reader.peekState() === CborReaderState.StartMap) {
+            // { + [ tag: redeemer_tag, index: uint ] => [ data: plutus_data, ex_units: ex_units ] }
+            reader.readStartMap();
+            while (reader.peekState() !== CborReaderState.EndMap) {
+              // Read key cbor
+              reader.readStartArray();
```
AngelCastilloB, line 258: We can assert here that the array has two elements, and throw otherwise as invalid CBOR for the redeemer key (same with the value)

packages/core/src/Serialization/TransactionWitnessSet/TransactionWitnessSet.ts
```diff
@@ -111,15 +123,38 @@ export class TransactionWitnessSet {
 
     if (this.#redeemers && this.#redeemers.length > 0) {
       writer.writeInt(5n);
-      writer.writeStartArray(this.#redeemers.length);
 
-      for (const data of this.#redeemers) {
-        writer.writeEncodedValue(Buffer.from(data.toCbor(), 'hex'));
+      if (CborWriter.useConwaySerialization) {
```
AngelCastilloB, line 127: Similar to the Set, we could create Reedemeer**s** class, which knowns how to serialize/deserialize the redeemers into an array or a map depending on the era

packages/core/src/Serialization/CBOR/CborTag.ts
```diff
@@ -48,6 +48,9 @@ export enum CborTag {
   /** Tag value for MIME messages (including all headers), as defined in RFC2045. */
   MimeMessage = 36,
 
+  /** Tag value for `set<a> = #6.258([* a]) / [* a]`, `nonempty_set<a> = #6.258([+ a]) / [+ a]`, `nonempty_oset<a> = #6.258([+ a]) / [+ a]` */
+  Set = 258,
```
AngelCastilloB, line 52: nitpick, this tag is formally called [MathematicalFinateSet](https://github.com/mofosyne/iana-headers/blob/main/c/src/cbor-constants.h#L184), and it was introduced by IOG 7 years ago to the IANA registry, I wonder by this wasn't initially introduced in the CDDL spec since it was already available

AngelCastilloB (changes requested):
Stellar job @mirceahasegan , other than a few minor comments, LGTM

packages/core/src/Serialization/TransactionBody/TransactionBody.ts
```diff
@@ -95,14 +97,9 @@ export class TransactionBody {
     //   }
     writer.writeStartMap(this.#getMapSize());
 
-    if (this.#inputs && this.#inputs.length > 0) {
+    if (this.#inputs && this.#inputs.values().length > 0) {
```
AngelCastilloB, line 100: Maybe you can add a size property to the CborSet itself, so you can simply call `this.#inputs.size()`

packages/core/src/Serialization/TransactionBody/TransactionBody.ts
```diff
@@ -506,31 +438,33 @@ export class TransactionBody {
    */
   static fromCore(coreTransactionBody: Cardano.TxBody): TransactionBody {
     const body = new TransactionBody(
-      coreTransactionBody.inputs.map((input) => TransactionInput.fromCore(input)),
+      new CborSet<TransactionInput>(coreTransactionBody.inputs.map((input) => TransactionInput.fromCore(input))),
```
AngelCastilloB, line 441: Why not implement in the CborSet class a fromCore and toCore, so you can abstract away this convertions

packages/core/src/Serialization/TransactionBody/TransactionBody.ts
```diff
@@ -10,32 +11,33 @@ import { TransactionInput } from './TransactionInput';
...
   #withdrawals: Map<Cardano.RewardAccount, Cardano.Lovelace> | undefined;
   #update: Update | undefined;
   #auxiliaryDataHash: Crypto.Hash32ByteBase16 | undefined;
   #validityStartInterval: Cardano.Slot | undefined;
   #mint: Cardano.TokenMap | undefined;
   #scriptDataHash: Crypto.Hash32ByteBase16 | undefined;
-  #collateral: Array<TransactionInput> | undefined;
-  #requiredSigners: Array<Crypto.Ed25519KeyHashHex> | undefined;
+  #collateral: CborSet<TransactionInput> | undefined;
+  #requiredSigners: CborSet<CborHexType<Crypto.Ed25519KeyHashHex>> | undefined;
```
AngelCastilloB, line 34: Why do we need CborHexType?

AngelCastilloB (commented):
Looks good @mirceahasegan just left a few more comments

packages/core/src/Serialization/Common/CborSet.ts
```diff
@@ -0,0 +1,94 @@
+import { CborReader, CborReaderState, CborTag, CborWriter } from '../CBOR';
+import { HexBlob } from '@cardano-sdk/util';
+
+export class CborHexType<T extends string> {
```
AngelCastilloB, line 4: We represent these types as Hex on core types because we want to have them in string format, but in the CDDL these are Bytestrings. I would move this to its own file and rename it. Looking at the CDDL and the way you are using it, it seems we can call this `Hash` i, e:

`CborSet<Hash<Crypto.Ed25519KeyHashHex>>)`

packages/core/src/Serialization/CBOR/CborTag.ts
mirceahasegan: Updated in the CborSet type doc

packages/core/src/Serialization/TransactionBody/TransactionBody.ts
mirceahasegan: Done in [here](https://github.com/input-output-hk/cardano-js-sdk/commit/bb6334982335b27733c25ea0b2dead9f5829bf99) and [here](https://github.com/input-output-hk/cardano-js-sdk/commit/0eee790392fe6336da9b5dc62f5f4b5ce1655a82)

packages/core/src/Serialization/TransactionBody/TransactionBody.ts
mirceahasegan: Done [here](https://github.com/input-output-hk/cardano-js-sdk/commit/eb316344a94eba2ca9b8fba73c1e78198e47da8b)

AngelCastilloB (approved):
Great work @mirceahasegan ⭐

iccicci (dismissed):
Awesome job! 🚀

AngelCastilloB (dismissed):

mkazlauskas (commented):
Great work! :muscle: bbf0bec5b4fb7d6fd0a5923af2a41f394f55207f should probably be squashed to a commit that introduced the change to certs

AngelCastilloB (approved):

PR #1156, opened 2024-03-18
Chore - dapp connector client build failing

VanessaPC:
# Context

# Proposed Solution


mkazlauskas (changes requested):
Please squash the 2 commits and change type to `chore`. `fix` would show up the CHANGELOGs, but it is not relevant to consumers of sdk packages. Also, if a package is imported in `src/`, then you need to add the tsconfig reference in `src/tsconfig.json`

VanessaPC:
> Please squash the 2 commits and change type to `chore`. `fix` would show up the CHANGELOGs, but it is not relevant to consumers of sdk packages. Also, if a package is imported in `src/`, then you need to add the tsconfig reference in `src/tsconfig.json`

No problem, I have amended the git history. 
I have not gone through the `dapp-connector-client` yet to match the packages in its own `src/tsconfig`. I opened this chore branch to decouple this work from the delegation work of the `dapp-connector`. I'll revisit it today to check it.

.github/workflows/continuous-integration-unit-tests.yaml
```diff
@@ -35,7 +35,7 @@ jobs:
       - name: 🔨 Generated files
         run: |
           git checkout yarn-project.nix
-          if [ ! -z "$(git status --porcelain)" ] ; then git status ; echo "The PR is missing some generated file" ; exit 1 ; fi
+          if [ ! -z "$(git status --porcelain)" ] ; then git status ; echo "The PR is missing some generated file" ; fi
```
VanessaPC, line 38: I am not sure why we want to check for `yarn-project.nix` but if I don't exit this exception then, the build runs fine.

mkazlauskas: > I am not sure why we want to check for yarn-project.nix

because it's inconsistent and changes on build depending on where the build is running

mkazlauskas: > exit 1 ;

Maybe you just need to run `yarn` locally and push the changes?

VanessaPC: I did this.. and there were no changes

VanessaPC: Yea, I just rebuild again locally, and normally I would have gotten a nix file, but not in this case. I am not sure why, and so I'm not sure whether I should have one.

mkazlauskas: It shouldn't matter what kind of branch it is. Please drop the `exit 1` commit and I'll take a look. Also please reword `fix: unfinished promises now fixed` to `test(e2e): await floating promises`

packages/e2e/test/web-extension/dapp-sdk/utils.ts
```diff
@@ -1,20 +1,40 @@
...
 }) => {
-  const txDetails = await builtTx.inspect();
-  textElement.textContent = `Built: ${txDetails.hash}`;
-  const signedTx = await builtTx.sign();
-  textElement.textContent = `Signed: ${signedTx.tx.id}`;
-  await connectedWallet.submitTx(signedTx);
-  textElement.textContent = `Submitted: ${signedTx.tx.id}`;
+  builtTx
+    .inspect()
+    .then((txDetails) => {
```
mkazlauskas, line 20: The original `await` syntax is preferable. Could you please revert this change (try-catch to handle errors)?

mkazlauskas (changes requested):

.github/workflows/continuous-integration-unit-tests.yaml
VanessaPC: I'll drop it, but I'd like to continue fixing the build, as it has issue created in the `dapp-connector` folder. Would you like to look at the nix issue first? otherwise, I can continue working this way, and will revert the change later.

packages/e2e/test/web-extension/dapp-sdk/utils.ts
VanessaPC: Yea, not too attached to this. Will revert.

VanessaPC:
@mkazlauskas the build now passes just fine, but I reverted the nix file change, so it will probably break there. We can address that here, or on another task. What do you prefer doing?

Dockerfile
```diff
@@ -29,6 +29,7 @@ COPY packages/cardano-services-client/package.json packages/cardano-services-cli
 COPY packages/core/package.json packages/core/package.json
 COPY packages/crypto/package.json packages/crypto/package.json
 COPY packages/dapp-connector/package.json packages/dapp-connector/package.json
+COPY packages/dapp-connector-client/package.json packages/dapp-connector-client/package.json
```
mkazlauskas, line 32: Dockerfile is for building the backend, it shouldn't need this package
```suggestion
```

mkazlauskas (changes requested):
I ran `yarn` to check in the changes that CI build is complaining about

Dockerfile
VanessaPC: Without it it breaks, and I can see the `dapp-connector` it's added. I assumed there must have been a reason for that?

mkazlauskas: I see it adds all packages, nevermind let's keep it like that, although seems like a hack

packages/e2e/test/web-extension/dapp-sdk/utils.ts
VanessaPC: This is reverted.

VanessaPC:
@mkazlauskas build is now working, and is ready for review.

PR #1157, opened 2024-03-18
chore: define live-sanchonet services

bernokl:
Migrate current live-sanchonet to the sdk repo and new cluster.


bernokl:
I think we still want clarity from @mchappell if we are ok just keeping short names for the new cluster or if we need to retain backend...eks. version as well

michalrus (approved):
LGTM, but I’d also like to know answers to Gytis' questions

PR #1160, opened 2024-03-19
fix(crypto): corrects Key-length error message

HashMapsData2Value:
# Context

While looking into how Cardano handles Bip32, specifically how it generates keys from Bip39-word entropy, I discovered that the error message in question kept correcting me for providing a 64-byte length key and not a 32-byte length key. However, the actual check is done against 96-bytes, i.e. the length of xprv after the pbkdf2 algorithm.

Perhaps the idea is that the original entropy should be 32-bytes and that's why NORMAL_ED25519_PRIVATE_KEY_LENGTH is set to that. However, that's not what the if-statement is actually checking against.

# Proposed Solution

Replace NORMAL_ED25519_PRIVATE_KEY_LENGTH with BIP32_ED25519_PRIVATE_KEY_LENGTH, the length that is actually checked against.

None.


rhyslbw (changes requested):
Thanks for the contribution @HashMapsData2Value 🤝 

In preparation for @AngelCastilloB's review, please squash the first two commits and add the scope, which in this case is the package name since the atomic change is limited to one package. i.e `fix(crypto): corrects Key-length error message`

AngelCastilloB (approved):
LGTM, thanks for the contribution! ❤️ Please squash commits as suggested by @rhyslbw and we can merge this

HashMapsData2Value:
Hey, please take a look at it now @AngelCastilloB

HashMapsData2Value:
https://github.com/input-output-hk/cardano-js-sdk/pull/1162 decided to create a new PR for expediency

rhyslbw:
Force pushing this branch would have been the way to go @HashMapsData2Value 😉

PR #1161, opened 2024-03-19
feat(hardware-ledger): miscellaneous improvements to establishing ledger connection

szymonmaslowski:
# Context

There is an ongoing lace onboarding flows revamp work. Changes in this PR are part of the improvements to the HW connection implementation

# Proposed Solution

- switched the ledger's transport lib to the one based on the `navigator.usb` Web API. Trezor supports only WebUSB API, so this way we will be able to show both device types in a single native picker.
- enabled LedgerKeyAgent to accept a device (`USBDevice` | `HID` depending on the `ConnectionType`) object and establish the connection using it.
- added a listener for a disconnection event which removes a particular connection object from the list of remembered active connections.
- method searching and picking matching active connection now tests the matched connection before returning it. In case of any problem the broken connection is removed from the remembered connections list.


packages/hardware-ledger/src/LedgerKeyAgent.ts
```diff
@@ -148,52 +148,101 @@ export class LedgerKeyAgent extends KeyAgentBase {
...
       (connection) => connection.communicationType === communicationType && connection.devicePath === devicePath
     );
+    if (!matchingConnectionData) return null;
+
+    try {
+      await this.testConnection(matchingConnectionData.deviceConnection);
+    } catch (error) {
+      if (
+        error instanceof errors.TransportError &&
+        error.message.includes('Cannot communicate with Ledger Cardano App')
```
mkazlauskas, line 165: Would be better to have this as a const defined in 1 place
```suggestion
        error.message === CONNECTION_FAILURE_ERROR_MESSAGE
```

packages/hardware-ledger/package.json
```diff
@@ -57,7 +57,7 @@
     "@cardano-sdk/util": "workspace:~",
     "@ledgerhq/hw-transport": "^6.28.1",
     "@ledgerhq/hw-transport-node-hid-noevents": "^6.27.12",
-    "@ledgerhq/hw-transport-webhid": "^6.27.12",
+    "@ledgerhq/hw-transport-webusb": "^6.27.12",
```
mkazlauskas, line 60: Please squash the 2 commits and describe the changes in commit message
```
feat(hardware-ledger): miscellaneous improvements to establishing ledger connection

describe other changes here

BREAKING CHANGE: replace LedgerKeyAgent webhid transport with webusb
```

packages/hardware-ledger/src/LedgerKeyAgent.ts
```diff
@@ -148,52 +148,101 @@ export class LedgerKeyAgent extends KeyAgentBase {
...
       throw new errors.TransportError('Cannot fetch device list', error);
     }
   }
 
+  private static attachDisconnectionCleanupHandler(transport: LedgerTransportType) {
+    const onDisconnect = () => {
+      this.deviceConnections = this.deviceConnections.filter(
+        ({ deviceConnection }) => deviceConnection.transport !== transport
+      );
+      transport.close();
```
mkazlauskas, line 195: This is an async function, linter complains here. Can it fail? Also, do we need to clean up the event listener?
```suggestion
      void transport.close();
      transport.off('disconnect', onDisconnect);
```

packages/hardware-ledger/src/LedgerKeyAgent.ts
```diff
@@ -202,35 +251,54 @@ export class LedgerKeyAgent extends KeyAgentBase {
...
+  private static async createDeviceConnection(activeTransport: LedgerTransport): Promise<LedgerConnection> {
+    const deviceConnection = new LedgerConnection(activeTransport);
+    await this.testConnection(deviceConnection);
+    return deviceConnection;
+  }
+
+  /**
+   * @throws TransportError
+   */
+  // eslint-disable-next-line complexity
```
mkazlauskas, line 263: This indicates that the function should *probably* be refactored

mkazlauskas (changes requested):

packages/hardware-ledger/src/LedgerKeyAgent.ts
tomislavhoracek: We shoul be careful here. I think you will get a disconnect event even when you open or close Cardano App since the device path will change. Could you check that @szymonmaslowski  🙏

tomislavhoracek:
@szymonmaslowski Tests in the wallet package are failing. 
https://github.com/input-output-hk/cardano-js-sdk/tree/master/packages/wallet/test/hardware/ledger
cmd: `yarn test:hw:ledger`

tomislavhoracek (changes requested):
@szymonmaslowski I left 2 comments. Could you please take a look

packages/hardware-ledger/src/LedgerKeyAgent.ts
szymonmaslowski: Definitely. I added the comment in the early stage and overlooked it when finalising the implementation. I will refactor that method :+1:

packages/hardware-ledger/src/LedgerKeyAgent.ts
szymonmaslowski: @mkazlauskas "Can it fail?" - good question. I checked the underlying implementation and the ledger's code cannot fail, but it calls the USBDevice API, which is a black box. There is another occurrence of closing the transport [here](https://github.com/input-output-hk/cardano-js-sdk/blob/514df4872b4cfe0364cba676e34f80b32ddfa7f2/packages/hardware-ledger/src/LedgerKeyAgent.ts#L310) which was implemented a long time ago and it doesn't have error handling. However, IMO the safest way would be to add error handling in both places with a simple error log.

@tomislavhoracek Yes the disconnect event will happen also on closing the cardano app. This is all fine IMO. Actually, this is the reason for this code. Do you remember the page reload on the lace side happening on error to the connection? I guess it was introduced because the connection stayed remembered and when you closed the cardano app or disconnected the cable and tried to reconnect the remembered connection was returned but that connection was already stale so lace was stuck and couldn't connect. At first, I wanted to remove this capability of remembering connections but it makes more sense to test the connection and eventually forget it. And when you forget it makes sense to close it because it won't be used.

szymonmaslowski:
> @szymonmaslowski Tests in the wallet package are failing. https://github.com/input-output-hk/cardano-js-sdk/tree/master/packages/wallet/test/hardware/ledger cmd: `yarn test:hw:ledger`

Thanks for pointing me there. I will handle it.

packages/hardware-ledger/src/LedgerKeyAgent.ts
szymonmaslowski: I created a `CARDANO_APP_CONNECTION_ERROR_MESSAGE` variable and used it when throwing an error and when catching it.

packages/hardware-ledger/src/LedgerKeyAgent.ts
```diff
@@ -29,17 +30,23 @@ import _LedgerConnection, {
...
-  : _TransportWebHID;
+const TransportWebUSB = (_TransportWebUSB as any).default
+  ? ((_TransportWebUSB as any).default as typeof _TransportWebUSB)
+  : _TransportWebUSB;
 const LedgerConnection = (_LedgerConnection as any).default
   ? ((_LedgerConnection as any).default as typeof _LedgerConnection)
   : _LedgerConnection;
 type LedgerConnection = _LedgerConnection;
 
+const isUsbDevice = (device: any): device is USBDevice =>
```
szymonmaslowski, line 44: The `USBDevice` class is not available in the nodejs runtime, so I decided to check a few properties.

packages/hardware-ledger/src/LedgerKeyAgent.ts
```diff
@@ -134,7 +145,13 @@ const isMultiSig = (tx: Transaction): boolean => {
 type LedgerConnectionWithCommunicationTypeAndDevicePath = {
   deviceConnection: LedgerConnection;
   communicationType: CommunicationType;
-  devicePath?: string;
+  nodeHidDevicePath?: string;
```
szymonmaslowski, line 148: Renamed `devicePath` to `nodeHidDevicePath` to indicate the device path exists only in the node js environment.

packages/hardware-ledger/src/LedgerKeyAgent.ts
```diff
@@ -134,7 +145,13 @@ const isMultiSig = (tx: Transaction): boolean => {
...
+  device?: LedgerDevice;
```
szymonmaslowski, line 149: Added a device object to the remembered connection data in order to retrieve existing connection for the particular device. Recognition is done by a shallow comparison of the device objects.

packages/hardware-ledger/src/LedgerKeyAgent.ts
```diff
@@ -148,98 +165,207 @@ export class LedgerKeyAgent extends KeyAgentBase {
...
+  }
+
   /**
    * @throws TransportError
    */
   static async createTransport({
     communicationType,
-    activeTransport,
-    devicePath = ''
+    nodeHidDevicePath = ''
```
szymonmaslowski, line 253: Breaking change here. I removed the `activeTransport` from params as I didn't find any usage. Is that ok? Are there any consumers of it that I'm not aware of?

packages/wallet/test/hardware/ledger/LedgerKeyAgent.test.ts
```diff
@@ -10,298 +10,382 @@ import {
...
+const cleanupEstablishedConnections = async () => {
+  if (LedgerKeyAgent.deviceConnections.length === 0) return;
+  for (const { deviceConnection } of LedgerKeyAgent.deviceConnections) {
+    await deviceConnection.transport.close();
+  }
+  LedgerKeyAgent.deviceConnections = [];
+};
+
 describe('LedgerKeyAgent', () => {
-  let ledgerKeyAgent: LedgerKeyAgent;
```
szymonmaslowski, line 23: I reorganized the structure slightly. I moved all the code and tests related to creating `LedgerKeyAgent` instance and using it together with `PersonalWallet` under the dedicated describe so it does not hang around while executing other tests.

packages/hardware-ledger/src/LedgerKeyAgent.ts
mkazlauskas: A union type based on `communicationType` would be better

packages/hardware-ledger/src/LedgerKeyAgent.ts
mkazlauskas: Same as https://github.com/input-output-hk/cardano-js-sdk/pull/1161/files#r1540811427

packages/hardware-ledger/src/LedgerKeyAgent.ts
```diff
@@ -148,98 +165,207 @@ export class LedgerKeyAgent extends KeyAgentBase {
...
       throw new errors.TransportError('Cannot fetch device list', error);
     }
   }
 
+  private static async closeTransport(transport: LedgerTransportType) {
+    try {
+      await transport.close();
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error('Failed to close the transport', transport, error);
```
mkazlauskas, line 213: Please add `Logger` as a dependency instead
```suggestion
      this.#logger.error('Failed to close the transport', transport, error);
```

packages/hardware-ledger/src/LedgerKeyAgent.ts
```diff
@@ -148,98 +165,207 @@ export class LedgerKeyAgent extends KeyAgentBase {
...
+    transport.on('disconnect', onDisconnect);
+  }
+
+  /**
+   * @throws TransportError
+   */
+  private static async openTransportForDevice({ communicationType, device }: OpenTransportForDeviceParams) {
+    let transport: LedgerTransportType;
+    try {
+      if (communicationType === CommunicationType.Node && device instanceof HID) {
```
mkazlauskas, line 234: I think it's better to be defensive when using globals, as it might not exist in some environments. Same for USBDevice
```suggestion
      if (communicationType === CommunicationType.Node && typeof HID === 'function' && device instanceof HID) {
```

packages/hardware-ledger/src/LedgerKeyAgent.ts
```diff
@@ -148,98 +165,207 @@ export class LedgerKeyAgent extends KeyAgentBase {
...
+    if (nodeHidDevicePathOrDevice instanceof HID || isUsbDevice(nodeHidDevicePathOrDevice)) {
+      device = nodeHidDevicePathOrDevice;
+    }
+    if (typeof nodeHidDevicePathOrDevice === 'string') {
+      nodeHidDevicePath = nodeHidDevicePathOrDevice;
+    }
+
+    return {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore TS2454 used before being assigned
```
mkazlauskas, line 303: I think it's better to obey to ts rules rather than disabling

```suggestion
    let device: LedgerDevice;
    let nodeHidDevicePath: string | undefined;

    if (nodeHidDevicePathOrDevice instanceof HID || isUsbDevice(nodeHidDevicePathOrDevice)) {
      device = nodeHidDevicePathOrDevice;
    }
    if (typeof nodeHidDevicePathOrDevice === 'string') {
      nodeHidDevicePath = nodeHidDevicePathOrDevice;
    }

    if (!device) throw ...
    
    return {
```

mkazlauskas (changes requested):

packages/hardware-ledger/src/LedgerKeyAgent.ts
szymonmaslowski: Right, I overlooked the usage of the `USBDevice` here. With the `HID` we are safe because it is imported.

packages/hardware-ledger/src/LedgerKeyAgent.ts
szymonmaslowski: I totally agree about obeying the rules. I just didn't come up with your solution 😉 
Regarding that throw you added - it is expected that both variables will stay undefined.

packages/hardware-ledger/src/LedgerKeyAgent.ts
mkazlauskas: > Breaking change here. I removed the activeTransport from params as I didn't find any usage. Is that ok?
It might be breaking `tsc` (I think it depends on tsconfig) but is not a breaking change in js, so it's very minor and I wouldn't consider it breaking

packages/hardware-ledger/src/LedgerKeyAgent.ts
mkazlauskas: Oh I see, didn't notice it's in a static method. In that case it wouldn't be as clean to inject. If this log isn't essential (seems like not?) then I suggest replacing this line with a comment about the implications of failing to close (if any)

packages/hardware-ledger/src/LedgerKeyAgent.ts
szymonmaslowski: I meant a breaking change on an API level. If there was someone using this method we would need to release this change under a new major version, to avoid breaking his app.

packages/hardware-ledger/src/LedgerKeyAgent.ts
szymonmaslowski: In this case I would prefer to keep a slightly loose type. If we go with a union this will affect other places where we will have to check for the value of the communicationType to access particular properties which will complicate implementation. IMO it is pretty safe to keep this type as is.

packages/hardware-ledger/src/LedgerKeyAgent.ts
szymonmaslowski: I decided to step back and remove this method. It is a low probability that the `transport.close` will fail and affect the logic, so IMO it is not worth making additional effort. If it fails we will notice that as a console error and that's enough.

packages/hardware-ledger/src/LedgerKeyAgent.ts
szymonmaslowski: Re: throw - Added a throw in case the user passed an invalid parameter against TS. In such case, it would fail anyway on further implementation but this approach is better.

packages/hardware-ledger/src/LedgerKeyAgent.ts
mkazlauskas: I don't see this as a complication, but as an assurance (by tsc) that code can handle different types of connection objects

packages/hardware-ledger/src/LedgerKeyAgent.ts
mkazlauskas: > I meant a breaking change on an API level

Me too, we are removing a property, not adding it, so if you have:
```ts
createTransport({ nodeHidDevicePath, ... })
```

It would still work after updating this package in most cases. For typescript users tsc might give you an error, but I believe not always (depends on tsconfig)

packages/hardware-ledger/src/LedgerKeyAgent.ts
szymonmaslowski: oki doki, let me attempt it and let's evaluate the result 😉

packages/hardware-ledger/src/LedgerKeyAgent.ts
szymonmaslowski: I improved it. To make it descriptive I embed the method name that the error message relates to. In order to ensure the message stays accurate in case the method gets renamed I created [a variable](https://github.com/input-output-hk/cardano-js-sdk/pull/1161/files#diff-19303657c074c5ebc3361bdd09aee5cf4d2c2e5b7de7a019d5bdb04f249e20bdR100) for that name and I am using it to [create the method](https://github.com/input-output-hk/cardano-js-sdk/pull/1161/files#diff-19303657c074c5ebc3361bdd09aee5cf4d2c2e5b7de7a019d5bdb04f249e20bdR310-R323) and in [the error message](https://github.com/input-output-hk/cardano-js-sdk/pull/1161/files#diff-19303657c074c5ebc3361bdd09aee5cf4d2c2e5b7de7a019d5bdb04f249e20bdR120). Let me know it this approach is fine for you.

packages/hardware-ledger/src/LedgerKeyAgent.ts
szymonmaslowski: @mkazlauskas Please take a look on those two approaches:
1. discriminated union which you suggested: https://github.com/input-output-hk/cardano-js-sdk/compare/feat/lw-9808-lace-HW-onboarding-revamp...feat/lw-9808-lace-HW-onboarding-revamp-a
2. simplification of the type: https://github.com/input-output-hk/cardano-js-sdk/compare/feat/lw-9808-lace-HW-onboarding-revamp...feat/lw-9808-lace-HW-onboarding-revamp-b

From the functionality perspective, they are equal - they give the same result reusing remembered connection
From the simplicity/maintainability perspective, I favor the second solution. IMO it is simpler for other so work with this implementation.

What are your thoughts?

mkazlauskas: I like the union because it is safer, less combinations to worry about when using the object

mkazlauskas (approved):
Looks good! ~~Please squash into 1 commit before merging~~

tomislavhoracek (approved):
🚀

tomislavhoracek:
https://github.com/input-output-hk/cardano-js-sdk/pull/1161/commits/a63795ac4987527738ed780433c0d9cab7214e17 ✅

mkazlauskas (approved):

tomislavhoracek (approved):

PR #1162, opened 2024-03-20
fix(crypto): replace variable in keylength error message

HashMapsData2Value:
# Context

While looking into how Cardano handles Bip32, specifically how it generates keys from Bip39-word entropy, I discovered that the error message in question kept correcting me for providing a 64-byte length key and not a 32-byte length key. However, the actual check is done against 96-bytes, i.e. the length of xprv after the pbkdf2 algorithm.

Perhaps the idea is that the original entropy should be 32-bytes and that's why NORMAL_ED25519_PRIVATE_KEY_LENGTH is set to that. However, that's not what the if-statement is actually checking against.

# Proposed Solution

Replace NORMAL_ED25519_PRIVATE_KEY_LENGTH with BIP32_ED25519_PRIVATE_KEY_LENGTH, the length that is actually checked against.

None.

### DIFF
--- a/packages/crypto/src/Bip32/Bip32PrivateKey.ts
+++ b/packages/crypto/src/Bip32/Bip32PrivateKey.ts
@@ -2,7 +2,7 @@
 import * as Bip32KeyDerivation from './Bip32KeyDerivation';
 import { Bip32PrivateKeyHex } from '../hexTypes';
 import { Bip32PublicKey } from './Bip32PublicKey';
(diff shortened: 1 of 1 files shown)


HashMapsData2Value:
@AngelCastilloB Hey decided to just close the other PR and open it here again with a clean branch.

AngelCastilloB (approved):
LGTM, thanks for the contribution!

rhyslbw (approved):
Thanks again @HashMapsData2Value

HashMapsData2Value:
Fixed so my commit is signed now 👍

HashMapsData2Value:
Looks like there's a failing check, probably because it's trying to reach out for AWS credentials which don't exist in my forked repo.

PR #1163, opened 2024-03-20
LW-9898 conway era tx

iccicci:
# Context

We need to make the `PersonalWallet/conwayTransactions` more robust, sending the DRep Reg Cert only if non already registered as DRep.

# Proposed Solution

Check the txs history to find relevant certificates in order to decide if sending or not the DRep Reg Cert.


mirceahasegan (approved):
Great work, @iccicci  🚀

mchappell (approved):
Great work, we'll use this as the basis for the drepTracker 🚀

PR #1166, opened 2024-03-21
feat(web-extension): added an optional parameter to WalletManager acivate to allow force reload

AngelCastilloB:
# Context

We want to be able to force the reloading of a wallet even if its properties haven't changed. this is useful when configuring wallet providers differently.

# Proposed Solution

Add a new optional field "force" that if set to true always re-activates the wallet.

A new param 'force' has been added to the wallet manager

### DIFF
--- a/packages/web-extension/src/walletManager/walletManager.ts
+++ b/packages/web-extension/src/walletManager/walletManager.ts
@@ -116,10 +116,16 @@ export class WalletManager<WalletMetadata extends { name: string }, AccountMetad
   /**
    * Create and activate a new ObservableWallet.
    *
-   * @param props The properties of the wallet to activate.
+   * @param props - An object containing the necessary properties and configurations to activate the wallet.
+   * @param force - Optional. A boolean flag that determines the activation behavior. If set to `true`,
+   *                the wallet will be activated regardless of whether its properties have changed since
+   *                the last activation. This is useful for scenarios where reinitialization is needed
+   *                without changes to the properties. Defaults to `false`, meaning the wallet will only
+   *                be activated if there have been changes in the `props`.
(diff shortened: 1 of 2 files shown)


packages/web-extension/src/walletManager/walletManager.ts
```diff
@@ -116,10 +116,16 @@ export class WalletManager<WalletMetadata extends { name: string }, AccountMetad
...
+   * @param force - Optional. A boolean flag that determines the activation behavior. If set to `true`,
+   *                the wallet will be activated regardless of whether its properties have changed since
+   *                the last activation. This is useful for scenarios where reinitialization is needed
+   *                without changes to the properties. Defaults to `false`, meaning the wallet will only
+   *                be activated if there have been changes in the `props`.
+   * @returns A Promise that resolves once the wallet has been successfully activated.
    */
-  async activate(props: WalletManagerActivateProps): Promise<void> {
-    if (this.#isActive(props)) {
+  async activate(props: WalletManagerActivateProps, force?: boolean): Promise<void> {
```
mkazlauskas, line 127: Could it be a property in `WalletManagerActivateProps`?
```suggestion
  async activate(props: WalletManagerActivateProps): Promise<void> {
```

AngelCastilloB: WalletManagerActivateProps is tied to a specific wallet/account, and also is stored on local storage, so if you ever activate with the flag on, everytime you reload this wallet this flag will be retrieve too (as in this particular wallet always force reload)

mkazlauskas (approved):
Looks good!

rhyslbw (approved):

PR #1168, opened 2024-03-21
feat(cardnao-services): txSubmitApiProvider now takes axios adapter

AngelCastilloB:
# Context

In order to enable custom TX Submit endpoints in Lace, we need to modify our TxSubmitApiProvider to take an axio adapter so it can work on browser environments.

# Proposed Solution

Add a new parameter to the TxSubmitApiProvider so it can take the Axio adapter for the browser.

TxSubmitApiProvider now takes an extra argument in its constructor

### DIFF
--- a/packages/cardano-services-client/src/TxSubmitProvider/TxSubmitApiProvider.ts
+++ b/packages/cardano-services-client/src/TxSubmitProvider/TxSubmitApiProvider.ts
@@ -2,18 +2,33 @@ import { Cardano, ProviderError, ProviderFailure, SubmitTxArgs, TxBodyCBOR, TxSu
 import { Logger } from 'ts-log';
 import { hexStringToBuffer } from '@cardano-sdk/util';
 import { mapCardanoTxSubmitError } from './cardanoTxSubmitErrorMapper';
-import axios, { AxiosInstance } from 'axios';
+import axios, { AxiosAdapter, AxiosInstance } from 'axios';
+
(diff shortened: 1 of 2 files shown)


ginnun (approved):
Clean job, lgtm.

packages/cardano-services-client/src/TxSubmitProvider/TxSubmitApiProvider.ts
```diff
@@ -25,6 +40,7 @@ export class TxSubmitApiProvider implements TxSubmitProvider {
       this.#logger.debug(`Submitting tx ${txId} ...`);
 
       await this.#axios({
+        adapter: this.#adapter,
```
mkazlauskas, line 43: A test like `it('supports custom axios adapter')` would be nice. Although I don't see any tests for this provider, in which case it's not needed and it's just tech debt.

mkazlauskas (approved):

mirceahasegan (approved):
Nice work, @AngelCastilloB 🚀

packages/cardano-services-client/src/TxSubmitProvider/TxSubmitApiProvider.ts
mkazlauskas: @iccicci @mirceahasegan was this merged from conway-era and we forgot to write tests? Don't remember why we added the provider :thinking:

PR #1174, opened 2024-03-25
LW-9422 committee certificate details

iccicci:
# Context

With **db-sync** `sancho-4.1.0` we have the ability to distinct between constitutional committee with or without script.

# Proposed Solution

Added the distinction in `DbSyncChainHistoryProvider`.


mirceahasegan (approved):
Nice work! 🚀

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
```diff
@@ -92,7 +92,7 @@ const getGovernanceAction = ({
   numerator,
   type
 }: ProposalProcedureModel): Cardano.GovernanceAction => {
-  const { contents } = JSON.parse(description);
+  const { contents } = description;
```
mkazlauskas, line 95: This doesn't seem to be validated with openapi schema. It would be good to check the shape of this object to catch db-sync schema changes

mkazlauskas (changes requested):

mkazlauskas (approved):

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
iccicci: LW-10179 should solve this

PR #1175, opened 2024-03-25
[LW-10090] Add SetValidityInterval to txBuilder()

lucas-barros:
# Context

Add setValidityInterval to txBuilder

# Proposed Solution


VanessaPC (approved):
I was going to ask what happened to `invalidBefore` but I see it's descoped, so in this case, it looks great to me!

renanvalentin (approved):
LGTM 👍

PR #1185, opened 2024-03-28
ci: add staging-preprod to std-release.yaml workflow

gytis-ivaskevicius:
# Context

Sorry guys, realized that I forgot to add this

PR #1186, opened 2024-03-29
ci: deploy staging-* only on demand

michalrus:
# Context

[LW-10177](https://input-output.atlassian.net/browse/LW-10177)

Tweaks input-output-hk/cardano-js-sdk#1185 as per [Rhys' comment](https://input-output.atlassian.net/browse/LW-10177?focusedCommentId=272226).

[LW-10177]: https://input-output.atlassian.net/browse/LW-10177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/std-release.yaml
+++ b/.github/workflows/std-release.yaml
@@ -9,7 +9,8 @@ jobs:
     if: startsWith(github.event.release.name, '@cardano-sdk/cardano-services@')
     uses: ./.github/workflows/std.yml
     with:
-      deploy-staging-preprod: true
+      # staging-* is deployed, with the versions and services identical to dev-*, at the start of the Lace release process:
+      deploy-staging-preprod: false
       deploy-dev-preprod: true
       deploy-dev-mainnet: true
     # Secrets have to be passed explicitly even if callee is in the same repo:


gytis-ivaskevicius (dismissed):

.github/workflows/std-release.yaml
```diff
@@ -9,7 +9,8 @@ jobs:
     if: startsWith(github.event.release.name, '@cardano-sdk/cardano-services@')
     uses: ./.github/workflows/std.yml
     with:
-      deploy-staging-preprod: true
+      # Only deploy staging-* on demand, otherwise they don’t differ from dev-*:
```
rhyslbw, line 12: Does this need to be explicit? Should we not just remove it from the list here? I don't feel strongly either way, but it seems a bit inconsistent since `dev-preview` is not here

michalrus: This is because `dev-preview` [is not in `workflow_call: inputs:`](https://github.com/input-output-hk/cardano-js-sdk/blob/e177883e406baafbac3aeb8802e8f8c2aa022e26/.github/workflows/std.yml#L26-L36).

Maybe for consistency, we should add it but set `required: false` for both `deploy-dev-preview` and `deploy-staging-preprod`, and then don't list them explicitly here?

Maybe set `required: false` for all, which would make `false` their default :)

michalrus: Or remove `deploy-staging-preprod` from the `workflow_call: inputs:`? 🤔 All of these are equivalent functionality-wise, but which is the most readable?

rhyslbw, line 12:
```suggestion
      # staging-* is deployed, with the versions and services identical to dev-*, at the start of the Lace release process:
```

rhyslbw: My first though is the correct semantics seems to be:

> remove deploy-staging-preprod from the workflow_call: inputs:

... as this says the workflow should only be invoked with a dispatch, right? However, if that's the case, why doesn't `dev-preview` need to be listed in `workflow_call: inputs`?

michalrus: Because `dev-preview` is deployed on each merge to `master` (or with a manual `workflow_dispatch:`), so it's most probably already been deployed once a release happens.

But I can add it to release's `workflow_call:` for consistency, it won't hurt

rhyslbw: I’ll look at it again in more depth. I’m more concerned about semantics here, so just do the minimum to make it work. If that’s what you already have, please ignore my comment 😬

michalrus: Applied in 9a6727a61e287514b3349afc53f47dba42bd4b98 ([compare](https://github.com/input-output-hk/cardano-js-sdk/compare/e177883e406baafbac3aeb8802e8f8c2aa022e26..9a6727a61e287514b3349afc53f47dba42bd4b98))

michalrus: Okay, so I think the minimal change regarding the diff is an explicit `s/true/false/`, which we have now.

I have implemented your comment suggestion below.

I'll be happy to implement the future results of your thinking on this :)

pczeglik-iohk (approved):

gytis-ivaskevicius (approved):

PR #1187, opened 2024-03-29
chore: add localhost as allowed origin to dev-*

gytis-ivaskevicius:
# Context

https://input-output.atlassian.net/browse/LW-9965

# Testing

Can someone deploy this to dev-preview and test it from a local machine? I am not sure how devs run the extension locally, if it is opened as a website - it should be good. If it is opened as extension, then it probably requires prefix but in that case i'd assume that it would have a custom hash in development mode as well

### DIFF
--- a/nix/cardano-services/deployments/default.nix
+++ b/nix/cardano-services/deployments/default.nix
@@ -12,6 +12,21 @@
     us-east-2 = readJsonFile ./tf-outputs/lace-prod-us-east-2.json;
   };
   oci = inputs.self.x86_64-linux.cardano-services.oci-images.cardano-services;
+  allowedOrigins =  [
+    # Represents Chrome production version
+    "chrome-extension://gafhhkghbfjjkeiendhlofajokpaflmk"
+    # Represents Edge production version
+    "chrome-extension://efeiemlfnahiidnjglmehaihacglceia"
+    # Represents midnights version of lace
+    "chrome-extension://bjlhpephaokolembmpdcbobbpkjnoheb"
+    # Represents Chrome dev preview version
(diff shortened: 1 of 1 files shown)


nix/cardano-services/deployments/default.nix
```diff
@@ -200,6 +201,8 @@ in
             ];
           };
 
+          backend.allowedOrigins = allowedOrigins ++ ["localhost"];
```
michalrus, line 204: I think a string (`allowedOrigins`) is added to a list `["localhost"]`, and `nix run .#⟨…⟩.plan` [is failing](https://github.com/input-output-hk/cardano-js-sdk/actions/runs/8477293822/job/23228036808?pr=1187#step:7:274):

```
       error: value is a string while a list was expected
```

nix/cardano-services/deployments/default.nix
```diff
@@ -12,6 +12,16 @@
     us-east-2 = readJsonFile ./tf-outputs/lace-prod-us-east-2.json;
   };
   oci = inputs.self.x86_64-linux.cardano-services.oci-images.cardano-services;
+  allowedOrigins =  [
+    # Represents Chrome production version
+    "chrome-extension://gafhhkghbfjjkeiendhlofajokpaflmk"
+    # Represents Edge production version
+    "chrome-extension://efeiemlfnahiidnjglmehaihacglceia"
+    # Represents midnights version of lace
```
pczeglik-iohk, line 20: `M`idnight :)

pczeglik-iohk (approved):

mkazlauskas (approved):

PR #1194, opened 2024-04-02
feat(wallet): add no wallet confirmation cb into the baseCip30WalletApi

vetalcore:
# Context

createWalletApi would be called with `null` as `wallet$` value in case DApp is authorised with `Always` settings (service worker won't be updated), as a result user will no see `no wallet` modal.

# Proposed Solution

provide additional `onNoWalletError`  confirmationCallback within `baseCip30WalletApi`

### DIFF
--- a/packages/wallet/src/cip30.ts
+++ b/packages/wallet/src/cip30.ts
@@ -1,3 +1,5 @@
+/* eslint-disable sonarjs/cognitive-complexity */
+/* eslint-disable sonarjs/no-duplicate-string */
 import {
   APIErrorCode,
   ApiError,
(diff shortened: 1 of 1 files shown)


packages/wallet/src/cip30.ts
```diff
@@ -272,6 +281,11 @@ const baseCip30WalletApi = (
     logger.debug('getting changeAddress');
     try {
       const wallet = await firstValueFrom(wallet$);
+      if (!wallet) {
```
vetalcore, line 284: @mkazlauskas maybe you could suggest if there is a way not to duplicate same checks in every method? maybe by wrapping it with some proxy handler and checking wallet observable before accessing the exact method?
if it makes sense of course*

mkazlauskas: I suggest a higher order function like `withWallet((wallet) = {...})`

vetalcore:
not a priority atm

PR #1195, opened 2024-04-03
fix(input-selection): greedy input selector now correctly distributes assets among addresses

AngelCastilloB:
# Context

There was a bug in the greedy input selector that was causing some of the change outputs to be discarded if the UTXO didn't have enough ADA to hold more native assets.

# Proposed Solution

When an UTXO can not hold any more native assets (due to its min value constraint), the UTXO needs to be removed from the list of UTXOs that can hold assets, but have to be later merged back with the rest of the assets before returning the selection.


mkazlauskas (approved):
Great work! :detective:

iccicci (approved):
LGTM!

PR #1201, opened 2024-04-05
chore: split smash url per env

bernokl:
Goal is to let us update smash url  per environment

### DIFF
--- a/nix/cardano-services/deployments/default.nix
+++ b/nix/cardano-services/deployments/default.nix
@@ -113,14 +113,12 @@ in
           resources.limits = mkPodResources "300Mi" "500m";
           resources.requests = mkPodResources "150Mi" "100m";
         };
-        pg-boss-worker =  let
-          smash-url = if values.network == "mainnet"
-                         then "https://smash.cardano-mainnet.iohk.io/api/v1"
-                         else "https://${values.network}-smash.world.dev.cardano.org/api/v1";
-        in {
+        pg-boss-worker = {
            enabled = false;
            metadata-fetch-mode = "smash";
-           smash-url = smash-url;
+           smash-url = if values.network == "mainnet"
+                         then "https://smash.cardano-mainnet.iohk.io/api/v1"
+                         else "https://${values.network}-smash.world.dev.cardano.org/api/v1";
            resources.limits = mkPodResources "300Mi" "300m";
            resources.requests = mkPodResources "150Mi" "200m";
         };
--- a/nix/cardano-services/deployments/default.nix
+++ b/nix/cardano-services/deployments/default.nix
@@ -117,6 +117,7 @@ in
         pg-boss-worker = {
           enabled = false;
           metadata-fetch-mode = "direct";
+          smash-url = "https://smash.cardano-mainnet.iohk.io/api/v1";
           resources.limits = mkPodResources "300Mi" "300m";
           resources.requests = mkPodResources "150Mi" "200m";
         };
@@ -211,6 +212,7 @@ in
           blockfrost-worker.enabled = true;
           pg-boss-worker.enabled = true;
           pg-boss-worker.metadata-fetch-mode = "smash";
+          pg-boss-worker.smash-url = "https://preview-smash.world.dev.cardano.org";
         };
       };
 
@@ -330,6 +332,7 @@ in
           blockfrost-worker.enabled = true;
           pg-boss-worker.enabled = true;
           pg-boss-worker.metadata-fetch-mode = "smash";
+          pg-boss-worker.smash-url = "https://smash.cardano-mainnet.iohk.io/api/v1";
         };
       };
 
@@ -347,6 +350,7 @@ in
         values = {
           network = "preprod";
           region = "us-east-1";
+
 
           backend.hostnames = ["${final.namespace}.${baseUrl}"];
           backend.passHandleDBArgs = false;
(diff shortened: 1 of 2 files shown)


iccicci (changes requested):
Many URLs lack the `/api/v1` path.

More than this there are some indentation problems.

nix/cardano-services/deployments/default.nix
```diff
@@ -211,6 +212,7 @@ in
           blockfrost-worker.enabled = true;
           pg-boss-worker.enabled = true;
           pg-boss-worker.metadata-fetch-mode = "smash";
+          pg-boss-worker.smash-url = "https://preview-smash.world.dev.cardano.org";
```
iccicci, line 215: I would say path `/api/v1` is required here as well.

Is there some reason I don't know or is it just a refuse which still needs to be addressed?

nix/cardano-services/deployments/default.nix
```diff
@@ -401,8 +405,11 @@ in
           backend.allowedOrigins = lib.concatStringsSep "," allowedOriginsDev;
           backend.hostnames = ["${final.namespace}.${baseUrl}"];
 
-          blockfrost-worker.enabled = true;
           pg-boss-worker.enabled = true;
+          pg-boss-worker.metadata-fetch-mode = "smash";
+          pg-boss-worker.smash-url = "https://preprod-smash.world.dev.cardano.org/";
```
iccicci, line 410: Same comment about path here.

nix/cardano-services/deployments/default.nix
```diff
@@ -476,6 +483,8 @@ in
 
           backend.hostnames = ["${final.namespace}.${baseUrl}"];
           blockfrost-worker.enabled = true;
+          pg-boss-worker.metadata-fetch-mode = "direct";
+          pg-boss-worker.smash-url = "https://preprod-smash.world.dev.cardano.org/";
```
iccicci, line 487: Same comment about path here.

nix/cardano-services/deployments/default.nix
```diff
@@ -556,6 +565,8 @@ in
 
           backend.hostnames = ["backend.${final.namespace}.eks.${baseUrl}" "${final.namespace}.${baseUrl}"];
           blockfrost-worker.enabled = true;
+          pg-boss-worker.metadata-fetch-mode = "direct";
+          pg-boss-worker.smash-url = "https://preprod-smash.world.dev.cardano.org/";
```
iccicci, line 569: Same comment about path here.

nix/cardano-services/deployments/default.nix
```diff
@@ -637,6 +648,8 @@ in
           backend.hostnames = ["backend.${final.namespace}.eks.${baseUrl}" "${final.namespace}.${baseUrl}"];
           blockfrost-worker.enabled = true;
           pg-boss-worker.enabled = true;
+          pg-boss-worker.metadata-fetch-mode = "direct";
+          pg-boss-worker.smash-url = "https://preview-smash.world.dev.cardano.org/";
```
iccicci, line 652: Same comment about path here.

nix/cardano-services/deployments/default.nix
```diff
@@ -705,6 +718,8 @@ in
           region = "us-east-2";
 
           blockfrost-worker.enabled = false;
+          pg-boss-worker.metadata-fetch-mode = "direct";
+					pg-boss-worker.smash-url = "https://sanchonet-smash.world.dev.cardano.org/";
```
iccicci, line 722: Same comment about path here.

iccicci (changes requested):
Sorry, I realized about another problem only now.

Since we are configuring all the environments to use smash servers, `pg-boss-worker.metadata-fetch-mode` should be set to `"smash"` for all the environments as well; maybe it worth to apply this setting as a global setting rather than having it on each environment.

nix/cardano-services/deployments/default.nix
bernokl: I only did manual testing of the url provided in slack. Thank you for confirming it needs the /api/v1

bernokl:
> Since we are configuring all the environments to use smash servers, pg-boss-worker.metadata-fetch-mode should be set to "smash" for all the environments as well; maybe it worth to apply this setting as a global setting rather than having it on each environment.


There is a default that used to be direct I set to smash and made the url a default as well so we can just override on deviation

bernokl:
Manually confirmed code only made url update as expected when ran against dev-preview

iccicci (approved):
LGTM!

nix/cardano-services/deployments/default.nix
```diff
@@ -113,12 +113,16 @@ in
...
-          resources.limits = mkPodResources "300Mi" "300m";
-          resources.requests = mkPodResources "150Mi" "200m";
+        pg-boss-worker =  let
+          smash-url = if values.network == "mainnet"
+                         then "https://smash.cardano-mainnet.iohk.io/api/v1"
+                         else "https://${values.network}-smash.world.dev.cardano.org/api/v1";
+        in {
+           enabled = false;
+           metadata-fetch-mode = "smash";
+           smash-url = smash-url;
```
gytis-ivaskevicius, line 123:
```suggestion
        pg-boss-worker =  {
           enabled = false;
           metadata-fetch-mode = "smash";
           smash-url = if values.network == "mainnet"
                         then "https://smash.cardano-mainnet.iohk.io/api/v1"
                         else "https://${values.network}-smash.world.dev.cardano.org/api/v1";```

gytis-ivaskevicius (approved):

michalrus (dismissed):

gytis-ivaskevicius (approved):

michalrus (approved):

PR #1202, opened 2024-04-08
LW-9889 Re-enable skipped tests

iccicci:
# Context

With the introduction of `cardano-node 8.8` some test was disabled: we need to re-enable them.

# Proposed Solution

- `buildAndQuery, queryPoolHashes & queryTotalCount`: was enough to re-enable it, most likely previously rebuild fixtures was enough;
- `finds transactions with address within outputs`: it uses `fixtureBuilder.getGenesisAddresses()`; I have no idea about what was changed from previous version to make genesis addresses no longer having ids starting from 1, but I know which is the value of their utxos: changed the `fixtureBuilder` to get genesis addresses regardless from their ids;
- `Ada handle`: it seems a bug was introduced in `cardano-submit-api`: submitting txs through `ogmios` fixes the problem.

### DIFF
--- a/packages/cardano-services/test/ChainHistory/ChainHistoryHttpService.test.ts
+++ b/packages/cardano-services/test/ChainHistory/ChainHistoryHttpService.test.ts
@@ -516,8 +516,7 @@ describe('ChainHistoryHttpService', () => {
           expect(response.pageResults[0].body.inputs).toMatchShapeOf(DataMocks.Tx.inputs);
         });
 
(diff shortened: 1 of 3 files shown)


mirceahasegan (approved):
Great work, @iccicci !

rhyslbw (commented):
Nice, that's comforting to see the minimal changes required to re-enable these tests. Just a minor note regarding commit scoping, each item from the list in the description is independent so committing separately (with the text you provided in the emphemeral PR 😄 ) is the preferred approach. We want to capture this information in the git log

mirceahasegan (approved):
LGTM 🚀

PR #1208, opened 2024-04-11
fix: the projection now always stores blocks when it reaches tip - k

AngelCastilloB:
# Context

We are getting an error in the projection service on rollback:

`Error: Could not rollback to bed4a6a2e9cf54d61cd56dd06d7b5e9990e7a51cebb137b48316df92df5ee439: tip block not found in stability window buffer`

This is because the projection services now selectively stores blocks based on which stores are enabled and certain conditions, however, for rollback purposes we should always store the blocks inside the stability window.

# Proposed Solution

Always stores block when the projection reaches tip - k height.


pparker:
Jira ticket created retrospectively for this: https://input-output.atlassian.net/browse/LW-10276

cc @AngelCastilloB

iccicci (dismissed):
LGTM!

packages/cardano-services/src/Projection/prepareTypeormProjection.ts
```diff
@@ -383,8 +385,10 @@ export const prepareTypeormProjection = (
     mappers: selectedMappers,
     stores: selectedStores,
     willStore: <T extends ProjectionEvent>(evt: T) =>
+      evt.eventType === ChainSyncEventType.RollBackward ||
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      evt.eventType === ChainSyncEventType.RollBackward || willStoreCheckers.some((check) => check(evt as any))
+      willStoreCheckers.some((check) => check(evt as any)) ||
+      willStoreBlockData(evt)
```
mkazlauskas, line 391: Very minor improvement: I suggest making this the 1st condition (before `evt.eventType === ChainSyncEventType.RollBackward`), because this predicate is very cheap, and once it reaches tip-k it won't have to ever run any other predicates. If you make this change, feel free to squash right away.

mkazlauskas (approved):
Great work! :rocket:

mkazlauskas (approved):

iccicci (approved):

PR #1212, opened 2024-04-15
Feat/lw 9927 review protocol params update

mirceahasegan:
# Context

Update Core types and serialization to account for the 2 new protocol parameters:
- [securityRelevantParamVotingThreshold](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/cddl-files/conway.cddl#L419)
- [minFeeRefScriptCostPerByte](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/cddl-files/conway.cddl#L411C11-L411C31)

# Proposed Solution
This will be covered in two PRs:
1. On `master` branch: update the core types and serialization.
2. On `conway-era` branch: update `DbSyncChainHistoryProvider` to get the values from db-sync

This PR is #1 from above.

Serialisation changes could not be tested as CSL does not support them yet.
I will attempt to validate via e2e tests as part of `conwa-era` e2e tests.


iccicci (dismissed):
Good enough till here!

We could make the serialization / de-serialization a bit more strict by checking for the absence of key `14` in `protocol_param_update` when it appears in the context of a `parameter_change_action` i.e. the transaction is from Conway era.

mirceahasegan:
> Good enough till here!
> 
> We could make the serialization / de-serialization a bit more strict by checking for the absence of key `14` in `protocol_param_update` when it appears in the context of a `parameter_change_action` i.e. the transaction is from Conway era.

I enforced serialisation for ConwayEra valid protocol params in [dd3a43b](https://github.com/input-output-hk/cardano-js-sdk/pull/1212/commits/dd3a43b5d217dad798a91a988d01f6a4001784e8).
Enforcement is done by enforcing protocol params update type when called from a parameter parameters action.

iccicci (approved):
Exceptional type safe implementation! 🚀

AngelCastilloB (approved):
Looks great @mirceahasegan 🚀

packages/core/src/Serialization/Update/ProtocolParamUpdate.ts
```diff
@@ -470,14 +472,10 @@ export class ProtocolParamUpdate {
       ? UnitInterval.fromFloat(Number(parametersUpdate.decentralizationParameter))
```
mirceahasegan, line 472: decentralizationParameter was also removed in conway era.
I am investigating this and a few other serialization issues on the conway-era branch.
I will update this PR once the implementation is stable.

AngelCastilloB: decentralizationParameter was actually removed in babbage era, but there shouldn't be a need to remove this, that way the serialization utils can be use to serialize/deserialize older transaction too. When this fields are deprecated their key is no longer used (in this case I believe is 12), but no other field can take its place, so there is no harm on leaving them there

mirceahasegan: In my particular case (e2e test), the raw txBuilder is passing a manually built protocol params to a conway params update proposal.
I am trying to prevent this by building the typesafe version of the `fromCore` when called via conway governance.
Otherwise, the effect is that txbuilder will build the tx for you, even if those protocol params were deprecated.
But the change will not cripple the ser/des part, it will only make it more restrictive when called from the gov path

AngelCastilloB: thats a pretty good solution 🚀

PR #1216, opened 2024-04-17
[LW-10288] - Fix unsecure comparison in order to mitigate CWE 208 vulnerability

danielmain:
# Description:

A recent security audit flagged potential CWE-208 vulnerabilities due to non-constant-time string comparisons. This PR mitigates those risks.


AngelCastilloB (approved):
LGTM @danielmain (Love the docs ❤️) (I will re-approve once the sonar warning is fixed)

VanessaPC (commented):
Looks good! Thanks for the comments. 
Could you fix the commit history so that the changes in the `util/` package are the first commit and describe what they do
`fix(util): ...description` 
and I'd probably add the changes to the other two packages in two commits, one per package.

packages/util/src/equals.ts
```diff
@@ -6,3 +6,50 @@ export const strictEquals = <T>(a: T, b: T) => a === b;
...
+
+  // Calculate the maximum length of the two strings. This value will be used to loop through each character.
+  const maxLength = Math.max(a.length, b.length);
+
+  // It loops through the maximum length of the two strings, ensuring that the iteration count doesn't leak information about the length difference.
+  const results: (0 | 1)[] = Array.from({ length: maxLength }, (_, i) => {
+    const charCodeA = i < a.length ? a.charCodeAt(i) : 0;
+    const charCodeB = i < b.length ? b.charCodeAt(i) : 0;
+    return charCodeA === charCodeB ? 1 : 0; // Return 1 if the characters are equal, otherwise return 0.
+  });
```
szymonmaslowski, line 45:
```suggestion
  const results: (0 | 1)[] = Array.from({ length: maxLength }, (_, i) => {
    return a.charCodeAt(i) === b.charCodeAt(i) ? 1 : 0; // Return 1 if the characters are equal, otherwise return 0.
  });
```

packages/util/src/equals.ts
```diff
@@ -6,3 +6,50 @@ export const strictEquals = <T>(a: T, b: T) => a === b;
...
+
+  // Reduce the results array to a single value, using bitwise AND, explicitly casting the result as binary 0 | 1.
+  // We do not use booleans here (true/false) instead 0 and 1 because bitwise operations in JavaScript operate on 32-bit integers.
+  const areAllCharactersEqual = results.reduce(
+    (accumulator: 0 | 1, currentValue: 0 | 1) => (accumulator & currentValue) as 0 | 1, // Cast to 0 | 1 to satisfy TypeScript's strict type checking.
+    1 as 0 | 1
+  );
```
szymonmaslowski, line 52: I think it is enough to cast just the result of the binary operation
```suggestion
  const areAllCharactersEqual = results.reduce(
    (accumulator, currentValue) => (accumulator & currentValue) as 0 | 1, // Cast to 0 | 1 to satisfy TypeScript's strict type checking.
    1
  );
```

szymonmaslowski (changes requested):
Great job! Leaving some suggestions :wink:

packages/util/src/equals.ts
danielmain: I did it only to improve readability :)

danielmain:
> Looks good! Thanks for the comments. There is an error in the hardware-ledger. Could you fix the commit history so that the changes in the `util/` package are the first commit and describe what they do `fix(util): ...description` and I'd probably add the changes to the other two packages in two commits, one per package.

I will rebase as soon I collect all suggestions ;) 
Thank you @VanessaPC for telling me in this repo there is no squas-merge as with lace

packages/util/src/equals.ts
```diff
@@ -6,3 +6,50 @@ export const strictEquals = <T>(a: T, b: T) => a === b;
...
+ *
+ * @param {string} a - The first string to compare.
+ * @param {string} b - The second string to compare.
+ * @returns {boolean} - Returns true if the strings are identical, false otherwise.
+ */
+export const areStringsEqualInConstantTime = (a: string, b: string): boolean => {
+  if (a.length !== b.length) return false;
+
+  // Calculate the maximum length of the two strings. This value will be used to loop through each character.
+  const maxLength = Math.max(a.length, b.length);
```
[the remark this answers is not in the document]
danielmain: I removed the unnecessary check @szymonmaslowski pls resolve 🙏🏻

packages/util/src/equals.ts
szymonmaslowski: Yep, we need to cast the binary op, but just it. Types of the `accumulator` and the `currentValue` and cast of the initial `accumulator` value are redundant.

szymonmaslowski (dismissed):

AngelCastilloB (approved):

mkazlauskas (approved):
Looks good! I think the commit type should be `refactor` instead of `chore`, because it has non-functional changes in `src/`. Please squash before merging

PR #1219, opened 2024-04-17
chore: split chain-history out

bernokl:
Goal is to split chain history into separate service to alleviate monolithic load


nix/cardano-services/deployments/backend-ingress.nix
```diff
@@ -63,6 +63,16 @@
...
+            ++ lib.optionals config.providers.chain-history-provider.enabled [
+              {
+                pathType = "Prefix";
+                path = "/v${lib.last (lib.sort lib.versionOlder values.cardano-services.versions.chainHistory)}/chain-history";
+                backend.service = {
+                  name = "${chart.name}-chain-history-provider";
+                  port.name = "http";
+                };
+              }
+            ]
```
gytis-ivaskevicius, line 75:
```suggestion
            ++ lib.optionals config.providers.chain-history-provider.enabled (
              (version: {
                pathType = "Prefix";
                path = "/v${version}/chain-history";
                backend.service = {
                  name = "${chart.name}-chain-history-provider";
                  port.name = "http";
                };
              }) values.cardano-services.versions.chainHistory
            )
```

also can you run treefmt?

nix/cardano-services/deployments/chain-history.nix
```diff
@@ -0,0 +1,58 @@
...
+    };
+
+    env = {
+      NETWORK = values.network;
+      OGMIOS_SRV_SERVICE_NAME = values.backend.ogmiosSrvServiceName;
+      LOGGER_MIN_SEVERITY = values.cardano-services.loggingLevel;
+      ENABLE_METRICS = "true";
+      DISABLE_STAKE_POOL_METRIC_APY = "true";
+      PAGINATION_PAGE_SIZE_LIMIT = "5500";
+      SERVICE_NAMES = "chain-history";
```
gytis-ivaskevicius, line 30: i assume this line is the only change compared to the rest of env variables from monolithic backend?

nix/cardano-services/deployments/default.nix
```diff
@@ -206,6 +212,19 @@ in
...
+                  inherit (oci.meta) versions;
+                 in
+                  lib.concatLists [
+                    (map (v: "/v${v}/chain-history") versions.chainHistory)
+                  ];
+                backend.service = {
+                  name = "${final.namespace}-chain-history-provider";
+                  port.name = "http";
+                };
+              }
```
gytis-ivaskevicius, line 227:
```suggestion
    
```

im pretty sure this is not necessary since we have a patch in ingress

gytis-ivaskevicius (changes requested):
Aside ingress tweaks i think this is good to be merged

nix/cardano-services/deployments/default.nix
bernokl: confirmed, after initial apply, removed this and re-apply shows no diff, removing

nix/cardano-services/deployments/chain-history.nix
bernokl: I just updated the env to match backend yes

nix/cardano-services/deployments/backend-ingress.nix
bernokl: I do not understand the comment, when I try to use the suggested "(version:" my working deploy fails, will leave it like this for now

michalrus: @bernokl I think there's a `map` missing in the suggestion from @gytis-ivaskevicius, so instead of:
```
((version: { … }) values.⟨…⟩.chainHistory)
```
it should be:
```
(map (version: { … }) values.⟨…⟩.chainHistory)
```

michalrus (dismissed):
If the last comment from @gytis-ivaskevicius is resolved ⇒ LGTM

michalrus (dismissed):

iccicci (changes requested):
Each commit (but the first one) seems to be a fixup of the previous one.

The final diff is thinner than the diff of some commits in this branch.

Please squash all of them to just one commit.

gytis-ivaskevicius (approved):

iccicci (approved):
Sorry for the delay, I missed this review request.

LGTM

gytis-ivaskevicius (commented):
Whats up with removal of bunch of providers/projectors? also I don't think that we have `chain-history-provider` merged in

iccicci (approved):
LGTM!

gytis-ivaskevicius (approved):
Looks good, tho could you run treefmt before merging this?

michalrus (commented):
LGTM 👍

PR #1233, opened 2024-04-29
Reduce logs

gytis-ivaskevicius:
# Context

As of creating this PR we have 3.29TB of logs since beginning of the month, sometimes they are getting discarded due to the Grafana rate limit

# Proposed Solution
Set logging severity to `info` instead of `debug` to optimize for cost and avoid discarding logs. This can be easily temporarily switched back if there is a need


iccicci (changes requested):
From the PR description I would have expected only the latest commit.

Could it be some more commits was pulled in this PR accidentally?

michalrus (approved):

PR #1236, opened 2024-05-01
Add pool-delist-schedule and pool-rewards queues to deployed pg-boss-worker

iccicci:
# Context

- LW-10311 - We have some code repetition in our repository.
- LW-10287 - We need to add `pool-delist-schedule` and `pool-rewards` to `pg-boss-worker`.

# Proposed Solution

- Cleaned some code repetition.
- Added required queues to deploy configuration.


Dockerfile
```diff
@@ -114,6 +114,8 @@ WORKDIR /app/packages/cardano-services
 CMD ["node", "dist/cjs/cli.js", "start-blockfrost-worker"]
 
 FROM cardano-services as pg-boss-worker
+WORKDIR /config
+COPY compose/schedules.json .
```
gytis-ivaskevicius, line 118: is this the default search location or how is the application finding this path?

Also maybe we can just change it to cli/env argument? Otherwise its annoyingly difficult to tweak it without rebuilding image

iccicci: The file must be added, somehow and somewhere...
The path can be provided to `pg-boss-worker` through `SCHEDULES` environment variable.
I guess the required changes to specify the filed on a different path from the default one should be added to this PR.

gytis-ivaskevicius:
```suggestion
WORKDIR /config
COPY compose/schedules.json .
ENV SCHEDULES=/config/schedules.json
```

iccicci: Which is the reason for this change?
We set environment variables through docker compose configuration.

packages/cardano-services/config/network/preprod/schedule.json
mchappell, line 1: Is this named correctly?

iccicci: It was an empty file with no reason to exist... I removed it

Dockerfile
iccicci: Errata corrige

> I guess the required changes to specify the file on a different path from the default one should be added to this PR.

I double checked this and there is no default: the `SCHEDULES` environment variable **must be specified** for deployed `pg-boss-worker`.

gytis-ivaskevicius: I think we should set it globally since it does not impact other services. This way it's less error-prone.
Also this is image is pg-boss-worker specific, thus I don't think that there are any issues with setting it here

iccicci:
> I assume we want to push these changes to production as soon as possible

No. In our DS we agreed a flow with several steps before promoting this in production. More than this we are tolerating the absence of ROS for months, we can tolerate it for some more day.

Better avoid any rush and follow the standard flow.

Dockerfile
iccicci: ba0838545e should solve this

gytis-ivaskevicius:
I was thinking that we are merging this to the release branch, syncing, testing it on dev environments and then off to production. What workflow does team have in mind? I believe that's the most appropriate hotfixes workflow until we switch to trunk based development

iccicci:
[This is the flow we agreed](https://input-output.atlassian.net/browse/LW-10287?focusedCommentId=282826) and this is not an hotfix, the bug in production is an end in itself: solving it could be considered as adding a new feature.

pparker:
Infra DS: let's merge this to master first, then (cherry-pick) to Release Branch

Dockerfile
```diff
@@ -114,6 +114,9 @@ WORKDIR /app/packages/cardano-services
...
```
mirceahasegan, line 118: Does this mean that the image must be rebuilt in case the `schedules.json` file changes?
Previosly it was a mapped volume, so any change would apply with a container restart.

mkazlauskas (approved):
Looks good, but @mirceahasegan made a good point about the volume. Probably a minor issue anyway as we're not likely to update it frequently in devenv.

Dockerfile
mirceahasegan: Resolving because updating `schedules.json` is not meat to be a frequest task (see comments below).

mirceahasegan (approved):
LGTM

michalrus (approved):

PR #1249, opened 2024-05-07
fix(cardano-services): stake pool fuzzy search empty result

iccicci:
# Context

The fuzzy stage pools search of `TypeormStakePoolProvider` throws an error in case of empty search result.

# Proposed Solution

Fixed the SQL in this edge case.


mchappell (approved):
Thanks for the quick response @iccicci !

mirceahasegan (approved):
LGTM!

PR #1257, opened 2024-05-08
feat(core): nft metadata parsing loose mode

mkazlauskas:
# Context

1. CIP-68 assets ([example](https://preprod.cexplorer.io/asset/asset14w53nf5ghakjcp37ruj7a8ffavjpn2cxamdvsl)) minted with NMKR are not compliant with [CIP-68](https://cips.cardano.org/cip/CIP-0068), as datum is missing the 3rd `extra` item (must be set to `#6.121([])` if not used

image

2. Some assets are minted without 'name' (only 'image')
3. [Current cip25 spec](https://cips.cardano.org/cip/CIP-0025) has an example of 'version' represented as a number (not string) and our current implementation expects a string. There are probably assets that we wouldn't be able to parse due to that. 

# Proposed Solution

- Loosen nft metadata validation by adding a 'strict' argument; loose by default
- Make version parsing more flexible, supporting both string and number

### DIFF
--- a/packages/core/src/Asset/NftMetadata/fromMetadatum.ts
+++ b/packages/core/src/Asset/NftMetadata/fromMetadatum.ts
@@ -10,6 +10,7 @@ import { isNotNil } from '@cardano-sdk/util';
 import difference from 'lodash/difference';
 
 const isString = (obj: unknown): obj is string => typeof obj === 'string';
+const VersionRegExp = /^\d+\.?\d?$/;
 
 const metadatumToString = (metadatum: Cardano.Metadatum | undefined): string | undefined => {
   if (Array.isArray(metadatum)) {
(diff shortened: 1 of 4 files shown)


packages/core/src/Asset/NftMetadata/fromMetadatum.ts
```diff
@@ -96,39 +96,57 @@ const getAssetMetadata = (policy: Cardano.MetadatumMap, assetName: Cardano.Asset
...
+};
+
+const parseVersion = (version: Cardano.Metadatum | undefined) => {
+  if (!version) return '1.0';
+  if (typeof version === 'bigint') {
+    return `${version}.0`;
+  }
+  const stringVersion = asString(version);
+  if (stringVersion) {
+    return `${Number(stringVersion)}.0`;
```
iccicci, line 119: I don't know which is the source of this data, so how can't tell how much my suggestion is valid or could be ignored.

Here may worth to validate the string (something like a RegExp to ensure it contains only digits...)

iccicci (approved):
Apart from a small doubt, LGTM

packages/core/src/Asset/NftMetadata/fromMetadatum.ts
mkazlauskas: Nice catch @iccicci ! Indeed it can be any string, added some validation 78adbbb556aa8dec8ac317d5ff3e431829830c8a

iccicci (approved):
LGTM

mchappell (approved):

PR #1259, opened 2024-05-09
feat(web-extension): add log of transaction id when signing

mkazlauskas:
# Context

It's difficult to confirm that we don't have roundtrip issues when troubleshooting

# Proposed Solution

Add log of transaction id that is being signed


AngelCastilloB (approved):
Good job @mkazlauskas 💪🏼

iccicci (approved):
LGTM

PR #1260, opened 2024-05-09
chore: register HTTP handlers with forward slashes on Windows

michalrus:
Found as a part of [LW-7619](https://input-output.atlassian.net/browse/LW-7619).

# Context

`path.join` is being used for constructing the HTTP paths when registering handlers. This is useful, because the developer doesn’t have to keep track of `../`, double slashes, no slashes, leading slashes, following slashes etc.

At the same time, Windows uses backslashes in its paths, and therefore in `path.join`.

# Proposed Solution

Let’s add a `.replaceAll`.

I tested this code on Windows and it works there.

_n/a_

[LW-7619]: https://input-output.atlassian.net/browse/LW-7619?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/cardano-services/src/Http/HttpServer.ts
+++ b/packages/cardano-services/src/Http/HttpServer.ts
@@ -175,7 +175,9 @@ export class HttpServer extends RunnableModule {
     };
 
     let handler: keyof typeof handlers;
-    for (handler in handlers) this.app.use(path.join(versionPath, handler), handlers[handler]);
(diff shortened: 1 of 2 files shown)


mkazlauskas (approved):

iccicci (approved):
LGTM

PR #1266, opened 2024-05-10
LW-10511 Increase pg-boss jobs retention to one yaer

iccicci:
# Context

pg-boss jobs have their own (well handled by pg-boss itself) retry strategy.
Recently we were warried about some edge cases:

-  what happens when the `pg-boss-worker` dies (so the final `UPDATE` is not performed)?
-  what happens when the DB server dies so `pg-boss-worker` has no chances to perform final `UPDATE`?

pg-boss handles these case as well: jobs have an expiry timeout, once expired, they are retried anyway.

Another interesting pg-boss option in this field is the job retention time (by default two weeks): a job, till not successfully completed, is retried for two weeks.

# Proposed Solution

Increased the retention time to one year, to have a chance to address problems even if spotted later than the default two weeks.

### DIFF
--- a/packages/projection-typeorm/src/pgBoss.ts
+++ b/packages/projection-typeorm/src/pgBoss.ts
@@ -48,6 +48,7 @@ export interface StakePoolRewardsJob {
 }
 
 export const defaultJobOptions: SendOptions = {
+  retentionDays: 365, // keep the job in a state that can be retried for one year
   retryDelay: 6 * 3600, // 6 hours
   retryLimit: 1_000_000 // retry forever
 };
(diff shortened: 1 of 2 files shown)


gytis-ivaskevicius (commented):
i dont think that keeping jobs for 1 year is a good idea, that will bloat database by a lot, I'm thinking something along the lines of 30-90 days tops

bernokl:
I would second the retention question, is there a use case for keeping older jobs? Otherwise LGTM

iccicci:
Hi @gytis-ivaskevicius ,

> that will bloat database by a lot

only jobs with not handled errors will be kept for one year, the ideal number of records we are speaking about is zero; in case we will be distracted and we loose some error, I think we are speaking about a dozen of records...

gytis-ivaskevicius (approved):
oh, then thats perfect. Thanks for clarifying, excellent work!!

bernokl:
> speaking about a dozen of records...
That seems reasonable to me, approved.

bernokl (approved):

mirceahasegan (approved):

michalrus (approved):

PR #1286, opened 2024-05-17
fix(tx-construction): fix an issue where the tx-builder was always using greedy input selector for multi delegation wallets

AngelCastilloB:
# Context

The transaction builder currently is using the greedy input selector when wallets are multi-delegating and performing regular transactions, more ever, the greedy input selector is being used to explicitly send all funds to the first stake address, this is due to a validation that should only apply to transaction that are changing the delegation state being apply to all transactions.

# Proposed Solution

The transaction builder now only perform the validation when the transaction is changing the delegation state and uses the greedy input selector accordingly.

Fixed an important issue in multi delegation


ginnun (dismissed):
lgtm

packages/tx-construction/test/tx-builder/TxBuilderDelegatePortfolio.test.ts
```diff
@@ -338,6 +358,20 @@ describe('TxBuilder/delegatePortfolio', () => {
...
 
+    it('does nothing and uses random improve input selector when the wallet is multi delegating and transaction is not changing delegation', async () => {
+      const tx = await multiDelegatingWalletTxBuilder.build().inspect();
+      expect(tx.body.certificates).toBeUndefined();
+      expect(GreedyInputSelector).not.toHaveBeenCalled();
+      expect(roundRobinRandomImprove).toHaveBeenCalled();
+    });
+
+    it('does nothing and uses random improve input selector when the wallet is single delegating and transaction is not changing delegation', async () => {
+      const tx = await singleDelegatingWalletTxBuilder.build().inspect();
```
mirceahasegan, line 369: Suggestion: you also include an output if the test fixtures allow it?
I see the input selection algorithm is invoked, so the test fulfils the purpose. Considering this, it is not mandatory to do this change, but it would make the test more meaningful, provided the fixtures support the change.

mirceahasegan (dismissed):
Great catch! :detective:

packages/tx-construction/test/tx-builder/TxBuilderDelegatePortfolio.test.ts
mirceahasegan: Done here: [9a3d65d](https://github.com/input-output-hk/cardano-js-sdk/pull/1286/commits/9a3d65d52f12474365cb96aaf039155d11708142)

mirceahasegan (approved):

iccicci (approved):
LGTM!

PR #1287, opened 2024-05-20
LW-10576 Set not zero value for pool and stake key deposit in local-network

iccicci:
# Context

In our `local-network` we have the dangerous setting of `0` for stake key deposit and stake pool deposit.
This makes us less confident about our tests.

# Proposed Solution

Changed the two settings as per `mainnet`.


AngelCastilloB (dismissed):
Awesome job @iccicci 🍰

packages/e2e/local-network/scripts/pools/update-node-utils.sh
```diff
@@ -129,7 +129,7 @@ updatePool() {
     --protocol-params-file ${SP_NODE_ID}/params.json | awk '{ print $1 }')
 
   initialBalance=$(getAddressBalance "$genesisAddr")
-  txOut=$((initialBalance - fee))
+  txOut=$((initialBalance - fee - 2000000))
```
mkazlauskas, line 132: Is this for keyDeposit that was previously 0?
```suggestion
  txOut=$((initialBalance - fee - keyDeposit))
```

packages/cardano-services/test/StakePool/StakePoolHttpService.test.ts
```diff
@@ -964,7 +964,7 @@ describe('StakePoolHttpService', () => {
             });
             it('pledgeMet false, status activating, and condition', async () => {
```
mkazlauskas, line 965: Not sure the original intent of the test, but that's what it's doing now
```suggestion
            it('pledgeMet false, status active, and condition', async () => {
```

mkazlauskas (changes requested):

packages/e2e/local-network/scripts/pools/update-node-utils.sh
iccicci: Yes. 001e4043df should solve this

packages/cardano-services/test/StakePool/StakePoolHttpService.test.ts
iccicci: It's hard to say which was the original intent of the test... it's inside a `identifier & status & pledgeMet filters` describe block, so I guess the intent of the test is to check that logical expression between filters work as expected.
Anyway the status of a stake pool in the fixtures is strongly dependent on the time when the snapshot was taken... probably recent changes introduced make no pools to have the right combination to match the filter in this tests so I had only two options: disabling the test or change it as I did.
In my opinion, tests about stake pool status should be done only as long running tests which prepare the pools as per their needs, so we can be sure about the time when they run.
Until that improvement is done... 94e3d9438a should solve this

mkazlauskas (approved):
Nice work! :rocket:

AngelCastilloB (approved):

PR #1291, opened 2024-05-21
Improve db selection through input parameters while connecting to deployed databases

iccicci:
# Context

The script we have to connect to deployed DBs (thank you very much @gytis-ivaskevicius) requires some explicit data as input parameter, some of them must be kept in mind (as the name of the regions) some could be determined by the script itself (as the region for `dev` environments).

# Proposed Solution

Added a new scripts which performs some checks and computations and accepts shortened input parameters.

I tried a first implementation through the `justfile` (nice tool!), but I had some problems with complex bash syntaxes as `cat << EOF`, so to make it fast, I added a new script outside of `justfile`.
We could improve this by moving the script inside the `justfile`.

### DIFF
--- a/.kube/eu-central-1
+++ b/.kube/eu-central-1
@@ -21,7 +21,7 @@ contexts:
     cluster: arn:aws:eks:eu-central-1:926093910549:cluster/lace-live-eu-central-1
     user: lace-ci
   name: lace-ci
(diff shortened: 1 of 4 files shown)


ginnun (approved):
nice improvement, lgtm.

PR #1294, opened 2024-05-22
fix: remove null characters from user-specified strings when storing …

mkazlauskas:
# Context

Projecting mainnet NFT metadata throws `invalid byte sequence for encoding "UTF8": 0x00`

NFT metadata is user input and string fields can have null characters, which makes postgres query to fail when inserting such data

# Proposed Solution

Add a new transformer to sanitize all user-specified string values when inserting NftMetadata entities

Also add a new chain sync data set that I generated for finding the cause of the issue in `util-dev` package


mirceahasegan (approved):
LGTM 🚀

iccicci (approved):
LGTM!

mchappell (approved):

PR #1295, opened 2024-05-22
LW-10615 Optimize e2e tests

iccicci:
# Context

Our `local-network` has an initial delay of about 12' in order to be operational. This has a huge development time cost as CI runs takes 32' and running e2e tests locally takes more or less the same time.

# Proposed Solution

### Changes impacting both CI and local development

- Changed the healthy logic of `local-testnet` container from _everything done_ to _the nodes started_ so other containers depending on it can perform their _init_ simultaneously to this container.
- Added an additional check to be aware when `local-testnet` _init_ is completed; with previous change, we can no longer trust the healthy logic with this purpose.
- Removed the _tricky_ `healthcheck` from `cardano-db-sync` as it is no longer needed.
- Split the e2e tests belonging to `yarn test:wallet` into two groups: those which can be run at epoch 0 from those which need to wait for epoch 3 in order to be run.
- Replaced `yarn test:wallet` with a wrapper to call `yarn test:wallet:epoch0`, `yarn test:wallet:epoch3` and relative wait script to run the tests at the right time.

### Changes dedicated to local development

- Parallelized the stake pool update procedures during `local-testnet` _init_ stage; this plus the change to healthy logic (with cached built docker images) gives the developer the ability to run locally `yarn test:wallet:epoch0` **about one single minute later than** `yarn local-network:up` **rather than after about fifteen minutes!**

### Changes dedicated to CI

- Added `--mode=skip-build` option to all `yarn install` calls to save about 20" each.
- Removed the installation of some unused `apt-get` packages. This actually has impact on local development as well, but has a very limited benefit due to local docker cache.

- Removed deprecated `version` property from `docker-compose.yml` like files.
- Moved the `healthcheck` definition of _projector_ related containers from `docker-compose.yml` to `Dockerfile` to remove some code repetition.
- Fixed a minor bug in `storePoolMetricsUpdateJob` operator: if it sent a job at block zero, it no longer sent other jobs.
- Reduced the `pool-metrics` jobs interval from `1000` blocks to `50` blocks in the `local-network`.
- This PR reduces the e2e CI run from 32' to 22': split the 28' unit tests CI flow into two flows to achieve an overall CI time reduction from 32' to 22'.


packages/e2e/docker-compose.yml
```diff
@@ -113,6 +108,7 @@ services:
   stake-pool-projector:
     environment:
       METADATA_JOB_RETRY_DELAY: 60
+      POOLS_METRICS_INTERVAL: '50'
```
mkazlauskas, line 111: For consistency with the existing number in line above
```suggestion
      POOLS_METRICS_INTERVAL: 50
```

mkazlauskas (dismissed):
Great work! :sunglasses: :mage:

packages/e2e/docker-compose.yml
iccicci: Thank you. 63c303f3e7 should fix.

AngelCastilloB (approved):
Awesome job @iccicci ⭐

mkazlauskas (approved):

mirceahasegan (approved):
Fantastic work @iccicci 🚀 
I'm so happy for this optimization 🥳

PR #1301, opened 2024-05-24
feat: add cip-1854 support to keyagent

greatertomi:
# Context

# Proposed Solution


packages/core/src/Cardano/Address/ScriptAddress.ts
```diff
@@ -0,0 +1,47 @@
+import { Address, AddressType } from './Address';
+import { HexBlob, InvalidStringError, OpaqueString, assertIsHexString } from '@cardano-sdk/util';
+import { isRewardAccount } from './PaymentAddress';
+
+export type ScriptAddress = OpaqueString<'ScriptAddress'>;
```
AngelCastilloB, line 5: I dont think we need this type, shared wallets should have PaymentAddress and RewardAddress (same as the regular wallet), this addresses can represent both pub key hash credentials and script hash credentials

packages/key-management/src/types.ts
```diff
@@ -38,10 +38,26 @@ export enum KeyRole {
   DRep = 3
 }
 
+export enum MultiSigKeyRole {
```
AngelCastilloB, line 41: We dont need this, you can use the same key role we already have defined as it can be the same

packages/key-management/src/types.ts
```diff
@@ -38,10 +38,26 @@ export enum KeyRole {
...
+  INDIVIDUAL = 1852,
+  MULTI_SIG = 1854
+}
+
 export interface AccountKeyDerivationPath {
   role: KeyRole;
   index: number;
 }
+
+export interface MultiSigAccountKeyDerivationPath {
```
AngelCastilloB, line 56: same here, we dont need a special case for this, you can use `AccountKeyDerivationPath`

packages/key-management/src/types.ts
```diff
@@ -38,10 +38,26 @@ export enum KeyRole {
...
 
+export enum MultiSigKeyRole {
+  Payment = 0,
+  Stake = 2
+}
+
+export enum KeyPurpose {
+  INDIVIDUAL = 1852,
+  MULTI_SIG = 1854
+}
```
AngelCastilloB, line 49: This is a nitpick, but following the wording of CIP-1854

> To associate multi-signature keys to a wallet, we reserve however purpose=1854' to distinguish multisig wallets from standard wallets.

Maybe we could rename these:

```
export enum KeyPurpose {
  STANDARD = 1852,
  MULTI_SIG = 1854
}
```

AngelCastilloB (changes requested):

packages/key-management/src/Bip32Account.ts
```diff
@@ -33,6 +38,11 @@ export class Bip32Account {
     return key.toRawKey();
   }
 
+  async deriveMultiSigPublicKey({ role, index }: MultiSigAccountKeyDerivationPath) {
+    const key = await this.extendedAccountPublicKey.derive([KeyPurpose.MULTI_SIG, role, index]);
```
mkazlauskas, line 42:
```suggestion
```
This is wrong: scheme in cip1854 is
image

While `this.extendedAccountPublicKey` is already at `purpose/coin_type/account_ix` - we can't use this key to derive new `purpose`. It should be done on another level, by creating a new `Bip32Account` with a different `extendedAccountPublicKey` that is derived with the 1854 `purpose`.

packages/key-management/src/KeyAgentBase.ts
```diff
@@ -57,7 +60,18 @@ export abstract class KeyAgentBase implements KeyAgent {
     return this.#account.deriveAddress(paymentKeyDerivationPath, stakeKeyDerivationIndex);
   }
 
+  async deriveSharedWalletAddress({
```
mkazlauskas, line 63:
```suggestion
```
There's no need to duplicate the functionality for key/address derivation as it's exactly the same as for regular wallets. Instead, add a `purpose` parameter to the constructor of KeyAgent.

packages/key-management/src/types.ts
mkazlauskas: Agree

mkazlauskas (changes requested):

packages/key-management/src/types.ts
greatertomi: Is there a way to enforce that a wrong KeyRole is not assigned during account key derivation? Because I see the possibility of assigning `Internal` or `DRep`

mkazlauskas: I don't think we need to prevent it on this (low) level. Just like there's nothing technically preventing you from using stake role for payment keys. Applications (Lace) is not deriving the keys directly. `deriveAddress` is called by address discovery and tx builder, which can have unit tests that assert derivation props.

greatertomi: Noted. Thank you very much.

packages/key-management/src/Bip32Account.ts
greatertomi: I looked around and could not find where the purpose of 1852 was defined for `this.extendedAccountPublicKey`. I also could not find the derivation path for the first three steps in the [keyagent](https://github.com/input-output-hk/cardano-js-sdk/blob/ed9caf09bf5f980a10733ce62b1a698ead884ef4/packages/key-management/test/Bip32Account.test.ts#L17-L25) when specifying a new `Bip32Account` or the derive function. Where can a new purpose (1854) be specified for `extendedAccountPublicKey `?

packages/core/src/Cardano/Address/ScriptAddress.ts
greatertomi: Removed on fac71bd

packages/key-management/src/types.ts
greatertomi: Fixed on fac71bd

packages/key-management/src/types.ts
greatertomi: Removed on fac71bd

packages/key-management/src/types.ts
greatertomi: renamed on fac71bd

packages/key-management/src/Bip32Account.ts
greatertomi: Fixed on fac71bd

packages/key-management/src/KeyAgentBase.ts
greatertomi: Removed on fac71bd

packages/key-management/src/types.ts
```diff
@@ -69,6 +75,11 @@ export interface AccountAddressDerivationPath {
   index: number;
 }
 
+export interface SharedWalletAddressDerivationPath {
```
AngelCastilloB, line 78: Do we still need this?

packages/key-management/src/util/key.ts
```diff
@@ -26,15 +34,17 @@ export interface DeriveAccountPrivateKeyProps {
...
   bip32Ed25519: Crypto.Bip32Ed25519;
+  purpose?: KeyPurpose;
 }
 
 export const deriveAccountPrivateKey = async ({
   rootPrivateKey,
   accountIndex,
-  bip32Ed25519
+  bip32Ed25519,
+  purpose = KeyPurpose.STANDARD
```
AngelCastilloB, line 44: We may need to also update hardware wallet key agents to take variable purpose:

https://github.com/input-output-hk/cardano-js-sdk/blob/fac71bd000a4fc4234373674e127960f3f8b500b/packages/key-management/src/util/key.ts#L58-L90

and here:

```
  static async getXpub({
    deviceConnection,
    communicationType,
    accountIndex
  }: GetLedgerXpubProps): Promise<Crypto.Bip32PublicKeyHex> {
    try {
      const recoveredDeviceConnection = await LedgerKeyAgent.checkDeviceConnection(communicationType, deviceConnection);
      const derivationPath = `${CardanoKeyConst.PURPOSE}'/${CardanoKeyConst.COIN_TYPE}'/${accountIndex}'`;
      const extendedPublicKey = await recoveredDeviceConnection.getExtendedPublicKey({
        path: str_to_path(derivationPath) // BIP32Path
      });
      const xPubHex = `${extendedPublicKey.publicKeyHex}${extendedPublicKey.chainCodeHex}`;
      return Crypto.Bip32PublicKeyHex(xPubHex);
    } catch (error: any) {
      if (error.code === 28_169) {
        throw new errors.AuthenticationError('Failed to export extended account public key', error);
      }
      throw transportTypedError(error);
    }
  }
  ```
  
  But I am not sure if hardware wallet would support a different purpose for the key path, we need to test this

**Review by AngelCastilloB — CHANGES_REQUESTED** [review]

File: packages/key-management/src/types.ts (hunk 7 above)
Comment (greatertomi) [reply packages/key-management/src/types.ts]: Leaving this was an oversight. Removed on c672ad7

File: packages/key-management/src/util/key.ts (hunk 8 above)
Comment (mkazlauskas) [reply packages/key-management/src/util/key.ts:44]: To generalize this change request: we need to review usages of every `CardanoKeyConst.PURPOSE`

image

File: packages/key-management/test/Bip32Account.test.ts [hunk 9]
```diff
@@ -121,4 +146,20 @@ describe('Bip32Account', () => {
       );
     });
   });
+
+  describe('derives CIP-1854 address', () => {
```
mkazlauskas, line 150: How is this test asserting CIP-1854? :thinking: There are no changes in Bip32Account, so we probably don't need to have any additional tests for it

packages/key-management/src/util/key.ts
mkazlauskas: Regarding HW wallets with 1854 purpose, maybe @tomislavhoracek can advise

greatertomi: I allowed for a variable purpose in derivation paths on 5d8313b

mkazlauskas: 1854 keys with HW might require a different [signing mode](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/hardware-ledger/src/LedgerKeyAgent.ts#L502). We need to test it to be sure.

mkazlauskas: Also, you seem to be just omitting `purpose` in many places ([example](https://github.com/input-output-hk/cardano-js-sdk/commit/5d8313b8232afab29e07b40890a86a08688b44e0#diff-39763719dec2cf7895e133f8f4b8f7a2ce0ed4a861a3c57d2efef4d8db919be9R107)), which means it will always use 1852 as it's the default. That's probably not what we want.

mkazlauskas: Yes I think that is a good idea

tomislavhoracek: @mkazlauskas @greatertomi Ledger `CardanoKeyConst.PURPOSE = 1854` checked and it works.
Here are results 
Screenshot 2024-05-29 at 20 29 59
IMG_4234

p.s. you can run this test to check your implementation with different "purpose"  ( command: `yarn test:hw:ledger`)
https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/wallet/test/hardware/ledger/LedgerKeyAgent.integration.test.ts#L51

mkazlauskas: Have you tested signing with 1854 keys @tomislavhoracek ?

tomislavhoracek: @mkazlauskas no, I just quickly checked is a different purpose whitelisted on Ledger by exporting xpub. 
But, I can check it, of course

tomislavhoracek: Ok, I tried. I'm able to sign tx with device but mapping fails with msg:
 HwMappingError: Hardware data mapping failure: Ledger computed a different transaction idLedgerKeyAgent.signTransaction`

greatertomi: Is that error SDK-related that can be resolved by us?

greatertomi: enforced derivation path on 82449e7

packages/wallet/src/Wallets/BaseWallet.ts
```diff
@@ -113,6 +121,7 @@ export interface BaseWalletProps {
   readonly name: string;
   readonly polling?: PollingConfig;
   readonly retryBackoffConfig?: RetryBackoffConfig;
+  readonly purpose: KeyPurpose;
```
mkazlauskas, line 124: `BaseWallet` shouldn't know about purpose, because it's not a bip32 wallet, it can also be a script wallet. `purpose` should probably only be passed to key management utils like the KeyAgent.

We might need to add optional `stakeKeyPurpose`/`paymentKeyPurpose` to [GroupedAddress](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/key-management/src/types.ts#L72).

@AngelCastilloB I think we should consider refactoring GroupedAddress to better model the addresses, maybe group some bip32 properties. WDYT?
```suggestion
```

mkazlauskas (changes requested):

packages/wallet/src/Wallets/BaseWallet.ts
greatertomi: Removed `purpose` from BaseWallet on 3487275

packages/key-management/src/types.ts
```diff
@@ -38,10 +38,17 @@ export enum KeyRole {
...
 
+export enum KeyPurpose {
+  STANDARD = 1852,
+  MULTI_SIG = 1854
+}
+
 export interface AccountKeyDerivationPath {
   role: KeyRole;
   index: number;
+  purpose: KeyPurpose;
```
mkazlauskas, line 49: This type is used to derive keys from account key that already has `purpose` derived
```suggestion
```

packages/key-management/src/InMemoryKeyAgent.ts
```diff
@@ -55,11 +57,15 @@ export class InMemoryKeyAgent extends KeyAgentBase implements KeyAgent {
...
-  async signBlob({ index, role: type }: AccountKeyDerivationPath, blob: HexBlob): Promise<SignBlobResult> {
+  async signBlob(
+    { index, role: type, purpose = KeyPurpose.STANDARD }: AccountKeyDerivationPath,
+    blob: HexBlob
+  ): Promise<SignBlobResult> {
     const rootPrivateKey = await this.#decryptRootPrivateKey();
     const accountKey = await deriveAccountPrivateKey({
       accountIndex: this.accountIndex,
       bip32Ed25519: this.bip32Ed25519,
+      purpose,
```
mkazlauskas, line 68:
```suggestion
      purpose: this.purpose,
```

packages/key-management/src/util/KeyAgentTransactionSigner.ts
```diff
@@ -32,6 +32,7 @@ export class KeyAgentTransactionSigner implements TransactionSigner {
       tx,
       {
         knownAddresses: [],
+        purpose: this.#account.purpose,
```
mkazlauskas, line 35: It doesn't seem like `Bip32Account.purpose` is used anywhere else, so we can probably undo changes in `Bip32Account.ts`
```suggestion
        purpose: this.#keyAgent.purpose,
```

mkazlauskas (changes requested):

packages/cardano-services/test/StakePool/TypeormStakePoolProvider/util.test.ts
```diff
@@ -4,7 +4,7 @@ import { validateFuzzyOptions, withTextFilter } from '../../../src/StakePool/Typ
 describe('TypeormStakePoolProvider utils', () => {
   describe('validateFuzzyOptions', () => {
     it('throws if value is not a valid JSON encoded string', () =>
-      expect(() => validateFuzzyOptions('test')).toThrow('Unexpected token e in JSON at position 1'));
+      expect(() => validateFuzzyOptions('test')).toThrow(/Unexpected token/));
```
mkazlauskas, line 7: This is an odd change for this PR, was this test flaky?

greatertomi: Yes it was. It fails locally as it receives this error message "Unexpected token 'e', \"test\" is not valid JSON". Maybe updating the expected error message to match the received message would work better than regex? wdyt?

packages/key-management/src/InMemoryKeyAgent.ts
greatertomi: Where is `this.purpose` defined? `KeyAgentBase` does not have a class variable `purpose`. Should I add a `purpose` property to [KeyAgent](https://github.com/input-output-hk/cardano-js-sdk/blob/82449e7dd25a6b291c855110d1487d1d4d3e3a85/packages/key-management/src/types.ts#L191)?

packages/key-management/src/types.ts
greatertomi: Fixed on 7c3a32f

packages/key-management/src/util/KeyAgentTransactionSigner.ts
greatertomi: Fixed on 7c3a32f

PR #1304, opened 2024-05-27
Feat/lw 10619 make local testnet slower for lace

mirceahasegan:
# Context

Have the option of running the local-network at a slower pace (slotLength=1), to use it for conway-era testing with Lace.

# Proposed Solution

For now, I cherry-picked the NETWORK_SPEED commit from master, because the merge is causing some issues.


iccicci (approved):
LGTM!

mkazlauskas (approved):
Looks good! But is 5x slower than local network slow enough? :sweat_smile:

mirceahasegan:
> Looks good! But is 5x slower than local network slow enough? 😅

It should be. It's the slotLength of mainnet.

PR #1306, opened 2024-05-28
fix(tx-construction): tx-builder now computes redeemer indices correctly everytime

AngelCastilloB:
# Context

Redeemer indices must be computed relatively to the lexicographical order of the inputs

# Proposed Solution

Sort the inputs before calculating the indices


mkazlauskas (approved):

iccicci (approved):
LGTM!

mirceahasegan (approved):

PR #1310, opened 2024-05-30
LW-10576 follow up

iccicci:
# Context

`conway-era` branch introduces several changes in the `local-network` startup scripts.

# Proposed Solution

This small PR is aimed to slightly reduce the diff between the two main branches in order to facilitate next merges reducing the chances of merge conflicts.

Changing the `--supply-delegated` option also helped to remove the ugly `sed` applied to the `genesis/shelley/genesis.json` file.

### DIFF
--- a/packages/e2e/local-network/scripts/make-babbage.sh
+++ b/packages/e2e/local-network/scripts/make-babbage.sh
@@ -154,7 +154,7 @@ cardano-cli genesis create-staked --genesis-dir "${ROOT}" \
   --testnet-magic "${NETWORK_MAGIC}" \
   --gen-pools ${NUM_SP_NODES} \
   --supply ${MAX_SUPPLY} \
-  --supply-delegated ${MAX_SUPPLY} \
+  --supply-delegated $((MAX_SUPPLY * 2)) \
   --gen-stake-delegs ${NUM_SP_NODES} \
   --gen-utxo-keys ${NUM_SP_NODES}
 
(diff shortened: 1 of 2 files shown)


mirceahasegan (approved):
LGTM!

packages/e2e/local-network/scripts/make-babbage.sh
```diff
@@ -178,10 +178,11 @@ jq --raw-output ".protocolConsts.protocolMagic = ${NETWORK_MAGIC}" "${ROOT}/gene
 
 rm "${ROOT}/genesis/byron/genesis-wrong.json"
 
-jq -M '
+jq -M "
 . + {
   activeSlotsCoeff: 0.1,
   epochLength: 1000,
+  maxLovelaceSupply: $((MAX_SUPPLY * 3)),
```
mkazlauskas, line 185: This variable is clearly not the max supply :sweat_smile:
```suggestion
  maxLovelaceSupply: $((??? * 3)),
```

mkazlauskas (approved):
Looks good, just 1 optional change request

packages/e2e/local-network/scripts/make-babbage.sh
iccicci: That's what I think every time I read that variable name 😄 

I'll merge the PR as is (otherwise it will increase the merge conflicts rather than reducing them).
Once this will be merged in `conway-era` as well the two branches will be aligned and we can change this variable name safely.
Next time I'll put my hands on these files, I'll fix this as well.
In the mean while, I'll seek for the right name for it.

PR #1313, opened 2024-05-31
[LW-10504]add timeout transaction inspectors

VanessaPC:
# Context

The `transactionSummaryInspector` and `tokenTransferInspector` do not return a timeout error currently, when no inputs are resolved.  

# Proposed Solution
We add a `timeout` prop so the consumer can set up the time in MS in which they want the timeout to return if no inputs are found. 

`transactionSummaryInspector` and `tokenTransferInspector` now require a `timeout` prop.


DominikGuzei (approved):

packages/core/src/util/tokenTransferInspector.ts
```diff
@@ -148,11 +157,37 @@ const toTokenTransferValue = async (
...
+ * Resolve inputs with a timeout so that if the inputs fail to resolve in n Miliseconds
+ * we throw an error so the UI consumer can handle situation accordingly.
+ * In this case we create a race condition in which we race to send the resolved inputs before the
+ * timeout, and timeout only if the inputs aren't there.
+ */
+export const resolveWithTimeout = <T>(promise: Promise<T>, timeout: Milliseconds): Promise<T> =>
+  Promise.race([
+    promise,
+    new Promise<T>((_, reject) =>
+      setTimeout(() => reject(new Error('Timeout: failed to resolve inputs in time')), timeout)
```
mkazlauskas, line 170: A more specifically-typed error would be nice
```suggestion
      setTimeout(() => reject(new TimeoutError('Timeout: failed to resolve inputs in time')), timeout)
```

packages/core/src/util/tokenTransferInspector.ts
```diff
@@ -148,11 +157,37 @@ const toTokenTransferValue = async (
...
+  ({ inputResolver, fromAddressAssetProvider, toAddressAssetProvider, timeout, logger }) =>
   async (tx) => {
-    const { resolvedInputs } = await resolveInputs(tx.body.inputs, inputResolver);
+    let resolvedInputs: ResolutionResult['resolvedInputs'];
+
+    try {
+      const inputResolution = await resolveWithTimeout(resolveInputs(tx.body.inputs, inputResolver), timeout);
+      resolvedInputs = inputResolution.resolvedInputs;
+    } catch (error) {
+      if (error === 'Timeout: failed to resolve inputs in time') {
```
mkazlauskas, line 184:
```suggestion
      if (error instanceof TimeoutError) {
```

packages/core/src/util/transactionSummaryInspector.ts
```diff
@@ -147,8 +151,34 @@ const toAssetInfoWithAmount = async (
  */
 export const transactionSummaryInspector: TransactionSummaryInspector =
   (args: TransactionSummaryInspectorArgs) => async (tx) => {
-    const { inputResolver, addresses, rewardAccounts, protocolParameters, assetProvider, dRepKeyHash } = args;
-    const resolvedInputs = await resolveInputs(tx.body.inputs, inputResolver);
+    const {
+      inputResolver,
```
mkazlauskas, line 155: We should memoize inputResolver in lace as both inspectors are calling it with same inputs

packages/core/test/util/tokenTransferInspector.test.ts
```diff
@@ -221,6 +223,8 @@ describe('txInspector', () => {
 
   const assetProvider = createMockAssetProvider(assetInfos);
 
+  const timeout = 6000 as Milliseconds;
```
mkazlauskas, line 226: Missing timeout test for both inspectors

packages/core/src/util/transactionSummaryInspector.ts
```diff
@@ -147,8 +151,34 @@ const toAssetInfoWithAmount = async (
...
+
+    let resolvedInputs: ResolutionResult;
+
+    try {
+      // eslint-disable-next-line sonarjs/prefer-immediate-return
+      resolvedInputs = await resolveWithTimeout(resolveInputs(tx.body.inputs, inputResolver), timeout);
+    } catch (error) {
+      if (error === 'Timeout: failed to resolve inputs in time') {
+        // Handle timeout error specifically
+        logger?.log('Error: Inputs resolution timed out');
```
mkazlauskas, line 173: Is `logger` optional? It shouldn't be
```suggestion
        logger.log('Error: Inputs resolution timed out');
```

mkazlauskas (changes requested):

packages/core/src/util/transactionSummaryInspector.ts
przemyslaw-wlodek: ✅

packages/core/test/util/tokenTransferInspector.test.ts
przemyslaw-wlodek: ✅

packages/core/src/util/tokenTransferInspector.ts
przemyslaw-wlodek: ✅

packages/core/test/util/tokenTransferInspector.test.ts
```diff
@@ -1049,5 +1073,75 @@ describe('txInspector', () => {
         ])
       );
     });
+
+    it.only('has no resolved fromAddresses on timeout', async () => {
```
mkazlauskas, line 1077:
```suggestion
    it('has no resolved fromAddresses on timeout', async () => {
```

packages/core/test/util/transactionSummaryInspector.test.ts
```diff
@@ -1017,4 +1043,81 @@ describe('Transaction Summary Inspector', () => {
       }
     });
   });
+
+  it('displays unresolved inputs on timeout', async () => {
+    const resolveDelay = 5;
```
mkazlauskas, line 1048: No a required change, just a nitpick for slightly better test readability
```suggestion
    const resolveDelay = Milliseconds(5);
```

packages/core/src/util/tokenTransferInspector.ts
```diff
@@ -150,9 +161,20 @@ const toTokenTransferValue = async (
 
 /** Inspect a transaction and return a map of addresses and their balances. */
 export const tokenTransferInspector: TokenTransferInspector =
```
mkazlauskas, line 163: I've just noticed that this PR only added timeout for input resolvers, but forgot to handle failures and timeouts of AssetProvider (e.g. https://github.com/input-output-hk/cardano-js-sdk/blob/b3f3280c9ddc0b9c0122c9c4be0eacc2201a7faf/packages/core/src/util/tokenTransferInspector.ts#L142 which can also fail/timeout) which is in scope of LW-10504

mkazlauskas (changes requested):

packages/core/test/util/transactionSummaryInspector.test.ts
przemyslaw-wlodek: Fixed by 87baffa8fd705826f0073bec5d318204a87344e5

packages/core/test/util/tokenTransferInspector.test.ts
przemyslaw-wlodek: Apologies, it should be fixed by 87baffa8fd705826f0073bec5d318204a87344e5 :)

packages/core/src/util/tokenTransferInspector.ts
przemyslaw-wlodek: Should be fixed by 6b56dce807feb76cff36b051a10c36da0d584d4c

mkazlauskas (approved):
Nice work! :rocket:

mkazlauskas (approved):

PR #1316, opened 2024-06-05
LW-10561 - Use new deposit fields from db-sync

iccicci:
# Context

Version `sancho-4-2-1` of **db-sync** introduced some new `deposit` fields.

# Proposed Solution

Used them to simplify queries performed by `DbSyncChainHistoryProvider`.


mkazlauskas (approved):
Nice work! :1st_place_medal:

Issue #1318, opened 2024-06-05
Empty List Using Indefinite Encoding


EzePze:
### Summary

When constructing an empty `PlutusList`, calling `toCbor` serialises it with indefinite encoding, which is not canonical and can cause problems with scripts, particularly the `serialise` builtin.

### Steps to reproduce the bug

```ts
console.log((new PlutusList()).toCbor())
console.log(PlutusList.fromCbor(HexBlob("80")).toCbor())
```

### Actual Result

```
9fff
9fff
```
Which is the encoding for an indefinite list

### Expected Result

```
80
80
```
Which is the encoding for a definite list

### SDK version

0.30.2

### Environment type

- [X] Node.js
- [ ] Browser
- [ ] Browser Extension
- [ ] Other

AngelCastilloB:
Thanks for the bug report @EzePze, I opened a PR with the fix here https://github.com/input-output-hk/cardano-js-sdk/pull/1319, should be fixed in the next version of the SDK

AngelCastilloB:
@EzePze we just published version 0.33.0 of the [@cardano-sdk/core](https://www.npmjs.com/package/@cardano-sdk/core) which should have this fixed, give it a try and let us know

AngelCastilloB:
It seems that the issue was not properly fixed, see https://github.com/input-output-hk/cardano-js-sdk/pull/1319#issuecomment-2151584168 . Is being addressed in this PR https://github.com/input-output-hk/cardano-js-sdk/pull/1322

AngelCastilloB:
You can try the latest release 0.33.1, that should have this fixed :)

PR #1319, opened 2024-06-05
fix(core): plutus list now encodes to canonical CBOR

AngelCastilloB:
# Context

PlutusList is not encoding following canonical CBOR (uses indefinite length array rather than the definite length form). see https://github.com/input-output-hk/cardano-js-sdk/issues/1318

# Proposed Solution

PlutusList should encode by default using definite length CBOR arrays.

### DIFF
--- a/packages/core/src/Serialization/PlutusData/PlutusList.ts
+++ b/packages/core/src/Serialization/PlutusData/PlutusList.ts
@@ -6,7 +6,7 @@ import { bytesToHex, hexToBytes } from '../../util/misc';
 /** A list of plutus data. */
 export class PlutusList {
   readonly #array = new Array<PlutusData>();
-  #useIndefiniteEncoding = true;
+  #useIndefiniteEncoding = false;
 
   /**
    * Serializes this PlutusList instance into its CBOR representation as a Uint8Array.
--- a/packages/core/test/Serialization/PlutusData.test.ts
+++ b/packages/core/test/Serialization/PlutusData.test.ts
@@ -142,7 +142,7 @@ describe('PlutusData', () => {
       data.add(Serialization.PlutusData.newInteger(4n));
       data.add(Serialization.PlutusData.newInteger(5n));
 
-      expect(data.toCbor()).toEqual('9f0102030405ff');
+      expect(data.toCbor()).toEqual('850102030405');
     });
 
     it('can encode a list of plutus list', () => {
@@ -162,7 +162,7 @@ describe('PlutusData', () => {
       outer.add(Serialization.PlutusData.newList(innerList));
       outer.add(Serialization.PlutusData.newInteger(5n));
 
-      expect(outer.toCbor()).toEqual('9f01029f0102030405ff9f0102030405ff05ff');
+      expect(outer.toCbor()).toEqual('85010285010203040585010203040505');
     });
   });
 
@@ -292,7 +292,7 @@ describe('PlutusData', () => {
 
       const data = new Serialization.ConstrPlutusData(0n, args);
 
-      expect(data.toCbor()).toEqual('d8799f0102030405ff');
+      expect(data.toCbor()).toEqual('d879850102030405');
     });
   });
 


mkazlauskas (approved):

iccicci (approved):

KtorZ:
Note: this PR is _wrong_ @AngelCastilloB @mkazlauskas @iccicci. The situation is a bit more subtle. 

Plutus encodes lists as _indefinite_ lists EXCEPT empty lists. Empty lists are encoded as _finite_ empty lists for a reason beyond human understanding.

Also, you might want to fix the serializer for maps: https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/core/src/Serialization/PlutusData/PlutusMap.ts#L10-L33. 

The actual behavior also subtle (but a different kind of subtle). Maps are serialised as finite maps if they have less than (or exactly) 23 elements. Starting at 24 elements, it switches to indefinite. 

Why 24? Because this is where the size starts being encoded on an additional byte (yet still 2 bytes, like indefinite). So the real cutoff should have been at 256 elements if it were really about saving bytes. But hey, I don't make the rules.

AngelCastilloB:
> Note: this PR is _wrong_ @AngelCastilloB @mkazlauskas @iccicci. The situation is a bit more subtle.
> 
> Plutus encodes lists as _indefinite_ lists EXCEPT empty lists. Empty lists are encoded as _finite_ empty lists for a reason beyond human understanding.
> 
> Also, you might want to fix the serializer for maps: https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/core/src/Serialization/PlutusData/PlutusMap.ts#L10-L33.
> 
> The actual behavior also subtle (but a different kind of subtle). Maps are serialised as finite maps if they have less than (or exactly) 23 elements. Starting at 24 elements, it switches to indefinite.
> 
> Why 24? Because this is where the size starts being encoded on an additional byte (yet still 2 bytes, like indefinite). So the real cutoff should have been at 256 elements if it were really about saving bytes. But hey, I don't make the rules.

Hello @KtorZ thanks for the clarification, BTW where is all this specified 😅 ? I cant find any of this in the CDDL files. I will make the changes to the serialization classes for lists and maps.

KtorZ:
> BTW where is all this specified 😅 ? 

Here: https://github.com/IntersectMBO/plutus/blob/f1f2e9fec188b3c22d02ff1d548fc18b432d25d6/plutus-core/plutus-core/src/PlutusCore/Data.hs#L145-L164

I know.. It's Haskell code. But that's the ground truth. 

Which now makes me realize: disregard my comment about _maps_, only the ledger does that. Plutus seems more reasonable here and only use definite maps. Though now it creates two different serialisation behaviors. Nice.

And for the list part, you need to dig as far as the default implementation of `encode` for lists, which is actually here:

https://hackage.haskell.org/package/serialise-0.2.6.1/docs/src/Codec.Serialise.Class.html#defaultEncodeList

AngelCastilloB:
> > BTW where is all this specified 😅 ?
> 
> Here: https://github.com/IntersectMBO/plutus/blob/f1f2e9fec188b3c22d02ff1d548fc18b432d25d6/plutus-core/plutus-core/src/PlutusCore/Data.hs#L145-L164
> 
> I know.. It's Haskell code. But that's the ground truth.
> 
> Which now makes me realize: disregard my comment about _maps_, only the ledger does that. Plutus seems more reasonable here and only use definite maps. Though now it creates two different serialisation behaviors. Nice.
> 
> And for the list part, you need to dig as far as the default implementation of `encode` for lists, which is actually here:
> 
> https://hackage.haskell.org/package/serialise-0.2.6.1/docs/src/Codec.Serialise.Class.html#defaultEncodeList

I did reference Data.hs when writing our serializers, but I didn't dig deep enough it seems:

```
defaultEncodeList :: Serialise a => [a] -> Encoding
defaultEncodeList [] = encodeListLen 0
defaultEncodeList xs = encodeListLenIndef
                    <> Prelude.foldr (\x r -> encode x <> r) encodeBreak xs
```

Its kind of random to be honest, anyways, I will make the change to the List. Thanks for the heads up!

KtorZ:
> Its kind of random to be honest

Completely. I suppose it's for saving one byte. Talk about premature optimization heh. It's really beyond me that a base library providing default serializer would not enforce a canonical encoding. And I can't also blame the Plutus team for not necessarily noticing.

PR #1320, opened 2024-06-05
fix: allow unauthorized postgres connections

gytis-ivaskevicius:
# Context

We would like to avoid force-specifying certificates in deployments since it is causing us some headaches, this change fixes our issue

This PR points to the release branch, but since the next release is coming soon I would be okay with changing it to the master

Tested on ops-preview-1
I noticed that `cardano-services` tests were failing, I think it's because I don't have Postgres running locally. I will see how CI run goes

### DIFF
--- a/packages/cardano-services/src/Program/services/postgres.ts
+++ b/packages/cardano-services/src/Program/services/postgres.ts
@@ -100,12 +100,12 @@ export const loadSecret = (path: string) => fs.readFileSync(path, 'utf8').toStri
 // according to documentation it is either a boolean or a compatible TlsOptions object.
 const mergeTlsOptions = (
   conn: connString.ConnectionOptions,
-  ssl: { ca: string } | undefined
+  ssl: { ca?: string; rejectUnauthorized?: boolean }
 ): boolean | TlsOptions | undefined =>
   typeof conn.ssl === 'object'
     ? {
         ...(conn.ssl as TlsOptions),
-        ca: ssl?.ca || (conn.ssl as TlsOptions).ca
+        ...ssl
       }
     : ssl || !!conn.ssl;
 
@@ -117,7 +117,7 @@ export const getConnectionConfig = <Suffix extends ConnectionNames>(
 ): Observable<PgConnectionConfig> => {
   const postgresConnectionString = getPostgresOption(suffix, 'postgresConnectionString', options);
   const postgresSslCaFile = getPostgresOption(suffix, 'postgresSslCaFile', options);
-  const ssl = postgresSslCaFile ? { ca: loadSecret(postgresSslCaFile) } : undefined;
+  const ssl = postgresSslCaFile ? { ca: loadSecret(postgresSslCaFile) } : { rejectUnauthorized: false };
 
   if (postgresConnectionString) {
     const conn = connString.parse(postgresConnectionString);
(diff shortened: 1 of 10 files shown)


iccicci (changes requested):
Why introducing `rejectUnauthorized`?
The purpose of setting it to `false` is to accept self signed certificates.
If we are removing at all the certificates, why applying changes to their verification strategy?

I would revert the changes in `postgres.ts` file.

In the case I'm wrong, and the changes I'm speaking about are required for some reasons, this PR should target `master` as introduces application changes, not only configuration changes.

gytis-ivaskevicius:
Without changes to `postgres.ts` application crashes with such error:
```
{"name":"provider-server","hostname":"ops-preview-1-cardanojs-backend-66fb5fb5bd-w8qlr","pid":1,"level":30,"msg":"[HttpServer] Started","time":"2024-06-05T16:09:20.519Z","v":0}
/nix/store/18121rn89iz0iihmpvki39nlb1kp22l5-cardano-sdk/libexec/incl/packages/cardano-services/dist/cjs/cli.js:38
    throw reason;
    ^

error: pg_hba.conf rejects connection for host "10.0.31.193", user "cardano_owner_user", database "cardano", no encryption
    at Parser.parseErrorMessage (/nix/store/18121rn89iz0iihmpvki39nlb1kp22l5-cardano-sdk/libexec/incl/node_modules/pg-protocol/dist/parser.js:287:98)
    at Parser.handlePacket (/nix/store/18121rn89iz0iihmpvki39nlb1kp22l5-cardano-sdk/libexec/incl/node_modules/pg-protocol/dist/parser.js:126:29)
    at Parser.parse (/nix/store/18121rn89iz0iihmpvki39nlb1kp22l5-cardano-sdk/libexec/incl/node_modules/pg-protocol/dist/parser.js:39:38)
    at Socket.<anonymous> (/nix/store/18121rn89iz0iihmpvki39nlb1kp22l5-cardano-sdk/libexec/incl/node_modules/pg-protocol/dist/index.js:11:42)
    at Socket.emit (node:events:513:28)
    at addChunk (node:internal/streams/readable:324:12)
    at readableAddChunk (node:internal/streams/readable:297:9)
    at Readable.push (node:internal/streams/readable:234:10)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23) {
  length: 178,
  severity: 'FATAL',
  code: '28000',
  detail: undefined,
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'auth.c',
  line: '477',
  routine: 'ClientAuthentication'
}
```

I believe this way traffic is still encrypted but certificate is not checked

packages/cardano-services/src/Program/services/postgres.ts
```diff
@@ -117,7 +117,7 @@ export const getConnectionConfig = <Suffix extends ConnectionNames>(
 ): Observable<PgConnectionConfig> => {
   const postgresConnectionString = getPostgresOption(suffix, 'postgresConnectionString', options);
   const postgresSslCaFile = getPostgresOption(suffix, 'postgresSslCaFile', options);
-  const ssl = postgresSslCaFile ? { ca: loadSecret(postgresSslCaFile) } : undefined;
+  const ssl = postgresSslCaFile ? { ca: loadSecret(postgresSslCaFile) } : { rejectUnauthorized: false };
```
mkazlauskas, line 120: If this option is needed, then we probably want to set it here as well?
```suggestion
  const ssl = postgresSslCaFile ? { ca: loadSecret(postgresSslCaFile, rejectUnauthorized: false) } : { rejectUnauthorized: false };
```

packages/cardano-services/src/Program/services/postgres.ts
```diff
@@ -170,7 +170,9 @@ export const getPool = async (
   logger: Logger,
   options?: PosgresProgramOptions<'DbSync'>
 ): Promise<Pool | undefined> => {
-  const ssl = options?.postgresSslCaFileDbSync ? { ca: loadSecret(options.postgresSslCaFileDbSync) } : undefined;
+  const ssl = options?.postgresSslCaFileDbSync
+    ? { ca: loadSecret(options.postgresSslCaFileDbSync) }
```
mkazlauskas, line 174:
```suggestion
    ? { ca: loadSecret(options.postgresSslCaFileDbSync), rejectUnauthorized: false }
```

PR #1321, opened 2024-06-06
Feat/lw 10575 fix roundtrip error in base wallet

AngelCastilloB:
# Context

We introduced a bug that was causing transaction submitted by dApps to be round tripped. Causing in some cases  differences in the CBOR representation, which lead to an invalid witness.

# Proposed Solution

Make sure the wallet only ever signs the original CBOR of the tx body.


packages/wallet/src/types.ts
```diff
@@ -45,6 +45,7 @@ export interface SyncStatus extends Shutdown {
 
 export type FinalizeTxProps = Omit<TxContext, 'signingContext'> & {
   tx: Cardano.TxBodyWithHash;
+  bodyCbor?: HexBlob;
```
mkazlauskas, line 48: This change seems to be in the wrong commit. Also, how breaking would it be to replace `tx` and `bodyCbor` with `Serialization.Transaction`?

packages/wallet/src/services/WalletUtil.ts
```diff
@@ -26,7 +26,7 @@ export interface WalletOutputValidatorContext {
 }
 
 export type WalletUtilContext = WalletOutputValidatorContext &
-  InputResolverContext & { chainHistoryProvider: ChainHistoryProvider };
+  InputResolverContext & { chainHistoryProvider?: ChainHistoryProvider };
```
mkazlauskas, line 29: Where is this used in Lace? Are we sure we don't need 'backend input resolver' there?

mkazlauskas (changes requested):

packages/wallet/src/services/WalletUtil.ts
AngelCastilloB: We dont need it, I added it in my plutus script PR, but made it requiered, and LAce doesnt have the chainHistoryProvider where it calls this function. Would be nice to have, but I didnt see an obvious way to get the provider since they have at those points only an observable wallet (rather than a base wallet). So I made the argument optional

packages/wallet/src/types.ts
AngelCastilloB: I can try. Lets see

mirceahasegan (dismissed):
Great work, @AngelCastilloB ! 💪

packages/wallet/src/types.ts
AngelCastilloB: This are the required changes: https://github.com/input-output-hk/cardano-js-sdk/pull/1321/commits/70246371e9f8c93b6d7f44ad1e8903cb7f6d2615

If we agree on this changes I will sort out the commits when squashing :)

mkazlauskas: Changes looks good, but I just realized it's not going to work :pensive: We also need to update the extension messaging serializer/reviver (toSerializableObj/fromSerializableObj). Ideally it would be extensible so that we can inject API-specific serialization behavior to it. How about we tag this commit in some branch and create a ticket in LW-7386?

AngelCastilloB: done, and moved this change to the right commit

mkazlauskas (approved):
Great work! :detective:

mirceahasegan (approved):

PR #1322, opened 2024-06-06
fix: plutus list now serializes using the correct encoding depending on its size

AngelCastilloB:
# Context

This PR fixes properly the CBOR encoding of Plutus Lists (see https://github.com/input-output-hk/cardano-js-sdk/pull/1319#issuecomment-2151584168 for context)

# Proposed Solution

Plutus List will be encoded as an empty array `80` when the list is empty, and as an indefinite length array when there are one or more elements.


mkazlauskas (approved):

pczeglik-iohk (approved):
Looks good to me. Thank you @AngelCastilloB

PR #1323, opened 2024-06-06
chore: fully qualified imports for esm compatibility

mkazlauskas:
# Context

ESM requires fully qualified imports, but tsc can't do that when transpiling.

Previously it was possible with --experimental-specifier-resolution=node, but this option has now been removed from node.js

image

# Proposed Solution

Add eslint rules to enforce esm-compliant source and fix existing code.

This makes it possible to use cardano-js-sdk in node.js with ESM.


mkazlauskas:
Closing in favor of #1325

PR #1325, opened 2024-06-07
LW-10722 chore: add tsc-alias to make dist/esm files imports complient with latest node.js versions

iccicci:
# Context

SDK is not using sully qualified imports which are required by latest Node.js version when using ESM,

# Proposed Solution

Added one more build step (`tsc-alias`) to change old fashion style import (still required for TypeScript development) into the new one.

# Other changes introduced

- Added a global `yarn build:cjs` to skip esm build.
- Used it in e2e CI flow.
- Removed `yarn` calling `yarn` calling `yarn` in the build process.

The combination of the last two reduced the e2e CI flow from 22' to 18'.


packages/cardano-services-client/package.json
```diff
@@ -20,11 +20,9 @@
   },
   "license": "Apache-2.0",
   "scripts": {
-    "build:esm": "tsc -p src/tsconfig.json --outDir ./dist/esm --module es2020",
-    "build:cjs": "tsc --build src",
-    "build:version": "bash -c 'if node ./scripts/createVersionSource.js --check ; then node ./scripts/createVersionSource.js --create && yarn lint:fix ; fi && cp ./package.json ./dist/package.json'",
-    "build": "yarn build:version && run-s build:cjs build:esm module-fixup",
-    "module-fixup": "cp ../../build/cjs-package.json ./dist/cjs/package.json && cp ../../build/esm-package.json ./dist/esm/package.json",
+    "build": "yarn build:cjs && tsc -p src/tsconfig.json --outDir ./dist/esm --module es2020 && tsc-alias -p src/tsconfig.json --outDir ./dist/esm && cp ../../build/esm-package.json ./dist/esm/package.json",
```
mkazlauskas, line 23: Not likely we'd need to run it separately so not a change request, but just wondering, why did you decide to merge the scripts?
```suggestion
    "build:esm": "tsc -p src/tsconfig.json --outDir ./dist/esm --module es2020 && tsc-alias -p src/tsconfig.json --outDir ./dist/esm && cp ../../build/esm-package.json ./dist/esm/package.json",
    "build": "yarn build:cjs && build:esm",
```

packages/cardano-services-client/src/HttpProvider.ts
```diff
@@ -2,8 +2,8 @@
 import { HttpProviderConfigPaths, Provider, ProviderError, ProviderFailure } from '@cardano-sdk/core';
 import { Logger } from 'ts-log';
 import { fromSerializableObject, toSerializableObject } from '@cardano-sdk/util';
+import { packageJson } from './version';
```
mkazlauskas, line 5: Can we revert this to the way it worked before? It does seem like a nice change on the surface, but I think it will be a bug because when publishing sdk packages I believe we first build, then bump version 
```suggestion
```

mkazlauskas (changes requested):

packages/cardano-services-client/src/HttpProvider.ts
iccicci: 0b9658d17b should solve this

mkazlauskas (approved):
Great work @iccicci ! A much less invasive solution than #1323 and even improves CI time! :heart:

AngelCastilloB (approved):
Great job guys!! @iccicci @mkazlauskas 🔥

PR #1330, opened 2024-06-12
Skip the flaky test suite

iccicci:
# Context

It seem the flakiness is in the test suite setup.

# Proposed Solution

Skipped the test suite


mkazlauskas (commented):
~~I don't think this will fix flakiness: I believe we already disabled some tests due to the same error. I believe it can happen to any test in this package.~~ @iccicci convinced me in PR thread :sweat_smile:

mkazlauskas (approved):

mirceahasegan (approved):

PR #1334, opened 2024-06-17
chore(util-dev): copy the correct esm-package.json

mkazlauskas:
# Context

This package has it's own variant of esm-package.json, because it has to set the 'browser' field to omit docker.js from browser builds.


AngelCastilloB (approved):
LGTM

mirceahasegan (approved):
Great catch, @mkazlauskas!

PR #1335, opened 2024-06-17
chore: whitelist all device models

tomislavhoracek:
# Context

This PR removes the device model's guard. All the possible Trezor and Ledger devices are whitelisted now. It is up to the end app to restrict usage.
[Jira](https://input-output.atlassian.net/browse/LW-10698)

- All Trezor and Ledger device models are whitelisted now

# TODO:
- [X] Remove Trezor device models guard
- [x] Remove Ledger device models guard

### DIFF
--- a/packages/hardware-ledger/src/LedgerKeyAgent.ts
+++ b/packages/hardware-ledger/src/LedgerKeyAgent.ts
@@ -98,8 +98,6 @@ const stakeCredentialCert = (cert: Certificate) =>
   cert.type === CertificateType.STAKE_REGISTRATION ||
   cert.type === CertificateType.STAKE_DEREGISTRATION ||
   cert.type === CertificateType.STAKE_DELEGATION;
-const isLedgerModelSupported = (deviceModelId: string): deviceModelId is 'nanoS' | 'nanoX' | 'nanoSP' =>
-  ['nanoS', 'nanoX', 'nanoSP'].includes(deviceModelId);
 
 const establishDeviceConnectionMethodName = 'establishDeviceConnection';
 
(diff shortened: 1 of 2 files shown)


rhyslbw:
Is it recommended practice from the manufacturer to be so defensive @tomislavhoracek ? Why not make it unrestricted and we can then list the devices that have been tested?

tomislavhoracek:
> recommended practice from the manufacturer to be so defensive @tomislavhoracek ? Why not make it unrestricted and we can then list the devices that have been tested?

@rhyslbw I'm not sure about the manufacturer but Vacuumlabs suggested and said there is a lack of Trezor Model One support. So we made it more defensive, even in Daedalus. 
But I get your point and I agree with you. Maybe we can enable it for all and exclude just Model One support. Then we will have support for the upcoming Safe 5 as well.

rhyslbw:
Does the VacuumLabs library make a similar guard on the Model One? If not, I think we can just drop it here and leave it up to  applications to implement it if desired.

tomislavhoracek:
> Does the VacuumLabs library make a similar guard on the Model One? If not, I think we can just drop it here and leave it up to applications to implement it if desired.

I'm pretty sure I saw a guard. But, let me double check that

tomislavhoracek:
@rhyslbw ok, there is no classic guard,  in the new @trezor/connect pkg, as before but I found this:
https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/utils/__tests__/deviceFeaturesUtils.test.ts#L31-L57
Seams like Cardano is not supported at all on Model One.
p.s. `T1B1` is `Model One` and `T2T1` is `Model T`

Shall we keep a guard then?

rhyslbw:
Nice. Please remove it @tomislavhoracek as it's not the SDK's concern. Do we also have similar guards in the Ledger package?

tomislavhoracek:
> ve similar guards in the Led

Yeah, we have it in SDK Ledger pkg. But we were just extending support how the devices were released. There is no concern about not supporting Cardano on one of them. I can remove a guard there as well

rhyslbw:
Great. This will simplify maintenance requirements. Thanks @tomislavhoracek!

tomislavhoracek:
> Great. This will simplify maintenance requirements. Thanks @tomislavhoracek!

Definitely. Thank you @rhyslbw 🙂

AngelCastilloB (approved):
LGTM @tomislavhoracek

rhyslbw (approved):

PR #1336, opened 2024-06-18
chore(cardano-services-client): LW-10778 allow openapiVersion override when configuring client providers

mirceahasegan:
LW-10778

openApi 3.1.0 version bump brought only two conway-era specific Redeemer purpose enum values: "vote" and "propose". Which makes it backwards compatible to 3.0.1 back-ends. But the version check does not allow querying 3.0.1 back-ends. 

This change will allow users to override the 3.1.0 version and perform requests with any custom version, in this case 3.0.1


iccicci (approved):

mchappell (approved):

mkazlauskas (approved):

iccicci (approved):
LGTM!

PR #1340, opened 2024-06-20
LW-10555 Bump versions

iccicci:
# Context

Due a [db-sync bug](https://github.com/IntersectMBO/cardano-db-sync/issues/1702), it was not possible to correctly test transactions with governance action proposals.

# Proposed Solution

- Bump dependencies versions to use `db-sync-sancho-5.0.0` which solves the bug.
- Reverted 0026e5a6ace to restore original test behavior.

- Reduced the initial delay in `local-network` from one to half epoch.


mirceahasegan (approved):
Fantastic work, @iccicci 💪

rhyslbw (approved):
Great work @iccicci 💪

PR #1345, opened 2024-06-24
LW-10739 Make db cache and cardano node config common options

iccicci:
# Context

Dba cache and cardano node configuration options are specific of the provider-server, but working LW-10739 I noticed they can be useful for other services.

# Proposed Solution

Moved the options to common options section.


mirceahasegan (approved):
LGTM 👍

mkazlauskas (approved):
LGTM, although db cache is probably not applicable for projector

PR #1346, opened 2024-06-25
feat: create Cip1854ExtendedAccountPublicKey core type

greatertomi:
# Context and Proposed Solution

Add a new `Cip1854ExtendedAccountPublicKey` type that encodes CIP 1854 extended public keys in bech32 following CIP 5 specification.

### DIFF
--- a/packages/core/src/Cardano/types/Cip1854ExtendedAccountPublicKey.ts
+++ b/packages/core/src/Cardano/types/Cip1854ExtendedAccountPublicKey.ts
@@ -1,0 +1,31 @@
+import * as BaseEncoding from '@scure/base';
+import { Bip32PublicKeyHex } from '@cardano-sdk/crypto';
+import { InvalidStringError, OpaqueString, assertIsBech32WithPrefix } from '@cardano-sdk/util';
+
+const MAX_BECH32_LENGTH_LIMIT = 1023;
+const bip32PublicKeyPrefix = 'acct_shared_xvk';
+
+/** This key is a bech32 encoded string with the prefix `acct_shared_xvk`. */
+export type Cip1854ExtendedAccountPublicKey = OpaqueString<'Cip1854PublicKey'>;
+
+export const Cip1854ExtendedAccountPublicKey = (value: string): Cip1854ExtendedAccountPublicKey => {
+  try {
+    assertIsBech32WithPrefix(value, [bip32PublicKeyPrefix]);
+  } catch {
+    throw new InvalidStringError(value, 'Expected key to be a bech32 encoded string');
+  }
+
+  return value as Cip1854ExtendedAccountPublicKey;
+};
+
+Cip1854ExtendedAccountPublicKey.fromBip32PublicKeyHex = (value: Bip32PublicKeyHex): Cip1854ExtendedAccountPublicKey => {
+  const words = BaseEncoding.bech32.toWords(Buffer.from(value, 'hex'));
+  return Cip1854ExtendedAccountPublicKey(
+    BaseEncoding.bech32.encode(bip32PublicKeyPrefix, words, MAX_BECH32_LENGTH_LIMIT)
+  );
+};
+
+Cip1854ExtendedAccountPublicKey.toBip32PublicKeyHex = (value: Cip1854ExtendedAccountPublicKey): Bip32PublicKeyHex => {
+  const { words } = BaseEncoding.bech32.decode(value, MAX_BECH32_LENGTH_LIMIT);
+  return Bip32PublicKeyHex(Buffer.from(BaseEncoding.bech32.fromWords(words)).toString('hex'));
+};
--- a/packages/core/test/Cardano/types/Cip1854ExtendedAccountPublicKey.test.ts
+++ b/packages/core/test/Cardano/types/Cip1854ExtendedAccountPublicKey.test.ts
@@ -1,0 +1,39 @@
+import { Bip32PublicKeyHex } from '@cardano-sdk/crypto';
+import { Cip1854ExtendedAccountPublicKey } from '../../../src/Cardano';
+
+const bip32PublicKeyPrefix = 'acct_shared_xvk';
+
+describe('Cardano/types/Cip1854ExtendedAccountPublicKey', () => {
+  const publicKeyHex = Bip32PublicKeyHex(
+    '979693650bb44f26010e9f7b3b550b0602c748d1d00981747bac5c34cf5b945fe01a39317b9b701e58ee16b5ed16aa4444704b98cc997bdd6c5a9502a8b7d70d'
+  );
+
+  describe('Cip1854ExtendedAccountPublicKey', () => {
+    it('Accepts a valid bech32-encoded CIP1854 public key', () => {
+      const bip32PublicKey =
+        'acct_shared_xvk1q395kywke7mufrysg33nsm6ggjxswu4g8q8ag7ks9kdyaczchtemd5d2armrfstfa32lamhxfl3sskgcmxm4zdhtvut362796ez4ecqx6vnht';
+      expect(Cip1854ExtendedAccountPublicKey(bip32PublicKey)).toEqual(bip32PublicKey);
+    });
+
+    it('Throws an error when an invalid bech32-encoded CIP1854 public key is passed', () => {
+      expect(() => Cip1854ExtendedAccountPublicKey(publicKeyHex)).toThrow();
+      expect(() =>
+        Cip1854ExtendedAccountPublicKey(
+          'addr_test1qpfhhfy2qgls50r9u4yh0l7z67xpg0a5rrhkmvzcuqrd0znuzcjqw982pcftgx53fu5527z2cj2tkx2h8ux2vxsg475q9gw0lz'
+        )
+      ).toThrow();
+      expect(() => Cip1854ExtendedAccountPublicKey('invalid')).toThrow();
+    });
+  });
+
+  it('fromBip32PublicKeyHex encodes a valid CIP1854 bip32 public key according to CIP5 specification', () => {
+    const cip1854PublicKey = Cip1854ExtendedAccountPublicKey.fromBip32PublicKeyHex(publicKeyHex);
+    expect(cip1854PublicKey.startsWith(bip32PublicKeyPrefix)).toBe(true);
+  });
+
+  it('toBip32PublicKeyHex decodes a bech32-encoded CIP1854 bip32 public key to a hex string', () => {
+    const cip1854PublicKey = Cip1854ExtendedAccountPublicKey.fromBip32PublicKeyHex(publicKeyHex);
+    const originalHex = Cip1854ExtendedAccountPublicKey.toBip32PublicKeyHex(cip1854PublicKey);
+    expect(originalHex).toEqual(publicKeyHex);
+  });
+});
--- a/packages/core/src/Cardano/types/index.ts
+++ b/packages/core/src/Cardano/types/index.ts
@@ -13,3 +13,4 @@ export * from './ProtocolParameters';
 export * from './PlutusData';
 export * from './UtilityTypes';
 export * from './Governance';
+export * from './Cip1854ExtendedAccountPublicKey';


packages/core/src/Cardano/types/Cip1854PublicKey.ts
```diff
@@ -0,0 +1,55 @@
+import * as BaseEncoding from '@scure/base';
+import { InvalidStringError, OpaqueString, assertIsBech32WithPrefix, assertIsHexString } from '@cardano-sdk/util';
+
+const MAX_BECH32_LENGTH_LIMIT = 1023;
+export type Cip1854PublicKey = OpaqueString<'Cip1854PublicKey'>;
+
+export const Cip1854PublicKey = (value: string): Cip1854PublicKey => {
+  try {
+    assertIsBech32WithPrefix(value, ['addr_shared_vk', 'stake_shared_vk']);
```
mkazlauskas, line 9: Are we not going to share the bip32 key instead of payment and stake credential (extended account public key, or xpub)? I believe we discussed this on Slack a while back and I thought we agreed on this, as it's more extensible (could potentially implement multi-delegation in the future). cc @danielmain

packages/core/src/Cardano/types/Cip1854PublicKey.ts
```diff
@@ -0,0 +1,55 @@
+import * as BaseEncoding from '@scure/base';
+import { InvalidStringError, OpaqueString, assertIsBech32WithPrefix, assertIsHexString } from '@cardano-sdk/util';
+
+const MAX_BECH32_LENGTH_LIMIT = 1023;
+export type Cip1854PublicKey = OpaqueString<'Cip1854PublicKey'>;
+
+export const Cip1854PublicKey = (value: string): Cip1854PublicKey => {
```
mkazlauskas, line 7: It's using prefixes similar to Cardano addresses (starting with `stake_` and `addr_`). Addresses are not keys, they are have a credential (e.g. key **hash**) and also network id.

Also, what is the intended use case for this util? If it's for sharing the keys, then see

```suggestion
export const Cip1854Address = (value: string): Cip1854PublicKey => {
```

mkazlauskas (changes requested):

packages/core/src/Cardano/types/Cip1854PublicKey.ts
AngelCastilloB: This is a bit confusing on the CIP, but the CIP actually specifies these as keys not key hashes (plus network ids) (notice the _vk at the end).

packages/core/src/Cardano/types/Cip1854PublicKey.ts
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @param value A Cip1854PublicKey value to convert.
+ * @returns A hex string.
+ */
+const fromBech32 = (value: Cip1854PublicKey): string => {
+  const { words } = BaseEncoding.bech32.decode(value, MAX_BECH32_LENGTH_LIMIT);
+  return Buffer.from(BaseEncoding.bech32.fromWords(words)).toString('hex');
+};
+
+Cip1854PublicKey.Stake = {
```
AngelCastilloB, line 47: I think its best to have two separate types, one for the payment key and one for the staking key:

Cip1854PaymentPublicKey and Cip1854StakePublicKey

packages/core/src/Cardano/types/Cip1854PublicKey.ts
AngelCastilloB: > I don't understand. What do you mean by the Bip32 key?

If we decide to share the bip32 keys instead of the actual keys then we dont need this utils, as those have already their own defined bech32 encoding:

`xpub1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq9h0rf4`

mkazlauskas: It's an account-level key that is used to derive payment and stake keys

packages/core/src/Cardano/types/Cip1854PublicKey.ts
mkazlauskas: Thanks, resolving this :)

packages/core/src/Cardano/types/Cip1854PublicKey.ts
```diff
@@ -0,0 +1,32 @@
+import * as BaseEncoding from '@scure/base';
+import { InvalidStringError, OpaqueString, assertIsBech32WithPrefix, assertIsHexString } from '@cardano-sdk/util';
+
+const MAX_BECH32_LENGTH_LIMIT = 1023;
+export type Cip1854PublicKey = OpaqueString<'Cip1854PublicKey'>;
```
AngelCastilloB, line 5: I think we need to rename this to:

```suggestion
export type Cip1854ExtendedAccountPublicKey = OpaqueString<'Cip1854ExtendedAccountPublicKey'>;
```
since we are using the prefix `acct_shared_xvk`, Its a bit verbose, but this is different than a PublicKey.

or maybe `Cip1854ExtAcctPublicKey` or `Cip1854Bip32AcctPublicKey`, what do you think @mkazlauskas ?

We should also assert the key is 64 bytes or 128 characters

AngelCastilloB (changes requested):

packages/core/src/Cardano/types/Cip1854PublicKey.ts
mkazlauskas: Agree, the regular xpub type is called `Bip32PublicKeyHex`, so this one should be `Cip1854Bip32AcctPublicKeyHex`

packages/core/test/Cardano/types/Cip1854PublicKey.test.ts
```diff
@@ -0,0 +1,32 @@
+import { Cip1854PublicKey } from '../../../src/Cardano';
+
+const paymentKeyPrefix = 'acct_shared_xvk';
+
+describe('Cip1854PublicKey', () => {
+  const publicKeyHex =
+    '979693650bb44f26010e9f7b3b550b0602c748d1d00981747bac5c34cf5b945fe01a39317b9b701e58ee16b5ed16aa4444704b98cc997bdd6c5a9502a8b7d70d';
+
+  describe('toBech32', () => {
+    it('Encodes a valid CIP1854 payment public key according to CIP5 specification', () => {
```
mkazlauskas, line 10:
```suggestion
    it('Encodes a valid CIP1854 bip32 public key according to CIP5 specification', () => {
```

packages/core/test/Cardano/types/Cip1854PublicKey.test.ts
```diff
@@ -0,0 +1,32 @@
+import { Cip1854PublicKey } from '../../../src/Cardano';
+
+const paymentKeyPrefix = 'acct_shared_xvk';
```
mkazlauskas, line 3:
```suggestion
const bip32KeyPrefix = 'acct_shared_xvk';
```

packages/core/test/Cardano/types/Cip1854PublicKey.test.ts
```diff
@@ -0,0 +1,32 @@
...
+      expect(cip1854PublicKey.startsWith(paymentKeyPrefix)).toBe(true);
+    });
+
+    it('Throws an error when an invalid hex string is passed', async () => {
+      expect(() => Cip1854PublicKey.toBech32('invalid')).toThrow();
+    });
+  });
+
+  describe('fromBech32', () => {
+    it('Decodes a bech32-encoded CIP1854 payment public key to a hex string', () => {
```
mkazlauskas, line 21:
```suggestion
    it('Decodes a bech32-encoded CIP1854 bip32 public key to a hex string', () => {
```

packages/core/src/Cardano/types/Cip1854PublicKey.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const Cip1854PublicKey = (value: string): Cip1854PublicKey => {
```
mkazlauskas, line 7: Missing tests for this

packages/core/src/Cardano/types/Cip1854PublicKey.ts
```diff
@@ -0,0 +1,32 @@
...
+  try {
+    assertIsBech32WithPrefix(value, ['acct_shared_xvk']);
+  } catch {
+    throw new InvalidStringError(value, 'Expected key to be a bech32 encoded string');
+  }
+
+  return value as Cip1854PublicKey;
+};
+
+Cip1854PublicKey.toBech32 = (value: string): Cip1854PublicKey => {
```
mkazlauskas, line 17: Don't need to validate length then as @AngelCastilloB requested, since `Bip32PublicKeyHex` type already does it
```suggestion
Cip1854PublicKey.fromBip32PublicKeyHex = (value: Bip32PublicKeyHex): Cip1854PublicKey => {
```

packages/core/src/Cardano/types/Cip1854PublicKey.ts
```diff
@@ -0,0 +1,32 @@
...
+  } catch {
+    throw new InvalidStringError(value, 'Expected key to be a hex-encoded bytes');
+  }
+
+  const prefix = 'acct_shared_xvk';
+  const words = BaseEncoding.bech32.toWords(Buffer.from(value, 'hex'));
+  return Cip1854PublicKey(BaseEncoding.bech32.encode(prefix, words, MAX_BECH32_LENGTH_LIMIT));
+};
+
+Cip1854PublicKey.fromBech32 = (value: Cip1854PublicKey): string => {
```
mkazlauskas, line 29:
```suggestion
Cip1854PublicKey.toBip32PublicKeyHex = (value: Cip1854PublicKey): Bip32PublicKeyHex => {
```

mkazlauskas (changes requested):

packages/core/src/Cardano/types/Cip1854PublicKey.ts
greatertomi: I'll go with `Cip1854ExtendedAccountPublicKey` since it's more readable than `Cip1854Bip32AcctPublicKeyBech32`

greatertomi: Fixed on c66a9c0

packages/core/test/Cardano/types/Cip1854PublicKey.test.ts
greatertomi: Fixed on c66a9c0

packages/core/test/Cardano/types/Cip1854PublicKey.test.ts
greatertomi: Fixed on c66a9c0

packages/core/test/Cardano/types/Cip1854PublicKey.test.ts
greatertomi: Fixed on c66a9c0

packages/core/src/Cardano/types/Cip1854PublicKey.ts
greatertomi: Fixed on c66a9c0

packages/core/src/Cardano/types/Cip1854PublicKey.ts
greatertomi: Fixed on c66a9c0

packages/core/src/Cardano/types/Cip1854ExtendedAccountPublicKey.ts
```diff
@@ -0,0 +1,36 @@
+import * as BaseEncoding from '@scure/base';
+import { Bip32PublicKeyHex } from '@cardano-sdk/crypto';
+import { InvalidStringError, OpaqueString, assertIsBech32WithPrefix, assertIsHexString } from '@cardano-sdk/util';
+
+const MAX_BECH32_LENGTH_LIMIT = 1023;
+const bip32PublicKeyPrefix = 'acct_shared_xvk';
+
+export type Cip1854ExtendedAccountPublicKey = OpaqueString<'Cip1854PublicKey'>;
```
mkazlauskas, line 8: Since it's not in the name, please add a doc comment about what encoding this type uses
```suggestion
/**
 * bech32-encoded... 
 */
export type Cip1854ExtendedAccountPublicKey = OpaqueString<'Cip1854PublicKey'>;
```

packages/core/test/Cardano/types/Cip1854ExtendedAccountPublicKey.test.ts
```diff
@@ -0,0 +1,61 @@
...
+describe('Cardano/types/Cip1854ExtendedAccountPublicKey', () => {
+  const publicKeyHex = Bip32PublicKeyHex(
+    '979693650bb44f26010e9f7b3b550b0602c748d1d00981747bac5c34cf5b945fe01a39317b9b701e58ee16b5ed16aa4444704b98cc997bdd6c5a9502a8b7d70d'
+  );
+
+  describe('Cip1854ExtendedAccountPublicKey', () => {
+    it('Accepts a valid bech32-encoded CIP1854 public key', () => {
+      const bip32PublicKey =
+        'acct_shared_xvk1q395kywke7mufrysg33nsm6ggjxswu4g8q8ag7ks9kdyaczchtemd5d2armrfstfa32lamhxfl3sskgcmxm4zdhtvut362796ez4ecqx6vnht';
+      expect(() => Cip1854ExtendedAccountPublicKey(bip32PublicKey)).not.toThrow();
```
mkazlauskas, line 15: This would be a better assertion
```suggestion
      expect(Cip1854ExtendedAccountPublicKey(bip32PublicKey)).toEqual(bip32PublicKey)
```

packages/core/test/Cardano/types/Cip1854ExtendedAccountPublicKey.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    it('Encodes a valid CIP1854 bip32 public key according to CIP5 specification', () => {
+      const cip1854PublicKey = Cip1854ExtendedAccountPublicKey.fromBip32PublicKeyHex(publicKeyHex);
+      expect(cip1854PublicKey.startsWith(bip32PublicKeyPrefix)).toBe(true);
+    });
+
+    it('Throws an error when an invalid hex string is passed', async () => {
+      expect(() => Cip1854ExtendedAccountPublicKey.fromBip32PublicKeyHex(Bip32PublicKeyHex('invalid'))).toThrow();
+      expect(() =>
+        Cip1854ExtendedAccountPublicKey.fromBip32PublicKeyHex(
+          Bip32PublicKeyHex('701e58ee16b5ed16aa4444704b98cc997bdd6c5a9502a8b7d70d')
```
mkazlauskas, line 39: `Bip32PublicKeyHex` is what throws here, not `Cip1854ExtendedAccountPublicKey.fromBip32PublicKeyHex`, so it's not testing the appropriate subject. I suggest removing hex length validation from `fromBip32PublicKeyHex` because `Bip32PublicKeyHex` is already validated

packages/core/test/Cardano/types/Cip1854ExtendedAccountPublicKey.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  describe('toBip32PublicKeyHex', () => {
+    it('Decodes a bech32-encoded CIP1854 bip32 public key to a hex string', () => {
+      const cip1854PublicKey = Cip1854ExtendedAccountPublicKey.fromBip32PublicKeyHex(publicKeyHex);
+      const originalHex = Cip1854ExtendedAccountPublicKey.toBip32PublicKeyHex(cip1854PublicKey);
+      expect(originalHex).toEqual(publicKeyHex);
+    });
+
+    it('Throws an error when an invalid bech32 string is passed', () => {
+      expect(() =>
+        Cip1854ExtendedAccountPublicKey.toBip32PublicKeyHex(Cip1854ExtendedAccountPublicKey('invalid'))
```
mkazlauskas, line 54: `Cip1854ExtendedAccountPublicKey` is what should be throwing here, not `Cip1854ExtendedAccountPublicKey.toBip32PublicKeyHex`. I suggest removing this test

mkazlauskas (changes requested):

packages/core/src/Cardano/types/Cip1854ExtendedAccountPublicKey.ts
greatertomi: Fixed on 7b583d8

packages/core/test/Cardano/types/Cip1854ExtendedAccountPublicKey.test.ts
greatertomi: Fixed on 7b583d8

packages/core/test/Cardano/types/Cip1854ExtendedAccountPublicKey.test.ts
greatertomi: Fixed on 7b583d8

packages/core/test/Cardano/types/Cip1854ExtendedAccountPublicKey.test.ts
greatertomi: Fixed on 7b583d8

packages/core/test/Cardano/types/Cip1854ExtendedAccountPublicKey.test.ts
mkazlauskas: It's still validating the length [here](https://github.com/input-output-hk/cardano-js-sdk/pull/1346/files#diff-f097170118dd093d5923cad70352a6070b0adc8ad213e5c0c06bbbc02b8badceR22-R26)

mkazlauskas: Also still testing `Bip32PublicKeyHex` [here](https://github.com/input-output-hk/cardano-js-sdk/pull/1346/files#diff-82112a97b58a6472f92ac1d7c91f6d618b21f04e545876d13c01e9e58ab918c7R36)

packages/core/test/Cardano/types/Cip1854ExtendedAccountPublicKey.test.ts
mkazlauskas: Still testing `Cip1854ExtendedAccountPublicKey('addr_test1qvw383738')` [here](https://github.com/input-output-hk/cardano-js-sdk/pull/1346/files#diff-82112a97b58a6472f92ac1d7c91f6d618b21f04e545876d13c01e9e58ab918c7R49), but describe block says it should be testing `describe('toBip32PublicKeyHex'`

packages/core/test/Cardano/types/Cip1854ExtendedAccountPublicKey.test.ts
greatertomi: Should I remove [this](https://github.com/input-output-hk/cardano-js-sdk/blob/7b583d8fa510972a8731f68fba1c7245e0be9b27/packages/core/src/Cardano/types/Cip1854ExtendedAccountPublicKey.ts#L22-L26) entire assertion block since it's already validated by `Bip32PublicKeyHex`?

packages/core/test/Cardano/types/Cip1854ExtendedAccountPublicKey.test.ts
greatertomi: Removed on db446c5

packages/core/test/Cardano/types/Cip1854ExtendedAccountPublicKey.test.ts
```diff
@@ -0,0 +1,39 @@
...
+      expect(() =>
+        Cip1854ExtendedAccountPublicKey(
+          'addr_test1qpfhhfy2qgls50r9u4yh0l7z67xpg0a5rrhkmvzcuqrd0znuzcjqw982pcftgx53fu5527z2cj2tkx2h8ux2vxsg475q9gw0lz'
+        )
+      ).toThrow();
+      expect(() => Cip1854ExtendedAccountPublicKey('invalid')).toThrow();
+    });
+  });
+
+  it('Encodes a valid CIP1854 bip32 public key according to CIP5 specification', () => {
```
mkazlauskas, line 29: Not clear what 'it' is here :sweat_smile: Please either add it to 'it' or wrap in 'describe'
```suggestion
  it('fromBip32PublicKeyHex encodes a valid CIP1854 bip32 public key according to CIP5 specification', () => {
```

packages/core/test/Cardano/types/Cip1854ExtendedAccountPublicKey.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    const cip1854PublicKey = Cip1854ExtendedAccountPublicKey.fromBip32PublicKeyHex(publicKeyHex);
+    expect(cip1854PublicKey.startsWith(bip32PublicKeyPrefix)).toBe(true);
+  });
+
+  it('Decodes a bech32-encoded CIP1854 bip32 public key to a hex string', () => {
```
mkazlauskas, line 34:
```suggestion
  it('toBip32PublicKeyHex decodes a bech32-encoded CIP1854 bip32 public key to a hex string', () => {
```

mkazlauskas (changes requested):

packages/core/test/Cardano/types/Cip1854ExtendedAccountPublicKey.test.ts
greatertomi: Fixed on 5fdbd2f

packages/core/test/Cardano/types/Cip1854ExtendedAccountPublicKey.test.ts
greatertomi: Fixed on 5fdbd2f

mkazlauskas (approved):
Looks good!

AngelCastilloB (approved):
Great work @greatertomi 💪🏼

PR #1352, opened 2024-06-28
LW-10739 Add web socket based network info provider

iccicci:
# Context

Due to HTTP architecture of our API, wallets are forced to perform heavy polling operations to stay synced with the blockchain.

# Proposed Solution

Implemented a **WebSocket based API PoC** to serve the _tip_ with an events based model API **completely compliant** with `NetworkInfoProvider`; i.e. **it can be used almost for free** by the Lace full stack!

- **Introduced the hot reload feature** even if ATM only for the newly added `ws-server` which allows us to check the effects of our changes in the running `local-network` **as soon as we save changed source files**!
- Since [`tsx`](https://github.com/privatenumber/tsx?tab=readme-ov-file#-) (the core of the hot reload feature) doesn't support [`emitDecoratorMetadata`](https://www.typescriptlang.org/tsconfig#emitDecoratorMetadata), fields type must be explicit in `TypeORM` entities definition.
- Used `CardanoWsClient.networkInfo` wherever `DbSyncNetworkInfoProvider` and `networkInfoHttpProvider` were used. 

# Still missing

- Tests.

# Next steps

## Required to be ready for production

- Add tests. Estimated development cost: 2 days.

## After this is ready for production

- Using `CardanoWsClient.networkInfoProvider` wherever `networkInfoHttpProvider` is used by Lace extension. Estimated development cost: a few hours for each provider substitution.
- Remove from SDK and Lace all the code we are using to transform the `async NetworkInfoProvider`'s methods in `Observable`s to use the `Observable`s natively exposed by `CardanoWsClient`. I can't do a valid development cost estimation as I don't know all the use cases we have.
  - This is the action which actually will let us to get all the efficiency benefits from this implementation.

## Unrelated (but strongly required) improvements

- Extends the hot reload feature to all other containers in the `local-network`. Estimated development cost: 2 days.
- Introducing the ability to run the tests without the need to run the build in order to.
  - I spent about 1 day trying to achieve this priceless improvement, but I wasn't able to. 😞
  - We could try asking support to the entire XSY tribe, next to whole IOG, next (if none of these are enough) also opening a _some K USD_ public bounty may worth the benefit we would get from this.
  - This is the last step to achieve the full hot reload feature in the SDK.


packages/cardano-services/src/Program/options/common.ts
```diff
@@ -92,8 +95,19 @@ export const withCommonOptions = (command: Command, apiUrl: URL) => {
       CommonOptionsDescriptions.EnableMetrics,
       'ENABLE_METRICS',
       (enableMetrics) =>
-        stringOptionToBoolean(enableMetrics, Programs.ProviderServer, CommonOptionsDescriptions.EnableMetrics),
-      ENABLE_METRICS_DEFAULT
+        stringOptionToBoolean(enableMetrics, Programs.ProviderServer, CommonOptionsDescriptions.EnableMetrics)
+    ),
+    newOption(
+      '--heartbeat-interval <heartbeatInterval>',
+      CommonOptionsDescriptions.HeartbeatTimeout,
```
mirceahasegan, line 102: small typo:
```diff
-      CommonOptionsDescriptions.HeartbeatTimeout,
+      CommonOptionsDescriptions.HeartbeatInterval,
```

.github/workflows/continuous-integration-e2e.yaml
```diff
@@ -15,8 +15,8 @@ env:
   HANDLE_PROVIDER_PARAMS: '{"baseUrl":"http://localhost:4011/"}'
   STAKE_POOL_CONNECTION_STRING: 'postgresql://postgres:doNoUseThisSecret!@localhost:5435/stake_pool'
   STAKE_POOL_TEST_CONNECTION_STRING: 'postgresql://postgres:doNoUseThisSecret!@localhost:5435/stake_pool_test'
-  NETWORK_INFO_PROVIDER: 'http'
-  NETWORK_INFO_PROVIDER_PARAMS: '{"baseUrl":"http://localhost:4000/"}'
+  NETWORK_INFO_PROVIDER: 'ws'
```
ginnun, line 18: Since this is a PoC, is it OK to change defaults to use it?

packages/cardano-services/src/Program/programs/providerServer.ts
```diff
@@ -137,17 +145,13 @@ const serviceMapFactory = (options: ServiceMapFactoryOptions) => {
     return new TypeormStakePoolProvider(args, { cache: getDbCache(), connectionConfig$, entities, logger });
   });
 
-  let networkInfoProvider: DbSyncNetworkInfoProvider | undefined;
-
   const getNetworkInfoProvider = (cardanoNode: CardanoNode, dbPools: DbPools) => {
-    if (networkInfoProvider) {
-      return networkInfoProvider;
-    }
+    if (args.useWebSocketApi) return getWebSocketClient().networkInfo;
```
ginnun, line 149: Maybe we can get more granularity here.  There might be cases where we have five websocket-enabled providers, but we only want to enable two.

I believe enabling the WebSocket server and using a WebSocket-enabled NetworkInfoProvider can be two different things. Something like:

- `enableWebSocketServer=true`
- `useWebSocketNetworkInfoProvider=true` 
- and provide urls

And, I think using good-old if-else syntax is readable in this kind of program flow. (I know not popular nowadays)

packages/cardano-services/src/WsServer/db.ts
```diff
@@ -0,0 +1,68 @@
...
+export const initDB = async (db: Pool, logger: Logger) => {
+  logger.info('Initializing the DB...');
+
+  // Wait for a block .i.e the schema is ready
+  let blocks: number;
+  do {
+    logger.debug('Checking for latest block...');
+    const { rows } = await db.query<{ blocks: number }>('SELECT COUNT(*) AS blocks FROM block');
+    blocks = rows[0].blocks;
+  } while (blocks === 0);
```
ginnun, line 37: For my own understanding, why do we want to have blocks before handling the triggers?

packages/cardano-services/src/WsServer/server.ts
```diff
@@ -0,0 +1,278 @@
...
+    });
+
+    this.wss.on('error', (error) => {
+      this.logger.error(error, 'Async error from WebSocket server');
+      this.close();
+    });
+
+    // Init the server
+    this.init(cfg.port).catch((error) => {
+      this.logger.error(error, 'Error in init sequence');
```
ginnun, line 63: It is not essential for a PoC, but we can fail all the services for a regular implementation.

packages/cardano-services/src/WsServer/server.ts
```diff
@@ -0,0 +1,278 @@
...
+
+        // eslint-disable-next-line unicorn/consistent-destructuring
+        client.on('notification', this.createOnNotification());
+
+        // Issue the LISTEN command to get the notification event
+        client.query('LISTEN sdk_tip', (e) => {
+          // If there was no errors in the statement, set the flag for the health check
+          if (!e) return (this.notifyConnected = true);
+
+          logger.error(e, 'Error while listening on sdk_tip NOTIFY');
```
ginnun, line 180: This is a fatal error, and we might want to fail the service altogether. (not essential for the PoC)

packages/projection-typeorm/src/entity/PoolMetadata.entity.ts
```diff
@@ -8,17 +8,17 @@ import { StakePoolEntity } from './StakePool.entity';
 export class PoolMetadataEntity {
   @PrimaryGeneratedColumn()
   id?: number;
-  @Column()
+  @Column('varchar')
```
ginnun, line 11: I am asking to learn; How are these and the other Column improvements related to the PoC? Maybe some serialization needs?

ginnun (commented):
Brilliant work @iccicci 
Please see my humble comments.

packages/cardano-services/src/WsServer/server.ts
```diff
@@ -0,0 +1,278 @@
...
+    this.db = dependencies.db;
+    this.logger = dependencies.logger;
+
+    // Create the HTTP and the WebSocket servers
+    this.server = this.createHttpServer();
+    this.wss = new WebSocketServer({ server: this.server });
+
+    // Attach the handlers to the servers events
+
+    this.wss.on('connection', this.createOnConnection(dependencies, (cfg.heartbeatTimeout || 60) * 1000));
```
mirceahasegan, line 49: Tiny nitpick: you could use `Seconds.toMilliseconds` for the transformation

packages/cardano-services/src/Program/programs/providerServer.ts
```diff
@@ -137,17 +145,13 @@ const serviceMapFactory = (options: ServiceMapFactoryOptions) => {
     return new TypeormStakePoolProvider(args, { cache: getDbCache(), connectionConfig$, entities, logger });
   });
 
-  let networkInfoProvider: DbSyncNetworkInfoProvider | undefined;
-
   const getNetworkInfoProvider = (cardanoNode: CardanoNode, dbPools: DbPools) => {
```
mirceahasegan, line 148: Is caching the `networkInfoProvider` not necessary anymore? There seemed to be a case where it was reused if already created.

packages/cardano-services-client/src/WebSocket.ts
```diff
@@ -0,0 +1,208 @@
...
+  logger: Logger;
+}
+
+const isEventError = (error: unknown): error is { error: Error } =>
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  typeof error === 'object' && !!error && (error as any).error instanceof Error;
+
+export class CardanoWsClient {
+  /** The client id, assigned by the server. */
+  clientId = 'not-connected';
```
mirceahasegan, line 38: Not for the POC. The `not-connected` string should be a constant.

mirceahasegan (dismissed):
Fantastic work @iccicci !
Please feel free to ignore my comments, as they are all minor changes that can be addressed post POC.

packages/cardano-services/src/Program/programs/wsServer.ts
```diff
@@ -0,0 +1,62 @@
...
+    const shutDown = (signal: string) => {
+      if (shuttingDown) return;
+      shuttingDown = true;
+
+      logger.info(`Shutting down Cardano SDK WebSocket Server due to ${signal} ...`);
+
+      server.close(() => {
+        Promise.all([db.end(), cardanoNode.shutdown()])
+          .then(() => logger.info('Cardano SDK WebSocket Server shutdown'))
+          .catch((error) => logger.error(error, 'Error while shutting down WebSocket Server'));
```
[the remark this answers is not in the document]
iccicci: I think it doesn't worth a retry procedure for shut down sequence. Usually problems in shutdown sequences require some human actions

.github/workflows/continuous-integration-e2e.yaml
iccicci: That's the default in development environment, this has no impact on deployed environments.

packages/cardano-services/src/Program/programs/providerServer.ts
iccicci: Adding configuration variables to be prepared for possible evolutions of the interface doesn't sounds as a good time investment: and if the interface will evolve in a not planned way?
I think it's better to improve the configuration when the interface evolves so the configuration improvement can reflect the interface evolution.

packages/cardano-services/src/WsServer/db.ts
iccicci: That's a way to wait for db-sync schema to be ready.
It's just a check to be sure the table is created before attaching a trigger to it.

packages/cardano-services/src/WsServer/server.ts
iccicci: Sorry, could you please elaborate a bit? Thank you,

packages/cardano-services/src/WsServer/server.ts
iccicci: It is actually a retriable error. The retry strategy is to close the connection to open a new 1 second later.

packages/projection-typeorm/src/entity/PoolMetadata.entity.ts
iccicci: This is not bound to the WebSocket PoC, this is related to adding the hot reload feature.
`tsx` doesn't support `emitDecoratorMetadata`, so TypeORM is not able to guess the column type, so I had to make then explicit.

packages/cardano-services/src/Program/programs/providerServer.ts
iccicci: I had the same feeling as well, but I double checked: `getNetworkInfoProvider` is called only once in the entire SDK source.

packages/cardano-services/src/WsServer/server.ts
iccicci: 70f3fe7bc1 should solve this

packages/cardano-services-client/src/WebSocket.ts
iccicci: 70f3fe7bc1 should solve this

packages/cardano-services/src/Program/options/common.ts
iccicci: 70f3fe7bc1 should solve this

packages/cardano-services-client/src/WebSocket.ts
```diff
@@ -0,0 +1,182 @@
...
+          if (eraSummaries) this.networkInfoSubjects.eraSummaries.next(eraSummaries);
+          if (genesisParameters) this.networkInfoSubjects.genesisParameters.next(genesisParameters);
+          if (lovelaceSupply) this.networkInfoSubjects.lovelaceSupply.next(lovelaceSupply);
+          if (protocolParameters) this.networkInfoSubjects.protocolParameters.next(protocolParameters);
+          if (stake) this.networkInfoSubjects.stake.next(stake);
+          // Emit ledgerTip as last one
+          if (ledgerTip) this.networkInfoSubjects.ledgerTip.next(ledgerTip);
+        }
+      } catch (error) {
+        this.logger.error(error, 'While parsing message', event.data, this.clientId);
```
mkazlauskas, line 124: Same as

packages/cardano-services-client/src/WebSocket.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    ws.onclose = () => {
+      this.logger.info('WebSocket client connection closed', this.clientId);
+      this.clientId = NOT_CONNECTED_ID;
+      this.retry();
+    };
+
+    ws.onerror = (error: unknown) => {
+      this.logger.error(isEventError(error) ? error.error : error, 'Async error from WebSocket client', this.clientId);
+      ws.close();
```
mkazlauskas, line 136: Same as

packages/cardano-services-client/src/WebSocket.ts
```diff
@@ -0,0 +1,182 @@
...
+  /**
+   * Sends a request through WS to server.
+   *
+   * @param request the request.
+   * @returns `true` is sent, otherwise `false`.
+   */
+  request(request: WSMessage) {
+    if (this.status !== 'connected') return false;
+
+    this.ws!.send(NJSON.stringify(request));
```
mkazlauskas, line 177: Can this throw? If yes, then I think it would interrupt the heartbeat interval unrecoverably

packages/cardano-services/package.json
```diff
@@ -116,6 +118,7 @@
     "json-bigint": "^1.0.0",
     "jsonschema": "^1.4.1",
     "lodash": "^4.17.21",
+    "next-json": "^0.3.2",
```
mkazlauskas, line 121: I don't mind using this if it's better, but then I think we should be consistent and replace `toSerializableObj`/`fromSerializableObj` utils within entire codebase. Probably better to do in a separate PR, good topic for tech talks too!

```suggestion
```

packages/cardano-services-client/src/WebSocket.ts
```diff
@@ -0,0 +1,182 @@
...
+  /** The client id, assigned by the server. */
+  clientId = NOT_CONNECTED_ID;
+
+  /** The `Observable` form of `NetworkInfoProvider`. */
+  networkInfo: {
+    [m in NetworkInfoMethods]: Observable<AsyncReturnType<NetworkInfoProvider[m]>>;
+  };
+
+  /** WebSocket based `NetworkInfoProvider` implementation. */
+  networkInfoProvider: NetworkInfoProvider;
```
mkazlauskas, line 56: Great idea for phased implementation! :heart:

packages/cardano-services/src/WsServer/server.ts
```diff
@@ -0,0 +1,319 @@
...
+  /** The heartbeat timeout in seconds. */
+  heartbeatTimeout?: number;
+
+  /** The port to listen. */
+  port: number;
+}
+
+export interface WsServerDependencies {
+  /** The `CardanoNode` object. */
+  cardanoNode: CardanoNode;
```
mkazlauskas, line 33: Wouldn't this be be a more suitable dependency, since it already emits some data that we would otherwise have to query? It might be missing some properties that we'd need to add though.
```suggestion
  cardanoNode: ObservableCardanoNode;
```

packages/projection/package.json
```diff
@@ -9,6 +9,7 @@
   "module": "dist/esm/index.js",
   "exports": {
     ".": {
+      "development": "./src/index.ts",
```
mkazlauskas, line 12: If not too difficult, would be great to extract hot reload changes into a separate commit

packages/cardano-services/src/WsServer/db.ts
```diff
@@ -0,0 +1,68 @@
...
+END;`;
+
+const notifyTipProcedure = `\
+CREATE OR REPLACE FUNCTION sdk_notify_tip()
+RETURNS TRIGGER AS
+$BODY$${notifyTipBody}$BODY$
+LANGUAGE plpgsql`;
+
+const notifyTipTrigger = `\
+CREATE TRIGGER sdk_notify_tip
```
mkazlauskas, line 24: Nice :heart:

mkazlauskas (changes requested):

packages/cardano-services-client/src/WebSocket.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  private connect() {
+    this.status = 'connecting';
+
+    const ws = new WebSocket(this.url);
+
+    // eslint-disable-next-line sonarjs/cognitive-complexity
+    ws.onmessage = (event) => {
+      try {
+        if (typeof event.data !== 'string') throw new Error('Unexpected data from WebSocket ');
```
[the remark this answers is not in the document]
iccicci: This should never happen. I think for now nothing more than a log is required.

In case it will actually never happen, nothing more is required, in case it happens, we should check what causes it in how to handle.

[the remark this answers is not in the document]
mkazlauskas: Should we `logger.error` instead of throw then?

packages/cardano-services-client/src/WebSocket.ts
iccicci: Also this should never happen. I suggest the same approach: let's spend time on it if it actually happens.

packages/projection/package.json
mkazlauskas: Maybe I misunderstand the purpose of hot reload: isn't it to restart the provider server when ts files change? If that's correct, then it could be used for non-websocket provider server development too. :thinking:

iccicci: Yes, it is my intention, but before applying to all other container I preferred to have the team approval (in this PR), next propagating to other containers with a distinct ticket and PR.

mkazlauskas: Ah I see, [this](https://github.com/input-output-hk/cardano-js-sdk/pull/1352/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R126) would belong in feature PR as it's adding a new type of server.

If we extract it, then maybe we can just add an npm script to run tsx for the existing provider server. Then team can review it in isolation.

packages/cardano-services/src/WsServer/server.ts
iccicci: I didn't checked it; ATM `CardanoNode` does its job, maybe we could decouple this as a separated improvement task?

mkazlauskas: Sure - not a required change for a POC. Do we have an epic/story for a production-ready ws provider yet where we could log this?

iccicci: Not yes, ATM it belongs to the BEv2 phase 1 LW-10676

packages/cardano-services/src/Program/programs/providerServer.ts
mkazlauskas: I actually thought about making a similar change request, but decided not to as it's a limited scope POC

packages/cardano-services-client/src/WebSocket.ts
iccicci: It can throw only in case or WebSocket disconnection, which has its own handling in the `ws.on('error', ... )` handler, heartbeat included.

packages/cardano-services-client/src/WebSocket.ts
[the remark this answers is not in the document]
iccicci: It is, in the `catch` block the error is logged; it is also thrown as being not a string the message can't be parsed so next steps can't be performed

packages/cardano-services-client/src/WebSocket.ts
```diff
@@ -0,0 +1,182 @@
...
+    };
+
+    ws.onopen = () => {
+      this.status = 'connected';
+      this.ws = ws;
+      this.heartbeat();
+    };
+  }
+
+  private heartbeat() {
```
mkazlauskas, line 146: Dumb question: what is the purpose of heartbeat? I'd expect tip notifications from the server to keep the connection alive :thinking:

packages/cardano-services-client/src/WebSocket.ts
mkazlauskas: I'm wondering if an uncaught error here could have any side effects though, it's being called in `setTimeout`. I'd feel safer if we did `try { send(); } catch { return false; }`

Btw, any reason for `request` to be a public method? :thinking:

packages/cardano-services-client/src/WebSocket.ts
iccicci: Here the logic is different, this is the handler for async errors from the WebSocket,
In such cases, the error is logged and the connection is closed.
Next the retry strategy, will reopen a new connection.

packages/cardano-services-client/src/WebSocket.ts
iccicci: It's purpose is on the reverse side.
Exactly thanks to the tip there is no need for a server to client heartbeat.
The purpose of this is to let the server know the client is still connected (and not disconnected with some hard disconnections which leaves the connection "open" at TCP-IP level but broken anyway).

packages/cardano-services-client/src/WebSocket.ts
iccicci: Nice catch! The Promise must be correctly handled in the caller. Thank you.

> any reason for request to be a public method?

Probably just a refuse....

packages/cardano-services-client/src/WebSocket.ts
mkazlauskas: Right, but as a consumer of `NetworkInfoProvider`, I would expect methods calls to reject with a `ProviderError` when there are issues. Similarly for `this.networkInfo.x` observable consumers - would expect it to error when there are issues. Otherwise there is no way for consumer to know if it's healthy/up-to-date.

packages/cardano-services-client/src/WebSocket.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  private heartbeatInterval: number;
+  private heartbeatTimeout: NodeJS.Timeout;
+  private logger: Logger;
+  private status: WSStatus = 'idle';
+  private url: URL;
+  private ws: WebSocket;
+
+  private networkInfoSubjects = {} as {
+    [m in NetworkInfoMethods]: ReplaySubject<AsyncReturnType<NetworkInfoProvider[m]>>;
```
mkazlauskas, line 66: Our convention for observable names is with `$` suffix
```suggestion
    [m in NetworkInfoMethods as `${m}$`]: ReplaySubject<AsyncReturnType<NetworkInfoProvider[m]>>;
```

packages/cardano-services-client/src/WebSocket.ts
mkazlauskas: Hm, are you sure it's required? I haven't worked with ws for a very long time, but I'd expect such failures to be handled by the protocol or lower level libraries

.github/workflows/continuous-integration-e2e.yaml
rhyslbw: That may be the case @iccicci but the dev or test env defaults should be aligned with the existing API

iccicci: We can revert this change before merging to `master`, but as long as this PR is open, this is a mandatory change in this branch to test the changes it contain...

packages/cardano-services-client/src/WebSocket.ts
iccicci: It's more a nice to have, we had some public discussion on this point as well.
It's true that underlaying layers **should have** the seme kind of logic, anyway each o pin the route has its own timeout...

packages/cardano-services/src/WsServer/server.ts
ginnun: If closing the WS causes service to shutdown, It covers what I was asking

packages/projection-typeorm/src/entity/PoolMetadata.entity.ts
ginnun: My personal preference would be to split the hotreload and the PoC work in an ideal world.

packages/cardano-services-client/src/WebSocket.ts
iccicci: 6ba4ef5a2d should solve this

packages/cardano-services/package.json
iccicci: 6ba4ef5a2d should solve this

packages/cardano-services-client/src/WebSocket.ts
iccicci: 6ba4ef5a2d should solve this

packages/cardano-services-client/src/WebSocket.ts
```diff
@@ -0,0 +1,182 @@
...
+  typeof error === 'object' && !!error && (error as any).error instanceof Error;
+
+export class CardanoWsClient {
+  /** The client id, assigned by the server. */
+  clientId = NOT_CONNECTED_ID;
+
+  /** The `Observable` form of `NetworkInfoProvider`. */
+  networkInfo: {
+    [m in NetworkInfoMethods]: Observable<AsyncReturnType<NetworkInfoProvider[m]>>;
+  };
```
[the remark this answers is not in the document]
iccicci: Yes, it would be nice to generalize the type. I spent about 1h, but I wasn't able to.
There is the problem of extracting the `$` from the key to access the methods of the `Provider` and also it would be nice to generalize the `Observable`: we have two types (one public one private) almost identical.

packages/cardano-services-client/src/WebSocket.ts
iccicci: One of the improvements of this PoC is the centralization of error handling with automatic retry. The caller no longer need to implement a retry logic for each call, the caller can just trust `CardanoWsClient` will retry autonomously.

Once said that, I agree some method to inform the caller about errors is required anyway.

With 6ba4ef5a2d I introduced `health$: Observable<HealthCheckResponse>` which emits a new state each time the state changes, so the caller can centralize error handling on their side as well.

packages/cardano-services-client/src/WebSocket.ts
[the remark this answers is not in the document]
mkazlauskas: Let's capture this in a ticket when planning follow-up work

packages/cardano-services-client/src/WebSocket.ts
mkazlauskas: Agree that this is an improvement from 'ease of use' standpoint, but not convinced that it's an improvement overall. Client still can't tell:
- which method/observable failed (or multiple?)?
- when did it fail?
- the error is also partially lost - we used to do `instanceof` and/or `failure` properties of the errors handle them

Happy to call this resolved for the POC, but please create a SPIKE ticket to review provider usages and see if we need more granular notifications of the errors. I suspect that a better solution is to just throw/reject on subscription/method call, but not 100% sure.

mkazlauskas (commented):
Great work @iccicci ! :muscle: :rocket: I approve the POC (but not merging this PR to `master` at this point).

Left some change requests open until they're captured in tickets.

packages/projection/package.json
iccicci: Decoupled #1354 to handle this

packages/cardano-services-client/src/WebSocket.ts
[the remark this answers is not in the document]
iccicci: LW-10837

packages/cardano-services/src/WsServer/server.ts
iccicci: LW-10840

packages/cardano-services-client/src/WebSocket.ts
iccicci: LW-10841

gytis-ivaskevicius (dismissed):

packages/e2e/test/ws-server/webSocket.test.ts
```diff
@@ -1,5 +1,195 @@
...
+
+    timeout = setTimeout(() => {
+      subscription.unsubscribe();
+      reject(new Error('WsProvider timeout'));
+    }, 10_000);
+
+    timeout.unref();
+  });
+
+describe('Web Socket', () => {
```
mkazlauskas, line 87: I'd like to see a test for some network info provider failure due to connection issue, because that is quite a lot of untested code/logic

mkazlauskas (changes requested):

packages/e2e/test/ws-server/webSocket.test.ts
iccicci: d39330353d and 7a3447e3a9 should solve this

ginnun (approved):
lgtm

mkazlauskas (approved):
Great work @iccicci ! :rocket:

PR #1354, opened 2024-07-03
LW-10739 Introduce preliminary changes for hot reload

iccicci:
Decoupled preliminary changes for hot reload from #1352 .


mkazlauskas (changes requested):
Looks good! Would be great to add package.json script(s) and/or some instructions in README about using this

packages/cardano-services/src/util/polling/DbSyncEpochPollService.ts
```diff
@@ -51,7 +51,7 @@ export class DbSyncEpochPollService implements EpochMonitor {
       name: 'current_epoch',
       text: findLastEpoch
     });
-    return Cardano.EpochNo(result.rows[0].no);
+    return Cardano.EpochNo(result.rowCount ? result.rows[0].no : 0);
```
mkazlauskas, line 54: This looks like a separate fix, is it?

mkazlauskas (dismissed):
Great work! :rocket:

packages/cardano-services/src/util/polling/DbSyncEpochPollService.ts
mkazlauskas: Would be best to do in a separate commit then

iccicci: a62d450e8f should solve this

ginnun (approved):
lgtm

packages/e2e/README.md
```diff
@@ -99,37 +99,10 @@ Address:    addr_test1qr0c3frkem9cqn5f73dnvqpena27k2fgqew6wct9eaka03agfwkvzr0zyq
...
-So, if while working on the `core` and `cardano-services` packages simultaneously, we need to restart the provider-server
-to check the effect of our ongoing changes, it is enough to issue the following command:
-
-```bash
-$ yarn workspace @cardano-sdk/core build && yarn workspace @cardano-sdk/cardano-services build && docker exec -it local-network-e2e-provider-server-1 kill 1
-```
-
-if it exits with error we need to fix our ongoing changes, if it exits silently, the provider-server is now running with our changes.
+The **local-network** runs with hot reload: once running, just saving a source file makes all the containers depending on that file
+are stopped and restarted loading the new version of the source files. No additional operations are required.
```
mirceahasegan, line 105: 🥳 🥇 👏 👏 👏 👏

mirceahasegan (approved):
Fantastic work @iccicci 🥇

PR #1356, opened 2024-07-05
LW-10816 Improve typeORM provider health check

iccicci:
# Context

SRE reported some not accurate health check.

# Proposed Solution

Added a query in the `healthCheck` method.

Fixed an unhandled error which was making rxjs to exit the program.

### DIFF
--- a/packages/cardano-services/src/util/TypeormProvider/TypeormProvider.ts
+++ b/packages/cardano-services/src/util/TypeormProvider/TypeormProvider.ts
@@ -10,6 +10,8 @@ export interface TypeormProviderDependencies {
   connectionConfig$: Observable<PgConnectionConfig>;
 }
 
+const unhealthy = { ok: false, reason: 'Provider error' };
+
 export abstract class TypeormProvider extends TypeormService implements Provider {
   health: HealthCheckResponse = { ok: false, reason: 'not started' };
 
@@ -17,11 +19,18 @@ export abstract class TypeormProvider extends TypeormService implements Provider
     super(name, { connectionConfig$, entities, logger });
     // We skip 1 to omit the initial null value of the subject
     this.dataSource$.pipe(skip(1)).subscribe((dataSource) => {
-      this.health = dataSource ? { ok: true } : { ok: false, reason: 'Provider error' };
+      this.health = dataSource ? { ok: true } : unhealthy;
     });
   }
 
   async healthCheck(): Promise<HealthCheckResponse> {
+    if (this.state === 'running')
+      try {
+        await this.withDataSource((dataSource) => dataSource.query('SELECT 1'));
+      } catch {
+        return unhealthy;
+      }
+
     return this.health;
   }
 }
--- a/packages/cardano-services/test/util/TypeormProvider.test.ts
+++ b/packages/cardano-services/test/util/TypeormProvider.test.ts
@@ -1,0 +1,45 @@
+import { PgConnectionConfig } from '@cardano-sdk/projection-typeorm';
+import { TypeormProvider } from '../../src/util';
+import { getEntities } from '../../src';
+import { logger } from '@cardano-sdk/util-dev';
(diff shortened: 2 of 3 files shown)


packages/cardano-services/src/util/TypeormProvider/TypeormProvider.ts
```diff
@@ -22,6 +22,19 @@ export abstract class TypeormProvider extends TypeormService implements Provider
   }
 
   async healthCheck(): Promise<HealthCheckResponse> {
+    try {
+      if (this.state === 'running') {
+        const dataSource = await firstValueFrom(this.dataSource$);
+
+        if (!dataSource) throw new Error('No DataSource');
```
mkazlauskas, line 29: Is there any point of calling `onError` in this case? It should already be (re)connecting when it's already `null`. This could potentially mess up the re-connection logic. I suggest returning unhealthy here instead of throwing.

packages/cardano-services/src/util/TypeormProvider/TypeormProvider.ts
```diff
@@ -22,6 +22,19 @@ export abstract class TypeormProvider extends TypeormService implements Provider
...
+
+        await dataSource.query('SELECT 1');
+      }
+    } catch (error) {
+      this.logger.error(error, 'Error while performing health check');
+      this.onError(error);
```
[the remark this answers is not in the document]
iccicci: Yes, I've checked it. `this.onError` updates `this.health`.

packages/cardano-services/src/util/TypeormProvider/TypeormProvider.ts
iccicci: a9ccfea1b0 should solve this

mkazlauskas (approved):
Nice work! :rocket:

mirceahasegan (dismissed):
Great work detective work @iccicci 🥇

packages/cardano-services/test/util/TypeormProvider.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    await provider.initialize();
+    await provider.start();
+  });
+
+  afterAll(() => provider.shutdown());
+
+  test('healthCheck', async () => {
+    expect((await provider.healthCheck()).ok).toBeTruthy();
+    expect((await provider.healthCheck()).ok).toBeFalsy();
+    expect((await provider.healthCheck()).ok).toBeTruthy();
```
mkazlauskas, line 39: Took a while to understand how this works, would be nice to add a comment
```suggestion
    expect((await provider.healthCheck()).ok).toBeTruthy();
    expect((await provider.healthCheck()).ok).toBeFalsy();
    // data source is re-created
    expect((await provider.healthCheck()).ok).toBeTruthy();
```

packages/cardano-services/test/util/TypeormProvider.test.ts
```diff
@@ -0,0 +1,43 @@
...
+
+        called = true;
+        return Promise.resolve(0);
+      }
+    });
+  })
+}));
+
+describe('TypeormProvider', () => {
+  const connectionConfig$ = new BehaviorSubject(projectorConnectionConfig);
```
mkazlauskas, line 25: Would be nice to make it clear that this doesn't matter for these tests
```suggestion
  const notUsedConnectionConfig = {} as ConnectionConfig;
  const connectionConfig$ = of(notUsedConnectionConfig)
```

mkazlauskas (changes requested):

packages/cardano-services/test/util/TypeormProvider.test.ts
iccicci: d8d401bd29 should solve this

mkazlauskas (approved):
Nice work! :rocket:

mirceahasegan (approved):
Awesome!

PR #1361, opened 2024-07-12
LW-10966 Remove "development" from package.json "exports" map in published packages

iccicci:
# Context

The recently added `package.json#exports["."].development` attribute (used for hot reload) is source of problems for build of client applications which use the published SDK packages.

# Proposed Solution

Added a command in the release procedure to remove given attribute right before publishing the SDK packages.


mirceahasegan (approved):
Looks good to me 👏

mkazlauskas (approved):
Looks good!

mirceahasegan (approved):

PR #1368, opened 2024-07-17
Fix lw-10852 Ledger hw sign voting proposals

mirceahasegan:
# Context

Signing a vote proposal with Ledger HW device returns MissingVKeyWitnessesUTXOW.
This was due to attempting to sign with DRep key hash which is not supported.

# Proposed Solution
Compare wallet DRep key hash to voter DRep key hash. If they are match, sign with `VoterType.DREP_KEY_PATH`


iccicci (commented):
Well written... apart from that, I'm sorry, I don't feel strong enough to do a valuable review

mkazlauskas (approved):
Great work! :rocket: 

Just maybe would be good to remove `!` from `fix(hardware-ledger)!: sign voting procedure with drep keypath` because `sign voting procedure with drep keypath` will be listed as a breaking change. It's just a fix, and breaking change will be picked up from the commit footer.

iccicci (approved):
Great job! 💪

PR #1370, opened 2024-07-17
Chore: LW-10279 cardano node upgrade to 9.0.0

mirceahasegan:
# Context

Please see commits.

Commit 24c0ce3a23611f25db3c8f318e516a5fca3ad55e: is meant as a temporary solution until the CardanoNode embedded cardano-cli
does not have [this issue]([https://github.com/IntersectMBO/cardano-cli/issues/825\](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html))
anymore, so cardano-cli 9.0.0.2 or higher.
Since cardano-cli 9.0.0.1 which has the fix cannot be used because it is dynamically
linked, and there is no newer release, I am forced to embed a customly built
cardano-cli to be used for bootstrapping the local-network.
We can remove the locally built cardano-cli when we have cli 9.0.0.2, and revert this
whole commit once it is embedded in cardano-node.


mkazlauskas (approved):
Great work @mirceahasegan !
Probably best to squash all bump commits (except the first and the last one).
Temporarily checking in 16.7M blob is not great, but I think we can do that to not be blocked.

iccicci (approved):
💪 🚀 🥇

PR #1374, opened 2024-07-18
Credential indexer

will-break-it:
# Context

# Proposed Solution


packages/core/src/Cardano/util/credentialsFromAddress.ts
```diff
@@ -0,0 +1,40 @@
+import { Hash28ByteBase16 } from '@cardano-sdk/crypto';
+import { Cardano } from '../../../dist/esm';
```
mkazlauskas, line 2: This is going to cause trouble. Best to have multiple imports from `src/`, e.g. `from '../Address'`.

packages/projection/src/operators/Mappers/withAddresses.ts
```diff
@@ -3,6 +3,7 @@ import { Hash28ByteBase16 } from '@cardano-sdk/crypto';
 import { WithUtxo } from './withUtxo';
 import { unifiedProjectorOperator } from '../utils';
 import uniq from 'lodash/uniq.js';
+import { credentialsFromAddress } from '@cardano-sdk/core/src/Cardano/util';
```
mkazlauskas, line 6: This is going to cause trouble; should import from index, e.g. `import { Cardano } from '@cardano-sdk/core'` then use it like `Cardano.util.credentialsFromAddress`

mkazlauskas (changes requested):

packages/projection-typeorm/src/operators/storeCredentials.ts
```diff
@@ -0,0 +1,29 @@
...
+    block: { body: txs },
+    eventType,
+    queryRunner
+  } = evt;
+
+  // produced credentials will be automatically deleted via block cascade
+  if (txs.length === 0 || eventType !== ChainSyncEventType.RollForward) return;
+  
+  const involvedOutputCredentials = uniq(evt.utxo.produced.map(([_, txOut]) => txOut.address).map((address) => {
+    const credential = credentialsFromAddress(address);
```
mkazlauskas, line 18: That's fine for POC but I consider this tech debt. It should depend on `WithAddresses` mapper instead (`typeormOperator<Mappers.WithUtxo & WithAddresses>`). We just need to add functionality to group addresses by tx id there.

mkazlauskas (changes requested):

compose/common.yml
```diff
@@ -176,7 +176,8 @@ services:
         '--node-config',
         '/config/cardano-node/config.json',
         '--log-level-websocket',
-        'Warning'
+        'Warning',
+        '--include-cbor'
```
mkazlauskas, line 180: We don't need the granular ones
```suggestion
        '--include-transaction-cbor'
```

PR #1378, opened 2024-07-19
Fix/backout cli blob

mirceahasegan:
# Context

- download cardano-cli build 9.1.0.0 instead of the binary blob
- fix protocol parameters issue found while merging conway-era to master
- install scripts need to be updated to cardano node 9.0.0 too

# Proposed Solution


iccicci (approved):
Good job! 🥇

AngelCastilloB (approved):
Great work @mirceahasegan 🚀

PR #1380, opened 2024-07-22
fix(input-selection): round robin random improve now produces valid change outputs when wallet has too many assets

AngelCastilloB:
# Context

Round Robin Random Improve produces a change output for each output in the transaction, however, if the the total asset selected as inputs do not fit in the change outputs, it will produce invalid outputs (violates the tokenBundleSizeExceedsLimit constraint).

# Proposed Solution

The RRRI selection strategy now splits change outputs if assets do not fit.


packages/input-selection/test/RoundRobinRandomImprove.test.ts
```diff
@@ -1,6 +1,15 @@
...
 import { TxTestUtil } from '@cardano-sdk/util-dev';
+import {
+  babbageSelectionParameters,
+  cborUtxoSetWithManyAssets,
+  getCoreUtxosFromCbor,
+  txBuilder,
+  txEvaluator
+} from './vectors';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { defaultSelectionConstraints } from '@cardano-sdk/tx-construction';
```
mkazlauskas, line 12: Is this a circular ref? If so, probably best to hoist this somewhere. And if not possible then duplicating some functionality in tests is ok.
```suggestion
import { defaultSelectionConstraints } from '?';
```

mkazlauskas (changes requested):

packages/input-selection/test/RoundRobinRandomImprove.test.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1380/commits/254e1af80399d849b1cab2a29e55591cf6418342

mkazlauskas (approved):
Nice work! :detective:

iccicci (approved):
LGTM!

PR #1381, opened 2024-07-23
fix: use drep key hash in stubSignTransaction

mirceahasegan:
# Context

This is a followup fix for "Cannot submit "vote on info action" with Ledger device".
`StubSignTransactionProps` still accepted `drepPublicKey`, but did not use it for witnessing.

# Proposed Solution
Remove `drepPublicKey` and rely on the `drepKeyHashHex` from the sign transactionContext.


iccicci (approved):
LGTM!

ginnun (approved):
lgtm

PR #1382, opened 2024-07-23
Add protected automated deployment to environments

ce0la:
chore: Remove kubeconfig dependency (@gytis-ivaskevicius commit)
chore: Add workflow for protected auto deployments (extra configuration done in repo > settings > environments)

# Context

# Proposed Solution

### DIFF
--- a/.github/workflows/deploy-env.yaml
+++ b/.github/workflows/deploy-env.yaml
@@ -47,5 +47,7 @@ jobs:
         run: |
           echo "${{ secrets.ENVRC }}" > .envrc.local
           source .envrc.local
+          which aws || true
           nix develop
+          which aws
           echo "yes" | nix run .#cardano-services.${{ inputs.environment || 'ops-preview-1@us-east-1' }}.apply
--- a/.github/workflows/deploy-env.yaml
+++ b/.github/workflows/deploy-env.yaml
@@ -49,5 +49,5 @@ jobs:
           source .envrc.local
           nix develop -L --command bash -c '
             set -euo pipefail
-            echo "yes" | nix run .#cardano-services.${{ inputs.environment || "ops-preview-1@us-east-1" }}.apply
+            echo "yes" | nix run .#cardano-services.${{ inputs.environment || \'ops-preview-1@us-east-1\' }}.apply
           '
--- a/.github/workflows/deploy-env.yaml
+++ b/.github/workflows/deploy-env.yaml
@@ -34,7 +34,7 @@ on:
 
 jobs:
   deploy:
-    environment: ${{ inputs.environment || ops-preview-1 }}
+    environment: ${{ inputs.environment || 'ops-preview-1' }}
     runs-on: ubuntu-latest
     steps:
       - name: 📥 Checkout repository
--- a/.github/workflows/deploy-env.yaml
+++ b/.github/workflows/deploy-env.yaml
@@ -34,7 +34,7 @@ on:
 
 jobs:
   deploy:
-    environment: ${{ inputs.environment || 'ops-preview-1@us-east-1' }}
+    environment: ${{ inputs.environment }}
     runs-on: ubuntu-latest
     steps:
       - name: 📥 Checkout repository
@@ -46,8 +46,7 @@ jobs:
       - name: 🚀 Deploy
         run: |
           echo "${{ secrets.ENVRC }}" > .envrc.local
-          source .envrc.local
-          nix develop -L --command bash -c "
+          nix develop -L --command bash -c '
             set -euo pipefail
-            echo "yes" | nix run .#cardano-services.${{ inputs.environment || 'ops-preview-1@us-east-1' }}.apply
-          "
+            echo "yes" | nix run .#cardano-services.${{ inputs.environment }}.apply
+          '
--- a/.github/workflows/deploy-env.yaml
+++ b/.github/workflows/deploy-env.yaml
@@ -47,7 +47,7 @@ jobs:
         run: |
           echo "${{ secrets.ENVRC }}" > .envrc.local
           source .envrc.local
-          nix develop -L --command bash -c '
+          nix develop -L --command bash -c "
             set -euo pipefail
-            echo "yes" | nix run .#cardano-services.${{ inputs.environment || \'ops-preview-1@us-east-1\' }}.apply
-          '
+            echo "yes" | nix run .#cardano-services.${{ inputs.environment || 'ops-preview-1@us-east-1' }}.apply
+          "
--- a/.github/workflows/deploy-env.yaml
+++ b/.github/workflows/deploy-env.yaml
@@ -1,7 +1,6 @@
 name: Deploy Environment
 
 on:
-  push:
   workflow_dispatch:
     inputs:
       environment:
@@ -10,10 +9,7 @@ on:
         required: true
         default: 'ops-preview-1'
         options:
-        - ALL
-        - LIVE
         - live-sanchonet@eu-central-1@v2
-        - DEV
         - live-mainnet@eu-central-1@v2
         - live-sanchonet@us-east-2@v1
         - dev-mainnet@us-east-1
@@ -21,7 +17,6 @@ on:
         - live-sanchonet@us-east-2@v2
         - dev-preprod@us-east-1@v2
         - live-preprod@eu-central-1@v2
-        - OPS
         - dev-preview@us-east-1
         - live-preprod@us-east-2@v2
         - ops-preprod-1@us-east-1
--- a/.github/workflows/deploy-env.yaml
+++ b/.github/workflows/deploy-env.yaml
@@ -10,7 +10,6 @@ on:
         required: true
         default: 'ops-preview-1'
         options:
-        options:
         - ALL
         - LIVE
         - live-sanchonet@eu-central-1@v2
--- a/.github/workflows/deploy-env.yaml
+++ b/.github/workflows/deploy-env.yaml
@@ -46,5 +46,6 @@ jobs:
       - name: 🚀 Deploy
         run: |
           echo "${{ secrets.ENVRC }}" > .envrc.local
+          source .envrc.local
           nix develop
           echo "yes" | nix run .#cardano-services.${{ inputs.environment || 'ops-preview-1@us-east-1' }}.apply
--- a/.github/workflows/deploy-env.yaml
+++ b/.github/workflows/deploy-env.yaml
@@ -1,6 +1,7 @@
 name: Deploy Environment
 
 on:
+  push:
   workflow_dispatch:
     inputs:
       environment:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-env.yaml
```diff
@@ -0,0 +1,48 @@
...
+        - ops-preview-1
+      region:
+        description: 'Region containing cluster to deploy to'
+        type: choice
+        required: true
+        default: 'us-east-1'
+        options:
+        - us-east-1
+        - us-east-2
+        - eu-central-1
```
gytis-ivaskevicius, line 22: We should have multiple listed environments such as `ops-preview-1@us-east-1` so people would not need to guess the environments, also some of the environments have `v2` postfix

.github/workflows/deploy-env.yaml
```diff
@@ -0,0 +1,48 @@
...
+jobs:
+  deploy:
+    environment: ${{ inputs.environment }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: 📥 Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Install QEMU
+        run: sudo apt-get install -y qemu-user-static
```
gytis-ivaskevicius, line 33: ?

.github/workflows/deploy-env.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - name: 🧰 Setup Nix
+        uses: cachix/install-nix-action@v21
+        with:
+          extra_nix_config: |
+            system = aarch64-linux
+
+      - name: 🚀 Deploy
+        run: |
+          echo "${{ secrets.ENVRC }}" > .envrc
+          source .envrc
```
gytis-ivaskevicius, line 44: this should be `.envrc.local` and does not need to be sourced

.github/workflows/deploy-env.yaml
```diff
@@ -0,0 +1,48 @@
...
+          nix develop
```
gytis-ivaskevicius, line 45: This will not provide you with `aws` executable if you don't do `-c`

.github/workflows/deploy-env.yaml
```diff
@@ -0,0 +1,48 @@
...
+          nix run .#cardano-services.${{ inputs.environment }}@${{ inputs.region }}.status
+          nix run .#cardano-services.${{ inputs.environment }}@${{ inputs.region }}.plan
+          nix run .#cardano-services.${{ inputs.environment }}@${{ inputs.region }}.apply -y
```
gytis-ivaskevicius, line 48: Why 3 commands?  This should be `echo yes | nix run .#cardano-services.${{ inputs.environment }}@${{ inputs.region }}.apply`

gytis-ivaskevicius (changes requested):

.github/workflows/deploy-env.yaml
ce0la: ✅

.github/workflows/deploy-env.yaml
ce0la: It needs to be sourced: https://github.com/input-output-hk/cardano-js-sdk/actions/runs/10089658475/job/27897524146

.github/workflows/deploy-env.yaml
ce0la: Taken out.

.github/workflows/deploy-env.yaml
ce0la: It provides `aws` executable without the flag.

.github/workflows/deploy-env.yaml
ce0la: ✅

.github/workflows/deploy-env.yaml
ce0la: https://github.com/input-output-hk/cardano-js-sdk/actions/runs/10089733209/job/27897743167

ce0la:
Successful deploy to `ops-preview-1@us-east-1`: https://github.com/input-output-hk/cardano-js-sdk/actions/runs/10089733209/job/27897743167

I am leaving the `push:` trigger to make it easy for anyone testing to do that. Once approved, I am going to take out the `push:` trigger and the `ops-preview-1@us-east-1` default value in instances of `{{ inputs.environment || 'ops-preview-1@us-east-1' }}`.

I would also create new environments for the repository and set protection rules as appropriate; to be discussed with @bernokl.

Cc: @michalrus @gytis-ivaskevicius

.github/workflows/deploy-env.yaml
```diff
@@ -0,0 +1,51 @@
...
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: 'Environment to deploy to'
+        type: choice
+        required: true
+        default: 'ops-preview-1'
+        options:
+        - ALL
+        - LIVE
```
michalrus, line 14: If, say, `live-preprod` is split between a few regions, should we also have a `LIVE-PREPROD` group? Or will they always be rolled out one by one?

.github/workflows/deploy-env.yaml
michalrus: I don't think so either, because `nix develop` drops you into a new bash. If it's not an interactive session (like here), it will just exit immediately with code 0.

If `aws` is available, it means it was previously on `PATH`, probably added there by GitHub as part of standard runner toolbox?

Can you change:

```sh
nix develop
```

to:

```sh
which aws || true
nix develop
which aws
```

Then we'll see if `aws` existed before `nix develop`, and whether the second path is in `/nix/store/…-aws/bin/aws` or provided by GitHub.

In the latter case you can do this (like `std.yml`), for reproducibility:

```sh
nix develop -L --command bash -c '
  set -euo pipefail
  echo "yes" | nix run .#cardano-services.${{ inputs.environment }}.apply
'
```

.github/workflows/deploy-env.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: 🧰 Setup Nix
+        uses: cachix/install-nix-action@v21
+
+      - name: 🚀 Deploy
+        run: |
+          echo "${{ secrets.ENVRC }}" > .envrc.local
+          source .envrc.local
+          nix develop
+          echo "yes" | nix run .#cardano-services.${{ inputs.environment || 'ops-preview-1@us-east-1' }}.apply
```
michalrus, line 51: I don't know how I feel about having this default? Can you say more?

michalrus: Ah, silly me, I only read the code, not your comments:

> I am going to take out the `push:` trigger and the `ops-preview-1@us-east-1` default value in instances of `{{ inputs.environment || 'ops-preview-1@us-east-1' }}`.

ce0la: I already addressed that here: https://github.com/input-output-hk/cardano-js-sdk/pull/1382#issuecomment-2249679228. The default values are to test the workflow. The approach is to supply this value using a `workflow_dispatch` interface which is not currently available for use since this workflow is not on `master`.

I would take out the default values before merge once the PR is approved.

.github/workflows/deploy-env.yaml
ce0la: Yes. Since the envs are in different clusters that act as backup for each other, it is a good idea to roll them out one by one; a disaster-recovery, blue-green deployment style, if you will.

.github/workflows/deploy-env.yaml
ce0la: Sillier me, I had the page already opened down and it did not update with your most recent comment. Kindly ignore my reply. 😅

.github/workflows/deploy-env.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - name: 📥 Checkout repository
+        uses: actions/checkout@v3
+
+      - name: 🧰 Setup Nix
+        uses: cachix/install-nix-action@v21
+
+      - name: 🚀 Deploy
+        run: |
+          echo "${{ secrets.ENVRC }}" > .envrc.local
+          source .envrc.local
```
gytis-ivaskevicius, line 49: I don't think this line is required, `nix develop` sources it in the background

gytis-ivaskevicius (approved):
Looks good, has this been tested?

.github/workflows/deploy-env.yaml
michalrus: Same :)

.github/workflows/deploy-env.yaml
```diff
@@ -0,0 +1,53 @@
...
+          nix develop -L --command bash -c "
```
michalrus, line 50: It works by accident, but not the way you think (I think ^^).

It should be a single quote here, because on L52 you close this double quote right before `yes`, and reopen it right after. Single quotes, OTOH, will pass everything to the bash running inside the devshell.

michalrus (changes requested):
Great! Only 2 small changes 🙏

ce0la:
> Looks good, has this been tested?

Yes, it works.

.github/workflows/deploy-env.yaml
ce0la: I agree. The purpose of using the double quotes here is to test the CI which initially failed on the single quotes due to setting a single-quoted default value for the environment variable here: `${{ inputs.environment || 'ops-preview-1@us-east-1' }}`. Updated now for final reviews.

ce0la: P.S: Escaping (`\`) the single quotes did not help.

michalrus: Wait, but `${{ inputs.environment || 'ops-preview-1@us-east-1' }}` is evaluated as a very first step, it's GitHub's YAML templating. So these single quotes have nothing to do with Bash, or?

michalrus (approved):
Awesome! Thank you :)

.github/workflows/deploy-env.yaml
```diff
@@ -0,0 +1,52 @@
...
+    environment: ${{ inputs.environment }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: 📥 Checkout repository
+        uses: actions/checkout@v3
+
+      - name: 🧰 Setup Nix
+        uses: cachix/install-nix-action@v21
+
+      - name: 🚀 Deploy
```
michalrus, line 46: Wait, but don't we have to build the images first? 🤔 Or at least make sure they're built and pushed to ECR?

gytis-ivaskevicius: I thought about it, it should not have `push` at the top and then it should be fine. We may encounter a race condition once in a while but that's probably not going to be even noticable

ce0la: Yes, I plan to take out `push:` once the PR is tested and cleared for merging.

Ref: https://github.com/input-output-hk/cardano-js-sdk/pull/1382#issuecomment-2249679228

michalrus: Okay, thanks. I remember I [added a check in dapp-store once](https://github.com/input-output-hk/dapps/blob/66875ff0e5a755b5477d913b2224b944620a6af2/.github/workflows/ci-cd.yml#L44-L54), to prevent such a race condition, but they use Git commits to tag images.

PR #1383, opened 2024-07-24
Fix/resubmit error handling

mirceahasegan:
# Context

1. `test:ws` requires configuring `OGMIOS_URL` and `DB_SYNC_CONNECTION_STRING` to the values that `local:network` was started with.
  - align these ports with the ones configured in .nix for deployments
2. Fix unexpected error in resubmit logic: `if` statement tried to access produced.coins for non-ValueNotConserved errors 
  - improve `ProviderError` type guard

# Proposed Solution

### DIFF
--- a/compose/common.yml
+++ b/compose/common.yml
@@ -187,7 +187,7 @@ services:
     healthcheck:
       retries: 2000
     ports:
-      - ${OGMIOS_PORT:-1337}:1337
+      - ${OGMIOS_PORT:-1340}:1337
     restart: on-failure
     volumes:
       - node-ipc:/ipc
@@ -241,7 +241,7 @@ services:
       timeout: 5s
       retries: 10
     ports:
-      - ${POSTGRES_PORT:-5432}:5432
+      - ${POSTGRES_PORT:-5435}:5432
     restart: on-failure
     secrets:
       - postgres_db_db_sync
--- a/packages/cardano-services-client/test/HttpProvider.test.ts
+++ b/packages/cardano-services-client/test/HttpProvider.test.ts
@@ -1,7 +1,7 @@
 /* eslint-disable sonarjs/cognitive-complexity */
 /* eslint-disable sonarjs/no-duplicate-string */
+import { CardanoNodeUtil, Provider, ProviderError, ProviderFailure } from '@cardano-sdk/core';
 import { HttpProviderConfig, createHttpProvider } from '../src';
-import { Provider, ProviderError, ProviderFailure } from '@cardano-sdk/core';
 import { Server } from 'http';
 import { fromSerializableObject, toSerializableObject } from '@cardano-sdk/util';
 import { getPort } from 'get-port-please';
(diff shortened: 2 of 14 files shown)


compose/common.yml
```diff
@@ -187,7 +187,7 @@ services:
     healthcheck:
       retries: 2000
     ports:
-      - ${OGMIOS_PORT:-1337}:1337
+      - ${OGMIOS_PORT:-1340}:1337
```
iccicci, line 190: Yesterday you noticed this and postgres port.

Is this change a refuse to me moved in a distinct PR or did you forgot the postgres port or what else?

Thank you

packages/cardano-services/test/jest-setup/rebuild-test-db.sh
```diff
@@ -10,7 +10,7 @@ DB_DB_SYNC=$(cat $SECRETS_DIR/postgres_db_db_sync)
 USER=$(cat $SECRETS_DIR/postgres_user)
 PASSWORD=$(cat $SECRETS_DIR/postgres_password)
 
-export DB_SYNC_CONNECTION_STRING="postgresql://${USER}:${PASSWORD}@localhost:5435/${DB_DB_SYNC}"
+export DB_SYNC_CONNECTION_STRING="postgresql://${USER}:${PASSWORD}@localhost:5432/${DB_DB_SYNC}"
```
iccicci, line 13: I would avoid to use the standard postgres port; if a developer has another postgres instance it may be source of problems,

packages/e2e/.env.example
```diff
@@ -29,9 +29,9 @@ STAKE_POOL_PROVIDER_PARAMS='{"baseUrl":"http://localhost:4000/"}'
 WS_PROVIDER_URL='ws://localhost:4100/ws'
 
 # Required by test:ogmios, test:blockfrost
-DB_SYNC_CONNECTION_STRING='postgresql://postgres:doNoUseThisSecret!@localhost:5435/cexplorer'
-STAKE_POOL_CONNECTION_STRING='postgresql://postgres:doNoUseThisSecret!@localhost:5435/stake_pool'
-STAKE_POOL_TEST_CONNECTION_STRING='postgresql://postgres:doNoUseThisSecret!@localhost:5435/stake_pool_test'
+DB_SYNC_CONNECTION_STRING='postgresql://postgres:doNoUseThisSecret!@localhost:5432/cexplorer'
+STAKE_POOL_CONNECTION_STRING='postgresql://postgres:doNoUseThisSecret!@localhost:5432/stake_pool'
+STAKE_POOL_TEST_CONNECTION_STRING='postgresql://postgres:doNoUseThisSecret!@localhost:5432/stake_pool_test'
```
iccicci, line 34: Same as above

iccicci (changes requested):

compose/common.yml
mirceahasegan: I did not forget. I changed the default port to 5435, but according to your comments that is not a good idea. I will update it.

mirceahasegan: done: 87faa2f7068fa9426a49798a4ecd8230a311607b

packages/cardano-services/test/jest-setup/rebuild-test-db.sh
mirceahasegan: Done: 87faa2f7068fa9426a49798a4ecd8230a311607b

packages/e2e/.env.example
mirceahasegan: Done: 87faa2f7068fa9426a49798a4ecd8230a311607b

iccicci (approved):
Excellent! 🚀

AngelCastilloB (approved):
LGTM! 🚀

PR #1385, opened 2024-07-24
LW-11028 Increase number of db connections in pool for asset provider

iccicci:
# Context

In some Lace e2e tests the `get-assets` call was failing and in the logs were present lines as:
```
error: sorry, too many clients already
```

# Proposed Solution

Increased the number of the connections in the pools from 10 to 50.


mkazlauskas (approved):

mirceahasegan (approved):
lgtm 🕵️

gytis-ivaskevicius (commented):
i dont think we can deploy this to production

bernokl:
Perhaps see how it does on load-test this weekend?

Issue #1372, opened 2024-07-18
ESM Support


joacohoyos:
### Summary

Hi Everoyne!

Creating this issue cause at jpg.store we were trying to use  [Blaze](https://github.com/butaneprotocol/blaze-cardano) (which internally uses cardano-js-sdk)and at the same time using the sdk directly for a lot of datum building and other cardano relate stuff and noticed several issues while using the library in ESM mode.
Talking with people at butane, they had issues with ESM too, so it seems its not an problem only for us.

@pocesar ( our js wizard) did an analysis of the codebase and even try fixing it with no luck. Here there is a rough summary of the findings he got and some output he gave us.

- If you try to use this library in a` "type": "module"` (ESM) codebase it will fail because the generated libraries in dist/esm folder aren't valid ESM modules.
- We tried to fix the ESM build step, but it's nearly impossible with code changes, since some files are mixing CJS and ESM dependencies, using` import { } from 'dep'` when it's a CJS and vice-versa. ESM modules doesn't have a require global function like CJS (the same goes to all CJS specific globals, like `require.resolve`,  `__dirname`, `__filename`). 
- Some dependencies got a mismatching `.d.ts` types for ESM and CJS, that makes the build break as well.
- Package manager vendor specific hacks, like `workspace:~` makes it even harder to make changes. 

@pocesar and also @micahkendall and @ezepze from butane recommended actually using a build step through something like esbuild, rollup, tsup instead of relying on correctness from the tsc compiler.

Think this is a major issue to try to address as ESM is starting to get more and more adoption and CJS is starting to almost be unproductive to create new codebases with given a lot of the newer libraries are starting to ship with esm support only.

### Steps to reproduce the bug

https://github.com/joacohoyos/cardano-sdk-esm-issue

Here there is a repo which you already start getting issue with if you try to run it.

It start with libsodium issues, but even if you fix those you then start getting different issues with different pieces of the library.

According to @pocesar another  quick way to see the issue is doing a  `--target ESNext --module ESNext --moduleResolution Bundler` inside each workspace module. 

### Actual Result

Getting multiple issues when using with ESM

### Expected Result

Library works both with CJS and ESM environemntes

### SDK version

^0.35.2

### Environment type

- [X] Node.js
- [ ] Browser
- [ ] Browser Extension
- [ ] Other

### Environment details

Node: 20.9.0

mkazlauskas:
[...]

PR #1387, opened 2024-07-25
fix: esm build now works correctly when imported in modules

AngelCastilloB:
# Context

If you try to use this library in a "type": "module" (ESM) codebase it will fail because the generated libraries in dist/esm folder aren't valid ESM modules.

This is becuase libsodium was being imported incorrectly, and we are using a package which is not a valid ESM module.

See

(https://github.com/input-output-hk/cardano-js-sdk/issues/1372)

# Proposed Solution

Fix how libsodium is imported and update ip-address package (latest version is a valid ESM module)


mkazlauskas (approved):
Nice work! :detective: :muscle:

mirceahasegan (approved):
Great detective work @AngelCastilloB 🥇

rvcas:
thank you so much

PR #1389, opened 2024-07-26
fix(wallet): tx not withdrawing all rewards

mirceahasegan:
# Context

simple-delegation-rewards e2e test started failing due to transaction not withdrawing all rewards. Issue was exposed after the websocket networkInfo provider was implemented and activated for the e2e tests, making responses instant (as opposed to slower HTTP requests). Wallet syncStatus updates were debounced causing the syncStatus not to update in case of many bursting requests, effectively causing the TxBuilder to consider the providers synced.

# Proposed Solution
updates the pending requests observable to emit any pending requests updates immediately, while debouncing only non-pending updates


iccicci (approved):
A W E S O M E ! ! ! 🚀 💪

mkazlauskas (approved):
Great work @mirceahasegan ! :muscle:

PR #1390, opened 2024-07-26
fix(cardano-services): use a single QueryRunner/connection for getAssets

mkazlauskas:
# Context

TypeormAssetProvider creates a query runner per asset in 'getAssets' call. It can be up to 100 assets per call, which creates a lot of query runners.

LW-11038

# Proposed Solution

This fix makes it use a single query runner.


mirceahasegan (approved):
Great work @mkazlauskas ! 🚀

PR #1391, opened 2024-07-26
fix(web-extension): do not include staking script if not required in the tx

mkazlauskas:
# Context

We're always including both staking and payment scripts, even if staking witness is not required

# Proposed Solution


iccicci (approved):
LGTM

AngelCastilloB (approved):

PR #1392, opened 2024-07-26
fix: remove redundant, uninitialized logger from TypeormService

will-break-it:
# Context

# Proposed Solution


iccicci (approved):
Usually we use context for commit messages of changes impacting only one package.
More, we use `fix` for commits which actually fix a bug; removing unused code is more a `refactor`...

`refactor(cardano-services): remove ...`

mkazlauskas (approved):
Thanks @will-break-it !

PR #1398, opened 2024-07-30
LW-11029 Use CBOR for projector

iccicci:
# Context

TBD

# Proposed Solution

TBD

TBD


packages/ogmios/src/ogmiosToCore/tx.ts
```diff
@@ -166,7 +167,7 @@ const mapCertificate = (certificate: Schema.Certificate): Cardano.Certificate =>
   //           : Cardano.MirCertificatePot.Treasury,
   //       quantity: certificate.moveInstantaneousRewards.value || 0n
   //       // TODO: update MIR certificate type to support 'rewards' (multiple reward acc map to coins)
-  //       // This is currently not compatible with core type (missing 'rewardAccount' which doesnt exist in ogmios)
+  //       // This is currently not compatible with core type (missing 'rewardAccount' which doesn't exist in ogmios)
```
rhyslbw, line 170: 👍

rhyslbw (changes requested):
I'm missing context on the need for https://github.com/input-output-hk/cardano-js-sdk/pull/1398/commits/7a0a5b5a1ba6d7a107a3346bafdb3749bb148cfe. It's best to embed a description in the commit message

packages/core/src/Asset/cip67.ts
```diff
@@ -1,4 +1,4 @@
-import * as Cardano from '../Cardano';
+import { AssetName } from '../Cardano';
```
mkazlauskas, line 1: Even better, as importing from index might still result in circular imports
```suggestion
import { AssetName } from '../Cardano/X';
```

iccicci: 6d080359ace should solve this

PR #1399, opened 2024-07-30
LW-11106 Add distinction between committe with or without script

iccicci:
# Context

Now https://github.com/input-output-hk/cardano-db-sync/issues/1571 is solved we need to add the same distinction in SDK.

# Proposed Solution

Changed the query accordingly.


mirceahasegan (approved):
Great work @iccicci 💪

mkazlauskas (approved):

PR #1401, opened 2024-07-31
fix(cardano-services): drep_hash can be null when delegating to abstain or no confidence

mirceahasegan:
# Context

LW-11110

DbSyncChainHistoryProvider will crash if it encounters an AlwaysAbstain or AlwaysNoConfidence vote delegation certificate.

# Proposed Solution
If drep_hash is null, check the value of `drep_view` which will indicate in which of the 2 scenarios we are in.

### DIFF
--- a/.github/workflows/std.yml
+++ b/.github/workflows/std.yml
@@ -112,7 +112,7 @@ jobs:
               echo '{"environment":"dev-preview", "target":"dev-preview@us-east-1",    "url": "https://dev-preview.lw.iog.io/"}'
             fi
             if [ "true" == ${{ inputs.deploy-dev-sanchonet || (github.event_name == 'push' && github.ref_name == 'conway-era') }} ] ; then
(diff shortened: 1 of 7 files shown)


iccicci (approved):
Very nice! Also I'm learning about delegating to always abstain and always no confidence right now...
It would be nice to add e2e tests as well in Conway transaction test; next to `'can delegate vote'` we could add two new test cases to check these two delegations.
Basically we could just replicate the test three times and changing only the vote delegation certificate.

mirceahasegan:
Added e2e test here: https://github.com/input-output-hk/cardano-js-sdk/pull/1401/commits/39bc55edd1fea1aabbc914fdd2db5085d4c00f79#diff-c4631d405c11d68c799be04111a87c91ab41c34b43a73f83394832b89cc9324fL412

iccicci (approved):
Exceptional! Well done @mirceahasegan !

iccicci (approved):

gytis-ivaskevicius (approved):

PR #1403, opened 2024-08-01
Refactor/small optimizations

iccicci:
# Context

- I noticed that in the `gov_action_proposal` table there is a data repetition.
- I noticed an optimization introduced with some tasks in the past was lost (probably due to some merges).

# Proposed Solution

- Reduced by one filed the query.
- Re-applied the lost optimization.


mirceahasegan (approved):
Nice work 🚀

PR #1404, opened 2024-08-01
LW-11136 Add circular deps check and fix all revealed issues

iccicci:
# Context

`core` package doesn't have the `circular-deps:check`! 😨 

# Proposed Solution

Added it and fixed all revealed issues.


mkazlauskas:
Closing in favor of #1406

PR #1405, opened 2024-08-02
feat!: add a new function to generate and track unused addresses in ObservableWallets

AngelCastilloB:
# Context

Our CIP-30 `getUnusedAddresses` implementation currently always returns an empty array. Particular DApps require wallets to use a freshly created address, as such we must enable this feature in our CIP-30 implementation.

# Proposed Solution

BaseWallet now implements a new `getNextUnusedAddress` method, which tries to find and track the next unused wallet, it does this by first listing all known addresses, picking the address with the highest derivation index and checking whether this is an empty address or not, if it is an unused address, it returns this address, otherwise, it generates a new unused address, adds it to the tracked addresses and returns it. This guarantees that we are always tracking our next unused address if it is ever requested.

- BaseWallet implementes a new  `getNextUnusedAddress` method that returns the next empty address.
- CIP-30 `getUnusedAddresses` now return an array with the next unused address.

### DIFF
--- a/packages/wallet/src/Wallets/BaseWallet.ts
+++ b/packages/wallet/src/Wallets/BaseWallet.ts
@@ -41,6 +41,7 @@ import {
   distinctBlock,
   distinctEraSummaries
 } from '../services';
+import { AddressType, Bip32Account, GroupedAddress, WitnessedTx, Witnesser, util } from '@cardano-sdk/key-management';
 import {
   AssetProvider,
   Cardano,
@@ -64,6 +65,7 @@ import {
   SignDataProps,
   SyncStatus,
   UpdateWitnessProps,
+  WalletAddress,
   WalletNetworkInfoProvider
 } from '../types';
 import { BehaviorObservable, TrackerSubject, coldObservableProvider } from '@cardano-sdk/util-rxjs';
@@ -86,7 +88,6 @@ import {
   tap,
   throwError
 } from 'rxjs';
-import { Bip32Account, GroupedAddress, WitnessedTx, Witnesser, util } from '@cardano-sdk/key-management';
 import { ChangeAddressResolver, InputSelector, roundRobinRandomImprove } from '@cardano-sdk/input-selection';
 import { Cip30DataSignature } from '@cardano-sdk/dapp-connector';
 import { Ed25519PublicKey, Ed25519PublicKeyHex } from '@cardano-sdk/crypto';
@@ -141,6 +142,27 @@ export const isScriptPublicCredentialsManager = (
 export const isBip32PublicCredentialsManager = (
   credManager: PublicCredentialsManager
 ): credManager is Bip32PublicCredentialsManager => !isScriptPublicCredentialsManager(credManager);
+
+/**
+ * Gets whether the given address has a transaction history.
+ *
+ * @param address The address to query.
+ * @param chainHistoryProvider The chain history provider where to fetch the history from.
+ */
+const addressHasTx = async (
+  address: Cardano.PaymentAddress,
+  chainHistoryProvider: ChainHistoryProvider
+): Promise<boolean> => {
+  const txs = await chainHistoryProvider.transactionsByAddresses({
+    addresses: [address],
+    pagination: {
+      limit: 1,
+      startAt: 0
+    }
+  });
+
+  return txs.totalResultCount > 0;
+};
 
 export interface BaseWalletDependencies {
   readonly witnesser: Witnesser;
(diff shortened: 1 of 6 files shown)


packages/wallet/src/Wallets/BaseWallet.ts
```diff
@@ -821,6 +843,41 @@ export class BaseWallet implements ObservableWallet {
     return firstValueFrom(this.addresses$);
   }
 
+  async getNextUnusedAddress(): Promise<WalletAddress | null> {
```
mkazlauskas, line 846: Should we derive 1 per stake key? :thinking: 
```suggestion
  async getNextUnusedAddresses(): Promise<WalletAddress[]> {
```

mkazlauskas (changes requested):

packages/wallet/src/Wallets/BaseWallet.ts
rhyslbw: Good point @mkazlauskas. This is probably the most appropriate thing to do but, with our current data API and fetching strategy to catch rollbacks, we need to be careful not to place too much additional load on the DB. We can always extend this with an improved API but being cautious for now seems appropriate.

packages/wallet/src/types.ts
```diff
@@ -130,6 +130,14 @@ export interface ObservableWallet {
    */
   discoverAddresses(): Promise<WalletAddress[]>;
 
+  /**
+   * Get the next unused address for the wallet.
+   *
+   * @returns Promise that resolves with the next unused address. Return null if there
+   * are no available unused addresses (I.E Single address wallets such as script wallets which already used up their only address).
+   */
+  getNextUnusedAddress(): Promise<WalletAddress | null>;
```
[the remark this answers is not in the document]
AngelCastilloB: As discussed, replaced the function in the interface with a cold observable:

done here https://github.com/input-output-hk/cardano-js-sdk/pull/1405/commits/3aaa516e3d8ab3cc43d5ce988f49da3f85107884

packages/wallet/src/Wallets/BaseWallet.ts
AngelCastilloB: Per our multi-delegation convention, secondary stake keys should only ever be paired with the first payment key, only the first stake key (index 0) is allowed to have more than one payment credential.

packages/web-extension/src/observableWallet/util.ts
```diff
@@ -156,6 +156,7 @@ export const observableWalletProperties: RemoteApiProperties<ObservableWallet> =
     },
     rollback$: RemoteApiPropertyType.HotObservable
   },
+  unusedAddresses$: RemoteApiPropertyType.HotObservable,
```
AngelCastilloB, line 159: @mkazlauskas unusedAddresses$ is actually a cold observable, but looking at the code, tagging this as RemoteApiPropertyType.HotObservable will also work, let me know if there is any extra consideration here.

mkazlauskas: Good point, indeed it's not going to work as expected, as messaging implementation [subscribes to underlying observable right away](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/web-extension/src/messaging/remoteApi.ts#L366-L367). So we have to either implement a new `RemoteApiPropertyType.ColdObservable`, or just go with a `MethodReturningPromise`. Happy to go either way since it's not trivial. Didn't realize that this would be an issue when I suggested to use an observable.

AngelCastilloB: Ok, if both options are equal, I just reverted the change of the observable (fixup commit dropped)

packages/wallet/src/Wallets/BaseWallet.ts
mkazlauskas, line 846: Based on 
> As for the second change request, I tought we commit to only having one payment crendetial (index 0) asociated with other stake keys, only stake key allowed to have more than one payment credential is the first one (the discovery assumes this)

There would be no use in the future for more than 1 unused address, right? https://github.com/input-output-hk/cardano-js-sdk/pull/1405#discussion_r1701347073
```suggestion
  async getNextUnusedAddresses(): Promise<WalletAddress[]> {
```

mkazlauskas (approved):
Great work! :rocket:

packages/wallet/src/Wallets/BaseWallet.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1405/commits/0b348d9124411f211994c69acc698a11c1468782

rhyslbw (approved):
Awesome work @AngelCastilloB 🦾

mkazlauskas (approved):

PR #1406, opened 2024-08-02
refactor: resolve circular references in core package

mkazlauskas:
# Context

Same as #1404

# Proposed Solution

add missing 'circular-deps:check' script

there are still some circular imports in Address and Serialization namespaces (see --exclude in package.json script). Resolving those would be a big refactor. They should not cause any issues at runtime, because there is no circular usage at the time of import

BREAKING CHANGES:
- remove Cardano.TransactionId.fromTxBodyCbor (replaced with Serialization.TransactionBody.hash)
- hoist getAssetNameAsText util to Asset.util namespace
- hoist TxCBOR and TxBodyCBOR under Serialization namespace


packages/core/package.json
```diff
@@ -23,6 +23,7 @@
   "scripts": {
     "build": "yarn build:cjs && tsc -p src/tsconfig.json --outDir ./dist/esm --module es2020 && tsc-alias -p src/tsconfig.json --outDir ./dist/esm && cp ../../build/esm-package.json ./dist/esm/package.json",
     "build:cjs": "tsc --build src && cp ../../build/cjs-package.json ./dist/cjs/package.json",
+    "circular-deps:check": "madge --circular dist/cjs --exclude \"/(?:Address)|(?:Serialization/(?:PlutusData)|(?:Certificates)|(?:Scripts)|(?:AuxiliaryData))/*\"",
```
iccicci, line 26: Why excluding these paths?
Or in other words: why allowing circular dependencies here?

mirceahasegan (approved):
Nice work, @mkazlauskas .
Is there a followup ticket for the excluded namespaces.

will-break-it (approved):
Looks good! 👍

PR #1409, opened 2024-08-05
fix(hardware-ledger): delegate vote drep is a hash not key path

mirceahasegan:
# Context
Self drep `Ledger.DRepParamsType.KEY_PATH` was used when mapping VoteDelegation certificate.
That is wrong, we have to pass the drep key hash, otherwise the ledger device will always calculate the DRep as our own, instead of the target one.

# Proposed Solution
VoteDelegation certificate should map drep to `DRepParamsType.KEY_HASH`.

### DIFF
--- a/packages/hardware-ledger/src/transformers/certificates.ts
+++ b/packages/hardware-ledger/src/transformers/certificates.ts
@@ -45,22 +45,16 @@ const credentialMapper = (
 
 const drepParamsMapper = (
   drep: Cardano.Credential,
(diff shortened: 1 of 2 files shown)


mkazlauskas (approved):
Nice catch! Is it good in `hardware-trezor`?

ginnun (approved):
lgtm

mirceahasegan:
> Nice catch! Is it good in hardware-trezor?

We don't have conway era support in Trezor.

PR #1411, opened 2024-08-07
feat!: replace updateWitness with addSignatures in observable wallet

AngelCastilloB:
# Context

The base wallet currently have a updateWitness method that adds signatures to a given transaction, however, this method does not preserve the original CBOR, we must modify this method to correctly preserve CBOR of the transaction.

# Proposed Solution

Modify updateWitness to take the CBOR of the transaction, add signatures to it, and return the transaction with the new signatures encoded as CBOR.

### DIFF
--- a/packages/wallet/src/Wallets/BaseWallet.ts
+++ b/packages/wallet/src/Wallets/BaseWallet.ts
@@ -1,6 +1,17 @@
 /* eslint-disable unicorn/no-nested-ternary */
 // eslint-disable-next-line import/no-extraneous-dependencies
 import {
+  AddSignaturesProps,
+  Assets,
+  FinalizeTxProps,
+  HandleInfo,
+  ObservableWallet,
+  SignDataProps,
+  SyncStatus,
+  WalletAddress,
+  WalletNetworkInfoProvider
+} from '../types';
+import {
   AddressDiscovery,
   AddressTracker,
   BalanceTracker,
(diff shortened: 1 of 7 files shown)


packages/wallet/src/Wallets/BaseWallet.ts
```diff
@@ -878,15 +878,33 @@ export class BaseWallet implements ObservableWallet {
...
-      witness: tx.witness
+      tx: { body, hash },
+      witness
     });
+
+    const coreWitness = witnessedTx.witness;
+    const witnessSet = serializableTx.witnessSet();
+
+    witnessSet.setVkeys(
+      Serialization.CborSet.fromCore([...coreWitness.signatures], Serialization.VkeyWitness.fromCore)
```
mkazlauskas, line 903: Is this preserving original signatures? :thinking: 
```suggestion
    witnessSet.setVkeys(
      Serialization.CborSet.fromCore([...coreWitness.signatures, ...witnessSet.vkeys().toCore()], Serialization.VkeyWitness.fromCore)
```

packages/wallet/src/Wallets/BaseWallet.ts
```diff
@@ -878,15 +878,33 @@ export class BaseWallet implements ObservableWallet {
     return isEmpty ? [knownAddresses[0]] : [];
   }
 
-  /** Update the witness of a transaction with witness provided by this wallet */
-  async updateWitness({ tx, sender }: UpdateWitnessProps): Promise<Cardano.Tx> {
-    return this.finalizeTx({
-      auxiliaryData: tx.auxiliaryData,
+  async addSignatures({ tx, sender }: AddSignaturesProps): Promise<Serialization.TxCBOR> {
```
mkazlauskas, line 881: We should add some test(s) in `methods.test.ts`

mkazlauskas (changes requested):

packages/wallet/src/Wallets/BaseWallet.ts
AngelCastilloB: Yes finalize preserves the original signatures, so coreWitness already has all the signatures

packages/wallet/src/Wallets/BaseWallet.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1411/commits/892c1e9ae82302eb4f07cfc033c3867eaee11c5a

mkazlauskas (approved):
Nice work! :rocket:

PR #1396, opened 2024-07-30
Blockfrost

ginnun:
# Context

This PR aims to add Blockfrost providers to the codebase by reverting a 2-year-old commit.
File content modifications can easily be reviewed in [this commit](https://github.com/input-output-hk/cardano-js-sdk/pull/1396/commits/1b46fc3d8398faf0213ae77dd7891835e40cdb49).

# Proposed Solution

- Revert commit e26037f6
- Rename folder `blockfrost` to `blockfrost-client` and replicate `cardano-services-client` folder structure
- Make it compile and tests pass

# Important Changes Introduced

- Instead of `"@blockfrost/blockfrost-js": "npm:blockfrost-js-ratelimited@2.0.4"`, now using using `"@blockfrost/blockfrost-js": "^5.5.0"` since it supports rate limiting.
- Not every provider has blockfrost-equivalent implementation at the moment.

PR #1412, opened 2024-08-09
LW-11077 perf tests asset provider server

mirceahasegan:
# Context

TypeormAssetProvider does not use any in-memory cache.
We must benchmark it’s performance and possibly optimize before we deploy it to live-mainnet.

# Proposed Solution

Prepare a large sample of mainnet asset IDs (e.g. 100k asset ids)
Some of them must have nft metadata (at least 30% of assets)
Some of them must have off-chain token metadata in the registry. Currently only 350 are included in the sample.
Each run queries (/get-assets) some # of assets from a big sample of possible asset IDs.
The number of each type of asset to include in the query is configurable.
Multiple test repetitions are possible


packages/e2e/test/k6/endpoints/asset/get-assets.test.js
```diff
@@ -1,278 +1,155 @@
...
+ *   - Threshold for FAIL responses not exceeded (1% for example)
+ *   - Threshold for response having the wrong number of assets compared to requested is not exceeded (1% for example)
+ *   - 95th percentile of response time is less than 500ms
+ */
+
+const ASSETS_NO_METADATA_PER_REQUEST = Number.parseInt(__ENV.ASSETS_NO_METADATA_PER_REQUEST || 1);
+const ASSETS_ON_CHAIN_METADATA_PER_REQUEST = Number.parseInt(__ENV.ASSETS_ON_CHAIN_METADATA_PER_REQUEST || 0);
+const ASSETS_OFF_CHAIN_METADATA_PER_REQUEST = Number.parseInt(__ENV.ASSETS_OFF_CHAIN_METADATA_PER_REQUEST || 0);
+const REQUESTS_COUNT = Number.parseInt(__ENV.REQUESTS_COUNT || 1);
+const TARGET_ENV = __ENV.TARGET_ENV || 'dev'; // only dev and live are supported
```
ginnun, line 25: It would be nice to prevent/fail trying to run unsupported envs and networks.

ginnun (approved):
Despite the failing tests, LGTM. I left a small comment.

packages/e2e/test/k6/endpoints/asset/get-assets.test.js
mirceahasegan: I've added the validation in 9e97339ec606635f256b581195b6a2fe45f08782 @ginnun

ginnun (approved):
Great work here!
I am planning to use this for testing the [new BlockfrostAsserProvider implementation](https://github.com/input-output-hk/cardano-js-sdk/pull/1396),

.github/workflows/k6-perf-test-cloud-assets.yaml
```diff
@@ -0,0 +1,55 @@
...
+        description: 'Number of assets ids with off-chain metadata to include in each request.'
+        type: number
+        required: true
+        default: 1
+        
+env:
+  # Only running on dev-mainnet for now
+  TARGET_ENV: 'dev'
+  TARGET_NET: 'mainnet'
+  REQUESTS_COUNT: ${{ inputs.requests-count || 10 }}
```
iccicci, line 30: Which is the purpose of `|| 10` when it is specified as default value for given input?

iccicci (approved):
Excellent! 🚀

.github/workflows/k6-perf-test-cloud-assets.yaml
mirceahasegan: In this case, where the only way to run the test is via the "Run action" button, there isn't much use for it.
But when we'll add a scheduled run (cron), the inputs and their defaults will be ignored, so this will ensure it's initialized.

iccicci: Nice to know: thank you!

I thought when run on schedule, default values were taken.

PR #1416, opened 2024-08-12
fix(ogmios): ignore byron signatures in ogmios mapper

mkazlauskas:
# Context

Mainnet projection crashes with `InvalidStringError: Invalid string: "expected length '64', got 128"` because it encounters a Byron tx with signatures map that have different length key (128 instead of the expected 64)

Our signatures key validation only supports praos txes

# Proposed Solution

We are not currently using signatures at all. Safe to ignore for the time being.


rhyslbw (approved):

michalrus (approved):
Thanks a lot, it got past 4% on Mainnet, and the error was somewhere around 0.2% :)

PR #1425, opened 2024-08-19
fix(wallet): cip30 signTx function now throws if not enough signatures before calling confirmationCallback

AngelCastilloB:
# Context

CIP-30 signTx was was throwing `DataSignErrorCode.ProofGeneration` when signTx was called with partialSign false and the transaction could not be signed entirely. This exception however was thrown after the confirmationCallback was invoked, which caused side effects on the client side (I.E Lace DappConnector) even when the transaction could not be signed. 

# Proposed Solution

Made the check and throw before confirmationCallback is invoked, so no side effects are triggered on the client side if the transaction cant be signed.

# Important changes

CIP-30 getUsedAddresses no longer include unused addresses


ginnun (dismissed):

rhyslbw (approved):

mkazlauskas (approved):

mirceahasegan (approved):
lgtm!

iccicci (approved):
LGTM!

PR #1426, opened 2024-08-20
fix(wallet): cip30 getUsedAddresses now returns an empty array if no used addresses are found

AngelCastilloB:
# Context

cip30 getUsedAddresses was throwing when no used addresses were found

# Proposed Solution

cip30 getUsedAddresses now returns an empty array instead of throwing

### DIFF
--- a/packages/wallet/src/cip30.ts
+++ b/packages/wallet/src/cip30.ts
@@ -401,11 +401,7 @@ const baseCip30WalletApi = (
     const unusedAddresses = await wallet.getNextUnusedAddress();
     const addresses = trackedAddresses.filter((address) => !unusedAddresses.includes(address));
 
-    if (addresses.length === 0) {
-      throw new ApiError(APIErrorCode.InternalError, 'could not get used addresses');
-    } else {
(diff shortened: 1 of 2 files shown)


packages/wallet/src/cip30.ts
```diff
@@ -401,11 +401,7 @@ const baseCip30WalletApi = (
     const unusedAddresses = await wallet.getNextUnusedAddress();
     const addresses = trackedAddresses.filter((address) => !unusedAddresses.includes(address));
 
-    if (addresses.length === 0) {
```
mkazlauskas, line 404: A unit test would be nice

mkazlauskas (changes requested):

mirceahasegan (approved):
lgtm

packages/wallet/src/cip30.ts
AngelCastilloB: done and squashed @mkazlauskas (since its a tiny PR)

mirceahasegan (approved):

mkazlauskas (approved):
Nice work! :rocket:

PR #1427, opened 2024-08-20
Feature: CIP20 helper functions + StringUtils

rhyslbw:
# Context
When fixing https://github.com/input-output-hk/lace/issues/978 I discovered there's more we can be doing in the SDK to help applications produce compliant CIP-20 transaction metadata. I've fixed the immediate issue here https://github.com/input-output-hk/lace/pull/1379, but there were no tests and I would rather put energy into removing the responsibility from Lace altogether.

https://github.com/input-output-hk/lace/blob/e877d4668c02ae647ed39293f05233d0ac513663/packages/cardano/src/wallet/lib/get-auxiliary-data.ts#L1-L74

N.B. This implementation is based on https://cips.cardano.org/cip/CIP-20 which defines a format to support multi-entry messages, whereas Lace currently only has good support for single entry (with a split to handle > 64 bytes). To adopt this solution we will be considering changing to a multi-line input field.

# Proposed Solution
See commit messages

See commits

### DIFF
--- a/packages/core/src/TxMetadata/cip20.ts
+++ b/packages/core/src/TxMetadata/cip20.ts
@@ -1,0 +1,65 @@
+import { CustomError } from 'ts-custom-error';
+import { StringUtils } from '@cardano-sdk/util';
+import { TxMetadata } from '../Cardano';
+
+export const CIP_20_METADATUM_LABEL = 674n;
+const MAX_BYTES = 64;
+
+export enum MessageValidationFailure {
+  wrongType = 'WrongType',
+  oversize = 'Oversize'
+}
+
+export type MessageValidationResult = {
+  valid: boolean;
+  failure?: MessageValidationFailure;
+};
+
+export type CIP20TxMetadataMessage = string;
+
+export type CIP20TxMetadataArgs = {
+  // An array of message strings, limited to 64 bytes each
+  messages: CIP20TxMetadataMessage[];
+};
+
+export type ValidationResultMap = Map<CIP20TxMetadataMessage, MessageValidationResult>;
+
+export class MessageValidationError extends CustomError {
+  public constructor(failures: ValidationResultMap) {
+    const m = [failures.entries()].map(
+      ([[message, result]]) => `${result.failure}: ${message.slice(0, Math.max(0, MAX_BYTES + 1))}`
+    );
+    super(`The provided message array contains validation errors | ${m}`);
+  }
+}
+
+/**
+ * Validate each message for correct type, in the case of JavaScript, and size constraint
+ *
+ * @param entry unknown
+ * @returns Validation result
+ */
+export const validateMessage = (entry: unknown): MessageValidationResult => {
+  if (typeof entry !== 'string') return { failure: MessageValidationFailure.wrongType, valid: false };
+  if (StringUtils.byteSize(entry) > MAX_BYTES) return { failure: MessageValidationFailure.oversize, valid: false };
+  return { valid: true };
+};
+
+/**
+ * Converts an object containing an array of individual messages into https://cips.cardano.org/cip/CIP-20 compliant
+ * transaction metadata
+ *
+ * @param args CIP20TxMetadataArgs or a string to be transformed into an array
+ * @returns CIP20-compliant transaction metadata
+ * @throws Message validation error containing details. Use validateMessage to independently check each message before calling this function
+ */
+export const toCIP20Metadata = (args: CIP20TxMetadataArgs | string): TxMetadata => {
+  const messages = typeof args === 'string' ? StringUtils.chunkByBytes(args, MAX_BYTES) : args.messages;
+  const invalidMessages: ValidationResultMap = new Map();
+  for (const message of messages) {
+    const result = validateMessage(message);
+    if (!result.valid) invalidMessages.set(message, result);
+  }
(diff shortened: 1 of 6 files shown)


mirceahasegan (dismissed):
Great work @rhyslbw 👏

AngelCastilloB (dismissed):
Great job @rhyslbw, love the comments ❤️

packages/core/src/TxMetadata/cip20.ts
```diff
@@ -0,0 +1,56 @@
...
+export type MessageValidationResult = {
+  valid: boolean;
+  failure?: MessageValidationFailure;
+};
+
+export type CIP20TxMetadataMessage = string;
+
+export type CIP20TxMetadataArgs = {
+  // An array of message strings, limited to 64 bytes each
+  messages: CIP20TxMetadataMessage[];
```
mkazlauskas, line 20: **(not a change req)** A higher level util that splits a single potentially long message into an array of 64 byte lines would be nice too

packages/core/src/TxMetadata/cip20.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @returns Validation result
+ */
+export const validateMessage = (entry: unknown): MessageValidationResult => {
+  if (typeof entry !== 'string') return { failure: MessageValidationFailure.wrongType, valid: false };
+  if (StringUtils.byteSize(entry) > 64) return { failure: MessageValidationFailure.oversize, valid: false };
+  return { valid: true };
+};
+
+export const isTxMetadata = (
+  value: TxMetadata | Map<CIP20TxMetadataMessage, MessageValidationResult>
```
mkazlauskas, line 38: **(not a change req)** a common Result type would be nice to have in the SDK for consistency. There are more places where this would be useful
```suggestion
  value: Result<TxMetadata, Map<CIP20TxMetadataMessage, MessageValidationResult>>
```

mkazlauskas (approved):
Great work!

packages/core/src/TxMetadata/cip20.ts
mkazlauskas: Actually maybe this is a change req :sweat_smile: Right now it will be difficult to differentiate between error and valid result, as both can be a `Map`

packages/core/src/TxMetadata/cip20.ts
```diff
@@ -0,0 +1,56 @@
...
+): value is TxMetadata => value.has(CIP_20_METADATUM_LABEL as never);
+
+/**
+ * Converts an object containing an array of individual messages into https://cips.cardano.org/cip/CIP-20 compliant
+ * transaction metadata
+ *
+ * @param args CIP20TxMetadataArgs
+ * @returns Transaction metadata or validation failures
+ */
+export const toCIP20Metadata = (args: CIP20TxMetadataArgs): TxMetadata | ValidationFailureMap => {
```
mkazlauskas, line 48: This is a usability issue, as both the result and the error are a Map - difficult to differentiate between the 2
```suggestion
export const toCIP20Metadata = (args: CIP20TxMetadataArgs): Result<TxMetadata, ValidationFailureMap> => {
```

mkazlauskas (changes requested):

packages/core/src/TxMetadata/cip20.ts
rhyslbw: Good point. That would cover Lace's current needs so we can replace the untested code with this

packages/core/src/TxMetadata/cip20.ts
rhyslbw: Alternative approach in https://github.com/input-output-hk/cardano-js-sdk/pull/1427/commits/af0c3f3a102259726ea5f541d3e57d6d6272c35c

packages/core/src/TxMetadata/cip20.ts
rhyslbw: 9e729201b5b77a2b3ecb81d832f6b636eb4f5bf4 + 8e217413d4967115d02f4075c936d43d56d2d858

packages/util/src/string.ts
```diff
@@ -0,0 +1,32 @@
+export const StringUtils = {
+  byteSize: (str: string) => new Blob([str]).size,
+  /**
+   * convert a string to an array of strings, items limited by the specified bytes
+   *
+   * @param str The string to slice up
+   * @param maxBytes number of bytes to impose as a limit
+   * @returns A sliced version of the str input
+   */
+  sliceByBytes: (str: string, maxBytes: number): string[] => {
```
mkazlauskas, line 10: Naming suggestion: Lodash has a similar function and calls it '[chunk](https://lodash.com/docs/4.17.15#chunk)'
```suggestion
  chunkByBytes: (str: string, maxBytes: number): string[] => {
```

mkazlauskas (approved):
Great work!

AngelCastilloB (approved):

packages/util/src/string.ts
rhyslbw: Thanks! https://github.com/input-output-hk/cardano-js-sdk/pull/1427/commits/439491317951a2936c9c777b4eb87efda5abe228

mkazlauskas (approved):

PR #1429, opened 2024-08-21
Bug: support overflowing pointer address slots with constraints

will-break-it:
# Context

There are invalid output addresses on mainnet, referencing stake credentials by invalid stake pointer address. These references consist in part of a absolute slot number, which is a location on the blockchain of the stake key registration certificate for that key. The issue arose during indexing when we've found large slot numbers that overflow the JS `Number.MAX_SAFE_INTEGER`.

The mainnet transaction hash, whose first transaction output that caused this issue is: `9a47e091709e4fe1a8ba1291bc8cfd0c0ed0bb9220688b3ccfad5f2d961e8d8e`

# Proposed Solution

We change the variable length encoding/ decoding for pointer addresses to use `bigint` for `Cardano.Pointer.slot` type to support correct serialization for pointer addresses.

We have exclusively changed the `Cardano.Pointer.slot` type to `bigint` and not the `Core/Cardano/PartialBlockHeader.slot`.

### DIFF
--- a/packages/core/src/Cardano/Address/PointerAddress.ts
+++ b/packages/core/src/Cardano/Address/PointerAddress.ts
@@ -3,7 +3,6 @@ import { Address, AddressProps, AddressType, Credential, CredentialType } from '
 import { Hash28ByteBase16 } from '@cardano-sdk/crypto';
 import { InvalidArgumentError, OpaqueNumber } from '@cardano-sdk/util';
 import { NetworkId } from '../ChainId';
-import { Slot } from '../types/Block';
 
 /**
  * Encodes the specified value with a variable number of bytes.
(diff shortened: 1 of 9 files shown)


iccicci (dismissed):
LGTM!

mkazlauskas (approved):
Great work! :rocket:

packages/core/src/Cardano/Address/PointerAddress.ts
```diff
@@ -65,7 +68,7 @@ export const CertIndex = (value: number): CertIndex => value as unknown as CertI
  * a stake key by a location on the blockchain of the stake key registration certificate.
  */
 export type Pointer = {
-  slot: Slot;
+  slot: bigint;
```
iccicci, line 71: Due to my low experience with pointer addresses I wasn't so sure about this change.
Since @mkazlauskas already approved it, I have no reasons to trust him.

iccicci (approved):
Good job!

mkazlauskas (approved):

PR #1438, opened 2024-08-22
fix(cardano-services): correctly load Mapper.withGovernanceActions

iccicci:
# Context

While supporting @will-break-it on LW-10737 we realized about this issue.

# Proposed Solution

Loaded the mapper.

### DIFF
--- a/packages/cardano-services/src/Projection/prepareTypeormProjection.ts
+++ b/packages/cardano-services/src/Projection/prepareTypeormProjection.ts
@@ -99,6 +99,7 @@ const createMapperOperators = (
     withAddresses: Mapper.withAddresses(),
     withCIP67: Mapper.withCIP67(),
     withCertificates: Mapper.withCertificates(),
+    withGovernanceActions: Mapper.withGovernanceActions(),
     withHandleMetadata,
     withHandles,
     withMint: Mapper.withMint(),
@@ -250,6 +251,7 @@ const mapperInterDependencies: Partial<Record<MapperName, MapperName[]>> = {
 const storeMapperDependencies: Partial<Record<StoreName, MapperName[]>> = {
   storeAddresses: ['withAddresses'],
   storeAssets: ['withMint'],
+  storeGovernanceAction: ['withGovernanceActions'],
   storeHandleMetadata: ['withHandleMetadata'],
   storeHandles: ['withHandles'],
   storeNftMetadata: ['withNftMetadata'],


rhyslbw (commented):
Curious about this not being caught in a test originally, and now the fix is being submitted without one. It's a hint things are a little disconnected. Can you justify the reason it's being submitted in isolation? Can @will-break-it not just pick this commit into his branch so it's submitted with more assurance we have a working system?

packages/cardano-services/src/Projection/prepareTypeormProjection.ts
```diff
@@ -92,13 +92,18 @@ const createMapperOperators = (
   const withHandleMetadata = handlePolicyIds
     ? Mapper.withHandleMetadata({ policyIds: handlePolicyIds }, logger)
     : passthrough();
+  const withGovernanceActions = projectionNames.includes(ProjectionName.ProtocolParameters)
+    ? Mapper.withGovernanceActions()
+    : passthrough();
+
```
mkazlauskas, line 98: This shouldn't be needed, as it's included by specifying it in `storeMapperDependencies`
```suggestion
```

packages/cardano-services/src/Projection/prepareTypeormProjection.ts
```diff
@@ -92,13 +92,18 @@ const createMapperOperators = (
...
   return {
     filterMint,
     filterProducedUtxoByAssetsPresence,
     filterUtxo,
     withAddresses: Mapper.withAddresses(),
     withCIP67: Mapper.withCIP67(),
     withCertificates: Mapper.withCertificates(),
+    withGovernanceActions,
```
mkazlauskas, line 106:
```suggestion
    withGovernanceActions: Mapper.withGovernanceActions(),
```

mkazlauskas (changes requested):

packages/cardano-services/src/Projection/prepareTypeormProjection.ts
iccicci: 855a05716f2 should solve this

packages/cardano-services/src/Projection/prepareTypeormProjection.ts
iccicci: 855a05716f2 should solve this

mkazlauskas (approved):

rhyslbw:
OK, I can see this is just impacting code that arguably shouldn't even live in this SDK, and it would take an e2e test to cover.

rhyslbw (approved):

PR #1444, opened 2024-08-26
fix(core): Uri type now supports IPFS CIDv1 format

AngelCastilloB:
# Context

IPFS has multiple [Content Identifiers (CIDs) | IPFS Docs](https://docs.ipfs.tech/concepts/content-addressing/#cid-versions) 

Cardano JS SDK currently only tries to validate suspected IPFS hashes against CIDv0 format

# Proposed Solution

CIDv1 formatted URLs are supported in the SDK


mkazlauskas (approved):
Great work!

iccicci (approved):
LGTM!

PR #1445, opened 2024-08-26
LW-11271 extract k6 utils for SDK http req and deployment under test selection

mirceahasegan:
# Context

As the number of K6 perf tests increases, we need to keep the code DRY and utilities tested.

# Proposed Solution

- Utils for selecting the deployment under test (DUT)
- Utils for sending SDK http requests.

- K6 utilities are part of `utils-dev` package to benefit from TS type checking, cjs build and unit test setup, but they are not exported as they are not meant to be used by any other code than K6 tests.
- Since K6 uses a custom javascript runtime, a direct import from the `dist` folder of `utils-dev` was necessary. Bundling would be another option, but it seems overkill at the moment. For the same reason, all the utils are included in the same .ts file, because K6 is unable to follow relative path imports.


packages/e2e/test/k6/endpoints/asset/get-assets.test.js
```diff
@@ -18,15 +19,17 @@ import http from 'k6/http';
...
+// eslint-disable-next-line no-undef
+const K6_ENV = __ENV;
+
+const ASSETS_NO_METADATA_PER_REQUEST = Number.parseInt(K6_ENV.ASSETS_NO_METADATA_PER_REQUEST || 1);
+const ASSETS_ON_CHAIN_METADATA_PER_REQUEST = Number.parseInt(K6_ENV.ASSETS_ON_CHAIN_METADATA_PER_REQUEST || 0);
+const ASSETS_OFF_CHAIN_METADATA_PER_REQUEST = Number.parseInt(K6_ENV.ASSETS_OFF_CHAIN_METADATA_PER_REQUEST || 0);
+const REQUESTS_COUNT = Number.parseInt(K6_ENV.REQUESTS_COUNT || 1);
+
+const dut = k6Utils.getDut(K6_ENV, { environments: ['dev'], networks: ['mainnet'] });
+console.log(`Domain under test is: ${dut}`);
```
iccicci, line 31: I see this line is repeated somewhere else; if it was a debug line we could remove it, otherwise we could insert it into `getDut`.

mirceahasegan: It is a valid log line, pinpointing the DUT the test ran against. I agree it should be part of `getDut` and I will move it there. Thank you!

mirceahasegan: Done and squashed: https://github.com/input-output-hk/cardano-js-sdk/compare/d21302f4a63641388fd9122455b633db333b344d..d79ead5e6d239cb7db952d154d98ee922d16fbcc

packages/util-dev/src/k6-utils.ts
```diff
@@ -0,0 +1,258 @@
+import { Cardano } from '@cardano-sdk/core';
+import http from 'k6/http';
```
mkazlauskas, line 2: Could you please test if `web-extension` e2e tests still work after this change?

iccicci: If I'm not wrong @mirceahasegan , you did some researches about running this test in a headless mode... but some times have passed.
Could it be a good idea to do a new attempt?
Hopefully some required tools have been upgraded in the mean while removing the blocker you found on last attempt.

iccicci (approved):
Exceptional work @mirceahasegan ! 🚀

packages/util-dev/src/k6-utils.ts
mirceahasegan: I did some research, but hit a dead end due to the web extension part which does not play well in headless mode.
> Could it be a good idea to do a new attempt?
It would be nice, but it's quite low on the priority list.

mirceahasegan: > Could you please test if web-extension e2e tests still work after this change?

They work.

mkazlauskas (approved):
Nice work! :rocket:

PR #1448, opened 2024-08-27
chore: enable walle_api_projector on all dev and staging environments

iccicci:
# Context

Follow up of #1447

# Proposed Solution

Enabled the new projector on all the `dev` and `staging` environments.

### DIFF
--- a/nix/cardano-services/deployments/default.nix
+++ b/nix/cardano-services/deployments/default.nix
@@ -220,9 +220,10 @@ in
           };
 
           projectors = {
+            asset.enabled = true;
             handle.enabled = true;
             stake-pool.enabled = true;
-            asset.enabled = true;
+            wallet-api.enabled = true;
           };
 
           values = {
@@ -272,9 +273,10 @@ in
           };
 
           projectors = {
+            asset.enabled = true;
             handle.enabled = true;
             stake-pool.enabled = true;
-            asset.enabled = true;
+            wallet-api.enabled = true;
           };
 
           values = {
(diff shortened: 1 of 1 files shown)


bernokl (dismissed):

gytis-ivaskevicius (commented):
Why is it enabled on all dev envs? I heard it does not work on mainnet

Also, dev-mainnet is still off-limits since it is used in prod

And if we wish to enable it everywhere, PR with database creation is required on lace-ops

iccicci:
> Why is it enabled on all dev envs?

Let's start thinking about `dev` environments as if they were `dev` environments 😉 
Be devs need the ability to test their job at their will on `dev` environments.

> I heard it does not work on mainnet

Correct, but the target is to test it against `mainnet`... at least till the projection of the block causing the issue.

> Also, dev-mainnet is still off-limits since it is used in prod

We are well aware of this, indeed (as suggested by @rhyslbw) we are going to perform this test on `staging-mainnet` rather then on `dev-mainnet`.
Until next Lace release we are forced to swap the meaning of `staging-mainnet` with `dev-mainnet`; nobody will touch the latter!

> And if we wish to enable it everywhere, PR with database creation is required on lace-ops

I'm learning this right now, please let me know when it is completed.

bernokl:
This PR has been applied to create the relevant db in dev-pre* and staging-preprod/mainnet https://github.com/input-output-hk/lace-ops/pull/178/files

bernokl (approved):

gytis-ivaskevicius (approved):

PR #1455, opened 2024-08-29
feat: dapp connector multi-api support

greatertomi:
# Context

# Proposed Solution


packages/web-extension/src/messaging/runContentScriptMessageProxy.ts
```diff
@@ -12,27 +12,39 @@ export interface MessagingApi {
   callApiMethod(request: MethodRequest): Promise<unknown>;
 }
 
+interface ChannelMessageEvent {
+  channelName: string;
```
mkazlauskas, line 16: Let's rename this everywhere to not confuse with Messenger's channelName
```suggestion
  apiName: ApiName;
```

packages/web-extension/src/cip30/initializeContentScript.ts
```diff
@@ -12,6 +12,7 @@ export const initializeContentScript = (
   { injectedScriptSrc, walletName }: InitializeContentScriptProps,
   dependencies: MessengerDependencies
 ) => {
+  // TODO: How to pass both with the channelName as apis {[channelName]: ApiObject}
```
mkazlauskas, line 15: This function should receive a record of APIs like `Record<ApiName, AnyApi>`

packages/web-extension/src/messaging/runContentScriptMessageProxy.ts
```diff
@@ -12,27 +12,39 @@ export interface MessagingApi {
...
     if (source !== window || !isRequestMessage(data)) return;
     logger.debug('[MessageProxy] from window', data);
-    const apiFunction = apis.find((api: any) => typeof api[data.request.method] === 'function')?.[data.request.method];
+
+    const { channelName, request, messageId } = data;
+    const api = apis[channelName];
+    if (!api) return;
+
+    // const apiFunction = apis.find((api: any) => typeof api[data.request.method] === 'function')?.[data.request.method];
+    const apiFunction = api[request.method];
```
mkazlauskas, line 36: It was probably checking like that for a reason
```suggestion
    const apiFunction = typeof api[request.method] === 'function' ? api[request.method] : null;
```

mkazlauskas (changes requested):

packages/web-extension/src/messaging/runContentScriptMessageProxy.ts
greatertomi: I assume we need to create `ApiName` as an opaque type. Are there any constraints on the type?

mkazlauskas: No constraints, just an opaque string. FIY in xsy-wallet the type is called `Tagged` and we use the type util from type-fest instead of SDK's `OpaqueString`. It's the same thing.

greatertomi: Since OpaqueString is already widely used in the SDK, I stuck with it on 8b69fdc

packages/web-extension/src/cip30/initializeContentScript.ts
greatertomi: Fixed on f401a49

packages/web-extension/src/cip30/initializeInjectedScript.ts
```diff
@@ -9,15 +9,21 @@ export interface InitializeInjectedDependencies {
 }
 
 // tested in e2e tests
-export const initializeInjectedScript = (props: WalletProperties, { logger }: InitializeInjectedDependencies) => {
-  const dependencies: MessengerDependencies = {
-    logger,
-    runtime: injectedRuntime
-  };
+// TODO: Should I create two different initialize Injected script functions? One for single API and another for multi-api
```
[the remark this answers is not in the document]
greatertomi: Fixed on f401a49

packages/dapp-connector/src/ScriptInitializers/types.ts
```diff
@@ -0,0 +1,6 @@
+import { OpaqueString } from '@cardano-sdk/util';
+
+const typedString = (value: string) => value;
+
+export type ApiName = OpaqueString<'ApiName'>;
+export const ApiName = (target: string) => typedString(target);
```
mkazlauskas, line 6: Is this returning an ApiName? :thinking: 
```suggestion
export const ApiName = (target: string) => target as ApiName;
```

packages/dapp-connector/src/ScriptInitializers/initializeContentScript.ts
```diff
@@ -0,0 +1,33 @@
...
+import { MessengerDependencies, cip30, runContentScriptMessageProxy } from '@cardano-sdk/web-extension';
+
+export interface InitializeContentScriptProps {
+  walletName: string;
+  injectedScriptSrc: string;
+}
+
+// tested in e2e tests
+export const initializeContentScript = (
+  props: Record<ApiName, InitializeContentScriptProps>,
```
mkazlauskas, line 11: Right now it will create a cip30 API for each item in props. We need to be able to inject different APIs. Also, right now it will append multiple <script> which is unnecessary, 1 is enough. 
```suggestion
  apis: Record<ApiName, AnyApi>,
```

greatertomi:
Moved [here](https://github.com/input-output-hk/xsy-wallet/pull/202)

PR #1457, opened 2024-09-02
LW-11068 Asset provider OOM fix attempt

iccicci:
# Context

`asset-provider` has memory consumption issues.

# Proposed Solution

Optimized `TypeormAssetProvider.getAssets` method.
- Used `withDataSource` rather than `withQueryRunner`.
- Aggregated the SQL queries to get all `assetInfo`s and `nftMetadata`s with one query (one query for each: two new queries) rather than performing one query for each `assetInfo` or `nftMetadata`.
- Parallelized the functions to get `assetInfo`s, `nftMetadata`s and `tokenMetadata`s.

# Why should that fix?
Before the change, `TypeormAssetProvider.getAssets` was performing several SQL queries on a singe `QueryRunner` _trying to parallelize them_ through `Promise.all`.
A `QueryRunner` represent a PostgreSQL client. Even if [pg](https://www.npmjs.com/package/pg) allows to perform multiple queries before the response for the first one is received, under the wood, PostgreSQL protocol doesn't allow that; by specifications, one PostgreSQL client can perform only one query at time.
The effect of the approach used before the change was to use an high amount of memory to handle the parallelization attempt of the SQL queries as soon as a `TypeormAssetProvider.getAssets` call was received which was gradually decreasing as soon as responses from SQL queries were received.
In my opinion, said behavior multiplied multiple times (once per simultaneous `TypeormAssetProvider.getAssets` call) was the root cause of the problem.

- Change `staging-mainnet` context to `eks-devs` as suggested by @gytis-ivaskevicius .
- Enabled SDK deploy on `staging-mainnet` in STD github actions flow.
- Enabled `asset-projector` and `asset-provider` in `staging-mainnet` and `dev-preprod`.
- Add `cSpell` configuration in `nix/cardano-services/deployments/default.nix` to avoid several warning messages from VSCode extension.

### DIFF
--- a/packages/cardano-services/src/Asset/TypeormAssetProvider/TypeormAssetProvider.ts
+++ b/packages/cardano-services/src/Asset/TypeormAssetProvider/TypeormAssetProvider.ts
@@ -8,8 +8,8 @@ import {
   ProviderError,
   ProviderFailure
 } from '@cardano-sdk/core';
-import { AssetEntity } from '@cardano-sdk/projection-typeorm';
-import { QueryRunner } from 'typeorm';
+import { AssetEntity, NftMetadataEntity } from '@cardano-sdk/projection-typeorm';
+import { DataSource, In, QueryRunner } from 'typeorm';
 import { TokenMetadataService } from '../types';
 import { TypeOrmNftMetadataService } from '../TypeOrmNftMetadataService';
 import { TypeormProvider, TypeormProviderDependencies } from '../../util';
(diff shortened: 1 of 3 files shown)


packages/cardano-services/src/Asset/TypeormAssetProvider/TypeormAssetProvider.ts
```diff
@@ -58,26 +58,31 @@ export class TypeormAssetProvider extends TypeormProvider implements AssetProvid
...
-          assetInfo.tokenMetadata = tokenMetadataList[index];
-        }
-      }
+    return assetIds.map((assetId) => {
+      const assetInfo = assetInfoMap.get(assetId);
+
+      if (!assetInfo) throw new ProviderError(ProviderFailure.NotFound, undefined, `Asset not found '${assetId}'`);
+
+      if (nftMetadata) assetInfo.nftMetadata = metadataMap.get(assetId) || null;
+      if (tokenMetadata) assetInfo.tokenMetadata = tokeMap.get(assetId);
```
mchappell, line 83: why does this line not need `|| null` but the previous does?

packages/cardano-services/src/Asset/TypeormAssetProvider/TypeormAssetProvider.ts
```diff
@@ -58,26 +58,31 @@ export class TypeormAssetProvider extends TypeormProvider implements AssetProvid
...
-          assetInfoList.map(async (assetInfo) => {
-            assetInfo.nftMetadata = await this.#getNftMetadata(assetInfo, queryRunner);
-          })
-        );
-      }
+    const { nftMetadata, tokenMetadata } = extraData || {};
 
-      if (extraData?.tokenMetadata) {
-        const tokenMetadataList = await this.#fetchTokenMetadataList(assetIds);
+    const [assetInfoMap, metadataMap, tokeMap] = await this.withDataSource(async (dataSource) =>
```
mchappell, line 63:
```suggestion
    const [assetInfoMap, metadataMap, tokenMap] = await this.withDataSource(async (dataSource) =>
```

packages/cardano-services/src/Asset/TypeormAssetProvider/TypeormAssetProvider.ts
mchappell, line 83:
```suggestion
      if (tokenMetadata) assetInfo.tokenMetadata = tokenMap.get(assetId);
```

mirceahasegan (dismissed):
Fantastic work @iccicci 🚀 
The commit separation was very useful 🥇 .

One optional nitpick: could you copy the PR comment relating to the getAssets refactoring in commit "refactor(cardano-services): optimize TypeormAssetProvider.getAssets" too?

packages/cardano-services/src/Asset/TypeormAssetProvider/TypeormAssetProvider.ts
iccicci: Honestly IDK the actual answer to the question.
What I can say is previous version was using directly the result from [`this.#fetchTokenMetadataList(assetIds)`](https://github.com/input-output-hk/cardano-js-sdk/blob/2da68e66341ecffe674009009a11d28d7e4a6d48/packages/cardano-services/src/Asset/TypeormAssetProvider/TypeormAssetProvider.ts#L73-L77) to populate `assetInfo.tokenMetadata` and in case of missing metadata in the DB, [`null`](https://github.com/input-output-hk/cardano-js-sdk/blob/2da68e66341ecffe674009009a11d28d7e4a6d48/packages/cardano-services/src/Asset/TypeOrmNftMetadataService.ts#L30-L32) was used to populate `assetInfo.nftMetadata`.
I didn't focus on this part of the logic, I focused on the core of the change paying attention to not change the produced result.

mkazlauskas (dismissed):
Great work! Before merging, please confirm that it's releasing the connections back to the pool by executing many requests. An automated test for that would be nice to have (not necessary).

packages/cardano-services/src/Asset/TypeormAssetProvider/TypeormAssetProvider.ts
iccicci: 4a75b1a4900 should solve this.

packages/cardano-services/src/Asset/TypeormAssetProvider/TypeormAssetProvider.ts
iccicci: 4a75b1a4900 should solve this.

mkazlauskas (approved):

mchappell (approved):

gytis-ivaskevicius (approved):

PR #1459, opened 2024-09-03
feat(cardano-services): improve health check response times by aggregating and caching

ginnun:
This PR aims to improve health check API response times  (to under 1 sec) by pre-fetching and caching the underlying service health.

# Context

Generating on-demand health check responses is taking a long time. 

# Proposed Solution

We want to aggregate and cache underlying health check responses in the background. On-demand health check API responses will be generated from the always-warm cache.

We implemented a WarmCache logic by extending InMemoryCache.

One cache can be shared with multiple providers.

Now, various providers will be aggregating health checks independently in the background.

### DIFF
--- a/packages/cardano-services/src/util/TypeormProvider/TypeormProvider.ts
+++ b/packages/cardano-services/src/util/TypeormProvider/TypeormProvider.ts
@@ -1,13 +1,17 @@
 import { HealthCheckResponse, Milliseconds, Provider } from '@cardano-sdk/core';
+import { InMemoryCache } from '../../InMemoryCache';
 import { TypeormService, TypeormServiceDependencies } from '../TypeormService';
 import { skip } from 'rxjs';
 
-export type TypeormProviderDependencies = Omit<TypeormServiceDependencies, 'connectionTimeout'>;
+export type TypeormProviderDependencies = Omit<TypeormServiceDependencies, 'connectionTimeout'> & {
+  healthCheckCache: InMemoryCache;
+};
 
 const unhealthy = { ok: false, reason: 'Provider error' };
 
 export abstract class TypeormProvider extends TypeormService implements Provider {
   health: HealthCheckResponse = { ok: false, reason: 'not started' };
+  healthCheckCache: InMemoryCache;
 
   constructor(name: string, dependencies: TypeormProviderDependencies) {
     super(name, { ...dependencies, connectionTimeout: Milliseconds(1000) });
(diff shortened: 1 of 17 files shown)


packages/cardano-services/src/util/TypeormProvider/TypeormProvider.ts
```diff
@@ -1,26 +1,34 @@
...
+
+    this.healthCheckCache = dependencies.healthCheckCache;
   }
 
   async healthCheck(): Promise<HealthCheckResponse> {
     if (this.state === 'running')
       try {
-        await this.withDataSource((dataSource) => dataSource.query('SELECT 1'));
+        await this.healthCheckCache.get('typeorm_db', () =>
+          this.withDataSource((dataSource) => dataSource.query('SELECT 1'))
```
iccicci, line 30: Sorry, I can't understand how possible timeouts here are caught.

ginnun: @iccicci The timeout will be thrown inside `.withDataSource` and caught by the `try-block` since we do `await` the returned `promise` inside it. Please let me know if I missed the point.

I extended the existing test case with a thrown error in f66478c45e886ceb793bce21c2be290647837d50

iccicci: I'm wondering what happens if the function `() => this.withDataSource((dataSource) => dataSource.query('SELECT 1'))` (or the `() => this.#checkHealth()` in `NodeTxSubmitProvider.ts`) takes (for example) 5 minutes.

Will the health check take 5 minutes to become unhealthy as well?
Due to the nature of `InMemoryCache`, in case of expired cache, the `get` method waits till the resolution (or rejection) of the promise; could it be that in case of expired cache AND a long time required for the callback make the health check to take 5 minutes to respond?

packages/cardano-services/src/util/TypeormProvider/TypeormProvider.ts
```diff
@@ -1,26 +1,34 @@
...
     });
+
+    this.healthCheckCache = dependencies.healthCheckCache;
   }
 
   async healthCheck(): Promise<HealthCheckResponse> {
     if (this.state === 'running')
       try {
-        await this.withDataSource((dataSource) => dataSource.query('SELECT 1'));
+        await this.healthCheckCache.get('typeorm_db', () =>
```
iccicci, line 29: If we use the static `typeorm_db` string all the providers share the same cache entry.
We need to create a unique cache item name (maybe a combination with the `name` attribute of the `RunnableModule` ancestor class) because two distinct `TypeOrmProviders`s connect to two distinct databases.

iccicci (changes requested):

packages/cardano-services/src/util/TypeormProvider/TypeormProvider.ts
ginnun: Good point @iccicci .
I changed the implementation, so the cache will be updated once the promise is completed. Except the first time calling, the caller will get already completed promises.
Here: a64610fe47ebff27ca351e56cc0dcaa1f55b4569

packages/cardano-services/src/util/TypeormProvider/TypeormProvider.ts
ginnun: Done in c076aaa95833353a83284964f8b1c4e57f28d7e7

iccicci (approved):
Good job @ginnun !

rhyslbw (approved):
Great work @ginnun. Happy to see this progress

PR #1463, opened 2024-09-06
feat(cardano-services): implement provider selection based on env variables and config

ginnun:
For most providers, we have multiple implementations.
This feature allows us to pick which implementation we run inside the related HttpService.

# Context

This PR is based on `blockfrost` branch and can be merged directly into the `master` once the `blockfrost` branch is merged.

# Proposed Solution
It's backward compatible by respecting or throwing errors in case of conflict with the old env vars or configs.
For example,

- if `USE_SUBMIT_API` is set to any value, setting the `TX_SUBMIT_PROVIDER` value will throw an error.

### DIFF
--- a/packages/cardano-services/src/Program/options/providers.ts
+++ b/packages/cardano-services/src/Program/options/providers.ts
@@ -1,0 +1,79 @@
+import { newOption } from './util';
+
+export enum ProviderImplementation {
+  TYPEORM = 'typeorm',
+  BLOCKFROST = 'blockfrost',
+  DBSYNC = 'dbsync',
+  // Below ones are specific to TxSubmitProvider
+  SUBMIT_NODE = 'submit-node',
+  SUBMIT_API = 'submit-api'
+}
+
+export type ProviderImplementations = {
+  assetProvider?: ProviderImplementation;
+  rewardsProvider?: ProviderImplementation;
+  networkInfoProvider?: ProviderImplementation;
+  utxoProvider?: ProviderImplementation;
+  txSubmitProvider?: ProviderImplementation;
+  chainHistoryProvider?: ProviderImplementation;
+  stakePoolProvider?: ProviderImplementation;
+};
+export const ProviderImplementationDescription = 'Select one of the available provider implementations';
+const argParser = (impl: string) => ProviderImplementation[impl.toUpperCase() as keyof typeof ProviderImplementation];
+export const providerSelectionOptions = [
+  newOption(
+    '--asset-provider <implementation>',
+    ProviderImplementationDescription,
+    'ASSET_PROVIDER',
+    argParser,
+    ProviderImplementation.DBSYNC
+  )
+    .conflicts('useTypeormAssetProvider')
+    .choices([ProviderImplementation.BLOCKFROST, ProviderImplementation.DBSYNC, ProviderImplementation.TYPEORM]),
+  newOption(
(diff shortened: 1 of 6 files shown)


packages/cardano-services/src/Program/options/providers.ts
```diff
@@ -0,0 +1,79 @@
+import { newOption } from './util';
+
+export enum ProviderImplementation {
+  TYPEORM = 'typeorm',
+  BLOCKFROST = 'blockfrost',
+  DBSYNC = 'dbsync',
+  // Below ones are specific to TxSubmitProvider
+  'SUBMIT_NODE' = 'submit-node',
```
mkazlauskas, line 8: Is there a difference? :thinking: 
```suggestion
  SUBMIT_NODE = 'submit-node',
```

packages/cardano-services/src/Program/programs/providerServer.ts
```diff
@@ -72,6 +88,18 @@ const connectionConfigs: { [k in ConnectionNames]?: Observable<PgConnectionConfi
 
 let sharedHandleProvider: HandleProvider;
 
+const selectProviderImplementation = <T extends Provider>(
+  impl: ProviderImplementation,
+  typeorm?: () => T & TypeormProvider,
+  dbsync?: () => T & Provider,
+  blockfrost?: () => T & BlockfrostProvider
```
mkazlauskas, line 95: Nitpick/suggestion: would be nice to group in order to keep lower # of args (we might also want to add a `logger` for example) and not have to pass `undefined` when some provider option is not implemented.
```suggestion
  providerFactories: {
    typeorm?: () => T & TypeormProvider,
    dbsync?: () => T & Provider,
    blockfrost?: () => T & BlockfrostProvider
  }
```

packages/cardano-services/src/Program/programs/providerServer.ts
```diff
@@ -248,73 +276,132 @@ const serviceMapFactory = (options: ServiceMapFactoryOptions) => {
...
-      return new ChainHistoryHttpService({ chainHistoryProvider, logger });
-    }, ServiceNames.ChainHistory),
+    [ServiceNames.Utxo]:
+      args.utxoProvider === ProviderImplementation.BLOCKFROST
+        ? getBlockfrostEnabledUtxoHttpService
+        : withDbSyncProvider(
+            async (dbPools, cardanoNode) =>
+              new UtxoHttpService({
+                logger,
+                utxoProvider: new DbSyncUtxoProvider({
```
mkazlauskas, line 341: All provider implementations use the same HttpService. I think it would be cleaner if we switch only the provider
```suggestion
                utxoProvider: selectProviderImplementation(args.utxoProvider, ...)
```

mkazlauskas (changes requested):

iccicci (dismissed):
LGTM

packages/cardano-services/src/Program/programs/providerServer.ts
ginnun: Nice suggestion, here 0d44798bfa387540e2db07bfab967eab043029dd

packages/cardano-services/src/Program/programs/providerServer.ts
ginnun: Done here 0d44798bfa387540e2db07bfab967eab043029dd

PR #1464, opened 2024-09-06
fix: expect disconnects during remote api method call

mkazlauskas:
# Context

it happens due to unknown reason, possibly due to the recently introduced BFCache feature in chromium)

also log runtime.lastError when port disconnects

# Proposed Solution

Do not throw if port re-connects within 10s (normally it takes ~1s)

BREAKING CHANGE: change return type of createWalletApi callbacks. 
See [changes in lace](https://github.com/input-output-hk/lace/commit/01b8b6e86faa2cf83a3ccc100277fa6f05840793)


mirceahasegan (approved):
Fantastic work @martynas!

AngelCastilloB (approved):
Great work @mkazlauskas  🧙🏼

PR #1465, opened 2024-09-06
chore: increase cpu/mem for live-mainnet history-provider

bernokl:
I see we have been getting cpu throttled with the 1.2 cores so I bumped it up to 2 full cores to see if that helps resolve the issue

### DIFF
--- a/nix/cardano-services/deployments/default.nix
+++ b/nix/cardano-services/deployments/default.nix
@@ -445,6 +445,8 @@ in
             chain-history-provider = {
               enabled = true;
               replicas = 2;
+              resources.limits = mkPodResources "500Mi" "2000m";
+              resources.requests = mkPodResources "150Mi" "1000m";
             };
             # asset-provider = {
             #   enabled = true;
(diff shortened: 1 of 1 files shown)


iccicci (approved):

nix/cardano-services/deployments/default.nix
```diff
@@ -445,6 +445,8 @@ in
             chain-history-provider = {
               enabled = true;
               replicas = 2;
+              resources.limits = mkPodResources "500Mi" "2000m";
+              resources.requests = mkPodResources "150Mi" "1000m";
```
gytis-ivaskevicius, line 449: I think you need `mkForce` for this. Also has this been tested? We might need to remove limit altogether to resolve this issue

gytis-ivaskevicius: Probably we will end up with something like this. Also please check grafana memory usage, 150m feels low
```suggestion
              resources = lib.mkForce {
                requests = mkPodResources "150Mi" "1000m";
              };
```

bernokl:
Replaced by planned removal of CPU limits from all providers.

PR #1466, opened 2024-09-09
Feat/lw 10717 conditional withdrawal in base wallet

AngelCastilloB:
# Context

After the Conway-intra-era pv10, hard fork, reward accounts will be blocked from withdrawing any rewards unless their associated stake credential is also delegated to a DRep. 

# Proposed Solution

Implement a wallet mechanism to ensure that reward retrieval operations are only executed when there is stake key is delegated to a dRep.

### DIFF
--- a/packages/core/src/Cardano/types/DelegationsAndRewards.ts
+++ b/packages/core/src/Cardano/types/DelegationsAndRewards.ts
@@ -1,3 +1,4 @@
+import { DelegateRepresentative } from './Governance';
 import { Lovelace } from './Value';
 import { Metadatum } from './AuxiliaryData';
 import { PoolId, PoolIdHex, StakePool } from './StakePool';
@@ -23,10 +24,13 @@ export enum StakeCredentialStatus {
   Unregistered = 'UNREGISTERED'
 }
 
+export type DRepDelegatee = { delegateRepresentative: DelegateRepresentative };
+
 export interface RewardAccountInfo {
   address: RewardAccount;
   credentialStatus: StakeCredentialStatus;
   delegatee?: Delegatee;
+  dRepDelegatee?: DRepDelegatee;
   rewardBalance: Lovelace;
   // Maybe add rewardsHistory for each reward account too
   deposit?: Lovelace; // defined only when keyStatus is Registered
--- a/packages/wallet/src/services/DelegationTracker/RewardAccounts.ts
+++ b/packages/wallet/src/services/DelegationTracker/RewardAccounts.ts
@@ -196,8 +196,38 @@ const accountCertificateTransactions = (
   );
 };
 
+const accountDRepCertificateTransactions = (
+  transactions$: Observable<TxWithEpoch[]>,
+  rewardAccount: Cardano.RewardAccount
+) => {
+  const stakeKeyHash = Cardano.RewardAccount.toHash(rewardAccount);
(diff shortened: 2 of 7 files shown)


packages/core/src/Cardano/types/DelegationsAndRewards.ts
```diff
@@ -27,6 +28,7 @@ export interface RewardAccountInfo {
   address: RewardAccount;
   credentialStatus: StakeCredentialStatus;
   delegatee?: Delegatee;
+  dRepDelegatee?: DelegateRepresentative;
```
mkazlauskas, line 31: This would make it easier to add some metadata in the future
```suggestion
type DRepDelegatee = { delegateRepresentative: DelegateRepresentative; }
...
    dRepDelegatee?: DRepDelegatee;
```

mkazlauskas (changes requested):

packages/core/src/Cardano/types/DelegationsAndRewards.ts
AngelCastilloB: done in https://github.com/input-output-hk/cardano-js-sdk/pull/1466/commits/2cafb87ae9fd6a481f0ed0b4b329e5381165375a

mkazlauskas (dismissed):
Great work! :muscle:

mirceahasegan (approved):
Great work @AngelCastilloB !

mkazlauskas (dismissed):

packages/wallet/src/services/DelegationTracker/RewardAccounts.ts
```diff
@@ -196,8 +196,32 @@ const accountCertificateTransactions = (
...
+  return transactions$.pipe(
+    map((transactions) =>
+      transactions
+        .map(({ tx, epoch }) => ({
+          certificates: (tx.body.certificates || [])
+            .map((cert) =>
+              Cardano.isCertType(cert, [...Cardano.VoteDelegationCredentialCertificateTypes]) ? cert : null
+            )
+            .filter(isNotNil)
+            .filter((cert) => (cert.stakeCredential.hash as unknown as Crypto.Ed25519KeyHashHex) === stakeKeyHash),
```
iccicci, line 213: Shouldn't we check if the voting delegation certificate is younger than the latest stake key registration certificate?
I mean, if we delegate our vote then we unregister and re-register our stake credentials, is the vote delegated as early than we unregistered the stake credential or is it in an _un-delegated_ status?

I understand that checking for stake de-registration certificates is not required: if a stake credential is no longer registered, it has no rewards to withdrawn, none of these checks will be performed.

AngelCastilloB: Good catch @iccicci , You are right, if we un-register and re-register the key, it returns to an un-delegated state, will add this check

AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1466/commits/829fd1f8fedc4bc1a2bf3702c95c8f318a9b731a

iccicci (approved):
As usual... excellent job @AngelCastilloB !

mirceahasegan (approved):
Great work @AngelCastilloB ! 💪

PR #1467, opened 2024-09-09
LW-11425 Misc fixes

iccicci:
# Context

I noticed a problem about certificates type; while fixing it I noticed some other issues... while working on them I discovered some more else... that chain originated this misc fixes PR.

Discovered issues.

- We are using an hard-code `Cardano.CredentialType.KeyHash` while building certificates.
- We are using a not correct sorting for transaction inputs.
- We are sorting `collateralReturn`: it's useless to sort an _up to 1 element_ array.
- We are using an hard-coded `null` for `poolParameters` while building `PoolRegistrationCertificate`s; this makes `getCertificatesByType` to not pick given certificates.
- We are loading assets for `collateralReturn` as it was an output but assets for it are stored in `multi_assets_descr` column not in `ma_tx_out` table.

# Proposed Solution

- Computed the `stakeCredential` directly from `certModel.address`.
  - The test may seems a bit chaotic or redundant: once `Cardano.CertificateType.StakeRegistration` will be deprecated in favor of `Cardano.CertificateType.Registration` the test can be simplified to check the entire certificate.
- Removed sorting by `index` (which is correct only for `outputs`) while producing `inputs`. This result in a slight optimization as well.
  - Checked the `collaterals` as well: it seems **db-sync** inserts them in a random order: we can't produce them with the original order. Removed the same sort as per `inputs` to apply the same slight optimization.
- Removed sorting of `collateralReturn`.
- Built a partial `Cardano.PoolParameters` fetching from DB the required data to fix the issue (the `rewardAccount`) plus those data already loaded by the DB but not fetched.
  - To make the structure complete `owners` and `relays` would be required as well. I considered that a _new feature implementation_ and not handled in this _fix_ PR.
- Skipped the call of `ChainHistoryBuilder.queryMultiAssetsByTxOut` as `ma_tx_out.tx_out` in case of fetching `collateralReturn` as it refers explicitly to `tx_out` table and not to `collateral_tx_out` table.
  - To add assets to `collateralReturn` would be required to fetch and parse `collateral_tx_out.multi_assets_descr` column. I considered that a _new feature implementation_ and not handled in this _fix_ PR.

- Based on a source from @mirceahasegan , added a script to get CBOR representation of on chain transactions. It's complexity may seems a bit redundant... it is required to give the script the most user friendly interface as possible:
  - once we have the interested network running in our local environment (and obviously synced at least up to the block containing the tx we are interested in) issuing `yarn tx-cbor <txId>` is enough, the script works regardless from local ports configuration (without the need to forward it given configuration).

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,18 @@ for you.
 nix run .#config-update
 ```
 
+## Get CBOR representation of an on chain transaction
+
+Once we have a [running network](packages/cardano-services/README.md#production) synced at least up to the block
+containing the transaction we are interested in, issue following command to get the CBOR representation of the
+transaction.
+
+```
+yarn tx-cbor <txId>
+```
+
+This works regardless of the local ports configuration through environment variables.
+
 ## Attic
 
 Previously supported features, no longer supported, but packed with a reference branch.
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
     "test": "yarn workspaces foreach -v run test",
     "test:build:verify": "yarn workspaces foreach -v run test:build:verify",
     "test:e2e": "yarn workspaces foreach -v run test:e2e",
-    "test:debug": "DEBUG=true yarn workspaces foreach -v run test"
+    "test:debug": "DEBUG=true yarn workspaces foreach -v run test",
+    "tx-cbor": "tsx packages/cardano-services/scripts/tx-cbor.js"
   },
   "repository": {
     "type": "git",
--- a/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
+++ b/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
@@ -1,3 +1,5 @@
+// cSpell:ignore descr
+
 import * as Queries from './queries';
 import {
   AuthorizeCommitteeHotCertModel,
@@ -261,7 +263,10 @@ export class ChainHistoryBuilder {
     if (result.rows.length === 0) return [];
 
     const txOutIds = result.rows.flatMap((txOut) => BigInt(txOut.id));
-    const multiAssets = await this.queryMultiAssetsByTxOut(txOutIds);
+    // In case of collateralReturn requests (collateral = true) assets in the output can't be read as for regular outputs:
+    // db-sync stores assets from collateral outputs in collateral_tx_out.multi_assets_descr column rather than in
+    // ma_tx_out table like for regular outputs. To have a complete collateralReturn, given column should be read and parsed.
+    const multiAssets = collateral ? new Map() : await this.queryMultiAssetsByTxOut(txOutIds);
     const referenceScripts = await this.queryReferenceScriptsByTxOut(result.rows);
 
     return result.rows.map((txOut) =>
--- a/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/DbSyncChainHistoryProvider.ts
+++ b/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/DbSyncChainHistoryProvider.ts
@@ -139,9 +139,7 @@ export class DbSyncChainHistoryProvider extends DbSyncProvider() implements Chai
       const txInputs = inputs.filter((input) => input.txInputId === txId).map(mapTxIn);
       const txCollaterals = collaterals.filter((col) => col.txInputId === txId).map(mapTxIn);
       const txOutputs = orderBy(outputs.filter((output) => output.txId === txId).map(mapTxOut), ['index']);
-      const txCollateralOutputs = orderBy(collateralOutputs.filter((output) => output.txId === txId).map(mapTxOut), [
-        'index'
-      ]);
+      const txCollateralOutputs = collateralOutputs.filter((output) => output.txId === txId).map(mapTxOut);
       const inputSource: Cardano.InputSource = tx.valid_contract
         ? Cardano.InputSource.inputs
         : Cardano.InputSource.collaterals;
--- a/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/DbSyncChainHistoryProvider.ts
+++ b/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/DbSyncChainHistoryProvider.ts
@@ -136,8 +136,8 @@ export class DbSyncChainHistoryProvider extends DbSyncProvider() implements Chai
 
     return txResults.rows.map((tx) => {
       const txId = tx.id.toString('hex') as unknown as Cardano.TransactionId;
-      const txInputs = orderBy(inputs.filter((input) => input.txInputId === txId).map(mapTxIn), ['index']);
-      const txCollaterals = orderBy(collaterals.filter((col) => col.txInputId === txId).map(mapTxIn), ['index']);
+      const txInputs = inputs.filter((input) => input.txInputId === txId).map(mapTxIn);
+      const txCollaterals = collaterals.filter((col) => col.txInputId === txId).map(mapTxIn);
       const txOutputs = orderBy(outputs.filter((output) => output.txId === txId).map(mapTxOut), ['index']);
       const txCollateralOutputs = orderBy(collateralOutputs.filter((output) => output.txId === txId).map(mapTxOut), [
         'index'
--- a/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/mappers.ts
+++ b/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/mappers.ts
@@ -231,12 +231,7 @@ export const mapCertificate = (
         : Cardano.CertificateType.Unregistration,
       cert_index: certModel.cert_index,
       deposit: BigInt(certModel.deposit),
-      stakeCredential: {
-        hash: Cardano.RewardAccount.toHash(
-          Cardano.RewardAccount(certModel.address)
-        ) as unknown as Crypto.Hash28ByteBase16,
-        type: Cardano.CredentialType.KeyHash
-      }
+      stakeCredential: Cardano.Address.fromBech32(certModel.address).asReward()!.getPaymentCredential()
     } as WithCertIndex<Cardano.NewStakeAddressCertificate>;
 
   if (isDelegationCertModel(certModel))
@@ -244,12 +239,7 @@ export const mapCertificate = (
       __typename: Cardano.CertificateType.StakeDelegation,
       cert_index: certModel.cert_index,
       poolId: certModel.pool_id as unknown as Cardano.PoolId,
-      stakeCredential: {
-        hash: Cardano.RewardAccount.toHash(
-          Cardano.RewardAccount(certModel.address)
-        ) as unknown as Crypto.Hash28ByteBase16,
-        type: Cardano.CredentialType.KeyHash
-      }
+      stakeCredential: Cardano.Address.fromBech32(certModel.address).asReward()!.getPaymentCredential()
     } as WithCertIndex<Cardano.StakeDelegationCertificate>;
 
   if (isDrepRegistrationCertModel(certModel))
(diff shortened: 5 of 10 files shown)


packages/e2e/test/wallet_epoch_3/SharedWallet/delegation.test.ts
```diff
@@ -204,6 +204,16 @@ describe('SharedWallet/delegation', () => {
...
 
+    // Check Registration and StakeDelegation certificate from ChainHistoryProvider
+    let gotTx = toSerializableObject(
+      (await firstValueFrom(aliceMultiSigWallet.transactions.history$)).find((t) => tx.id === t.id)!
+    );
+    // These are required because txBuilder still uses StakeRegistration
+    (gotTx as Cardano.HydratedTx).body.certificates![0].__typename = Cardano.CertificateType.StakeRegistration;
+    delete ((gotTx as Cardano.HydratedTx).body.certificates![0] as Partial<Cardano.NewStakeAddressCertificate>).deposit;
+    expect((gotTx as Cardano.HydratedTx).body.certificates?.length).toEqual(2);
+    expect((gotTx as Cardano.HydratedTx).body.certificates).toEqual(toSerializableObject(tx.body.certificates));
```
mkazlauskas, line 215: Nitpick/suggestion: this could be DRYed and also made more readable like this. The function can patch **all** stake registration and deregistration certificates it finds in the tx before comparing.
```suggestion
    await expect(walletHistoryHasAllCertificatesFrom(aliceMultiSigWallet, tx)).resolves.toBe(true);
```

package.json
```diff
@@ -37,7 +37,8 @@
     "test": "yarn workspaces foreach -v run test",
     "test:build:verify": "yarn workspaces foreach -v run test:build:verify",
     "test:e2e": "yarn workspaces foreach -v run test:e2e",
-    "test:debug": "DEBUG=true yarn workspaces foreach -v run test"
+    "test:debug": "DEBUG=true yarn workspaces foreach -v run test",
+    "tx-cbor": "tsx packages/cardano-services/scripts/tx-cbor.js"
```
mkazlauskas, line 41: Would be nice to add a paragraph in README about how to use this script. Also, not sure about the location of the script: we do have `scripts/` in monorepo root. Maybe it would be better to hoist it there?

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/DbSyncChainHistoryProvider.ts
```diff
@@ -136,8 +136,8 @@ export class DbSyncChainHistoryProvider extends DbSyncProvider() implements Chai
 
     return txResults.rows.map((tx) => {
       const txId = tx.id.toString('hex') as unknown as Cardano.TransactionId;
-      const txInputs = orderBy(inputs.filter((input) => input.txInputId === txId).map(mapTxIn), ['index']);
-      const txCollaterals = orderBy(collaterals.filter((col) => col.txInputId === txId).map(mapTxIn), ['index']);
+      const txInputs = inputs.filter((input) => input.txInputId === txId).map(mapTxIn);
```
mkazlauskas, line 139: Was `index` here the utxo output index?

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
```diff
@@ -261,7 +261,7 @@ export class ChainHistoryBuilder {
     if (result.rows.length === 0) return [];
 
     const txOutIds = result.rows.flatMap((txOut) => BigInt(txOut.id));
-    const multiAssets = await this.queryMultiAssetsByTxOut(txOutIds);
+    const multiAssets = collateral ? new Map() : await this.queryMultiAssetsByTxOut(txOutIds);
```
mkazlauskas, line 264: Could you please elaborate on the reason of this change? It might be an issue when inspecting transaction on front-end. We're computing the difference of `consumed` and `produced` coins and assets that come from and goes to our own addresses. This could make it look like we've spent some assets on a failed transaction.

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/mappers.ts
```diff
@@ -209,7 +221,7 @@ export const mapCertificate = (
       __typename: Cardano.CertificateType.PoolRegistration,
       cert_index: certModel.cert_index,
       deposit: BigInt(certModel.deposit),
-      poolParameters: null as unknown as Cardano.PoolParameters
+      poolParameters: mapPoolParameters(certModel)
```
mkazlauskas, line 224: This looks like a fairly cheap change, so just wondering: are we using/planning to use it for something in Lace?

mkazlauskas (changes requested):

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
iccicci: This can't be source of any issue: we load `collaterReturn` just for additional info, today with assets from a random output from a random transaction from a random address, after this change simply without any assets.
Transaction with failed phase 2 validation have their `collaterReturn` stored in db-sync as a standard output, so they already have the correct map of assets regardless from this fix.

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/mappers.ts
iccicci: IDK, I just saw `getCertificatesByType` is not able to pick stake pool registration certificates and I fixed it.

package.json
iccicci: > Would be nice to add a paragraph in README about how to use this script.

Going to add it.

> Also, not sure about the location of the script: we do have scripts/ in monorepo root. Maybe it would be better to hoist it there?

That was my fist attempt, but I got some import issues (I don't remember the details) so I chosen this option as fallback to save some time.

packages/e2e/test/wallet_epoch_3/SharedWallet/delegation.test.ts
mkazlauskas: Thanks for explanation, I just noticed duplicate logic for both types of certificates (registration and de-registration) which could be hoisted to a util. Anyway, it's just a suggestion, happy to keep as-is if you prefer it this way. Resolving :)

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
mkazlauskas: Not sure I understand. Are we still returning `collateralReturn: { value: { coins, assets } }` with assets for chain history requests?

iccicci: We are returning `collateralReturn: Cardano.TxOut` anyway, but today the asset property contains random assets after the fix it is empty.
Anyway this can't be source of any issue as `collateralReturn`
- is never used for transaction with valid contract
- is stored (in db-sync) as an output so is correctly read with the right assets

package.json
iccicci: 7347e1b7ab8 should solve this

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
iccicci: d1ead2edd4a should solve this

README.md
```diff
@@ -137,6 +137,18 @@ for you.
 nix run .#config-update
 ```
 
+## Get CBOR representation of an in chain transaction
```
mkazlauskas, line 140:
```suggestion
## Get CBOR representation of an on chain transaction
```

README.md
```diff
@@ -137,6 +137,18 @@ for you.
...
+
+Once we have a [running network](packages/cardano-services/README.md#production) synced at least up to the block
+containing the transaction we are interested in, issue following command to get the CBOR representation of the
+transaction.
+
+```
+yarn tx-cbor <txId>
+```
+
+This works regardless from the local ports configuration through environment variables.
```
mkazlauskas, line 150:
```suggestion
This works regardless of the local ports configuration through environment variables.
```

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
```diff
@@ -261,7 +263,10 @@ export class ChainHistoryBuilder {
     if (result.rows.length === 0) return [];
 
     const txOutIds = result.rows.flatMap((txOut) => BigInt(txOut.id));
-    const multiAssets = await this.queryMultiAssetsByTxOut(txOutIds);
+    // In case or collateralReturn requests (collateral = true) assets in the output can't be read as for regular outputs:
+    // db-sync stores assets from collateral outputs in collateral_tx_out.multi_assets_descr column rather than in
+    // ma_tx_out table like for regular outputs. To have a complete collateralReturn, given column should be red and parsed.
```
mkazlauskas, line 268:
```suggestion
    // ma_tx_out table like for regular outputs. To have a complete collateralReturn, given column should be read and parsed.
```

mkazlauskas (dismissed):
Just some minor/spelling change requests, feel free to squash those (no need for fixups)

packages/e2e/test/wallet_epoch_3/SharedWallet/delegation.test.ts
```diff
@@ -204,6 +204,16 @@ describe('SharedWallet/delegation', () => {
     await waitForTx(aliceMultiSigWallet, tx.id);
     const tx1ConfirmedState = await getWalletStateSnapshot(aliceMultiSigWallet);
 
+    // Check Registration and StakeDelegation certificate from ChainHistoryProvider
+    let gotTx = toSerializableObject(
+      (await firstValueFrom(aliceMultiSigWallet.transactions.history$)).find((t) => tx.id === t.id)!
+    );
+    // These are required because txBuilder still uses StakeRegistration
```
mirceahasegan, line 211: We can flip the `isConwayEra` switch, or drop it. All networks are in conway era now, and we can use the newer certificates in TxBuilder too.
This can be another task if you want 😄

mkazlauskas: Let's do it separately, and a little bit later to give people some time to update HW wallet apps

mirceahasegan (dismissed):
As usual, great work @iccicci !
Love the split of changes into commits too 👏

packages/e2e/test/wallet_epoch_3/SharedWallet/delegation.test.ts
iccicci: Yep, the intention behind the complexity I added here is:
IDK when will be the right moment to consider the Conway era stable, I'll write a test in a wat that once the PR to remove pre-Conway stuff will be done:

1. this test will fail
2. the action to fix the test is the same action to reduce it to a single line

Please try to tolerate my obsessive compulsive disorders. 😜

packages/cardano-services/src/ChainHistory/DbSyncChainHistory/ChainHistoryBuilder.ts
iccicci: c7707dd88f1 should solve this

README.md
iccicci: 649430b44b4 should solve this

README.md
iccicci: 649430b44b4 should solve this

mkazlauskas (approved):
Great work! :muscle:

mirceahasegan (approved):
Great work!

PR #1468, opened 2024-09-11
fix: correctly set handle owner when handle is upgraded to cip68

mkazlauskas:
# Context

Found some suspicious logs in mainnet handle provider:
```
[TypeOrmHandleProvider] silverfox has no associated address, which could be the result of a double mint.
```

It [is a valid handle](https://cexplorer.io/asset/asset1lafyy7j4kaa09wrsg34jgz3r3d60ryr3daqth2/tx#data) that was minted just once by upgrading a cip25 handle into cip68.

The bug is caused by [withHandles mapper](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/projection/src/operators/Mappers/withHandles.ts#L163-L164) overwriting handles found in outputs with the ones found in mint (burned), therefore setting address to `null`. 

# Proposed Solution

- [x] create a new chain-sync dataset with problematic block and add an integration test in `projection-typeorm`
- [ ] fix the bug


rhyslbw (approved):
LGTM

iccicci (approved):
LGTM

PR #1469, opened 2024-09-11
fix(wallet): dynamicChangeResolver gives preference to lower derivation indices

AngelCastilloB:
# Context

User used Lace to manage his ada handle settings ( a lot of those ) and Eternl wouldn't show his balance anymore ( because of the discovery gap )

Then after quite a few transactions, it suddenly used a newly created (index 557) address for change and sticked to it for the following 3 transactions.

Lace is probably picking the newly created unused address as change address sometimes. This changes makes sure it always picks the same addresses per reward accounts and its always the lowest derivation index.


mkazlauskas (approved):
Great work! :rocket:

rhyslbw (approved):
Great work @AngelCastilloB :mechanical_arm:

PR #1479, opened 2024-09-18
test(e2e): abort k6 wallets test if too many failed requests

mirceahasegan:
# Context

wallets K6 test does not fail regardless of how many requests fail.
P95, P99 response times seem to include the failed response times, which are usually fast.

# Proposed Solution
abort k6 wallets test if too many failed requests
- show that test is failing, otherwise failed requests are considered successful with small response time
- To save on K6 VUh
-


gytis-ivaskevicius (approved):

iccicci (approved):
LGTM!

PR #1482, opened 2024-09-23
Fix signed/unsigned bug in Value.fromCBOR

AngelCastilloB:
It looks like the author meant to do:

    if(reader.peek() === UNSIGNED_INT) { reader.readUInt() }

instead of

    if(reader.peek() === UNSIGNED_INT) { reader.readInt() }


rhyslbw (approved):
image

mkazlauskas (approved):

PR #1486, opened 2024-09-24
refactor!: hoist bufferChainSyncEvent to projection package

AngelCastilloB:
# Context

@cardano-sdk/core requires rxjs but it seems not having it in its dependencies

# Proposed Solution

Hoist bufferChainSyncEvent (the only dependent on rxjs) to projection package

### DIFF
--- a/packages/core/src/CardanoNode/types/CardanoNode.ts
+++ b/packages/core/src/CardanoNode/types/CardanoNode.ts
@@ -1,5 +1,5 @@
 import type { HealthCheckResponse } from '../../Provider';
-import type { Lovelace, PoolId, VrfVkHex } from '../../Cardano';
+import type { Lovelace, PoolId, Tip, VrfVkHex } from '../../Cardano';
 import type { Milliseconds } from '../../util';
 
 export interface EraSummary {
@@ -48,3 +48,13 @@ export interface CardanoNode {
    */
   healthCheck(): Promise<HealthCheckResponse>;
 }
+
+// Similar to Ogmios.Point, but using Cardano.BlockId opaque string for hash
+export type Point = Pick<Tip, 'hash' | 'slot'>;
+export type Origin = 'origin';
+export type TipOrOrigin = Tip | Origin;
+export type PointOrOrigin = Point | Origin;
+export type Intersection = {
+  point: PointOrOrigin;
+  tip: TipOrOrigin;
+};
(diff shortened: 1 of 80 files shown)


packages/projection/package.json
```diff
@@ -49,7 +49,6 @@
     "tsc-alias": "^1.8.10"
   },
   "devDependencies": {
-    "@cardano-sdk/ogmios": "workspace:~",
```
mkazlauskas, line 52: Looks like my vscode sorted `package.json` :facepalm:

packages/core/src/CardanoNode/types/CardanoNode.ts
```diff
@@ -48,3 +48,38 @@ export interface CardanoNode {
...
+  block: Block;
+}
+
+export interface ChainSyncRollBackward extends WithRequestNext {
+  eventType: ChainSyncEventType.RollBackward;
+  point: PointOrOrigin;
+  tip: TipOrOrigin;
+}
+
+export type ChainSyncEvent = ChainSyncRollForward | ChainSyncRollBackward;
```
mkazlauskas, line 85: Consider hoisting chain sync types to `projection` too

mkazlauskas (approved):
Nice work! Just not sure about the commit message:
- typically if you add both `!` in the header and `BREAKING CHANGE:` in the footer it would create 2 entries for breaking changes in the CHANGELOG, so 1 of the 2 is enough
- Not sure the footer will be picked up in this case, because:
  - it doesn't have commit message body
  - footer is the co-authored message
```
<header>
<body>
<footer>
```
image

packages/core/src/CardanoNode/types/CardanoNode.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1486/commits/a7d9646e8d50d0c7487588371eb2e007e8395c12

iccicci (approved):
Nice catch!

I also noticed the required dependency not listed in `package.json` file a couple of days ago... but I had no time yet to investigate it

packages/projection/package.json
AngelCastilloB: This is also fixed (but I forget about this change as is mix with all the other changes in the fixup 😬 )

mkazlauskas (approved):
Great work! :rocket: Please update the commit message as discussed on Slack so that I'm not listed a breaking change :sweat_smile:

PR #1487, opened 2024-09-24
fix(core): protocol parameters update now correctly deserializes extra entropy when it set as empty

AngelCastilloB:
# Context

We were incorrectly deserializing extra entropy in protocol params update when the entropy bytes were empty (which is a valid use case). Even tho this parameter is deprecated we should be able to deserialize transactions with this field properly as there are some on mainnet:

See:

https://cardanoscan.io/transaction/fc863a441b55acceebb7d25c81ff7259e4fc9b92fbdf6d594118fb8f1110a78c

# Proposed Solution

Fix serialization/de-serialization logic


ginnun (approved):
lgtm

iccicci (approved):
LGTM!

PR #1488, opened 2024-09-24
test(e2e): reduce test flakiness

iccicci:
# Context

Out e2e test often uses a piec of code like following
```
    await wallet.submitTx(signedTx);

    // Search chain history to see if the transaction is there.
    const txFoundInHistory = await firstValueFrom(
      wallet.transactions.history$.pipe(
        map((txs) => txs.find((tx) => tx.id === signedTx.id)),
        filter(isNotNil),
        take(1)
      )
    );
```
next perform the checks.
The problem is this code is based only on `ChainHistoryProvider` while some checks are performed on `Observable`s sometimes based on other providers (`AssetProvider`, `UtxoProvider`, etc...).
This is a source of flakiness as check result may be dependent on which provider completes earlier.

# Proposed Solution

Substitute given piece of code (where required) with `submitAndConfirm` utility, which internally also waits for the wallet to be settled, i.e. all the providers finished their tasks.
This also slightly reduce code repetition.

### DIFF
--- a/packages/e2e/test/wallet_epoch_0/PersonalWallet/byron.test.ts
+++ b/packages/e2e/test/wallet_epoch_0/PersonalWallet/byron.test.ts
@@ -2,10 +2,8 @@
 import { BaseWallet } from '@cardano-sdk/wallet';
 import { Cardano } from '@cardano-sdk/core';
 import { createLogger } from '@cardano-sdk/util-dev';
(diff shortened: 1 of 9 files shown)


mkazlauskas (approved):
Great improvement! :muscle:

mkazlauskas (approved):

mirceahasegan (approved):
Nice! Thank you @iccicci

PR #1491, opened 2024-09-27
fix(core): plutus integers are now decoded correctly from indefinite length array form

AngelCastilloB:
# Context

There was a bug in the CborReader that would cause incorrect reading of cbor bytestings encoded in indefinite form if they were inside another CBOR element as the offset was not being taken into account.

The PlutusData was no chunking integers if their serialized form did not fit in the 64 bytes chunk limit.

# Proposed Solution

- CborReader now correctly takes into account the current reader offset.
- PlutusData now perform chunking on integers that do not fit in a 64 bytestring.


packages/core/src/Serialization/CBOR/CborReader.ts
```diff
@@ -886,7 +886,7 @@ export class CborReader {
       const { length: chunkLength, bytesRead } = CborReader.#peekDefiniteLength(nextInitialByte, data.slice(i));
       const payloadSize = bytesRead + Number(chunkLength);
 
-      concat = Buffer.concat([concat, this.#data.slice(i + (payloadSize - chunkLength), i + payloadSize)]);
```
AngelCastilloB, line 889: 🥲

rhyslbw (approved):
Tests look good 👍

mkazlauskas (approved):

PR #1492, opened 2024-09-27
refactor: hoist ws provider to cardano-services-client package

iccicci:
# Context

`WsProvider` class require rxjs making it a dependency for the `core` package.

# Proposed Solution

Hoisted the class to `cardano-services-client`.


mirceahasegan (approved):
lgtm 👏

packages/core/src/WebSocket.ts
```diff
@@ -13,32 +12,3 @@ export interface WSMessage {
   /** Latest value(s) for the `NetworkInfoProvider` methods.*/
   networkInfo?: Partial<NetworkInfoResponses>;
 }
-
```
mkazlauskas, line 16: Maybe we should also hoist `WsMessage` as it's intended to be used with `WsProvider`
Also, please reword the commit to indicate breaking change

mkazlauskas (changes requested):

packages/core/src/WebSocket.ts
iccicci: 72e42cebb0a should solve this

mirceahasegan (approved):

mkazlauskas (approved):

PR #1495, opened 2024-10-02
LW-11642 Remove github bot token from checkout step of release github action

iccicci:
# Context

The `release` github action flow no longer works.

# Proposed Solution

As @mirceahasegan discovered removing the `token` fixes the issue


mirceahasegan (approved):
lgtm

gytis-ivaskevicius (changes requested):
we need to set `permissions` with `contents: write` (at least), otherwise it will crash.

iccicci:
Refreshing the token fixed the issue.
This is no longer required.

PR #1496, opened 2024-10-02
feat(dapp-connector): custom inject key, extensions and api

mirceahasegan:
Add support for:
- inject api under custom key, instead of wallet name
- configure the supported extensions when constructing the Cip30Wallet object
- deviation from standard cip30 api. Currently overrides getCollateral to return empty array when no collaterals are found


vetalcore (approved):

AngelCastilloB (approved):
LGTM @mirceahasegan ⭐

PR #1497, opened 2024-10-02
LW-11640 Fix queries performances issue

iccicci:
# Context

SRE discovered two queries causing high CPU workload on DB server.

# Proposed Solution

- Optimized the `redeemer` query hoisting the sort at Node.js level.
- Removed meaningless `ORDER BY` for collateral outputs query.

- Removed some dead code I detected while working on this

### DIFF
--- a/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/queries.ts
+++ b/packages/cardano-services/src/ChainHistory/DbSyncChainHistory/queries.ts
@@ -37,25 +37,9 @@ export const findTxCollateralsByIds = `
 	WHERE tx.id = ANY($1)
 	ORDER BY tx_in.id ASC`;
 
-export const findTxInputsByAddresses = `
-  ${selectTxInput()}
-	JOIN block ON tx.block_id = block.id
-  WHERE tx_out.address = ANY($1)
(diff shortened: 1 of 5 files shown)


packages/cardano-services/src/ChainHistory/DbSyncChainHistory/queries.ts
```diff
@@ -37,31 +37,14 @@ export const findTxCollateralsByIds = `
...
-  WHERE tx_out.address = ANY($1)
-	AND block.block_no >= $2
-	AND block.block_no <= $3
-	ORDER BY tx_out.id ASC`;
-
 export const findCollateralOutputsByTxIds = `
 	${selectTxOutput(true)}
-	WHERE tx.id = ANY($1)
-  ORDER BY tx_out.id ASC`;
+	WHERE tx.id = ANY($1)`;
```
gytis-ivaskevicius, line 47: No need to add sorting in the js code? FYI: even if query seems to return sorted results all the time, one can not rely on that with postgres

gytis-ivaskevicius (approved):
Not sure why 2 other queries were removed, since builds are passing I assume that they are unused

PR #1503, opened 2024-10-04
test: update deployed envs e2e test to latest sdk version

iccicci:
# Context

- The e2e tests against the deployed environments need to be updated to check the `ws-server`.
- The github action has some legacy custom behavior.
- The logic to build the `.env` file is split between the github action and the `generate-dotenv.sh` file.

# Proposed Solution

Refactored the github action and the `generate-dotenv.sh` file.

The github action now dumps the `.env` file for clarity; github is smart enough to hide the secret mnemonics.

# Test

The tests pass both in [dev-preview](https://github.com/input-output-hk/cardano-js-sdk/actions/runs/11180116614/job/31081165350) and [dev-preprod](https://github.com/input-output-hk/cardano-js-sdk/actions/runs/11180371379/job/31081957121).

### DIFF
--- a/.github/workflows/test-deploy-e2e.yaml
+++ b/.github/workflows/test-deploy-e2e.yaml
@@ -4,21 +4,25 @@ on:
   workflow_dispatch:
     inputs:
       environment:
-        description: 'Environment to run e2e tests against'
+        description: 'Target environment'
         type: choice
         required: true
-        default: 'dev-preprod'
         options:
-          - live-preprod
-          - live-preview
-          - dev-preprod
-          - dev-preview
-          - staging-preprod
-      cluster:
-        description: 'Specific cluster to run e2e tests against'
+          - 'dev'
+          - 'staging'
+          - 'live'
+      network:
+        description: 'Target network'
         type: choice
         required: true
-        default: 'any'
+        options:
+          - 'preview'
+          - 'preprod'
+          - 'sanchonet'
+      cluster:
+        description: 'Target cluster'
+        type: choice
+        required: true
         options:
           - any
           - eu-central-1
@@ -27,7 +31,6 @@ on:
         description: 'Log level'
         type: choice
         required: true
-        default: 'fatal'
         options:
           - fatal
           - error
@@ -35,10 +38,6 @@ on:
           - info
           - debug
           - trace
-
-env:
-  TL_DEPTH: ${{ github.event.pull_request.head.repo.fork && '0' || fromJson(vars.TL_DEPTH) }}
-  TL_LEVEL: ${{ github.event.pull_request.head.repo.fork && 'info' || vars.TL_LEVEL }}
 
 jobs:
   build_and_test:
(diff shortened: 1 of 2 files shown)


packages/e2e/src/scripts/generate-dotenv.sh
```diff
@@ -1,48 +1,79 @@
...
 echo "$envFileContent" > .env
+
+# Dump inputs and outputs
+echo "
+Target environment: ${ENVIRONMENT}
+Target network:     ${NETWORK}
+Target cluster:     ${CLUSTER}
+
+Result .env:"
+cat .env
```
gytis-ivaskevicius, line 79: Lets avoid dumping .env, at some point it may have actual secrets there

gytis-ivaskevicius (approved):

.github/workflows/test-deploy-e2e.yaml
```diff
@@ -4,21 +4,25 @@ on:
...
-        default: 'dev-preprod'
         options:
-          - live-preprod
-          - live-preview
-          - dev-preprod
-          - dev-preview
-          - staging-preprod
+          - 'dev'
+          - 'staging'
+          - 'live'
```
gytis-ivaskevicius, line 13: maybe we can have field for url instead?

iccicci: Based on my experience this is the more convenient way to give this input from github actions: some choices must be taken based on the networks, some other from the environment.

packages/e2e/src/scripts/generate-dotenv.sh
iccicci: It already has actual secrets, and github hides them. What is the problem you see?

.github/workflows/test-deploy-e2e.yaml
iccicci: More than this, maintaining the three dimensions (network, environment, cluster) is much more easy than maintaining each element of the resulting matrix of possible URLs.

packages/e2e/src/scripts/generate-dotenv.sh
gytis-ivaskevicius: oh right, github takes care of it. Thanks

.github/workflows/test-deploy-e2e.yaml
gytis-ivaskevicius: I was thinking more of a string field for URL, maybe an optional field that takes preference over this matrix?

Use case: when we temporarily deploy something under a different domain

we can leave it as it is for the time being until it causes us problems

iccicci: Nice suggestion! Thank you. I'm working to make this section of input standard and shared between all the test github actions. I agree leaving this as is, since this is not an urgent requirement, but I'll implement it when I'll do the centralized input.
Thank you

PR #1504, opened 2024-10-06
Feat/lw 11589 make auxiliary data and witness available in sign request

mirceahasegan:
# Context

`wallet.finalizeTx` has optional params for `auxiliaryData, witnessSet and isValid`.
`cip30.signTx` did extract this data from the input transaction CBOR, to pass it down to the `finalizeTx` method.
Further more, even if they are included, serializing the transaction body (passed as CBOR) with the `auxiliaryData and witnessSet` (as deserialised core types), might not result in the same CBOR, due to differences in serialization implementation.

# Proposed Solution
cip30.signTx always receives the entire transaction cbor, not just the transaction body.
Extend wallet.finalizeTx() to accept CBOR in the `tx` parameter. When provided, the other optional parameters (witnessSet, auxiliaryData, isValid) are ignored, and the transaction CBOR is passed as is to the witnesser.

### DIFF
--- a/packages/wallet/src/cip30.ts
+++ b/packages/wallet/src/cip30.ts
@@ -465,10 +465,10 @@ const baseCip30WalletApi = (
   },
   signTx: async ({ sender }: SenderContext, tx: Cbor, partialSign?: Boolean): Promise<Cbor> => {
     const scope = new ManagedFreeableScope();
-    logger.debug('signTx');
-    const txDecoded = Serialization.Transaction.fromCbor(Serialization.TxCBOR(tx));
+    logger.debug('signTx', tx);
+    const txCbor = Serialization.TxCBOR(tx);
+    const txDecoded = Serialization.Transaction.fromCbor(txCbor);
     const wallet = await firstValueFrom(wallet$);
-    const hash = txDecoded.getId();
     const coreTx = txDecoded.toCore();
 
     const needsForeignSignature = await requiresForeignSignatures(coreTx, wallet);
@@ -493,9 +493,8 @@ const baseCip30WalletApi = (
           witness: { signatures }
         } = await signOrCancel(
           wallet.finalizeTx({
-            bodyCbor: txDecoded.body().toCbor(),
             signingContext: { sender },
-            tx: { ...coreTx, hash }
+            tx: txCbor
           }),
           confirmationResult,
           () => new TxSignError(TxSignErrorCode.UserDeclined, 'user declined signing tx')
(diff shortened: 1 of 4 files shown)


packages/wallet/src/cip30.ts
```diff
@@ -465,10 +465,10 @@ const baseCip30WalletApi = (
   },
   signTx: async ({ sender }: SenderContext, tx: Cbor, partialSign?: Boolean): Promise<Cbor> => {
     const scope = new ManagedFreeableScope();
-    logger.debug('signTx');
-    const txDecoded = Serialization.Transaction.fromCbor(Serialization.TxCBOR(tx));
+    logger.debug('signTx', tx);
```
mirceahasegan, line 468: @rhyslbw , @AngelCastilloB . Is there any issue if we log the cbor we're about to sign? It has been very helpful in my debugging.

AngelCastilloB (commented):
@mirceahasegan The finalize method only passes the body as CBOR because its the only thing it needs to sign the transaction properly (from CIP-30 endpoint). The transaction body has the hash of the original aux data and the script integrity hash, hashes the relevant parts of the witness set (that would alter the hash in the body), so its safe to pass the aux data and witness set in non CBOR, the caller of the sign transaction endpoint has the original aux data and witness set anyways, and we only return the signatures. When the wallet signs an internal transaction, the transaction is in Core types so it does reconstruct them but at that point is okay because there are not signatures yet.

But, were you having any particular issue? what problem does this PR tries to solve?

packages/wallet/src/cip30.ts
rhyslbw: No issues, this is suggested

AngelCastilloB (approved):
LGTM

packages/core/src/Cardano/types/Transaction.ts
```diff
@@ -11,6 +11,7 @@ import { PlutusData } from './PlutusData';
 import { ProposalProcedure, VotingProcedures } from './Governance';
 import { RewardAccount } from '../Address';
 import { Script } from './Script';
+import { Serialization } from '../..';
```
mkazlauskas, line 14: This might cause circular import issues. Anyway I think a better place for the new `isTxBodyWithHash` util is in `wallet` package because that where the `Serialization.TxCBOR | TxBodyWithHash` is used
```suggestion
import type { Serialization } from '../..';
```

mkazlauskas (changes requested):

packages/core/src/Cardano/types/Transaction.ts
mirceahasegan: Moved the implementation in the wallet package: 84396c6

mkazlauskas (approved):
Nice work!

PR #1505, opened 2024-10-07
feat(hardware-ledger): the Ledger key agent now picks the first avaialble device if possible

AngelCastilloB:
# Context

We want to remove the need from the user to pick the ledger device on every use, the Ledger key agent will now first check if there is a connected authorized ledger device to use, if not it will prompt the user to select/authorize one.


mkazlauskas (approved):

iccicci (approved):
LGTM

PR #1508, opened 2024-10-09
fix(cardano-services-client): do not call timout id unref in browsers

iccicci:
# Context

`CardanoWsClient` is calling `timeoutID.unref()`, but it is not supported on browsers.

# Proposed Solution

Check if it is a function before calling it to make it being called only under **Node.js**.


AngelCastilloB (approved):
LGTM

rhyslbw (approved):

PR #1509, opened 2024-10-11
LW-11610 Increase ws client logging verbosity

iccicci:
# Context

Working with @AngelCastilloB on LW-11610 we noticed that the web socket activity can't be inspected in the network tab of devTools.

# Proposed Solution

Increased web socket client log verbosity.


AngelCastilloB (approved):
LGTM

iccicci:
> Using contextLogger would be a slight improvement

Nice suggestion @mkazlauskas , thank you!

Following the RERO philosophy, I'm going to merge this as is, but I'll keep this in mind for the future.

PR #1510, opened 2024-10-11
fix: project cip68 metadata when datum resides in tx witness

mkazlauskas:
# Context

Some CIP-68 NFT metadata is not being projected LW-11682

Some IPFS URIs are not recognized LW-11673

# Proposed Solution

Some CIP-68 NFTs are created by using transaction witness datum instead of inline datum.

Mappers will now look up witness datum by hash.

https://input-output.atlassian.net/browse/LW-11682


iccicci (dismissed):
LGTM

mchappell (approved):
Looks great!

iccicci (approved):

PR #1513, opened 2024-10-15
chore: update enpoint to ready

bernokl:
# Context

Updating backend and chain-history health endpoings to use `/ready`


bernokl:
The testing on dev-preview worked as expected.
a) Shutting down core, caused backend and chain-history to restart pods.
b) Starting core back up. Both chain-history and backend restarted correctly.

Note: I significantly reduced the timout on the backend, current timout value was extended because of load issues in the past. We will need to monitor this closely once it starts taking live-mainnet traffic.

iccicci (approved):
LGTM

gytis-ivaskevicius (approved):

PR #1514, opened 2024-10-19
fix(wallet): one ledger-tip req per pollInterval

mirceahasegan:
# Context

More ledger-tip requests per polling interval occur.

# Proposed Solution

WHAT: replaced `delay` operator with `auditTime` operator, which starts a timer once the first isSettled$ emission, and emits the last value, ignoring previous ones.
WHY: isSettled$ observable emissions triggered ledger-tip requests, delayed by pollingInterval.
Flapping isSettled$ caused multiple requests to be done, because the flapping was delayed.
Throttling is in place for the case where the ledger-tip request is not resolved yet.

### DIFF
--- a/packages/wallet/src/services/TipTracker.ts
+++ b/packages/wallet/src/services/TipTracker.ts
@@ -5,9 +5,9 @@ import {
   EMPTY,
   Observable,
   Subject,
+  auditTime,
   combineLatest,
   concat,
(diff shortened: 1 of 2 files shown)


mkazlauskas (approved):
Great work! :muscle:

rhyslbw (approved):

PR #1516, opened 2024-10-22
LW-11470 Blockfrost based backend

iccicci:
# Context

In order to facilitate the Blockfrost migration we need new Blockfrost providers based `provider-server`.

# Proposed Solution

Added the `blockfrost-provider-server` which loads Blockfrost providers rather than DbSync providers..

Slightly changed same deserialization function to not add some `undefined` keys to make transaction deserialization compliant with `ChainHistoryProvider.transactionsByAddresses` output.


ginnun:
@iccicci Does this include my whole PR?

ginnun (approved):

packages/core/src/Serialization/AuxiliaryData/AuxiliaryData.ts
```diff
@@ -228,10 +228,13 @@ export class AuxiliaryData {
    */
   toCore(): Cardano.AuxiliaryData {
     const scripts = this.#getCoreScripts();
-    return {
-      blob: this.#metadata ? this.#metadata.toCore() : undefined,
-      scripts: scripts.length > 0 ? scripts : undefined
+    const auxiliaryData: Cardano.AuxiliaryData = {
+      blob: this.#metadata ? this.#metadata.toCore() : undefined
```
mirceahasegan, line 232: shouldn't blob also be omitted when undefined?
```diff
```

ginnun: @mirceahasegan `blob: undefined` is the expected behavior compared to the existing Providers.

packages/cardano-services/src/ChainHistory/BlockrostChainHistoryProvider/BlockfrostChainHistoryProvider.ts
```diff
@@ -87,65 +107,101 @@ export class BlockfrostChainHistoryProvider extends BlockfrostProvider implement
...
+          pledge: BigInt(pledge),
+          relays: [],
+          rewardAccount: reward_account as Cardano.RewardAccount,
+          vrf: vrf_key as Cardano.VrfVkHex
+        }
+      }))
+    );
+  }
+
+  async fetchCBOR(hash: string): Promise<string> {
```
mirceahasegan, line 140: just nitpicking proposal: use the `TransactionId` type instead of `string`

packages/cardano-services/package.json
```diff
@@ -29,7 +29,7 @@
     "cleanup": "rm -rf dist node_modules",
     "cli": "ts-node --transpile-only src/cli.ts",
     "compose:single:up": "yarn compose:up cardano-node ogmios postgres",
-    "compose:up": "__FIX_UMASK__=$(chmod -R a+r ../../compose/placeholder-secrets) docker compose --env-file environments/.env.$NETWORK -p cardano-services-$NETWORK -f docker-compose.yml -f ../../compose/common.yml -f ../../compose/pg-agent.yml -f ../../compose/$(uname -m).yml ${FILES:-}  --profile ${DOCKER_COMPOSE_PROFILE:-none} up",
```
mirceahasegan, line 32: Is the strategy to always start blockfrost locally when starting cardano-services?

iccicci: Yes, it is required to test Blockfrost integration also in real networks.

packages/cardano-services/src/ChainHistory/BlockrostChainHistoryProvider/BlockfrostChainHistoryProvider.ts
iccicci: I tried... but this code is based on Blockfrost client package, not on our core types: doing that would require to add a lot of type casts.

PR #1518, opened 2024-10-25
fix(wallet): preserve pouchdb collection documents when bulkDocs fails

mkazlauskas:
# Context

PouchDbCollectionStore.setAll used to
1. delete all documents
2. insert the new documents

This approach has a problem that when step 2. fails, all documents from db are gone and not re-created. For some collections such as transactions or utxo this is not a problem, because it can recover, but for it is a big problem for WalletRepository, because there is no way to recover the wallets if they are lost.

LW-11760

# Proposed Solution

This fix updates setAll to perform the following steps:
1. delete *only* the documents that are intended to be deleted
2. upsert all new documents

### DIFF
--- a/packages/wallet/package.json
+++ b/packages/wallet/package.json
@@ -81,7 +81,8 @@
     "pouchdb": "^7.3.0",
     "rxjs": "^7.4.0",
     "ts-custom-error": "^3.2.0",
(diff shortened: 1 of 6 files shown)


mirceahasegan (approved):
nice catch!

PR #1520, opened 2024-10-31
fix(cardano-services-client): txSubmitApiProvider now computes the tx id correctly

AngelCastilloB:
# Context

TxSubmitApiProvider was trying to deserialize the transaction CBOR as transaction body, which was causing a failure in deserialization.

# Proposed Solution

TxSubmitApiProvider now deserializes the CBOR as a transaction when computing the tx hash.


vetalcore (approved):

iccicci (approved):
My fault! Thank you @AngelCastilloB

PR #1521, opened 2024-11-04
fix(cardano-services-client): comply HttpProvider with 'normal' Object behavior

mkazlauskas:
# Context

HttpProvider is implemented as a Proxy object, which is currently not compatible with SDK's web-extension messaging utilities that are checking exposed/underlying object properties with 'in' operator

LW-11796

# Proposed Solution

This commit has 2 fixes:
- do not throw an error when trying to access an undefined property of an object
- return 'true' with with 'in' operator for defined properties

### DIFF
--- a/packages/cardano-services-client/src/HttpProvider.ts
+++ b/packages/cardano-services-client/src/HttpProvider.ts
@@ -78,11 +78,9 @@ export const createHttpProvider = <T extends Provider>({
   new Proxy<T>({} as T, {
     // eslint-disable-next-line sonarjs/cognitive-complexity
     get(_, prop) {
-      if (prop === 'then') return;
       const method = prop as keyof T;
       const urlPath = paths[method];
-      if (!urlPath)
-        throw new ProviderError(ProviderFailure.NotImplemented, `HttpProvider missing path for '${prop.toString()}'`);
+      if (!urlPath) return;
       return async (...args: any[]) => {
         try {
           const req: AxiosRequestConfig = {
@@ -123,5 +121,9 @@ export const createHttpProvider = <T extends Provider>({
           throw new ProviderError(ProviderFailure.Unknown, error);
         }
       };
+    },
+
+    has(_, prop) {
+      return prop in paths;
     }
   });
--- a/packages/cardano-services-client/test/HttpProvider.test.ts
+++ b/packages/cardano-services-client/test/HttpProvider.test.ts
@@ -63,10 +63,16 @@ describe('createHttpProvider', () => {
     if (closeServer) await closeServer();
   });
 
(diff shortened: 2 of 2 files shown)


pczeglik-iohk (approved):
Code looks good. Happy to approve

packages/cardano-services-client/src/HttpProvider.ts
```diff
@@ -78,11 +78,9 @@ export const createHttpProvider = <T extends Provider>({
   new Proxy<T>({} as T, {
     // eslint-disable-next-line sonarjs/cognitive-complexity
     get(_, prop) {
-      if (prop === 'then') return;
```
mirceahasegan, line 81: why is this not needed anymore? I suppose it was handling waiting for a promise to resolve

packages/cardano-services-client/src/HttpProvider.ts
```diff
@@ -78,11 +78,9 @@ export const createHttpProvider = <T extends Provider>({
...
       const method = prop as keyof T;
       const urlPath = paths[method];
-      if (!urlPath)
-        throw new ProviderError(ProviderFailure.NotImplemented, `HttpProvider missing path for '${prop.toString()}'`);
+      if (!urlPath) return;
```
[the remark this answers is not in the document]
mkazlauskas: We don't want to remove a function, because we want this behavior
```ts
let obj = {};
typeof obj.undefinedProp === 'undefined'
// instead of
typeof obj.undefinedProp === 'function'
```

packages/cardano-services-client/src/HttpProvider.ts
mirceahasegan: oh, got it. It's going to be handled by the missing prop code path

mkazlauskas: A couple lines below this one does the job, this is redundant now that it doesn't throw (returns undefined instead)

packages/cardano-services-client/test/HttpProvider.test.ts
```diff
@@ -63,10 +63,16 @@ describe('createHttpProvider', () => {
     if (closeServer) await closeServer();
   });
 
-  it('attempting to access unimplemented method throws ProviderError', async () => {
+  it('attempting to access unimplemented method returns undefined', async () => {
     const provider = createTxSubmitProviderClient();
+    expect('doesNotExist' in provider).toBe(false);
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    expect(() => (provider as any).doesNotExist).toThrowError(ProviderError);
+    expect((provider as any).doesNotExist).toBeUndefined();
```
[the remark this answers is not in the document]
iccicci: Previous comment answers this as well

PR #1523, opened 2024-11-07
fix(wallet): include voteDelegation certs in delegationTracker

mirceahasegan:
# Context

Delegation tracker was recently updated to follow voting delegations in addition to stake delegations, but the transactions were filtered only for staking certificates.

# Proposed Solution
Include voting certificates in the transaction filter.

### DIFF
--- a/packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
+++ b/packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
@@ -139,7 +139,13 @@ export const createDelegationTracker = ({
     transactionsTracker,
     rewardAccountAddresses$,
     slotEpochCalc$,
-    [...Cardano.RegAndDeregCertificateTypes, ...Cardano.StakeDelegationCertificateTypes]
+    [
+      ...new Set([
+        ...Cardano.RegAndDeregCertificateTypes,
+        ...Cardano.StakeDelegationCertificateTypes,
+        ...Cardano.VoteDelegationCredentialCertificateTypes
+      ])
+    ]
   ).pipe(tap((transactionsWithEpochs) => logger.debug(`Found ${transactionsWithEpochs.length} staking transactions`)));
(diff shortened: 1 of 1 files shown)


packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
```diff
@@ -139,7 +139,13 @@ export const createDelegationTracker = ({
     transactionsTracker,
     rewardAccountAddresses$,
     slotEpochCalc$,
-    [...Cardano.RegAndDeregCertificateTypes, ...Cardano.StakeDelegationCertificateTypes]
+    [
+      ...new Set([
+        ...Cardano.RegAndDeregCertificateTypes,
+        ...Cardano.StakeDelegationCertificateTypes,
+        ...Cardano.VoteDelegationCredentialCertificateTypes
```
mkazlauskas, line 146: In general `fix:` should have some test (or a change in existing test) and this looks like something testable. Let me know if I'm mistaken and it's not practical

mkazlauskas (changes requested):

packages/wallet/src/services/DelegationTracker/DelegationTracker.ts
mirceahasegan: I agree and spent some time on it. 
What would be needed is a set of unit tests for `createDelegationTracker`, but given the complexity of the dependencies, one was never created.
The bits and pieces, like `certificateTransactionsWithEpochs` are unit tested, but in this case it's just a matter of certificates being filtered on, passed as an array to `certificateTransactionsWithEpochs`.
So I found it impractical and out of scope to add unit tests for `createDelegationTracker` for this small change.

mkazlauskas (approved):
Nice work! :detective:

AngelCastilloB (approved):

PR #1526, opened 2024-11-12
feat(wallet): input resolver now searches TX history if input cant be found in current UTXO set

AngelCastilloB:
# Context

Input resolver currently can only resolve inputs as long as they are part of the UTXO set of the wallet. However if we want to resolve inputs that were already spent (I.E to render tx history), we need to also search in TX history.

# Proposed Solution

Input resolver now searches tx history.

### DIFF
--- a/packages/wallet/src/services/WalletUtil.ts
+++ b/packages/wallet/src/services/WalletUtil.ts
@@ -14,6 +14,7 @@ export interface InputResolverContext {
     available$: Observable<Cardano.Utxo[]>;
   };
   transactions: {
+    history$: Observable<Cardano.HydratedTx[]>;
     outgoing: {
       signed$: Observable<WitnessedTx[]>;
     };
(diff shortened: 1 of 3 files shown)


packages/wallet/src/services/WalletUtil.ts
```diff
@@ -63,6 +65,16 @@ export const createInputResolver = ({ utxo, transactions }: InputResolverContext
         return tx.body.outputs[input.index];
       }
     }
+
+    // Input resolver is also used to resolve inputs for historical view, so it is possible that client code is trying to resolve an input that
+    // was already spent, we need to look at the tx history to find it. We search here as last resort since this is more expensive
+    // that searching through the UTXO set for wallets with big TX History.
+    const historyTx = txHistory.find((entry) => entry.id === input.txId);
```
AngelCastilloB, line 72: We probably also want to search here in reverse order (most recent TXs first), since most of the time users will be inspecting recent transactions, but I am not sure if this history is sorted/how it is sorted, also sorting it could be more expensive than simply searching

mkazlauskas (approved):
Nice work! :rocket:

packages/wallet/src/services/WalletUtil.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1526/commits/fb0f200f4f5f65fc2c60c23f0e704c9037b34b4f

mkazlauskas (approved):

rhyslbw (approved):
🙌

PR #1528, opened 2024-11-13
feat: browser compatible BlockfrostAssetProvider

mkazlauskas:
# Context

BlockfrostAssetProvider only works in nodejs

# Proposed Solution

replace BlockfrostAssetProvider in cardano-services with a new one that works in the browser

implementation was hoisted from 'lace' repo

BREAKING CHANGE: hoist BlockfrostAssetProvider to cardano-services-client


rhyslbw (approved):

packages/cardano-services/src/util/BlockfrostProvider/BlockfrostClientFactory.ts
mirceahasegan, line 1: shouldn't these helpers be part of cardano-services-client? Users of the provider would then only have to import `cardano-services-client` to use it

mkazlauskas: Right, clients are not supposed to import `cardano-services`. This is really application config so I don't see much value for other consumers. Maybe the constants could be useful to export from `cardano-services-client` (default urls/rate limiter config).

mkazlauskas: Hoisted the constants https://github.com/input-output-hk/cardano-js-sdk/pull/1528/commits/60d11905fce0d4d313b038cf7923f1a713a72a60

mirceahasegan (approved):
Fantastic work @mkazlauskas 🚀

PR #1530, opened 2024-11-14
Feat/lw 11842 performance improvements

AngelCastilloB:
# Context

Currently transactions are always compared (in quadratic time) and stored. This one is probably the main culprit of the performance issues in wallets with high TX history count. The Wallet fetches the transaction from last known slot up to tip, then it has some heuristic to determine if there were changes (such as rollbacks or new transactions), and if so applies the changes, after that it stores the transactions and returns them. It has a guard to avoid emitting the same transactions several time by comparing them to last emitted. It is using an algorithm of complexity O(N x M).

Additionally, Lace fetches the whole UTXO set of the wallet on every tip update.

- [x] JIRA - \<[link](https://input-output.atlassian.net/browse/LW-11849)>
- [x] JIRA - \<[link](https://input-output.atlassian.net/browse/LW-11850)>

# Proposed Solution

- Only compare new transactions with the latest known transaction of the same block. The comparison between the transactions is now done in linear time O(N).
- Only fetch UTXOs when there are changes to the transaction history.

### DIFF
--- a/packages/wallet/src/services/TransactionsTracker.ts
+++ b/packages/wallet/src/services/TransactionsTracker.ts
@@ -38,9 +38,7 @@ import { distinctBlock, signedTxsEquals, transactionsEquals, txInEquals } from '
 import { WitnessedTx } from '@cardano-sdk/key-management';
 import { newAndStoredMulticast } from './util/newAndStoredMulticast';
 import chunk from 'lodash/chunk.js';
-import intersectionBy from 'lodash/intersectionBy.js';
 import sortBy from 'lodash/sortBy.js';
-import unionBy from 'lodash/unionBy.js';
 
 export interface TransactionsTrackerProps {
   chainHistoryProvider: ChainHistoryProvider;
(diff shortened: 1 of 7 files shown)


packages/wallet/src/services/TransactionsTracker.ts
```diff
@@ -172,8 +173,16 @@ export const createAddressTransactionsProvider = ({
                   // Rollback by 1 block, try again in next loop iteration
                   localTransactions = localTransactions.filter(({ blockHeader: { blockNo } }) => blockNo < lowerBound);
                 } else {
-                  localTransactions = unionBy(localTransactions, newTransactions, (tx) => tx.id);
-                  store.setAll(localTransactions);
+                  const lastLocalTxs = localTransactions.slice(-newTransactions.length);
+                  const areTransactionsSame =
+                    lastLocalTxs.length === newTransactions.length &&
+                    lastLocalTxs.every((tx, index) => tx.id === newTransactions[index].id);
```
mirceahasegan, line 179: This comparison could be skipped entirely by checking the length of the already computed `duplicateTransactions`.
If it has the same length as newTransactions, it means that all new transactions are already part of the localTransactions.
In this case, it's not even necessary to compare the length anymore.
```diff
- lastLocalTxs.length === newTransactions.length &&
- lastLocalTxs.every((tx, index) => tx.id === newTransactions[index].id);
+ newTransactions.length === duplicateTransactions.length;
```

packages/wallet/src/services/TransactionsTracker.ts
```diff
@@ -160,6 +160,7 @@ export const createAddressTransactionsProvider = ({
                 );
                 const duplicateTransactions =
                   lastStoredTransaction && intersectionBy(localTransactions, newTransactions, (tx) => tx.id);
```
mirceahasegan, line 162: intersectionBy will be O(n*m)

mirceahasegan: `newTransactions` are transactions starting from the `lastStoredTransaction` blockNo, up to the tip.
So to detect a rollback it is sufficient to check if `lastStoredTransaction` is the first transaction in `newTransactions`. There is no need for intersectionBy between `localTransactions` and `newTransactions`.
```diff
-                const duplicateTransactions =
-                  lastStoredTransaction && intersectionBy(localTransactions, newTransactions, (tx) => tx.id);
+                const lastTxRolledBack =
+                  lastStoredTransaction && newTransactions[0].id !== lastStoredTransaction.id;
```

packages/wallet/src/services/TransactionsTracker.ts
```diff
@@ -160,6 +160,7 @@ export const createAddressTransactionsProvider = ({
...
+
                 if (typeof duplicateTransactions !== 'undefined' && duplicateTransactions.length === 0) {
                   const rollbackTransactions = localTransactions.filter(
```
mirceahasegan, line 165: Following the logic from prev comment, we rollback one transaction at a time, so:
```diff
- const rollbackTransactions = localTransactions.filter(
+ const rollbackTransactions = [lastStoredTransaction]
```

packages/wallet/src/services/TransactionsTracker.ts
```diff
@@ -172,8 +173,16 @@ export const createAddressTransactionsProvider = ({
                   // Rollback by 1 block, try again in next loop iteration
                   localTransactions = localTransactions.filter(({ blockHeader: { blockNo } }) => blockNo < lowerBound);
```
mirceahasegan, line 174: And...
```diff
- localTransactions = localTransactions.filter(({ blockHeader: { blockNo } }) => blockNo < lowerBound);
+ localTransactions = localTransactions.filter(({ blockHeader: { blockNo } }) => blockNo < lowerBound);
```

packages/wallet/src/services/TransactionsTracker.ts
```diff
@@ -127,6 +125,7 @@ export const createAddressTransactionsProvider = ({
         )
       ),
       combineLatest([addresses$, storedTransactions$.pipe(defaultIfEmpty([] as Cardano.HydratedTx[]))]).pipe(
+        // eslint-disable-next-line sonarjs/cognitive-complexity
```
mkazlauskas, line 128: Can we hoist some logic to separate functions?

mirceahasegan: Done here: [d243142](https://github.com/input-output-hk/cardano-js-sdk/pull/1530/commits/d243142eda345212e22f501f14dac1dce2f9df30)

mirceahasegan:
Added more unit tests for rollback behaviour here: [6d0e216](https://github.com/input-output-hk/cardano-js-sdk/pull/1530/commits/6d0e21681fbbc1819dcbd462c13951dec72febe4)

packages/wallet/src/services/TransactionsTracker.ts
```diff
@@ -152,13 +152,15 @@ const findIntersectionAndUpdateTxStore = ({
           lowerBound !== undefined && `since block ${lowerBound}`
         );
 
-        const txRolledBack = lastStoredTransaction && newTransactions[0]?.id !== lastStoredTransaction.id;
+        const txRolledBack =
+          lastStoredTransaction && newTransactions[newTransactions.length - 1]?.id !== lastStoredTransaction.id;
```
mirceahasegan, line 156: I think this should be looking only at the first block from `newTransactions`
```diff
- lastStoredTransaction && newTransactions[newTransactions.length - 1]?.id !== lastStoredTransaction.id;
+ const newTxFirstBlock = newTransactions.filter(tx => tx.blockHeader.blockNo === lowerBound);
+ lastStoredTransaction && newTxFirstBlock[newTxFirstBlock.length - 1]?.id !== lastStoredTransaction.id;
```

AngelCastilloB: I am rewriting this logic here, sec

packages/wallet/test/services/TransactionsTracker.test.ts
```diff
@@ -280,6 +280,42 @@ describe('TransactionsTracker', () => {
         pagination: { limit: 25, startAt: 0 }
       });
     });
+
+    describe('distinct transaction sets in latest stored block vs new blocks', () => {
```
mirceahasegan, line 284: @mkazlauskas , @AngelCastilloB , please review these proposed scenarios to cover in unit tests.
If they make sense and improve coverage I will implement them.

mkazlauskas: Looks good, just don't quite understand when/how this could happen `reversed order transactions plus new tx are re-emitted, but not considered rollbacks`

mirceahasegan: It is far fetched, but I was thinking of a scenario where a competing SP has txId 3 in the mempool, then gets a mempool update with txId 2, then another with txId 1, and creates a fork with the 3 transactions <3 2 1>.
I'm not sure if that is even possible tbh 😅

AngelCastilloB: Looks great @mirceahasegan , I think we just need to add the case where there are no changes:

```
// latestStoredBlock  <1 2 3>
// fetched Txs            <1 2 3>
```

Or is it that this is already covered by the already existing tests?

mkazlauskas: Thanksk @mirceahasegan , better safe than sorry :sweat_smile:

mirceahasegan: I would say it's already covered in the other tests, for single transaction blocks.
I'll add one for the multi-transaction block scenario too, just to be safe.

mirceahasegan: Unit tests are implemented in ef247eaf2c51547162ec4f997ec4d34b600970c0.
The commit also contains a fix for an issue discovered in tx tracker.
There still are ~3 tests that are failing and we need to investigate if they're bugs or test issues after updating implementation

mirceahasegan (approved):
Fantastic work @AngelCastilloB 🚀

pczeglik-iohk (approved):

PR #1535, opened 2024-11-23
fix(wallet): transaction tracker now process transaction in chronological order

AngelCastilloB:
# Context

There was a wrong assumption that transactions would return from the backend sorted by slot. However, this is not the case. We rely on transaction order for the new optimized algorithm, which operates under the assumption that transactions are sorted in ascending order.

# Proposed Solution

After fetching all transactions, sort them by slot.


rhyslbw (dismissed):
Nice work @AngelCastilloB

rhyslbw (approved):

PR #1537, opened 2024-11-25
test(e2e): fix cip25 v2 test

mkazlauskas:
# Context

it was setting version as part of the map under <asset_name> but the spec specifies that it has to be under the '721' map

image

# Proposed Solution


rhyslbw (approved):

rhyslbw:
This can be merged now right @mkazlauskas?

PR #1541, opened 2024-12-05
Feat/lw 11939 improve error reporting on providers

AngelCastilloB:
# Context

Currently coldObservableProvider stays in an infinity loop retrying when there are provider errors. Errors are also not logged.

# Proposed Solution

coldObservableProvider now stops retrying after maxRetries has been reached (max 10), and logs errors.


packages/util-rxjs/test/coldObservableProvider.test.ts
```diff
@@ -8,11 +9,24 @@ jest.mock('backoff-rxjs', () => ({
   retryBackoff: jest.fn().mockImplementation((...args) => jest.requireActual('backoff-rxjs').retryBackoff(...args))
 }));
 
+const createMockLogger = (): Logger => ({
```
mirceahasegan, line 12: There's a test logger you can use in `util-dev`. One usage example is in [ogmios](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/ogmios/test/util.test.ts#L27) tests

packages/util-rxjs/test/coldObservableProvider.test.ts
```diff
@@ -8,11 +9,24 @@ jest.mock('backoff-rxjs', () => ({
...
+  trace: jest.fn(),
+  warn: jest.fn()
+});
+
 describe('coldObservableProvider', () => {
   it('returns an observable that calls underlying provider on each subscription and uses retryBackoff', async () => {
     const underlyingProvider = jest.fn().mockResolvedValue(true);
     const backoffConfig: RetryBackoffConfig = { initialInterval: 1 };
-    const provider$ = coldObservableProvider({ provider: underlyingProvider, retryBackoffConfig: backoffConfig });
+    const logger = createMockLogger();
```
mirceahasegan, line 24: logger object should be created in `beforeEach`

packages/util-rxjs/test/coldObservableProvider.test.ts
rhyslbw, line 12: You could use https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/util-dev/src/TestLogger.ts

rhyslbw (commented):
Nice work @AngelCastilloB

packages/util-rxjs/src/coldObservableProvider.ts
```diff
@@ -58,16 +62,33 @@ export const coldObservableProvider = <T>({
...
+
+                  if (!shouldRetry) {
+                    return false;
+                  }
+                }
 
                 if (error instanceof InvalidStringError) {
                   onFatalError?.(error);
                   cancelOnFatalError$.next(true);
+                  return false;
```
mirceahasegan, line 81: This is not be required (unless there's a bug).
When `cancelOnFatalError$` emits, it causes `internalCancel$` to emit, which will complete the observable.

mirceahasegan: I think we can drop the `cancelOnFatalError$` observable. It will cancel anyway because we return false, which will cause the retry mechanism to throw.

packages/wallet/src/Wallets/BaseWallet.ts
```diff
@@ -426,6 +428,7 @@ export class BaseWallet implements ObservableWallet {
       coldObservableProvider({
         cancel$,
         equals: deepEquals,
+        logger: this.#logger,
```
mirceahasegan, line 431: Just a suggestion: add context so we know which tracker is printing the errors
```diff
-        logger: this.#logger,
+        logger: contextLogger(this.#logger, 'eraSummaries$'),
```

mirceahasegan (dismissed):

packages/util-rxjs/test/coldObservableProvider.test.ts
mirceahasegan: done here: 867945df0b58d37b9a1955b794e8950cfcf5b743

mirceahasegan: done here: 867945df0b58d37b9a1955b794e8950cfcf5b743

packages/util-rxjs/test/coldObservableProvider.test.ts
mirceahasegan: done here: https://github.com/input-output-hk/cardano-js-sdk/commit/867945df0b58d37b9a1955b794e8950cfcf5b743

mkazlauskas (approved):
Looks good! Please update commit message and PR title to indicate that this is only about logging before merging

rhyslbw (approved):

PR #1542, opened 2024-12-05
Explain why it is not necessary to call setStatInitialized() for TrackedDrepProvider.

mirceahasegan:
# Context

Explain why it is not necessary to call setStatInitialized() for TrackedDrepProvider.

### DIFF
--- a/packages/wallet/src/Wallets/BaseWallet.ts
+++ b/packages/wallet/src/Wallets/BaseWallet.ts
@@ -525,6 +525,10 @@ export class BaseWallet implements ObservableWallet {
       utxoProvider: this.utxoProvider
     });
 
+    // When the rewardAccounts delegation tracker is initialized, it will invoke drepInfo$
+    // with an empty array, since there aren't any drepDelegatees yet.
+    // This will not perform network requests, but will initialize the TrackedDrepProvider.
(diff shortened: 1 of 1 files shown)


rhyslbw (approved):

packages/wallet/src/services/DrepInfoTracker.ts
```diff
@@ -1,28 +1,35 @@
...
-export const createDrepInfoColdObservable =
-  ({ drepProvider, retryBackoffConfig, refetchTrigger$, logger }: DrepInfoObservableProps) =>
-  (drepIds: Cardano.DRepID[]) =>
+export const createDrepInfoColdObservable = ({
+  drepProvider,
+  retryBackoffConfig,
+  refetchTrigger$,
+  logger
+}: DrepInfoObservableProps) => {
+  drepProvider.setStatInitialized(drepProvider.stats.getDRepInfo$);
```
mkazlauskas, line 24: It would be better to set it as initialized only if this tracker makes 0 requests during initialization (no drep ids). This fix is better than nothing though, happy to merge this first

mkazlauskas (dismissed):

mkazlauskas (approved):

rhyslbw (approved):

PR #1545, opened 2024-12-10
perf(wallet): cache mapped addresses and utxo

mkazlauskas:
# Context

wallets with large # of utxo were performing poorly, because every subscription to utxo ran some computation

notably, each call to input resolver was running `firstValueFrom(utxo.available$)` which triggered a series of computations

# Proposed Solution


rhyslbw (approved):
LGTM, nice work @mkazlauskas 🕵️

AngelCastilloB (approved):
Great work @mkazlauskas 🧙🏼

PR #1548, opened 2024-12-13
fix(core): include withdrawals in txSummary spent coins

mirceahasegan:
Withdrawals are a type of "own input", which must be accounted for when computing the wallet spent amount

# Context

txSummaryInspector should factor in `withdrawals` when calculating the spent `coins`.

# Proposed Solution
`coins = ownOutputs - (ownInputs + ownWithdrawals)

### DIFF
--- a/packages/core/src/util/transactionSummaryInspector.ts
+++ b/packages/core/src/util/transactionSummaryInspector.ts
@@ -37,6 +37,11 @@ interface TransactionSummaryInspectorArgs {
 
 export type TransactionSummaryInspection = {
   assets: Map<Cardano.AssetId, AssetInfoWithAmount>;
+  /**
+   * Spent amount from the wallet's perspective, computed as `own outputs - (own inputs + withdrawals)`.
+   * Withdrawals are a type of "own input", and are accounted for when computing the wallet spent amount.
+   * Positive when the wallet is receiving funds, negative when the wallet is sending funds.
+   */
   coins: Cardano.Lovelace;
   collateral: Cardano.Lovelace;
   deposit: Cardano.Lovelace;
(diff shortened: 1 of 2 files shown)


packages/core/src/util/transactionSummaryInspector.ts
```diff
@@ -201,14 +201,19 @@ export const transactionSummaryInspector: TransactionSummaryInspector =
...
     const totalOutputValue = await totalAddressOutputsValueInspector(addresses)(tx);
     const totalInputValue = await totalAddressInputsValueInspector(addresses, inputResolver)(tx);
-    const implicitCoin = (implicit.withdrawals || 0n) + (implicit.reclaimDeposit || 0n) - (implicit.deposit || 0n);
+    const implicitCoin = withdrawals + (implicit.reclaimDeposit || 0n) - (implicit.deposit || 0n);
     const implicitAssets = await getImplicitAssets(tx);
 
     const diff = {
       assets: subtractTokenMaps([totalOutputValue.assets, totalInputValue.assets]),
-      coins: totalOutputValue.coins - totalInputValue.coins
+      // Withdrawals are a type of "own input", which must be accounted for when computing the wallet spent amount.
```
rhyslbw, line 212: Nice description. My first though here is that it would be better added to a type definition, so developers will see it when implementing, but it looks like we're just relying on inference here?

rhyslbw (approved):
Great work @mirceahasegan!

packages/core/src/util/transactionSummaryInspector.ts
mirceahasegan: good point. I'll create an explicit type and add jsdocs for it

mirceahasegan: I amended the commit, the only addition being the jsdoc: https://github.com/input-output-hk/cardano-js-sdk/pull/1548/commits/6166bfc973ddfbd947f99f30e69c9aa31eeec29b#diff-a9bcb5ce8c787fb688c6ba5c62595afae666260d468bd5b2ee1bd9a22e868b51R40

packages/core/test/util/transactionSummaryInspector.test.ts
```diff
@@ -271,8 +271,15 @@ describe('Transaction Summary Inspector', () => {
             ]),
             coins: 6_000_000n
           }
+        },
+        {
+          address: addresses[0],
+          value: {
+            coins: 3_000_000n
+          }
```
mkazlauskas, line 279: It would be nice to demonstrate how a withdrawal can change the amount of `coins` in the inspection
```suggestion
          address: addresses[0],
          value: {
            coins: 2_000_000n
          }
        },
        {
          address: externalAddress1,
          value: { coins: 1_000_000n }
```

mkazlauskas (changes requested):

packages/core/test/util/transactionSummaryInspector.test.ts
mirceahasegan: Updated here: [3897f94](https://github.com/input-output-hk/cardano-js-sdk/pull/1548/commits/3897f945eea3f4da83fb9e9d52aa99884cb54338)

mchappell (approved):

PR #1549, opened 2024-12-16
fix(cardano-services-client): correct blockfrost tx submit request body

mkazlauskas:
# Context

it has to be binary, otherwise Blockfrost responds with 400

this bug wasn't caught initially because we're not using BlockfrostTxSubmitProvider in our e2e test suite

LW-11695


pczeglik-iohk (approved):
Looks good!

mirceahasegan (approved):

PR #1552, opened 2025-01-03
fix: map Blockfrost ValueNotConserved and OutsideOfValidityInterval errors

mkazlauskas:
# Context

BaseWallet has some logic based on TxSubmissionError. Therefore we need to map errors from BlockfrostTxSubmitProvider

# Proposed Solution

Map 2 most important errors:
- ValueNotConserved
- OutsideOfValidityInterval

**Note: for completeness we should follow-up with more error mappings, see [BaseWallet implementation](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/wallet/src/Wallets/BaseWallet.ts#L719-L724)**

LW-11912


rhyslbw (approved):
Nice work @mkazlauskas

PR #1553, opened 2025-01-05
fix(web-extension): send large messages in smaller chunks

mkazlauskas:
# Context

Wallets with a large tx history causes the background process to crash. There is a message size limit in web extension messaging.

# Proposed Solution

Split large messages into into smaller chunks


rhyslbw:
Should we close this @mkazlauskas ?

PR #1555, opened 2025-01-09
fix(tx-constructor): txBuilder ctor must be sync to be used with remo…

mirceahasegan:
…teApi

RemoteApi does not support async factory methods.
Promise rejection will not be handled.

# Context

firstValueFrom throws emptyValue due to observable completing without emitting any values.
The exception is not caught and forwarded as `uncaughtError`.

# Proposed Solution
Ensure TxBuilder constructor is synchronous to be compatible with remoteApi


mkazlauskas (approved):
🕵️

AngelCastilloB (approved):

PR #1558, opened 2025-01-14
feat: remove async from crypto API

### LINKED ISSUE #1529: remove the use of await sodium.ready
### Describe the feature you'd like

To make keys functions non-async.

Rewrite `packages/crypto/src/Bip32/Bip32PrivateKey.ts` and `packages/crypto/src/Bip32/Bip32PublicKey.ts` to remove async when `derive(derivationIndices: number[]): Bip32PublicKey` and `toPublic(): Bip32PublicKey`

### Describe alternatives you've considered
- https://github.com/StricaHQ/bip32ed25519
- https://github.com/HarmonicLabs/bip32_ed25519/tree/main

AngelCastilloB:
# Context

The crypto interface currently is async because it uses libsodium underneath which requires an async call to sodium.ready. We can remove the async from the interface and hoist it to the strategy constructor or a global method async Crypto.ready, that can be called by the client code at their convenience.

Resolves #1529

### DIFF
--- a/packages/core/test/Cardano/util/computeImplicitCoin.test.ts
+++ b/packages/core/test/Cardano/util/computeImplicitCoin.test.ts
@@ -7,6 +7,8 @@ describe('Cardano.util.computeImplicitCoin', () => {
   let dRepPublicKey: Crypto.Ed25519PublicKeyHex;
   let dRepKeyHash: Crypto.Ed25519KeyHashHex;
 
+  beforeAll(() => Crypto.ready());
+
   beforeEach(async () => {
     rewardAccount = Cardano.RewardAccount('stake_test1uqfu74w3wh4gfzu8m6e7j987h4lq9r3t7ef5gaw497uu85qsqfy27');
     stakeCredential = {
@@ -14,7 +16,7 @@ describe('Cardano.util.computeImplicitCoin', () => {
       type: Cardano.CredentialType.KeyHash
     };
     dRepPublicKey = Crypto.Ed25519PublicKeyHex('deeb8f82f2af5836ebbc1b450b6dbf0b03c93afe5696f10d49e8a8304ebfac01');
-    dRepKeyHash = (await Crypto.Ed25519PublicKey.fromHex(dRepPublicKey).hash()).hex();
+    dRepKeyHash = Crypto.Ed25519PublicKey.fromHex(dRepPublicKey).hash().hex();
   });
 
   describe('calculates deposit', () => {
--- a/packages/crypto/src/Bip32/Bip32PrivateKey.ts
+++ b/packages/crypto/src/Bip32/Bip32PrivateKey.ts
@@ -71,6 +71,8 @@ export class Bip32PrivateKey {
    *   - 32 bytes: Ed25519 curve scalar from which few bits have been tweaked according to ED25519-BIP32
    *   - 32 bytes: Ed25519 binary blob used as IV for signing
    *
+   * NOTE: You must await `Crypto.ready()` at least once before calling this function.
+   *
    * @param entropy Random stream of bytes generated from a BIP39 seed phrase.
    * @param password The second factor authentication password for the mnemonic phrase.
    * @returns The secret extended key.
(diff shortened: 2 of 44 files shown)


packages/crypto/src/index.ts
```diff
@@ -7,3 +8,5 @@ export * from './Ed25519e';
 export * from './strategies';
 export * from './hexTypes';
 export * from './types';
+
+export const ready = async (): Promise<void> => await sodium.ready;
```
mkazlauskas, line 12:
```suggestion
export const sodiumReady = async (): Promise<void> => await sodium.ready;
```

packages/crypto/src/strategies/SodiumBip32Ed25519.ts
```diff
@@ -12,77 +12,79 @@ import {
...
 import { HexBlob } from '@cardano-sdk/util';
+import sodium from 'libsodium-wrappers-sumo';
 
 const EXTENDED_KEY_HEX_LENGTH = 128;
 
 export class SodiumBip32Ed25519 implements Bip32Ed25519 {
+  // Prevent instantiation
+  private constructor() {
+    // Empty
+  }
```
mkazlauskas, line 23: Why do we need to export `ready` from index then? Wouldn't it always be initialized as the only way to create is the `async create`?

mkazlauskas (changes requested):

packages/crypto/src/strategies/SodiumBip32Ed25519.ts
AngelCastilloB: There are two ways to use the package, one is to use the strategy, in that case the factory method will take care of the initialization, however, you can use the types directly without the strategy instance, in that case, you must make use its initialized before using those. I mean types such as `Ed25519PrivateKey`

packages/crypto/src/index.ts
AngelCastilloB: I would rather hide the specific implementation detail (that we are using sodium underneath), so ready seems more generic

mkazlauskas: I see, ok that makes sense since sodium is the default one

mkazlauskas (commented):
Looks good! Please update commit message to indicate breaking changes

packages/core/test/Cardano/util/computeImplicitCoin.test.ts
```diff
@@ -7,14 +7,18 @@ describe('Cardano.util.computeImplicitCoin', () => {
   let dRepPublicKey: Crypto.Ed25519PublicKeyHex;
   let dRepKeyHash: Crypto.Ed25519KeyHashHex;
 
+  beforeAll(async () => {
+    await Crypto.ready();
+  });
```
mkazlauskas, line 12: Just FIY: this can be simplified as you just need to return a promise
```suggestion
  beforeAll(() => Crypto.ready());
```

packages/crypto/src/Ed25519e/Ed25519PublicKey.ts
```diff
@@ -30,8 +30,7 @@ export class Ed25519PublicKey {
    * @param message The original message the signature was computed from.
    * @returns true if the signature is valid; otherwise; false.
    */
-  async verify(signature: Ed25519Signature, message: HexBlob): Promise<boolean> {
-    await sodium.ready;
```
mirceahasegan, line 34: Should we mention that `await Crypto.ready()` is a prerequisite before calling this function?

mirceahasegan (approved):
Fantastic work @AngelCastilloB  🚀

mkazlauskas (approved):
Nice work! 🚀

packages/crypto/src/Ed25519e/Ed25519PublicKey.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1558/commits/0d664a6fb44ce29b61a0de3d456d676a8f43bd78

packages/crypto/src/Bip32/Bip32PrivateKey.ts
```diff
@@ -123,11 +125,12 @@ export class Bip32PrivateKey {
    * This is why deriving the private key should not fail while deriving
    * the public key may fail (if the derivation index is invalid).
    *
+   * NOTE: You must call `Crypto.ready()` at least once before calling this function.
```
mkazlauskas, line 128:
```suggestion
   * NOTE: You must await `Crypto.ready()` at least once before calling this function.
```

mkazlauskas (approved):

packages/crypto/src/Bip32/Bip32PrivateKey.ts
AngelCastilloB: done in https://github.com/input-output-hk/cardano-js-sdk/pull/1558/commits/171910859b1703c0958db59676555b1aec52965a

mkazlauskas (approved):

PR #1559, opened 2025-01-16
feat!: fee calculation now takes into account reference script size

AngelCastilloB:
Our fee calculation algorithm currently doesn't take into account script ref byte sizes. This PR makes a few changes to the fee calculation algorithm to include ref script into this calculation


mkazlauskas (approved):
Nice work! 🚀

rhyslbw (approved):
Great work @AngelCastilloB 🧙

PR #1562, opened 2025-01-16
Quick fix for #1561

Quantumplation:
# Context

See #1561

# Proposed Solution

This is a quick fix, which matches the conditionals in `toCbor` to the conditionals in `getMapSize`; See #1561 and commit message for a more robust approach.

# Important Changes Introduced

PR #1563, opened 2025-01-17
fix(core): zero value fields CBOR serialization

mirceahasegan:
In some cases, CBOR serialization not including fields having value 0. Fixed and added unit tests

# Context

A continuation of [SundaeSwap-finance:pi/1561/cbor-issues](https://github.com/input-output-hk/cardano-js-sdk/pull/1562)

# Proposed Solution


mkazlauskas (approved):
Great work! 💪

PR #1568, opened 2025-01-25
Feat/transaction history gap issue

mirceahasegan:
# Context

Fetching a limited number of transactions from all addresses and displaying them aggregated could lead to gaps in the tx history when one transaction has been very active (i.e. 100 transactions in the past year, with 10 of them in the past day), and another transaction is less active (one transaction one year ago).
Fetching 10 per addres and aggregating the results will lead to a history of one transaction one year ago, and 10 transactions in the past day, having a gap of 90 transactions from the active address. 

# Proposed Solution
Update `fetchInitialTransactions`:
- If there is a single address, it returns the most recent ones up to `historicalTransactionsFetchLimit`.
- If there are more than one address, it returns all transaction from all addresses, one month back from the most recent transaction.

Fixed an issue with fetchSequentially which ignored the `PagingOptions.page` arg, leading to an infinite loop.
Removing the positional arg `page` is a breaking change.

### DIFF
--- a/packages/cardano-services-client/src/blockfrost/util.ts
+++ b/packages/cardano-services-client/src/blockfrost/util.ts
@@ -32,17 +32,20 @@ export const fetchSequentially = async <Response, Item = Response>(
     haveEnoughItems?: (allItems: Item[], lastResponseBatch: Response[]) => boolean;
     paginationOptions?: PaginationOptions;
   },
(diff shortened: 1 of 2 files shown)


rhyslbw (approved):
Nice work @AngelCastilloB @mirceahasegan

PR #1569, opened 2025-01-27
Feat/cache data from blockfrost locally

AngelCastilloB:
# Context

We want to reduce the amount of request we make to blockfrost. Now ChainHistoryProvider and UtxoProvider use a local volatile cache to reuse when requesting the same data several times.

### DIFF
--- a/packages/cardano-services-client/src/ChainHistoryProvider/BlockfrostChainHistoryProvider.ts
+++ b/packages/cardano-services-client/src/ChainHistoryProvider/BlockfrostChainHistoryProvider.ts
@@ -34,6 +34,7 @@ type BlockfrostTx = Pick<Responses['address_transactions_content'][0], 'block_he
 const compareTx = (a: BlockfrostTx, b: BlockfrostTx) => a.block_height - b.block_height || a.tx_index - b.tx_index;
 
 export class BlockfrostChainHistoryProvider extends BlockfrostProvider implements ChainHistoryProvider {
+  private readonly cache: Map<string, Cardano.HydratedTx> = new Map();
   private networkInfoProvider: NetworkInfoProvider;
 
   constructor(client: BlockfrostClient, networkInfoProvider: NetworkInfoProvider, logger: Logger) {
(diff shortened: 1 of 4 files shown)


packages/cardano-services-client/src/ChainHistoryProvider/BlockfrostChainHistoryProvider.ts
```diff
@@ -475,11 +476,31 @@ export class BlockfrostChainHistoryProvider extends BlockfrostProvider implement
...
+
+        for (const tx of fetchedTransactions) {
+          this.cache.set(tx.id, tx);
+        }
+      } catch (error) {
+        throw this.toProviderError(error);
+      }
     }
+
+    return [...cachedTransactions, ...fetchedTransactions];
```
mirceahasegan, line 503: Maybe it would be better to return in the same order from the function call.

mirceahasegan (changes requested):

packages/cardano-services-client/src/ChainHistoryProvider/BlockfrostChainHistoryProvider.ts
mirceahasegan:
```
Promise.all(
ids.map(id => {
  if (this.cache.has(id)) return Promise.resolve(this.cache.get(id));
  try { return this.fetchTransaction(id)} catch (error) { throw this.toProviderError(error); }
}
)

Comment (AngelCastilloB) [reply packages/cardano-services-client/src/ChainHistoryProvider/BlockfrostChainHistoryProvider.ts]: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1569/commits/4dce93bdc29ffae9f6e0f3b3f173dce0dc52186f

File: packages/cardano-services-client/src/ChainHistoryProvider/BlockfrostChainHistoryProvider.ts [hunk 2]
```diff
@@ -34,6 +34,7 @@ type BlockfrostTx = Pick<Responses['address_transactions_content'][0], 'block_he
 const compareTx = (a: BlockfrostTx, b: BlockfrostTx) => a.block_height - b.block_height || a.tx_index - b.tx_index;
 
 export class BlockfrostChainHistoryProvider extends BlockfrostProvider implements ChainHistoryProvider {
+  private readonly cache: Map<string, Cardano.HydratedTx> = new Map();
```
mkazlauskas, line 37: It would be better if cache objects were injected (e.g. `type Cache<T> = { get: (key) => Promise<T>, set: (key: string, value: T) => Promise<void> }`), then it would be easier to update/swap it with a more sophisticated cache (lru, persistent etc.) and control it with FFs. Anyway happy to merge as-is to mitigate the current issue quickly.

mirceahasegan (approved):

mkazlauskas (approved):

PR #1574, opened 2025-01-29
fix: [lw-12138] handle disconnected ledger wallet error

vetalcore:
# Context

There are multiple `Failed to execute 'releaseInterface' on 'USBDevice'` issues reported by sentry.
Further investigation confirmed that `hw-transport-webusb` package also attaches a disconnect event listener while creating a Ledger transport, so we might end up trying to handle disconnection multiple times. 

# Proposed Solution
Wrap `transport.close` call in the on disconnect handler with a try catch clause.

### DIFF
--- a/packages/hardware-ledger/src/LedgerKeyAgent.ts
+++ b/packages/hardware-ledger/src/LedgerKeyAgent.ts
@@ -377,12 +377,20 @@ export class LedgerKeyAgent extends KeyAgentBase {
   }
 
   private static attachDisconnectionCleanupHandler(transport: LedgerTransportType) {
-    const onDisconnect = () => {
+    const onDisconnect = async () => {
       transport.off('disconnect', onDisconnect);
       this.deviceConnections = this.deviceConnections.filter(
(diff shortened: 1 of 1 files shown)


mirceahasegan (approved):

packages/hardware-ledger/src/LedgerKeyAgent.ts
```diff
@@ -377,12 +377,18 @@ export class LedgerKeyAgent extends KeyAgentBase {
...
+    const onDisconnect = async () => {
       transport.off('disconnect', onDisconnect);
       this.deviceConnections = this.deviceConnections.filter(
         ({ deviceConnection }) => deviceConnection.transport !== transport
       );
-      void transport.close();
+
+      try {
+        await transport.close();
+      } catch {
```
rhyslbw, line 388: Can the error be inspected to ensure it's the expected failure before ignoring?

vetalcore: addressed

rhyslbw (approved):

mirceahasegan (approved):

PR #1576, opened 2025-01-30
fix: [lw-12169] map pool id in blockfrost rewards provider response

vetalcore:
# Context

Pool id is not mapped in the response of blockfrost rewards providers requests.

# Proposed Solution


rhyslbw (dismissed):

mkazlauskas (approved):
Nice find 🕵️

mirceahasegan (approved):
🚀

PR #1579, opened 2025-01-31
Fix/sort initially fetched txes - LW-12194

mirceahasegan:
# Context

initially fetched trasactions were returned in asc order, but sliced from array start when we actually want the latest txes

# Proposed Solution

### DIFF
--- a/packages/wallet/src/services/TransactionsTracker.ts
+++ b/packages/wallet/src/services/TransactionsTracker.ts
@@ -224,8 +224,8 @@ const fetchInitialTransactions = async (
   if (transactions.length === 0) {
     return [];
   }
-
-  return transactions.slice(0, historicalTransactionsFetchLimit);
+  // allTransactionsByAddresses returns in `asc` order
+  return transactions.slice(-1 * historicalTransactionsFetchLimit);
 };
 
 const findIntersectionAndUpdateTxStore = ({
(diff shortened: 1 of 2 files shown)


rhyslbw (approved):

AngelCastilloB (approved):
LGTM @mirceahasegan

packages/wallet/src/services/TransactionsTracker.ts
```diff
@@ -335,11 +335,12 @@ const findIntersectionAndUpdateTxStore = ({
...
             }
           }
-          // Skip overlapping transactions to avoid duplicates
+          // Skip overlapping transactions to avoid duplicates.
+          // Limit # of stored transactions to last `historicalTransactionsFetchLimit`
           localTransactions = deduplicateSortedArray(
             [...localTransactions, ...newTransactions.slice(localTxsFromSameBlock.length)],
             txEquals
-          );
+          ).slice(-1 * historicalTransactionsFetchLimit);
```
mirceahasegan, line 343: @mkazlauskas . I did not add extra changes from your initial impl here. Please let me know if that is ok

mkazlauskas: Not sure if this change has test coverage, but I'm ok merging this as is due to time pressure

PR #1585, opened 2025-02-05
feat: introduce persistent cache for providers

szymonmaslowski:
# Context

We want to improve caching mechanism by making it persistent and therefore decrease the number of requests made to blockfrost

# Proposed Solution

Allow utxo and chain history providers to use persistent cache

- define a contract of the cache
- enable chain history and utxo providers to use injected cache
- create persistent cache storage
  - utilizing web extension storage
  - implementing the cache contract
  - having data evicting capability

### The eviction mechanism
 - checks how much space is occupied by all of the collection's keys
 - if the allowed (specified in the parameters) space is exceeded it removes most rarely accessed items
 - the set of items to be removed is chosen based on the item's `accessTime` stored in metadata
 - the size of items to be removed is 10% of their count.
 - a fallback mechanism in case the `getBytesInUse` is not present - it is not present in the MDN docs neither in the polyfill lib we use (recent version)
 - the fallback checks if the total count of cached items exceeded the amount provided in `fallbackMaxCollectionItemsGuard` parameter

Example of data managed by the persistent cache implementation. Resouce name: `utxo-provider-cache`.
Result of `storage.local.get()`:
```
{
  utxo-provider-cache-metadata: {
    utxo-provider-cache-item-x: {
      accessTime: 1738765316652
    },
    utxo-provider-cache-item-y: {
      accessTime: 1738765316653
    },
    utxo-provider-cache-item-z: {
      accessTime: 1738765316654
    }
  },
  utxo-provider-cache-item-x: {/* data */},
  utxo-provider-cache-item-y: {/* data */},
  utxo-provider-cache-item-z: {/* data */},
  /* other stored data */
}
```

Reasons behind that structure:
1. Having metadata in a separate object we can safely access all the cache items for envicting.
2. Storing each item flat allows to:
   - easily count occupied size (`storage.local.getBytesInUse([...allKeys])`)
   - easy removal and access to particular items.


szymonmaslowski:
I need to finish unit tests but the implementation might already get the initial reviews

packages/util/src/types.ts
```diff
@@ -28,3 +28,8 @@ type Impossible<K extends keyof any> = {
   [P in K]: never;
 };
 export type NoExtraProperties<T, U> = U & Impossible<Exclude<keyof U, keyof T>>;
+
+export type Cache<T> = {
+  get(key: string): Promise<T | undefined>;
```
szymonmaslowski, line 33: I parametrised only the value type. The extension storage allows only for string keys.

packages/web-extension/src/storage/persistentCacheStorage/persistentCacheStorage.ts
```diff
@@ -0,0 +1,118 @@
...
+        return value;
+      },
+      async set(key: string, value: T) {
+        const itemKey = getItemKey(key);
+        loaded.set(itemKey, value);
+        await extensionLocalStorage.set({ [itemKey]: value });
+
+        void (async () => {
+          await updateAccessTime(itemKey);
+          if (await isQuotaExceeded()) await evict();
```
pczeglik-iohk, line 110: Should we check exceeded quota before setting item, so we know we have room to insert it?
Or maybe we can do try ... catch and handle `evict` in catch, and after re do loaded.set in `finally`?

pczeglik-iohk (dismissed):
Nice work @szymonmaslowski 👏

przemyslaw-wlodek (approved):
👏🏻 Great job @szymonmaslowski!

pczeglik-iohk (approved):

PR #1590, opened 2025-02-10
fix: [lw-12235] handle expected blockfrost 404 errors

vetalcore:
# Context

Blockfrost 404s are already handled in the [fetchSequentially util](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services-client/src/blockfrost/util.ts#L52), but these errors are still logged in the [request method of BlockfrostProbider class](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/cardano-services-client/src/blockfrost/BlockfrostProvider.ts#L49) via logger.error.

# Proposed Solution
Change log level to `debug` instead of `error` in case this is blockfrost not found error.


mirceahasegan (approved):
lgtm

PR #1591, opened 2025-02-11
refactor(cardano-services-client): use CIP129 when querying dreps/:drepId - LW-12316

mirceahasegan:
# Context

In the inception of governance DReps ids were bech32 encoded using CIP105 spec.

Later a new standard, CIP129 because the defacto bech32 way to encode DRep ids

Blockfrost supports querying for specific DReps using both standards.

We should aim to use a single format throughout the code, to avoid confusion and to be consistent

# Proposed Solution

### DIFF
--- a/compose/common.yml
+++ b/compose/common.yml
@@ -121,7 +121,7 @@ x-sdk-environment: &sdk-environment
 services:
   blockfrost-ryo:
     build:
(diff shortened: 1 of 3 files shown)


greatertomi (approved):

mkazlauskas (approved):
Nice work!

rhyslbw (approved):
🙌

PR #1593, opened 2025-02-12
feat: make metadata update operation atomic in persisten cache storage

szymonmaslowski:
# Context

In order to make sure the cache metadata collection tracks properly all the stored cache items it is required for read and write operations to be performed in sequence. Otherwise, some parallel metadata writing might cause data loss.

# Proposed Solution

Making read and write operations atomic by maintaining a queue of those operations and performing next one only if previous one finished.


pczeglik-iohk (approved):

przemyslaw-wlodek (approved):
💪🏻 🚢

PR #1599, opened 2025-02-13
NativeScript signing

Riley-Kilgore:
# Context

Lace currently fails to identify when it is being requested to sign a transaction on behalf of a NativeScript where it's signature may be one of the signatures required by that NativeScript.

# Proposed Solution

Evaluate NativeScript(s) used as witnesses to determine keypaths which must be used to sign (if any).

Added the capability described above to enhance Lace's functionality with CIP-30 dApps that leverage NativeScripts.

### DIFF
--- a/packages/key-management/src/InMemoryKeyAgent.ts
+++ b/packages/key-management/src/InMemoryKeyAgent.ts
@@ -134,7 +134,7 @@ export class InMemoryKeyAgent extends KeyAgentBase implements KeyAgent {
 
   async signTransaction(
     txBody: Serialization.TransactionBody,
-    { txInKeyPathMap, knownAddresses }: SignTransactionContext,
+    { txInKeyPathMap, knownAddresses, scripts }: SignTransactionContext,
     { additionalKeyPaths = [] }: SignTransactionOptions = {}
   ): Promise<Cardano.Signatures> {
     // Possible optimization is casting strings to OpaqueString types directly and skipping validation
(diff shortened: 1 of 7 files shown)


rhyslbw:
Thanks for submitting this @Riley-Kilgore. @AngelCastilloB, who is the best candidate for review, will be back on Monday. In the mean-time, would you mind re-working your commits to follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)? This product has an automatic release process that requires formatting in this way to generate the changelog. Sorry, we're missing a contributor guide

AngelCastilloB (commented):
Thanks for the PR @Riley-Kilgore ❤️ . I made a small change request, additionally, for this change to be effective we need to:

add a new property here:

https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/key-management/src/types.ts#L166-L170

probably 'scripts' ?,

Then we need to use that new property to pass into `ownSignatureKeyPaths` in this two places:

https://github.com/input-output-hk/cardano-js-sdk/blob/f89fba9aaebc7aa95f3ecb513d9b5a3b653a9b77/packages/key-management/src/util/stubSignTransaction.ts#L27-L29
https://github.com/input-output-hk/cardano-js-sdk/blob/f89fba9aaebc7aa95f3ecb513d9b5a3b653a9b77/packages/key-management/src/InMemoryKeyAgent.ts#L146

Additionally would be great to add some unit test to the change to  `ownSignatureKeyPaths` here:

https://github.com/input-output-hk/cardano-js-sdk/blob/f89fba9aaebc7aa95f3ecb513d9b5a3b653a9b77/packages/key-management/test/util/ownSignaturePaths.test.ts#L82

As a side note, this will only enable signing transactions with native scripts with in-memory wallets. For hardware wallets, we need to somehow tell the devices to sign with these keys found on the scripts (currently they only parse the body data), I think they allow to specify additional key paths for signing as options, but I am not sure, we need to check. There is a workaround for this, and is to add the key hashes present on the scripts on the DApp side to the requieredExtraSigners field in the transaction body:

https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/cddl-files/conway.cddl#L531

AngelCastilloB (approved):
LGTM @Riley-Kilgore , thanks for the contribution. adding @mkazlauskas and @mirceahasegan for review

mkazlauskas (approved):
Thanks for contribution! ❤️

mirceahasegan (approved):
Fantastic work @Riley-Kilgore 👏 . Thank you

PR #1602, opened 2025-02-18
fix(core): withdrawal calculation in computeImplicitCoin [LW-12198]

przemyslaw-wlodek:
# Context
See [LW-12198](https://input-output.atlassian.net/browse/LW-12198)

# Proposed Solution
Calculate withdrawals only for own reward accounts.

[LW-12198]: https://input-output.atlassian.net/browse/LW-12198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/Cardano/util/computeImplicitCoin.ts
+++ b/packages/core/src/Cardano/util/computeImplicitCoin.ts
@@ -146,6 +146,11 @@ const getTxDeposits = (
   };
 };
 
+const getOwnWithdrawalsTotal = (withdrawals: Cardano.Withdrawal[], rewardAccounts: Address.RewardAccount[]) =>
+  BigIntMath.sum(
+    withdrawals.filter(({ stakeAddress }) => rewardAccounts.includes(stakeAddress)).map(({ quantity }) => quantity)
+  );
+
 /**
  * Computes the implicit coin from the given transaction.
  * If rewardAccounts is provided, it will only count the deposits from
@@ -166,7 +171,7 @@ export const computeImplicitCoin = (
   {
     certificates,
     proposalProcedures,
-    withdrawals
+    withdrawals = []
   }: Pick<Cardano.HydratedTxBody, 'certificates' | 'proposalProcedures' | 'withdrawals'>,
   rewardAccounts?: Address.RewardAccount[],
   dRepKeyHash?: Crypto.Ed25519KeyHashHex
@@ -179,7 +184,9 @@ export const computeImplicitCoin = (
     proposalProcedures
   );
 
-  const withdrawalsTotal = (withdrawals && BigIntMath.sum(withdrawals.map(({ quantity }) => quantity))) || 0n;
+  const hasRewardAccount = !!rewardAccounts?.length;
+  const allWithdrawalsTotal = BigIntMath.sum(withdrawals.map(({ quantity }) => quantity));
+  const withdrawalsTotal = hasRewardAccount ? getOwnWithdrawalsTotal(withdrawals, rewardAccounts) : allWithdrawalsTotal;
 
   return {
     deposit,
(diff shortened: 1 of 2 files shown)


mirceahasegan (approved):
Excellent work @przemyslaw-wlodek 👏

vetalcore (approved):

packages/core/src/Cardano/util/computeImplicitCoin.ts
```diff
@@ -179,7 +194,9 @@ export const computeImplicitCoin = (
     proposalProcedures
   );
 
-  const withdrawalsTotal = (withdrawals && BigIntMath.sum(withdrawals.map(({ quantity }) => quantity))) || 0n;
+  const hasRewardAccount = !!rewardAccounts?.length;
```
AngelCastilloB, line 197: ❤️

AngelCastilloB (dismissed):
LGTM @przemyslaw-wlodek

vetalcore (dismissed):

packages/core/src/Cardano/util/computeImplicitCoin.ts
```diff
@@ -179,7 +194,9 @@ export const computeImplicitCoin = (
...
+  const withdrawalsAll = BigIntMath.sum(withdrawals.map(({ quantity }) => quantity));
+  const withdrawalsTotal = hasRewardAccount ? getOwnWithdrawals(withdrawals, rewardAccounts) : withdrawalsAll;
```
mkazlauskas, line 199: Shouldn't this be 0? 🤔 
```suggestion
  const withdrawalsTotal = hasRewardAccount ? getOwnWithdrawals(withdrawals, rewardAccounts) : 0n;
```

przemyslaw-wlodek: Initially we always filtered own withdrawals, but this is also used in txBuilder and due to some constraints, this logic should be returning all withdrawals like it does now. @AngelCastilloB could elaborate probably.

przemyslaw-wlodek: Done

mkazlauskas (approved):
Nice work! 🚀

vetalcore (approved):

packages/core/src/Cardano/util/computeImplicitCoin.ts
```diff
@@ -30,6 +30,14 @@ const stakeCredentialInRewardAccounts = (
   return rewardAccounts.includes(Address.RewardAccount.fromCredential(stakeCredential, networkId));
 };
 
+const stakeAddressInRewardAccounts = (
+  stakeAddress: Address.RewardAccount,
+  rewardAccounts: Address.RewardAccount[]
+): boolean => {
+  if (rewardAccounts.length === 0) return true;
```
mirceahasegan, line 37: Sorry for nitpicking. I find this line confusing. What I expect from this function is: if there are no reward accounts, then stake address is not included. But this line makes it: if there are no reward accounts, then stake address is included.
My suggestion would be to remove this line, as the empty reward accounts scenario is already captured at line 199:
`const withdrawalsTotal = hasRewardAccount ? getOwnWithdrawals(withdrawals, rewardAccounts) : withdrawalsAll;`

przemyslaw-wlodek: Completely valid comment, not a nit. Thank you! It was a mistake. This utility became useless in the end, so I inlined it.

mirceahasegan (approved):
Great work @przemyslaw-wlodek 🚀

vetalcore (approved):

PR #1604, opened 2025-02-21
feat: sign shared wallet tx with ledger

greatertomi:
# Context

Currently, it signs successfully, but there are no signatures in the Witness Tx. 

# Proposed Solution


packages/hardware-ledger/src/transformers/txOut.ts
```diff
@@ -20,8 +30,9 @@ const toDestination: Transform<Cardano.TxOut, Ledger.TxOutputDestination, Ledger
   context
 ) => {
   const knownAddress = context?.knownAddresses.find((address) => address.address === txOut.address);
+  const isScriptWallet = isScriptAddress(txOut.address);
 
-  if (knownAddress) {
+  if (knownAddress && !isScriptWallet) {
```
greatertomi, line 35: Ledger does not allow outputs for Multisig transactions to be of type "device-owned".

AngelCastilloB: Hi @greatertomi , this is not enough to make the ledger sign the transactions. This just tesll the ledger the transactions is multisig (as in other parties also need to sign it), but it doesnt tell it how to sign it. The API just parses transaction body to look for signature, not the witness set, is it doesnt know there is an script with a key it owns. We need to ask the team that made the ledger cardano app (and Trezor), if there is a way to tell the ledger to sign the transaction with any specific key (not persent in any field of the transaction body)

packages/hardware-trezor/src/TrezorKeyAgent.ts
```diff
@@ -116,6 +122,21 @@ export class TrezorKeyAgent extends KeyAgentBase {
         // Show Trezor Suite popup. Disabled for node based apps
         popup: communicationType !== CommunicationType.Node && !silentMode
       });
+
+      trezorConnect.on(Trezor.UI_EVENT, (event) => {
```
greatertomi, line 126: According to #1461, this should be improved and extended. I don't know what improvements we would need to make, but without this addition, the trezor wallet is stuck in "Enter your passphrase" prompt in tests.

AngelCastilloB: I think we shouldnt add this in this PR, I dont know if this will have any effect on Trezor for Lace users, best to deal this separately/anoterh PR (maybe add something on the constructor of the KeyAgent to enable this)

packages/hardware-trezor/src/TrezorKeyAgent.ts
```diff
@@ -235,14 +256,20 @@ export class TrezorKeyAgent extends KeyAgentBase {
...
+        }),
         signingMode
       });
 
-      const expectedPublicKeys = await Promise.all(
-        util
-          .ownSignatureKeyPaths(body, knownAddresses, txInKeyPathMap)
-          .map((derivationPath) => this.derivePublicKey(derivationPath))
-      );
+      const expectedPublicKeys =
```
AngelCastilloB, line 265: We can probably generalize this, something like https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/key-management/src/util/ownSignatureKeyPaths.ts#L335-L370 this would also allow dApps to build transactions with scripts and have Lace sign them correctly

packages/hardware-trezor/src/transformers/txOut.ts
```diff
@@ -10,8 +10,9 @@ const toDestination: Transform<Cardano.TxOut, TrezorTxOutputDestination, TrezorT
   context
 ) => {
   const knownAddress = context?.knownAddresses.find((address: GroupedAddress) => address.address === txOut.address);
+  const isScriptWallet = Cardano.util.isScriptAddress(txOut.address);
 
-  if (!knownAddress) {
+  if (!knownAddress || isScriptWallet) {
```
AngelCastilloB, line 15: This can be renamed to `isScriptAddress`, we want to skip it if the address is script, but this applies to all transactions regardless of the type of wallet

AngelCastilloB (changes requested):
Hi @greatertomi I made a few change request, main point is, this change should allow hardware wallets to sign transactions with native scripts, which will work for shared wallet, but also for dApps, rather than making this specific to sharedWallets

packages/hardware-trezor/src/TrezorKeyAgent.ts
greatertomi: Fixed on 8030c8f

packages/hardware-trezor/src/transformers/txOut.ts
greatertomi: Fixed on 8030c8f

packages/hardware-trezor/src/TrezorKeyAgent.ts
greatertomi: Addressed on 7a7ccf2. Would this work?

AngelCastilloB (dismissed):
Great work @greatertomi ⭐

packages/core/src/Cardano/util/isScriptAddress.ts
```diff
@@ -0,0 +1,9 @@
+import { Address, CredentialType } from '../Address';
+
+// TODO: Would like to verify if this is needed before writing test for it
```
rhyslbw, line 3: Unresolved todo

rhyslbw (changes requested):
Nice work @greatertomi. Just a minor change req

packages/key-management/src/types.ts
```diff
@@ -92,6 +92,7 @@ export interface TrezorConfig {
     email: string;
     appUrl: string;
   };
+  shouldHandlePassphrase?: boolean;
```
mkazlauskas, line 95: Please add a doc comment with some explanation about what this does

mkazlauskas (changes requested):

packages/core/src/Cardano/util/isScriptAddress.ts
```diff
@@ -0,0 +1,9 @@
...
+export const isScriptAddress = (address: string): boolean => {
```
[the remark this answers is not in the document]
greatertomi: Fixed and merged with b418fff707eb71f5ef71fc68bf61502e5915ca31

packages/key-management/src/types.ts
greatertomi: Added and merged with b418fff707eb71f5ef71fc68bf61502e5915ca31

rhyslbw (dismissed):
Approval is for resolution of https://github.com/input-output-hk/cardano-js-sdk/pull/1604#discussion_r1975363978.

packages/key-management/src/types.ts
mkazlauskas:
```
/**
 * please use this format comment, as IDEs detect it as docs
 */
```

greatertomi: Fixed and squashed with b658c825190ef18849941d49af0a55e4e2e3749c

mkazlauskas (approved):
Great work! 🚀

rhyslbw (approved):

PR #1606, opened 2025-02-27
fix(wallet): base wallet will now immediately throw if tx is outside validity interval

AngelCastilloB:
# Context

Transaction can get stuck as 'inFlight' if they are rejected due to invalidHereAfter, this change will make it so the base wallet now throws before ever queuing the transaction.

# Proposed Solution

- BaseWallet throws before ever queuing the transaction if the transaction is already expired.


mkazlauskas (approved):
Nice work!

rhyslbw (approved):
Nice work @AngelCastilloB 👏

PR #1608, opened 2025-03-10
feat: discover shared wallet by metadata label

greatertomi:
# Context

This PR adds a provider to fetch transactions with 1854 label from the blockchain using Blockfrost

# Proposed Solution


packages/cardano-services-client/src/SharedWalletProvider/BlockfrostSharedWalletProvider.ts
```diff
@@ -0,0 +1,28 @@
+import * as Crypto from '@cardano-sdk/crypto';
+import { BlockfrostClient, BlockfrostProvider } from '../blockfrost';
+import { Logger } from 'ts-log';
+import { MultiSigTransaction, SharedWalletProvider } from '@cardano-sdk/core';
+
+const MULTI_SIG_LABEL = '1854';
```
vetalcore, line 6: same as `KeyPurpose.MULTI_SIG` ?

packages/core/src/Provider/SharedWalletProvider/types.ts
```diff
@@ -0,0 +1,30 @@
...
+  participants?: MultiSigParticipants;
+}
+
+export interface MultiSigTransaction {
+  tx_hash: string;
+  json_metadata: MultiSigRegistration;
+}
+
+export interface SharedWalletProvider {
+  discoverWallets: (pubKey: Ed25519KeyHashHex) => Promise<any[]>;
```
vetalcore, line 29:
```suggestion
  discoverWallets: (pubKey: Ed25519KeyHashHex) => Promise<MultiSigTransaction[]>;
```

vetalcore (changes requested):

packages/cardano-services-client/src/SharedWalletProvider/BlockfrostSharedWalletProvider.ts
greatertomi: Fixed.

packages/core/src/Provider/SharedWalletProvider/types.ts
greatertomi: Fixed and squashed on ee69b57b4da58161614d8e1c6c7dc79c74ba4f4d

packages/cardano-services-client/src/SharedWalletProvider/BlockfrostSharedWalletProvider.ts
greatertomi: I can't use "@cardano-sdk/key-management" within the core package. The module is still not found even when I add it to the package.json and install.

mkazlauskas: You shouldn't add this dependency, as it would be a circular dependency. You can hoist new provider types to cardano-services-client

packages/core/src/Provider/SharedWalletProvider/types.ts
```diff
@@ -0,0 +1,32 @@
...
+  types: ScriptType[];
+  name?: string;
+  description?: string;
+  icon?: string;
+  participants?: MultiSigParticipants;
+}
+
+export interface MultiSigTransaction {
+  tx_hash: string;
+  json_metadata: MultiSigRegistration;
```
mkazlauskas, line 27: - Please update field names to be consistent with cardano-js-sdk style
- Please use opaque string types where applicable
- Please add native scripts, as it will be required for creating the wallet in `lace`
```suggestion
  txId: Cardano.TransactionId;
  metadata: MultiSigRegistration;
  nativeScripts: Cardano.NativeScript[];
```

packages/core/src/Provider/SharedWalletProvider/types.ts
```diff
@@ -0,0 +1,32 @@
+import { Ed25519KeyHashHex } from '@cardano-sdk/crypto';
+
+type ScriptType = number;
```
mkazlauskas, line 3: From cip:
> This corresponds to the key derivation path role

We probably have a type for that in key-management package

packages/core/src/Provider/SharedWalletProvider/types.ts
```diff
@@ -0,0 +1,32 @@
...
+export const MULTI_SIG_LABEL = 1854;
+
+interface MultiSigParticipant {
+  name: string;
+  description?: string;
+  icon?: string;
+}
+
+interface MultiSigParticipants {
+  [key: string]: MultiSigParticipant;
```
mkazlauskas, line 14: Please find and use an appropriate type for this. Create a new one if it doesn't exist.
```suggestion
  [key: MoreSpecificType]: MultiSigParticipant;
```

packages/cardano-services-client/src/SharedWalletProvider/BlockfrostSharedWalletProvider.ts
```diff
@@ -0,0 +1,26 @@
...
+    const pageSize = 100;
+    let allTransactions: MultiSigTransaction[] = [];
+
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
+      const response = await this.request<MultiSigTransaction[]>(`metadata/txs/labels/${MULTI_SIG_LABEL}?page=${page}`);
+      allTransactions = [...allTransactions, ...response];
+      if (response.length < pageSize) break;
+      page++;
+    }
```
mkazlauskas, line 22: We have a util that does this. Search for usages for examples.
```suggestion
const allTransactions = await fetchSequentially(...)
```

mkazlauskas (changes requested):

packages/core/src/Provider/SharedWalletProvider/types.ts
greatertomi: `nativeScripts` is not returned from `/metadata/txs/labels/{Metadata_num}`. Where would that be gotten from?

packages/core/src/Provider/SharedWalletProvider/types.ts
mkazlauskas: Don't do it in core package, see https://github.com/input-output-hk/cardano-js-sdk/pull/1608#discussion_r1989226192

packages/core/src/Provider/SharedWalletProvider/types.ts
mkazlauskas: CIP says
> Auxiliary data with a non-empty native scripts array.

Which means you need tx->auxiliaryData->scripts. I don't see Blockfrost endpoint for auxiliary scripts. If there is none, it means you need to fetch entire tx cbor `https://docs.blockfrost.io/#tag/cardano--transactions/GET/txs/{hash}/cbor``

packages/core/src/Provider/SharedWalletProvider/types.ts
greatertomi: Fixed on cfcbae8

packages/core/src/Provider/SharedWalletProvider/types.ts
greatertomi: Fixed on cfcbae8

packages/cardano-services-client/src/SharedWalletProvider/BlockfrostSharedWalletProvider.ts
greatertomi: Fixed on cfcbae8

packages/core/src/Provider/SharedWalletProvider/types.ts
greatertomi: Fixed and squashed with 3492029c1dfb6e4160065e24d3445016ae4000d7

vetalcore (dismissed):

AngelCastilloB (approved):

vetalcore (dismissed):

packages/cardano-services-client/src/SharedWalletProvider/BlockfrostSharedWalletProvider.ts
rhyslbw: I'm not seeing this feedback addressed in the reference commit @greatertomi. It would need to be something like:

```ts
// from core
export const MULTISIG_CIP_ID = `1854`;

// from key-management
KeyPurpose.MULTI_SIG  = MULTISIG_CIP_ID;

// from cardano-services-client
const MULTI_SIG_LABEL = MULTISIG_CIP_ID;

```

but I think it's also fine to reject the suggestion

greatertomi: Fixed on da1b65f

rhyslbw (approved):

rhyslbw:
Can squash now @greatertomi

PR #1616, opened 2025-04-11
LW-12624: Fix output mapping on hardware wallets

AngelCastilloB:
# Context

We had a round-trip issue in our hardware wallet output mappings. This PR introduces a change to check for the actual format on the raw CBOR and passes as as context to the type mappers

# Proposed Solution

Analyze the raw CBOR to detect the output format and pass it as part of the transformation context.

PR #1618, opened 2025-04-16
Fix/output mapping on hw

AngelCastilloB:
# Context

The change introduced in https://github.com/input-output-hk/cardano-js-sdk/pull/1616 was not completely accurate. It assumes all otuputs on the transaction will be encoded in the same format, but this is not necesarily the case, Legacy format is more compact, dApps can take this to their advantage and use legacy encoding for outputs that do not need babbage fields, reducing transaction size. So we should assume outputs could (and probably will) be encoded using different formats.

# Proposed Solution

check raw CBOR of all outputs and determined their format.


packages/hardware-ledger/src/transformers/txOut.ts
```diff
@@ -56,27 +56,33 @@ const getScriptHex = (output: Serialization.TransactionOutput): HexBlob | null =
...
+  Ledger.TxOutput,
+  LedgerTxTransformerContext
+> = (elem, context) => {
+  const { txOut, index, isCollateral } = elem;
   const output = Serialization.TransactionOutput.fromCore(txOut);
   const scriptHex = getScriptHex(output);
+  const format = isCollateral ? context?.collateralReturnFormat : context?.outputsFormat[index];
 
-  return context?.useBabbageOutputs
+  return format === Ledger.TxOutputFormat.MAP_BABBAGE
```
iccicci, line 69: Probably out of scope for this PR, but here I see some code repetition. I'd refactor in something like

```
return {
  amount: <...>,
  destination: <...>,
  format,
  tokenBundle,
  ...(format === Ledger.TxOutputFormat.MAP_BABBAGE ?
    { datum: <...>, referenceScriptHex: <...> } :
    { datumHashHex: <...> })
};
```

packages/hardware-trezor/src/transformers/txOut.ts
```diff
@@ -42,17 +42,23 @@ const getScriptHex = (output: Serialization.TransactionOutput): HexBlob | undefi
...
+  TrezorTxTransformerContext
+> = (elem, context) => {
+  const { txOut, index, isCollateral } = elem;
   const destination = toDestination(txOut, context);
   const output = Serialization.TransactionOutput.fromCore(txOut);
   const scriptHex = getScriptHex(output);
+  const format = isCollateral ? context?.collateralReturnFormat : context?.outputsFormat[index];
 
-  return context?.useBabbageOutputs
+  return format === Trezor.PROTO.CardanoTxOutputSerializationFormat.MAP_BABBAGE
```
iccicci, line 56: Same as before

iccicci (dismissed):
Apart from a couple of side comments (probably outside of the scope of this PR),
LGTM!

packages/hardware-ledger/src/transformers/txOut.ts
AngelCastilloB: done here @iccicci https://github.com/input-output-hk/cardano-js-sdk/pull/1618/commits/d00d80a567431a1ed734502745b8b9793fd465fc for some reason the type checker complain if I use the spread and not set the format specifically, so had to duplicate that bit

packages/hardware-trezor/src/transformers/txOut.ts
AngelCastilloB: done here https://github.com/input-output-hk/cardano-js-sdk/pull/1618/commits/3687665ab60d993070a62d2a5d0243ccb99dd330, Trezor target structure reuses more filed so this one could be reduced even more

packages/hardware-ledger/src/transformers/txOut.ts
iccicci: Nice!

packages/hardware-trezor/src/transformers/txOut.ts
iccicci: Very nice!

mkazlauskas (approved):
Great work! 💪

PR #1623, opened 2025-05-15
Feat/lw 12753 add retry mechanism for inputselection errors

AngelCastilloB:
# Context

Some users with complex wallets that have hundreds (or even thousand) of UTxOs with small amounts can make our current input selection strategy fail due to selecting more inputs than max allowed, we need alternative input selection strategies for such wallets

# Proposed Solution

Add a retry mechanism that fallback to large first when random improve fails to select

### DIFF
--- a/packages/tx-construction/src/tx-builder/TxBuilder.ts
+++ b/packages/tx-construction/src/tx-builder/TxBuilder.ts
@@ -26,6 +26,13 @@ import {
   UnwitnessedTx
 } from './types';
 import { GreedyTxEvaluator } from './GreedyTxEvaluator';
+import {
+  InputSelectionError,
+  InputSelectionFailure,
+  LargeFirstSelector,
+  SelectionSkeleton,
+  StaticChangeAddressResolver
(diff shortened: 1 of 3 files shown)


iccicci (approved):
LGTM!

packages/tx-construction/src/tx-builder/TxBuilder.ts
```diff
@@ -396,30 +404,51 @@ export class GenericTxBuilder implements TxBuilder {
...
+              // Fallback to large first if we are not using the greedy selector
+              if (
+                usingGreedySelector ||
+                !(error instanceof InputSelectionError) ||
+                error.failure === InputSelectionFailure.UtxoBalanceInsufficient
+              ) {
+                throw error;
+              }
+
+              dependencies.inputSelector = new LargeFirstSelector({
```
mkazlauskas, line 443: Ideally fallback inputSelector would be a separate dependency, but this is just a minor nit

mkazlauskas (approved):
Great work! 💪

PR #1630, opened 2025-05-29
feat: bip32account bip32ed25519 dependency injection LW-12941

mkazlauskas:
# Context

We need to be able to replace the key derivation function, because the one that is using libsodium is performing poorly in mobile environment

# Proposed Solution

Main commit: 7faaab4c77ef1e90cbe114e9902e96735727c8f9

See commit messages


AngelCastilloB (commented):
Changes to `Bip32Account.ts` look good to me @mkazlauskas

AngelCastilloB (approved):

mirceahasegan (approved):
Great work @mkazlauskas 🚀

PR #1632, opened 2025-06-04
fix(wallet): resolve only the own signed portions of the witness set

### LINKED ISSUE #1631: CIP-30 signTx() returns existing signatures
### Summary

In other CIP-30 wallets such as Eternl, when calling signTx(), it will return only the signatures that the wallet itself has added. When there are existing signatures added on the server side and Lace is then asked to add its own signature, it will return ALL signatures instead of just its own signature.  This is a deviation from the CIP-30 standard and should be fixed so Lace operates the same as other CIP-30 wallets. Currently, my dApp has to have special code for Lace just to de-duplicate the signatures so that the transaction can be properly built.

### Steps to reproduce the bug

1.) Call signTx() with a transaction that already has a signature attached to it.
2.) See that pre-existing signatures are returned.

### Actual Result

signTx() returns pre-existing signatures

### Expected Result

signTx() returns only new signatures applied by lace wallet itself.

### SDK version

Whatever the latest in Lace wallet is

### Environment type

- [ ] Node.js
- [ ] Browser
- [ ] Browser Extension
- [x] Other

### Environment details

Lace wallet running in latest google chrome.

mkazlauskas:
# Context

Fixes #1631 

image

# Proposed Solution

See commit

### DIFF
--- a/packages/wallet/src/cip30.ts
+++ b/packages/wallet/src/cip30.ts
@@ -469,7 +469,10 @@ const baseCip30WalletApi = (
           );
         }
 
-        const cbor = Serialization.TransactionWitnessSet.fromCore({ signatures }).toCbor();
+        const ownSignatures = new Map(
+          [...signatures.entries()].filter(([pubKey]) => !coreTx.witness.signatures.has(pubKey))
+        );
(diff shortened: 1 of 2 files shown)


AngelCastilloB (approved):
Good catch @mkazlauskas 🧙🏼

iccicci (approved):
LGTM!

PR #1633, opened 2025-06-05
fix: fix native script singing

AngelCastilloB:
# Context

Our integration with hardware wallets and transactions with native scripts had a few gaps/bugs that was preventing us from signing correctly multiSig transactions with ledger or trezor.

### DIFF
--- a/packages/wallet/jest.setup.js
+++ b/packages/wallet/jest.setup.js
@@ -1,0 +1,18 @@
+// Polyfill for Array.prototype.findLast on unit tests
+if (!Array.prototype.findLast) {
+  // eslint-disable-next-line no-extend-native
+  Array.prototype.findLast = function (predicate, thisArg) {
+    if (!Array.isArray(this)) {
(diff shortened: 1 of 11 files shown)


packages/wallet/jest.setup.js
```diff
@@ -0,0 +1,18 @@
+// Polyfill for Array.prototype.findLast on unit tests
+if (!Array.prototype.findLast) {
+  // eslint-disable-next-line no-extend-native
+  Array.prototype.findLast = function (predicate, thisArg) {
+    if (this === null) {
```
iccicci, line 5:
```suggestion
    if (! (this instanceof Array)) {
```
Could this make the check a bit stronger?

iccicci (commented):
I feel not strong enough on this topic to provide a valuable review. I just left a minor comment.

mkazlauskas (approved):
Nice work! 🚀

mirceahasegan (approved):
Fantastic work @AngelCastilloB 🚀

PR #1636, opened 2025-06-25
fix: fix long message signing on ledger

AngelCastilloB:
# Context

We are unable to sign long messages with the CIP-08 sign message function on ledger, this is because ledger has limited memory. For longer messages (as per CIP-08 spec) we must switch to sign the hashed message.

# Proposed Solution

Detect the length of the message being signed and switch to hashed message if >= 100 in length.


iccicci (approved):
Minimal and effective: excellent!

PR #1637, opened 2025-06-25
fix(crypto): remove setImmediate from async calls

AngelCastilloB:
# Context

setImmediate is not available in browser environments, and we dont really need it.

# Proposed Solution

Remove calls to setImmediate


iccicci (commented):
IDK if `blake2b.hash<T>(message, outputLengthBytes)` returns a `Promise`, I guess yes, otherwise we would have had a type error.
If that's the case, wouldn't be better to remove the `async` attribute from the function definition?

iccicci (approved):
LGTM!

PR #1640, opened 2025-07-11
chore: remove npm from dependencies

szymonmaslowski:
# Context

The `crypto` and `tx-construction` packages have `npm` as dependencies, which seems incorrect. Moreover, it causes some problems in the lace-platform. The `npm` gets installed in the `node_modules` of the project and then when running `npx` from the terminal it resolves the npm from the `node_modules` instead of the one installed globally. As a consequence npm v9 gets used instead of v10.

# Proposed Solution


iccicci (approved):
Excellent! ❤️

PR #1649, opened 2025-09-01
[LW-12690] fix: ledger hardware wallets can now create byron era outputs

AngelCastilloB:
# Context

Currently hardware-ledger has a bug that prevents Byron era addresses to be used as destination. This PR fixes this bug.

# Proposed Solution

- Fix address parsing so support all Cardano address formats


greatertomi (approved):
🚀

PR #1650, opened 2025-09-02
feat: support all Trezor derivation types

rhyslbw:
Based on the master key generation algorithm variations


packages/hardware-trezor/MASTER_KEY_GENERATION.md
```diff
@@ -0,0 +1,126 @@
...
+      appUrl: 'https://myapp.com'
+    },
+    derivationType: 'ICARUS'
+  }
+}, dependencies);
+```
+
+
+
+## Discovery & UX Guidelines
```
AngelCastilloB, line 95: I am not sure if this discovery process is feasible given our current UX. Since we would need to export the root pub key for all three derivation schemes, and execute the discovery on each. Additionally is technically possible for the user to be using more than one derivation scheme and have access on more than one. I think the best/easiest solution here is to simply add the drop down during on-boarding and let the user select which one he wants

PR #1651, opened 2025-09-05
fix(hardware-trezor): support derivationType in getXpub method

rhyslbw:
Found some issues during integration. Fixes, improved test coverage, improved docs contained within


packages/wallet/HARDWARE_TESTING.md
```diff
@@ -0,0 +1,110 @@
+# Hardware Testing Environment
```
mkazlauskas, line 1: The new `./scripts/*` are not referenced anywhere in the docs, please either reference or delete

mkazlauskas (changes requested):

packages/wallet/HARDWARE_TESTING.md
rhyslbw: https://github.com/input-output-hk/cardano-js-sdk/pull/1651/commits/918568cfcc56b8f21812e348bb14908fefe3027e

mkazlauskas (approved):

rhyslbw:
We have a couple of pre-existing test failures in the Trezor suite, but I won't fix them here

PR #1665, opened 2025-09-15
feat(util): return identical fromSerializableObject if already serialized

mirceahasegan:
Update `fromSerializableObject` function to correctly identify and return native types (like Date, Error, Map, Set, and ArrayBuffer views) without modification.


mkazlauskas (dismissed):
Looks good! 🚀

iccicci (approved):

mkazlauskas (approved):

PR #1667, opened 2025-09-24
fix(hardware-ledger): fix ledger hardware wallet mapping when there are reference inputs and no collateral [LW-13641]

AngelCastilloB:
# Context

Is valid to add reference input to transaction that are not executing plutus scripts, however, ledger hw can only sign those transaction as plutus transactions


packages/hardware-ledger/src/LedgerKeyAgent.ts
```diff
@@ -703,6 +703,7 @@ export class LedgerKeyAgent extends KeyAgentBase {
    * Ledger has certain limitations due to which it cannot sign arbitrary combination of all transaction features.
    * The mode specifies which use-case the user want to use and triggers additional validation on `tx` field.
    */
+  // eslint-disable-next-line complexity
```
rhyslbw, line 706: I'm OK with approving this, however we should either look to remove the rule and use other quality gates to catch overly-complex logic or stick with it and refactor to conform.

rhyslbw (approved):
Nice work @AngelCastilloB

PR #1676, opened 2026-01-26
test(key-management): add tests for accountKeyDerivationPathToBip32Path

semsuddin:
## Summary

- Add test coverage for the `accountKeyDerivationPathToBip32Path` function which constructs BIP32 derivation paths following CIP-1852
- Remove the `// TODO: test` comment from `key.ts` since tests are now in place

## Test Cases

- Default purpose (1852) with external key role
- Stake key role derivation
- Custom account index
- Custom purpose (1854 for multi-sig)

## Test Plan

- [x] All new tests pass locally
- [x] Existing tests continue to pass

### DIFF
--- a/packages/key-management/test/util/key.test.ts
+++ b/packages/key-management/test/util/key.test.ts
@@ -1,6 +1,10 @@
 import { AddressType, GroupedAddress, KeyRole } from '../../src';
 import { Cardano } from '@cardano-sdk/core';
-import { paymentKeyPathFromGroupedAddress, stakeKeyPathFromGroupedAddress } from '../../src/util';
+import {
+  accountKeyDerivationPathToBip32Path,
+  paymentKeyPathFromGroupedAddress,
+  stakeKeyPathFromGroupedAddress
+} from '../../src/util';
 
 export const paymentAddress = Cardano.PaymentAddress(
   'addr1qxdtr6wjx3kr7jlrvrfzhrh8w44qx9krcxhvu3e79zr7497tpmpxjfyhk3vwg6qjezjmlg5nr5dzm9j6nxyns28vsy8stu5lh6'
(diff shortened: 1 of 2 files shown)


packages/key-management/test/util/key.test.ts
```diff
@@ -44,4 +48,35 @@ describe('key utils', () => {
...
+      expect(path).toEqual([2_147_485_500, 2_147_485_463, 2_147_483_648, 3, 0]);
+    });
+
+    it('returns correct path with custom account index', () => {
+      const path = accountKeyDerivationPathToBip32Path(5, { role: KeyRole.External, index: 3 });
+      expect(path).toEqual([2_147_485_500, 2_147_485_463, 2_147_483_653, 0, 3]);
+    });
+
+    it('returns correct path with custom purpose', () => {
+      const path = accountKeyDerivationPathToBip32Path(0, { role: KeyRole.External, index: 0 }, 1854);
```
rhyslbw:
```suggestion
      const path = accountKeyDerivationPathToBip32Path(0, { role: KeyRole.External, index: 0 }, KeyPurpose.MULTI_SIG);
```

packages/key-management/test/util/key.test.ts
```diff
@@ -44,4 +48,35 @@ describe('key utils', () => {
       expect(stakeKeyPathFromGroupedAddress(knownAddress)).toEqual(knownAddressStakeKeyPath);
     });
   });
+  describe('accountKeyDerivationPathToBip32Path', () => {
+    it('returns correct path with default purpose', () => {
+      const path = accountKeyDerivationPathToBip32Path(0, { role: KeyRole.External, index: 0 });
+      expect(path).toEqual([2_147_485_500, 2_147_485_463, 2_147_483_648, 0, 0]);
```
rhyslbw: These path fixtures — plus `knownAddressKeyPath`/`knownAddressStakeKeyPath` above — all repeat the hardened prefix `[2_147_485_500, 2_147_485_463, 2_147_483_648, …]`. Please extract named constants so intent is legible and the duplication collapses, and use the `KeyRole` enum for the role element instead of raw `0/1/2/3`:

```ts
// CIP-1852: m / 1852' / 1815' / account' / role / index  (hardened = n + 0x8000_0000)
const HARDENED_PURPOSE_1852  = 2_147_485_500;
const HARDENED_PURPOSE_1854  = 2_147_485_502;
const HARDENED_COINTYPE_1815 = 2_147_485_463;
const HARDENED_ACCOUNT_0     = 2_147_483_648;
const HARDENED_ACCOUNT_5     = 2_147_483_653;
```

If you extract, fold the existing `knownAddress*` fixtures into the same constants so it's a net reduction in duplication, not just the new block.

rhyslbw (commented):
Verified the path arithmetic against the implementation and the `KeyRole`/`CardanoKeyConst` enums — all six expectations are correct, and pinning the role/index-not-hardened behaviour is exactly the right thing to lock in. Good coverage across all four roles plus custom account/purpose. One readability ask before merge (below). Thanks for clearing the TODO.

rhyslbw (approved):

packages/key-management/test/util/key.test.ts
semsuddin: done, pulled the hardened prefix into named constants with the cip-1852 comment and folded knownAddressKeyPath/knownAddressStakeKeyPath into the same ones, role element now uses the KeyRole enum instead of raw 0/1/2/3

packages/key-management/test/util/key.test.ts
semsuddin: done, swapped the magic 1854 for KeyPurpose.MULTI_SIG

AngelCastilloB (approved):

PR #1680, opened 2026-04-13
feat: add native cip8 serialization to ledger package

AngelCastilloB:
# Context

This PR removes the wasm dependency of `@emurgo/cardano-message-signing-nodejs` from hardware wallet ledger package as this was breaking cip8 signing on mobile.

# Proposed Solution

Implement the cip8 header serialization natively in typescript

### DIFF
--- a/packages/key-management/src/cip8/coseUtils.ts
+++ b/packages/key-management/src/cip8/coseUtils.ts
@@ -1,0 +1,62 @@
+import { Serialization } from '@cardano-sdk/core';
+
+// COSE algorithm identifiers
+export const ALGORITHM_EDDSA = -8;
+export const KEY_TYPE_OKP = 1;
+export const CURVE_ED25519 = 6;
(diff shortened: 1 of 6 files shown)


packages/key-management/src/cip8/coseUtils.ts
mkazlauskas, line 1: These utils should probably reside in `key-management` package (or a separate new one) as it's also being used for in-memory data signing

AngelCastilloB: Yes but we will need a bigger refactor for that, I can do it in a separate PR, and just remove the wasm completely, this is only adding a few of the headers

mkazlauskas (dismissed):
Nice work! 🚀

mkazlauskas (approved):

iccicci (approved):
LGTM!

PR #1683, opened 2026-05-23
fix(crypto): bump libsodium wrappers for ESM imports

### LINKED ISSUE #1682: @cardano-sdk/crypto pins libsodium-wrappers-sumo@0.7.x — broken ESM import on consumer install
## Summary

`@cardano-sdk/crypto` pins `libsodium-wrappers-sumo` to a `0.7.x` version (exact `0.7.10` in `0.4.5`, range `^0.7.5` in `0.2.3`). Versions `0.7.x` of `libsodium-wrappers-sumo` ship with a broken ESM import that fails to resolve on a fresh install, crashing any consumer at runtime.

The fix landed upstream in `libsodium-wrappers-sumo@0.8.0` (changed from a relative `import e from "./libsodium-sumo.mjs"` to a bare `import e from "libsodium-sumo"`). Bumping the range to `^0.8.0` (or `>=0.7.5 <0.9.0` if `0.7.x` was working at some point) would unblock the whole transitive consumer base.

## Repro

```bash
mkdir /tmp/repro && cd /tmp/repro
pnpm init
pnpm add @cardano-sdk/crypto@0.4.5
node --input-type=module -e \"import('@cardano-sdk/crypto').then(m => console.log('ok'))\"
```

Result:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module
  '/tmp/repro/node_modules/.pnpm/libsodium-wrappers-sumo@0.7.10/node_modules/libsodium-wrappers-sumo/dist/modules-sumo-esm/libsodium-sumo.mjs'
imported from
  '/tmp/repro/node_modules/.pnpm/libsodium-wrappers-sumo@0.7.10/node_modules/libsodium-wrappers-sumo/dist/modules-sumo-esm/libsodium-wrappers.mjs'
```

The file `libsodium-sumo.mjs` does NOT exist in `0.7.10`'s `modules-sumo-esm/` directory. `0.7.16` is the same (we hit this through `@cardano-sdk/crypto@0.2.3` → `^0.7.5` resolving to `0.7.16`). In `0.8.0+` the import line was rewritten to use a bare specifier (`libsodium-sumo`) which resolves correctly via node's normal package lookup.

## Why this matters now

The whole Lucid Evolution stack pulls in `@cardano-sdk/crypto`:

```
consumer
└── @lucid-evolution/lucid@0.4.29
    └── @lucid-evolution/utils@0.1.66
        └── @cardano-sdk/core@0.45.10
            └── @cardano-sdk/crypto@0.2.3
                └── libsodium-wrappers-sumo@0.7.16  ← BROKEN
```

Same story on Lucid 0.4.31 (latest):

```
└── @lucid-evolution/lucid@0.4.31
    └── ... → @cardano-sdk/crypto@0.4.5
        └── libsodium-wrappers-sumo@0.7.10  ← BROKEN
```

Currently consumers work around this with a `pnpm.overrides` / npm `overrides` block forcing `libsodium-wrappers-sumo: ^0.8.2`. That works but every downstream library has to document it independently.

## Suggested fix

In `packages/crypto/package.json`, bump:

```diff
-  \"libsodium-wrappers-sumo\": \"0.7.10\"
+  \"libsodium-wrappers-sumo\": \"^0.8.2\"
```

(or whatever range you're comfortable with that's `>= 0.8.0`). The API surface between 0.7.x and 0.8.x looks unchanged for the parts cardano-sdk uses — only the internal import path changed.

I'm filing this from the cardano402 project (https://github.com/MorganOnCode/cardano402) where we hit it in our published `@cardano402/mcp-server@0.1.1` after Lucid pulled it in transitively. Happy to test a fix branch if useful.

## Context

- Both #1372 and #1529 touched the broader ESM-support story but didn't address the version pin specifically.
- libsodium-wrappers-sumo current versions on npm: `0.8.0, 0.8.1, 0.8.2, 0.8.3, 0.8.4` (`0.8.4` is latest).

Chris79OG:
## Summary
- bump `@cardano-sdk/crypto` from `libsodium-wrappers-sumo@0.7.10` to `^0.8.4`
- refresh the Yarn/Nix lock entries for `libsodium-wrappers-sumo` and `libsodium-sumo`
- pass an explicit `null` key to `crypto_generichash` so the package builds against the 0.8 bundled types

## Verification
- `node .yarn/releases/yarn-3.2.1.cjs install --mode=skip-build`
- `node .yarn/releases/yarn-3.2.1.cjs workspace @cardano-sdk/crypto exec tsc --build src`
- `node .yarn/releases/yarn-3.2.1.cjs workspace @cardano-sdk/crypto exec tsc -p src/tsconfig.json --outDir ./dist/esm --module es2020 --noEmit`
- `node -e "import('libsodium-wrappers-sumo').then(async m => { const s=m.default||m; await s.ready; const h=s.crypto_generichash(28, new Uint8Array([1,2,3]), null); console.log(h.length === 28 ? 'hash ok' : 'bad length '+h.length) })"`

Note: `yarn workspace @cardano-sdk/crypto build` reaches the TypeScript build, then fails on Windows because the package script uses Unix `cp`.

Closes #1682


AngelCastilloB:
Thanks @Chris79OG , we need to audit the new package version before we can merge this, cc @rhyslbw

packages/crypto/package.json
```diff
@@ -70,7 +70,7 @@
     "@cardano-sdk/util": "workspace:~",
     "blake2b": "^2.1.4",
     "i": "^0.3.7",
-    "libsodium-wrappers-sumo": "0.7.10",
+    "libsodium-wrappers-sumo": "^0.8.4",
```
rhyslbw:
```suggestion
    "libsodium-wrappers-sumo": "0.8.4",
```

rhyslbw (changes requested):
## Audit: `libsodium-wrappers-sumo` 0.7.10 → 0.8.4

Ran a two-pass audit (static review + tarball/runtime inspection). Clean, narrowly-scoped fix for a real ESM-loader bug (#1682), against a dep with a single trusted maintainer, with byte-identical crypto outputs at the SDK's hash/sign call sites. Small change request to pin the version exactly, otherwise good to merge.

### Provenance
- Same npm publisher across all 0.7.x and 0.8.x versions: `jedisct1` (Frank Denis, author of upstream libsodium-C). No ownership change.
- All five 0.8.x tarballs are npm-signed with the same key (`SHA256:DhQ8wR5APBvFHLF/+Tc+AYvPOdTpcIDqOhxsBHRwC7U`); the keyid delta vs 0.7.10 is npm's own registry signing-key rotation, not a user-controlled change.
- 7 distinct upstream contributors between tags `0.7.10` and `0.8.4` (163 commits), all real-named; every non-jedisct1 PR was reviewed and merged by Frank Denis. Sampled non-maintainer diffs match their commit messages.

### Tarball contents
- Both `libsodium-wrappers-sumo@0.8.4` and `libsodium-sumo@0.8.4` contain only `package.json`, `README.md`, `LICENSE`, and four JS/dts files. **No `scripts` field, no pre/post-install hooks, no native blobs (`.node`/`.so`/`.dylib`), no `bin` entries.**
- Only network/eval-adjacent strings are canonical Emscripten boilerplate (`require("crypto")` for Node `randomBytes` fallback; `XMLHttpRequest`/`fetch` in the wasm/asm.js loader). Zero `eval()`, `new Function()`, or `child_process`.
- No transitive surface expansion — `libsodium-sumo` remains the only runtime dep, same as 0.7.10.

### Vulnerabilities
- OSV reports no advisories for either `libsodium-wrappers-sumo` or `libsodium-sumo` at any version.

### Crypto correctness (the part that matters for a key-derivation/signing dep)
- Bundled native libsodium in 0.8.4 is **1.0.22-RELEASE** (legitimate latest upstream, signed by jedisct1).
- `crypto_generichash` symbol diff between 0.7.10 and 0.8.4: only change is `hash_length` typing `uint → ranged_uint` (stricter input validation; identical underlying C call). `crypto_generichash(28, pk)` and `crypto_generichash(28, pk, null)` produce **bit-identical** 28-byte output — the `null` arg is libsodium's canonical "unkeyed BLAKE2b" sentinel.
- Verified known-answer vectors against 0.8.4: BLAKE2b-256(empty) = `0e5751c0…12fe3a8`, BLAKE2b-256("abc") = `bddd813c…8d52319`, Ed25519 RFC 8032 seed-1 keypair and detached signature both byte-match the spec.
- All other sodium symbols `packages/crypto/src` imports (`crypto_scalarmult_ed25519_base_noclamp`, `crypto_core_ed25519_*`, `crypto_hash_sha512`, `crypto_sign_seed_keypair`, `crypto_sign_detached`, `crypto_sign_verify_detached`, `crypto_auth_hmacsha512`) remain call-compatible with current usage — no other source changes required.

AngelCastilloB (dismissed):
I pulled the 0.8.4 tarballs and checked their SHA-512 hashes against what the registry reports, both matched exactly, and npm's own signature on the packages verified. I also re-ran our own yarn to recompute the lockfile checksums from a fresh download, and they matched the values in yarn.lock and yarn-project.nix byte-for-byte.

AngelCastilloB (approved):

rhyslbw (approved):

PR #1685, opened 2026-06-10
feat(core): opt-in strict CBOR deserialization mode

### LINKED ISSUE #1684: Opt-in strict decoding mode for Serialization (detect unknown TransactionBody keys / constructs)
## Problem

`Serialization.TransactionBody.fromCbor` (in `@cardano-sdk/core`) deserializes the transaction body by switching on the known map keys, with no `default` branch — a key it does not recognize is **silently skipped**. The same lenient pattern applies to other discriminated constructs (certificates, governance actions, etc.).

Two consequences for consumers that render a summary from the decoded model:

1. A transaction containing constructs from a newer era than the client understands is **under-rendered** — the omitted parts simply don't appear in the decoded model, with no signal that anything was skipped.
2. Because `originalBytes` is cached and reused on re-serialization, a re-encode round-trip comparison does **not** detect the omission either.

## Use case

Wallet UIs render a human-readable summary of a transaction before signing. During a hard-fork window, a not-yet-updated client receiving a transaction that uses new-era constructs would prefer to detect "this tx contains something this build does not understand" and refuse or warn, rather than silently render a partial summary while the signature covers the full body bytes.

## Proposal

An **opt-in strict decoding mode** — e.g. `fromCbor(bytes, { strict: true })`, or a parallel strict entry point — that throws (or returns a structured "unknown constructs" report) on:

- unknown `TransactionBody` map keys
- unknown certificate / governance-action / voter / credential discriminants
- unknown era-specific tags where applicable

Default behavior stays lenient (unchanged for existing consumers); strict is opt-in for signing/summary paths.

mkazlauskas:
# Context

Closes #1684

`Serialization.TransactionBody.fromCbor` switches on known map keys with no `default` branch, so a transaction containing constructs from a newer era than the client understands is under-rendered with no signal that anything was skipped. Wallet UIs rendering a summary before signing would prefer to detect this and refuse/warn, since the signature covers the full body bytes. Re-encode round-trips don't catch it either because `originalBytes` is cached.

# Proposed Solution

Two parts:

**Always-on discriminant validation (fix).** Unknown discriminants cannot be parsed leniently — the old code misinterpreted them rather than skipping them:
- `DRep.fromCbor` silently decoded any unknown kind as `AlwaysNoConfidence`
- credential types and votes were accepted as-is via numeric casts
- an unknown `datum_option` kind produced an empty datum

These now throw unconditionally, consistent with `Certificate`, `Voter` and `ProposalProcedure` which already throw on unknown kinds. Credential type validation is centralized in a new `readCredentialType` helper (`Common/Credential.ts`) used by all certificate types, `Committee` and `Credential`.

**Opt-in strict mode for extensible maps (feat).** `Transaction`, `TransactionBody`, `TransactionOutput` and `TransactionWitnessSet` `fromCbor` accept an optional `{ strict?: boolean }`. When strict, an unknown map key throws `SerializationError` with the new `SerializationFailure.UnknownField` reason. Default behavior stays lenient for existing consumers.

- Lenient mode now skips unknown map keys correctly: previously the unknown key's value was left unconsumed, so the reader misaligned and parsed garbage (or threw a confusing error) on the next field.
- Known follow-up gaps (unchanged in this PR): `ProtocolParamUpdate`, `Update` and `AuxiliaryData` maps still have no `default` branch, so they retain the misalignment behavior and no strict support.

### DIFF
--- a/packages/core/src/Serialization/Certificates/AuthCommitteeHot.ts
+++ b/packages/core/src/Serialization/Certificates/AuthCommitteeHot.ts
@@ -3,6 +3,7 @@ import { CborReader, CborWriter } from '../CBOR';
 import { CertificateKind } from './CertificateKind';
 import { CertificateType } from '../../Cardano/types/Certificate';
 import { HexBlob, InvalidArgumentError } from '@cardano-sdk/util';
+import { readCredentialType } from '../Common/Credential';
 import type * as Cardano from '../../Cardano';
 
 const EMBEDDED_GROUP_SIZE = 2;
@@ -86,7 +87,7 @@ export class AuthCommitteeHot {
         `Expected an array of ${EMBEDDED_GROUP_SIZE} elements, but got an array of ${length} elements`
       );
 
-    const coldType = Number(reader.readInt()) as Cardano.CredentialType;
+    const coldType = readCredentialType(reader);
     const coldHash = Crypto.Hash28ByteBase16(HexBlob.fromBytes(reader.readByteString()));
 
     reader.readEndArray();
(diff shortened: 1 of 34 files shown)


AngelCastilloB (approved):
Great work @mkazlauskas

mirceahasegan (approved):
🚀

PR #1686, opened 2026-06-17
feat(tx-construction): expose tx-builder utils and add fluent minting

rhyslbw:
Closes gaps in `@cardano-sdk/tx-construction` that forced consumers (e.g. Lace's full-domain Cardano transaction builder) to reimplement SDK internals or drop to lower-level APIs.

## 1. Export tx-builder utils

`buildRedeemers`, `buildWitness` and `computeCollateral` are defined in [`tx-builder/utils.ts`](packages/tx-construction/src/tx-builder/utils.ts) but the barrel never re-exported `./utils`, so they were absent from the public API. Fixed with `export * from './utils'` (pure addition, no collisions).

## 2. Fluent minting on the transaction builder

Add `TxBuilder.addMint({ policy, assets, redeemer })`. The policy script is attached to the witness set, the asset quantities are added to the body, and for Plutus policies the redeemer is included and its execution units evaluated during `build()`. Previously `GenericTxBuilder` had **no minting support** — `mint` was never forwarded to `initializeTx`.

## 3. Canonical redeemer indexing (review follow-up)

Redeemer pointers must reference an item's position in the **canonically-sorted transaction-body collection**, not insertion order. `RedeemersByType` now keys redeemers by on-chain identifier (mint → `PolicyId`, withdrawal → `RewardAccount`, vote → `Voter`), and `computeMinimumCost` derives each index from the body:

- **mint**: rank of the policy id in the sorted minting policies.
- **withdrawal**: rank among script-credential reward accounts sorted by hash; key-hash withdrawals carry no redeemer and don't shift the index (node quirk).
- **vote**: position in the canonically-sorted voters (voter group, then script-before-keyhash credential, then hash).
- **certificate / propose**: caller-provided index (item position in body order).

The `@cardano-sdk/web-extension` remote-api property map gains the new `addMint` method.

## Tests

- Unit (`TxBuilderPlutusScripts.test.ts`): plutus mint with redeemer, **burn** (negative quantity), **native-script mint without a redeemer**, and **multi-policy** redeemer indexing.
- Canonical-index vector (`selectionConstraints.test.ts`): withdrawal indices computed among script reward accounts by hash, with a key-hash withdrawal interleaved and the body out of canonical order. Full `@cardano-sdk/tx-construction` suite green (221).
- E2E (`PersonalWallet/mint.test.ts`): the mint scenario now builds through `createTxBuilder().addMint(...).build().sign()` instead of `initializeTx`, validated on-chain by the e2e workflow (`wallet_epoch_0`, runs on PR).

## Open

Vote redeemer ordering is derived from the ledger Voter Ord (the cardano-c reference for votes was past the readable excerpt) and lacks a dedicated vector — flagged in-thread for confirmation against the reviewer's vote vectors. Withdrawal-with-redeemer is otherwise reachable via the documented `customize` hook; a first-class "build unsigned tx for external signers" path remains a follow-up.

### DIFF
--- a/packages/tx-construction/src/input-selection/selectionConstraints.ts
+++ b/packages/tx-construction/src/input-selection/selectionConstraints.ts
@@ -17,13 +17,26 @@ export const MAX_U64 = 18_446_744_073_709_551_615n;
 
 export type BuildTx = (selection: SelectionSkeleton) => Promise<Cardano.Tx>;
 
+/**
+ * Redeemers grouped by purpose. Each purpose is keyed by the on-chain identifier its redeemer
+ * points at, so the canonical (ledger) redeemer index can be derived from the transaction body
+ * during {@link computeMinimumCost}:
+ *
+ * - `spend` by `txId#index`
+ * - `mint` by minting `PolicyId`
+ * - `withdrawal` by script `RewardAccount`
+ * - `vote` by `Voter`
+ *
+ * `certificate` and `propose` redeemers are sequence-indexed in transaction-body order, which is
+ * what the ledger expects for those purposes, so they remain ordered lists.
+ */
 export interface RedeemersByType {
   spend?: Map<TxIdWithIndex, Cardano.Redeemer>;
-  mint?: Array<Cardano.Redeemer>;
+  mint?: Map<Cardano.PolicyId, Cardano.Redeemer>;
+  withdrawal?: Map<Cardano.RewardAccount, Cardano.Redeemer>;
+  vote?: Array<{ voter: Cardano.Voter; redeemer: Cardano.Redeemer }>;
   certificate?: Array<Cardano.Redeemer>;
-  withdrawal?: Array<Cardano.Redeemer>;
   propose?: Array<Cardano.Redeemer>;
-  vote?: Array<Cardano.Redeemer>;
 }
 
 export interface DefaultSelectionConstraintsProps {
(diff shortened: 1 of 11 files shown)


packages/tx-construction/src/input-selection/selectionConstraints.ts
```diff
@@ -75,6 +75,23 @@ const updateRedeemers = (
...
 };
 
+/**
+ * `buildRedeemers` stamps builder-provided redeemers with a sentinel index. Spend redeemers are
+ * reindexed by input position elsewhere; the remaining (mint/withdrawal/certificate/...) redeemers
+ * carrying the sentinel are reindexed here by their position within their purpose, matching the
+ * canonical ordering in which they were added (e.g. mint redeemers in sorted-policy order).
+ * Redeemers that already carry a concrete index are left untouched.
+ */
+const reindexSentinelRedeemers = (redeemers: Cardano.Redeemer[]): void => {
```
AngelCastilloB: @rhyslbw This isn't canonically accurate, it can (and will) produce the wrong indices for `mint`, `withdrawal` and `vote`. The ledger doesnt index these by insertion order, it indexes by the items position in the canonically sorted collection (sorted policy IDs for mint, sorted RewardAccount/Voter maps for withdrawal/vote).

But there is a quirk in the sort: reward accounts and voters are keyed by `Credential`, whose derived `Ord` puts `ScriptHashObj` before `KeyHashObj`, so all `scripthash` credentials sort ahead of all `keyhash` ones regardless of their hash bytes. So with a mix of `keyhash` and `scripthash` credentials, appearance order won't match what the node computes and validation fails with a redeemer pointer mismatch.

Right now we are relying on the caller having added them in the correct order, but given these quirks that's really finicky, we should derive the index here from the sorted body collections (the way `reorgRedeemers` already does for spend inputs) instead of trusting insertion order.

We can probably get rid of this function entirely and fold the logic into `reorgRedeemers`. The blocker is that `Cardano.Redeemer` only carries `{ index, purpose, data, executionUnits }`, it has no back reference to its `policy/account/voter`. `Spend` works because it's keyed (`Map<TxIdWithIndex, Redeemer>`), so the fix is to give the other purposes the same treatment: 

key mint by `PolicyId`, withdrawal by `RewardAccount`, vote by `voter`, then in `reorgRedeemers` set each index to the position of that key in the canonically sorted body collection (with the script-before-keyhash rule for withdrawal/vote). cert/propose stay sequence-indexed, which is what the ledger does for those, as long as we emit them in body order

If you give an LLM this context plus the following links, it should be able implement the correct behaviour + use same vectors:

cardano-ledger:

  - getAlonzoScriptsNeeded (from this point onwards the LLM should be ablke to piece the whole flow):
  https://github.com/IntersectMBO/cardano-ledger/blob/8dc1c431e06db2c8b5d44fb4b3cca6419197d763/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/UTxO.hs#L303

I implemented this manually in `Biglup/cardano-c` so you should be able to reference this too and is probably easier to port to JS/TS than haskell:

  - transaction_builder dir: https://github.com/Biglup/cardano-c/tree/238fe346192f2a2d54997fd98ba3bb576e3c51ea/lib/src/transaction_builder
  - sort-by-key + reassign indices on every insert: https://github.com/Biglup/cardano-c/blob/238fe346192f2a2d54997fd98ba3bb576e3c51ea/lib/src/transaction_builder/internals/blake2b_hash_to_redeemer_map.c#L272-L294
  - compare_by_bytes comparator: https://github.com/Biglup/cardano-c/blob/238fe346192f2a2d54997fd98ba3bb576e3c51ea/lib/src/transaction_builder/internals/blake2b_hash_to_redeemer_map.c#L124-L134
  - withdrawal script-first quirk:
  https://github.com/Biglup/cardano-c/blob/238fe346192f2a2d54997fd98ba3bb576e3c51ea/lib/src/transaction_builder/transaction_builder.c#L2726-L2729
  - vote script-first quirk:
  https://github.com/Biglup/cardano-c/blob/238fe346192f2a2d54997fd98ba3bb576e3c51ea/lib/src/transaction_builder/transaction_builder.c#L3610-L3613

rhyslbw: You're right, insertion order was wrong.

`RedeemersByType` now keys redeemers by their on-chain identifier (mint → `PolicyId`, withdrawal → `RewardAccount`, vote → `Voter`); `reindexSentinelRedeemers` is gone and indices are derived from the canonically-sorted body collections in `computeMinimumCost`:

- **mint**: rank of the `PolicyId` in the sorted minting policies.
- **withdrawal**: rank among the **script-credential** reward accounts sorted by credential hash — key-hash withdrawals carry no redeemer and don't shift the index (the node quirk you flagged). Vector test added: a body listing scripts out of canonical order with a key-hash withdrawal interleaved; the script redeemers resolve to 0/1 by hash.
- **vote**: position in the canonically-sorted voters (group: Committee < DRep < StakePool, then script-before-keyhash credential, then hash).
- **certificate/propose**: caller-provided index (the item's position in body order), which is what the ledger uses for those.

One thing I'd value your eyes on: the **vote** ordering. The `transaction_builder.c` vote section was past the excerpt I could read, so I derived the Voter Ord from the ledger constructor order + Credential Ord rather than from your implementation — and unlike mint/withdrawal it has no dedicated vector test yet. Could you point me at your vote vectors (or confirm the rule)? In particular: do votes get the same script-first quirk as withdrawals, or is the index the position among *all* voters?

Used `getAlonzoScriptsNeeded` + your cardano-c links throughout — thanks for those.

AngelCastilloB: Implemenattion is correct 👍🏼

packages/tx-construction/test/tx-builder/TxBuilderPlutusScripts.test.ts
```diff
@@ -282,6 +282,101 @@ describe('TxBuilder/plutusScripts', () => {
...
+    const tx = await txBuilder
+      .addMint({ assets: new Map([[Cardano.AssetName('41'), 1n]]), policy: policyA, redeemer: 1n })
+      .addMint({ assets: new Map([[Cardano.AssetName('42'), 1n]]), policy: policyB, redeemer: 2n })
+      .build()
+      .inspect();
+
+    const mintRedeemers = (tx.witness?.redeemers ?? []).filter((r) => r.purpose === Cardano.RedeemerPurpose.mint);
+    expect(mintRedeemers).toHaveLength(2);
+    // Reindexed to canonical sequential positions, each with evaluated execution units.
+    expect(mintRedeemers.map((r) => r.index).sort()).toEqual([0, 1]);
```
AngelCastilloB: This doesn't really exercise the reindexing. Sorting the indices and dropping the data into a Set loses which policy maps to which index, so it still passes with plain insertion order. Add the two policies in reverse canonical order and assert that the one whose policy id sorts first gets index 0.

AngelCastilloB (dismissed):
Great work @rhyslbw

packages/tx-construction/test/tx-builder/TxBuilderPlutusScripts.test.ts
rhyslbw: Good point. Now adds the two policies in reverse canonical order and asserts the policy whose id sorts first gets index 0 (matched via redeemer data) — insertion order would fail it.

AngelCastilloB (dismissed):

AngelCastilloB (approved):

Issue #1707, opened 2026-06-19
Remaining medium-severity vulnerabilities (major bumps + exact pins + no-fix)


rhyslbw:
## Context
After the in-range medium relock (commit `chore(deps): patch medium-severity transitive vulnerabilities via in-range relock`), these medium alerts remain. None are closable by a safe in-range relock.

## Consumer-facing — coordinated/major bumps
| Package | Stuck | Patched | Owner | Notes |
|---|---|---|---|---|
| `uuid` | 3.4.0 / 8.x / 9.0.0 / 10.0.0 | >= 11.1.1 | **direct deps**: cardano-services `^10.0.0`, projection-typeorm `^9.0.0`, wallet/web-extension/e2e `^8.3.2` | Coordinated **major** bump (8/9/10 -> 11) across 5 packages; align on one major. |
| `ip-address` | 9.0.5 | >= 10.1.1 | `@cardano-sdk/core` direct `^9.0.5` (and `socks`) | major 9 -> 10. |
| `postcss` | 7.0.39 | >= 8.4.31 | dev/build transitive | 7.x has no fix; needs parent off postcss 7 (8.x copy already patched). |
| `tough-cookie` | 2.5.0 | >= 4.1.3 | transitive (request/jsdom era) | major 2 -> 4 (4.x copy already patched). |

## Exact / narrow pins (need parent bump or scoped resolution)
- `qs` 6.5.5 (`~6.5.2`, low-major line) — the 6.11.0/6.13.0 exact pins were resolved in PR #1709; the rest of qs is at 6.15.2.
- `js-yaml` 4.1.0 (exact, dev) — `^4.1.0` consumers already at 4.2.0.
- `nanoid` 3.3.3 (exact) — `^3.3.4` consumers already at 3.3.13.
- `esbuild` 0.21.5 (`~0.21.4`, dev) — patched 0.25.0; needs the pinning parent bumped.

## No upstream fix available (monitor / risk-accept)
- `pkg` (<= 5.8.1) — archived/deprecated project, no patch. Pulled by `@cardano-sdk/golden-test-generator` (`^5.3.0`). Consider replacing `pkg` (vercel/pkg is EOL) with an alternative bundler/SEA, or risk-accept (build-time tool).

## Already covered by existing issues
- `axios` medium alerts -> #1704
- `fast-xml-parser`, `vm2` -> #1702
- `serialize-javascript`, `tar` -> #1706

## Acceptance criteria
[...]

PR #1718, opened 2026-06-19
chore(deps): bump uuid to ^11.1.1 (minimal node-16-safe fix)

rhyslbw:
## Summary

Bumps the direct `uuid` dependency from `^8`/`^9`/`^10` to **`^11.1.1`** in `cardano-services`, `e2e`, `projection-typeorm`, `wallet`, and `web-extension`.

This is the **minimal** version that clears the advisory (uuid `< 11.1.1`), chosen deliberately over Dependabot's `uuid@14` (PR #1710, closed) which **requires Node 20+** and conflicts with this repo's `engines: >=16.20.2`. `uuid@11` has no `engines` constraint and is Node-16-safe.

Part of #1707.

## Why it's safe
- We only use named imports (`import { v4 } from 'uuid'`) across all 5 packages — no removed deep imports (`uuid/v4`) — so v11 is API-compatible.
- `uuid@11` ships its own TypeScript types, so the now-redundant `@types/uuid@^8` is removed from the 3 packages that declared it.

## Scope / remaining
Closes the **5 direct-manifest** uuid alerts. The transitive copies (`uuid@8.3.2`/`9.0.0` via `pg-boss`, `pouchdb`, `jayson`, `rpc-websockets`, `@lerna-lite/version`, `devtools`) remain — clearing them would need a global `uuid` resolution that risks those packages, so it stays deferred under #1707.

## Test plan
- [x] full `yarn build` green
- [x] `wallet` 440/440 (uses `uuid.v4` in PouchDbCollectionStore)
- [x] `web-extension` build green; uuid-using messaging tests pass (the 3 `NonBackgroundMessenger` reconnect-timing failures are **pre-existing on master** in this environment and unrelated to uuid)
- [ ] CI: confirm `build_and_test` green

### DIFF
--- a/packages/cardano-services/package.json
+++ b/packages/cardano-services/package.json
@@ -69,7 +69,6 @@
     "@types/express-prometheus-middleware": "^1.2.1",
     "@types/lodash": "^4.14.182",
     "@types/pg": "^8.6.5",
-    "@types/uuid": "^8.3.4",
     "@types/wait-on": "^5.3.1",
     "@types/ws": "^8.5.10",
     "axios-mock-adapter": "^2.0.0",
@@ -132,7 +131,7 @@
     "ts-custom-error": "^3.2.0",
     "ts-log": "^2.2.4",
     "typeorm": "^0.3.15",
-    "uuid": "^10.0.0",
+    "uuid": "^11.1.1",
     "ws": "^8.17.1"
   },
   "files": [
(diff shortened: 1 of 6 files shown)


rhyslbw:
Superseded by #1723, which consolidates all the pure dependency bumps into a single follow-up to the Node 22 upgrade (#1719).

Issue #1704, opened 2026-06-19
Bump axios to >=1.16.0 (high-severity, consumer-facing — needs CI validation)


rhyslbw:
## Context
High-severity Dependabot alerts affect `axios` (SSRF / credential-leak / DoS family, ~23 alerts). It is **consumer-facing**, a direct `dependencies` entry in `cardano-services`, `cardano-services-client`, and `util-dev` (`axios: ^1.7.4`).

The in-range relock to `axios@1.18.0` was prepared but **deliberately excluded** from the low-risk security batch: axios is the HTTP client behind `HttpProvider`, and its HTTP behavior cannot be validated in the sandbox used for the batch (the `HttpProvider` tests require local sockets that aren't available there — they fail pre-existingly, independent of axios version). This bump must be validated by CI with real networking.

## The fix
- `^1.7.4` already permits `1.18.0`; a lockfile refresh moves the 1.x copy. Optionally bump the manifest range to `^1.16.0` to make the floor explicit.
- Two legacy **0.x** copies remain (`0.25.0`, `0.27.2`) pulled by old transitive deps; these need their parents bumped (axios 0.x -> 1.x is a major bump) and may not be fully closable without those.

## Acceptance criteria
- [ ] axios 1.x resolves to >= 1.16.0
- [ ] `cardano-services-client` `HttpProvider` suite green in CI
- [ ] `cardano-services` + provider integration tests green in CI
- [ ] Remaining axios 0.x copies tracked/closed where a parent bump exists

rhyslbw:
Resolved by #1723 (merged to master as 97da539). `axios` relocked to **1.18.0** (within `^1.7.4`), above the >=1.16.0 fix line. Validated by the full CI/e2e suite on the merged commit.

PR #1721, opened 2026-06-19
chore(deps): relock axios to 1.18.0

rhyslbw:
Relocks axios so the `^1.7.4` consumers (cardano-services, cardano-services-client, util-dev) resolve to **1.18.0**, clearing the axios 1.x advisories (`< 1.16.x`). In-range relock — no manifest change. Resolves the consumer-facing part of #1704; now viable on the Node 22 base.

**Scope:** a legacy `axios@0.25.0` remains via an old transitive (0.x→1.x is a major for that parent) — tracked in #1704.

**Test plan:**
- [x] full clean build green on axios 1.18 (types)
- [ ] CI: `HttpProvider` + provider HTTP tests (the local sandbox can't run the in-process socket tests)

> Stacked on #1719 (Node 22). Retarget to `master` once it merges.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -8954,13 +8954,14 @@ __metadata:
   linkType: hard
 
 "axios@npm:^1.7.4":
-  version: 1.11.0
-  resolution: "axios@npm:1.11.0"
+  version: 1.18.0
+  resolution: "axios@npm:1.18.0"
   dependencies:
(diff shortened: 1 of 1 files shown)


rhyslbw:
Superseded by #1723, which consolidates all the pure dependency bumps into a single follow-up to the Node 22 upgrade (#1719).

Issue #1726, opened 2026-06-20
Protocol-param string values are degraded to float before rational conversion (precision loss)


rhyslbw:
## Summary

Several protocol parameters are modelled as exact decimal **strings** in the domain types, but `core`'s serialization degrades them to a float and then *re-approximates* a rational from that float. The round-trip is lossy and feeds **consensus-affecting CBOR** (protocol-param updates, `UnitInterval` tag-30 rationals).

## The bug

In `packages/core/src/Serialization/Update/ProtocolParamUpdate.ts`, e.g.:

```ts
// parametersUpdate.poolInfluence is typed `string` (see Cardano/types/ProtocolParameters.ts)
UnitInterval.fromFloat(Number(parametersUpdate.poolInfluence))
```

`UnitInterval.fromFloat` (`Serialization/Common/UnitInterval.ts`) calls `new Fraction(number)`, whose single-arg path is a **Stern-Brocot best-rational-approximation** (`fraction.js`), not an exact parse. So the pipeline is:

`"0.51"  →  Number("0.51")  →  Stern-Brocot approximation  →  n/d`

The input string is already exact, so the `Number()` hop throws away precision and the approximation can land on a different rational than the value denotes — which then serializes to CBOR.

### Affected string params (degraded via `Number(...)` → `fromFloat`)
- `poolInfluence`, `monetaryExpansion`, `treasuryExpansion`, `decentralizationParameter`, `minFeeRefScriptCostPerByte` — all typed `string` in `Cardano/types/ProtocolParameters.ts`.

## Proposed fix

Parse the decimal string **directly** to a reduced rational (`"0.51" → 51/100`) instead of `Number()` → `fromFloat`. This is exact, deterministic, trivial, and self-contained to `core` (`UnitInterval` + `ProtocolParamUpdate`). Add tests demonstrating the exact-vs-approximate difference.

Bonus: removes the `fraction.js` dependency from this path.

## Related: float-sourced rationals (separate, larger scope)

Other rational params genuinely arrive as `number` at the boundary and still need conversion:

| Source | Field(s) | Type | Can it be made exact? |
|---|---|---|---|
| db-sync | `dvt_*` voting thresholds, prices | `number` (Postgres `numeric` coerced to JS float) | **Yes** — return the `numeric` columns as **strings** from the query (lossless), then parse string→rational. Touches `cardano-services`. |
[...]

PR #1735, opened 2026-07-07
feat: update serialization and core types to Dijkstra era [LW-15103]

AngelCastilloB:
# Context

This PR updates the serialization layer and core types in `@cardano-sdk/core` to support the upcoming Dijkstra era (protocol version 12). Everything is based on the dijkstra CDDL and the reference implementation in the cardano-ledger repo, pinned at commit [`9b81d994`](https://github.com/IntersectMBO/cardano-ledger/tree/9b81d9943ce79b8531e9d8e2c5f9adcb158677cb):

- CDDL: [eras/dijkstra/impl/cddl/data/dijkstra.cddl](https://github.com/IntersectMBO/cardano-ledger/blob/9b81d9943ce79b8531e9d8e2c5f9adcb158677cb/eras/dijkstra/impl/cddl/data/dijkstra.cddl)
- Reference implementation: [eras/dijkstra/impl/src/Cardano/Ledger/Dijkstra](https://github.com/IntersectMBO/cardano-ledger/tree/9b81d9943ce79b8531e9d8e2c5f9adcb158677cb/eras/dijkstra/impl/src/Cardano/Ledger/Dijkstra)

Note the era is still under development upstream (peras/leios certs are placeholders, plutus v4 script context is stubbed), so we will most likely need a follow up PR to re sync with the final CDDL before the hardfork.

# Proposed Solution

Same approach we took for conway, no era abstractions, eras are backward compatible on the wire so we just extend the integer keyed maps and enum variants. Almost everything here is additive, there are only two deliberate breaking changes (see below).

- Guards (CIP-112): body key 14 now accepts the legacy key hash set or a credential ordered set. Key hash only bodies still serialize byte identical to conway, the credential form is only emitted when script hash guards are present. The decoder disambiguates by peeking at the first element, same as [decodeGuards](https://github.com/IntersectMBO/cardano-ledger/blob/9b81d9943ce79b8531e9d8e2c5f9adcb158677cb/eras/dijkstra/impl/src/Cardano/Ledger/Dijkstra/TxBody.hs) in the ledger.
- Nested transactions (CIP-118): new `SubTransactionBody` and `SubTransaction` types, wired into body key 23 as an ordered set keyed by sub tx id (the body decoder rejects empty sets and duplicate ids for key 23; the generic CborSet stays permissive). Sub bodies have no fee/collateral keys and add key 24 `required_top_level_guards`, the decoder is strict about wrong level keys, mirroring the ledger `decoderByKey`.
- Plutus V4: script union tag 4 (hash prefix 0x04), cost models key 3, auxiliary data key 5. There is no V4 witness set key in this CDDL revision, thats intentional, upstream still has `eraMaxLanguage = PlutusV3`.
- New protocol params 34-37 (maxRefScriptSizePerBlock, maxRefScriptSizePerTx, refScriptCostStride, refScriptCostMultiplier).
- Direct deposits (body key 25) and account balance intervals (body key 26).
- RequireGuard native script (kind 6) and the guarding redeemer tag (6).
- Wire tightenings: bootstrap witness chain_code fixed at 32 bytes, protocol_version minor is now uint32, values never emit an empty multiasset map.
- Legacy stake certs (kinds 0/1) are marked deprecated, the ledger's dijkstra decoder rejects them outright, new transactions should use kinds 7/8 with explicit deposit.

Every type has byte exact round trip tests plus toCore/fromCore symmetry tests. Vectors were taken from the ledger golden files where they exist (the full golden tx is vendored verbatim), the rest were constructed and validated against the pinned CDDL with the ruby cddl tool. There is also an integration test with a maximal transaction exercising every new construct in a single blob.

Two breaking changes:

- `Transaction.setIsValid` was removed and the flag is now hardcoded to true. The is_valid flag is deprecated in dijkstra, the mempool only accepts true during the grace period and the flag is stripped once the tx is in a block, so there was no valid use case left for setting it. We still emit the 4 element frame for byte compatibility, decoding a false flag now throws.
- `inConwayEra` now defaults to true. Dijkstra removes the flat array redeemer encoding entirely (the ledger rejects it from PV 12), so map form redeemers and tag 258 sets are now the default output. `setInConwayEra(false)` still works if you need pre conway bytes.


packages/core/src/Serialization/TransactionBody/Guards.ts
```diff
@@ -0,0 +1,128 @@
...
+   */
+  static fromCbor(cbor: HexBlob): Guards {
+    const reader = new CborReader(cbor);
+
+    if (reader.peekState() === CborReaderState.Tag && reader.peekTag() === CborTag.Set) reader.readTag();
+
+    reader.readStartArray();
+
+    if (reader.peekState() === CborReaderState.EndArray)
+      throw new InvalidArgumentError('cbor', 'Guards must contain at least one element');
```
mkazlauskas: Alonzo/[Babbage](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/babbage/impl/cddl/data/babbage.cddl#L587) CDDL: `required_signers = set<addr_keyhash>` - empty legal, on-chain data may contain it. Master decoded it (CborSet accepts empty). Could be a problem for some use cases like projector/indexers

mkazlauskas (changes requested):

packages/core/src/Serialization/TransactionBody/Guards.ts
AngelCastilloB: good catch, fixed here https://github.com/input-output-hk/cardano-js-sdk/pull/1735/changes/9c3637107e9c096d1e90f0ac6ed1ccdb4da4b1a4, will only throw with the use of credentials but allow empty set<addr_keyhash>

mkazlauskas (dismissed):
Great work! 💪

packages/core/src/Serialization/Transaction.ts
```diff
@@ -100,11 +104,15 @@ export class Transaction {
...
-    let isValid = true;
 
-    // The isValid flag was added in Alonzo era (onwards), mary era transactions only have three fields.
-    if (length === ALONZO_ERA_TX_FRAME_SIZE) {
-      isValid = reader.readBoolean();
+    // The is_valid flag was added in Alonzo; Mary era transactions only have three fields. Dijkstra
+    // deprecates the flag and its decoder rejects false outright, mirroring cardano-ledger
+    // decodeDijkstraTopTx: 'Value `false` not allowed for `isValid`'.
+    if (length === ALONZO_ERA_TX_FRAME_SIZE && !reader.readBoolean()) {
+      throw new SerializationError(
```
rhyslbw: **Does rejecting `is_valid=false` here break reads of historical chain data?**

This throw isn't era-gated, and `Transaction.fromCbor` is the SDK's universal decoder for txs from every era. Phase-2-validation-failed transactions (the collateral-collecting case) are recorded on mainnet from Alonzo onward with `is_valid=false`, so this looks like it rejects legitimate historical data, not just newly-authored Dijkstra txs. The added test that decodes a Babbage frame with the flag flipped to false and expects a throw seems to lock that in on purpose — is that the intent?

Two consumer paths I traced that feed arbitrary chain txs through here:
- `packages/ogmios/src/ogmiosToCore/tx.ts:469` — `mapBlockBody` calls `Serialization.Transaction.fromCbor(tx.cbor).toCore()` with no catch, so it would throw while mapping any block containing a phase-2-failed tx.
- `packages/cardano-services-client/src/ChainHistoryProvider/BlockfrostChainHistoryProvider.ts:211` — `fetchDetailsFromCBOR` wraps it in `.catch → return null`, so such a tx would be silently dropped from history (the `valid_contract === false` collateral case is handled just below, at :438).

Am I reading this right, or is there an era-gating / ingest plan that keeps historical decode lossless? If the deprecation really only applies to *authoring* new Dijkstra txs, would it work to keep decode permissive (surface the real flag through `toCore`) and enforce `true` only on `fromCore` / `toCbor`? Related: since `toCore` now hardcodes `isValid: true`, `Cardano.Tx` can no longer represent a phase-2 failure at all — is that acceptable for downstream consumers that read chain history?

packages/core/test/Serialization/dijkstraVectors.ts
```diff
@@ -0,0 +1,169 @@
...
+    name: 'ledger-golden-full',
```
rhyslbw, line 125: This is the only byte-vendored ledger golden in the set — the strongest correctness oracle in the PR — but as far as I can tell it's never actually decoded. `vectorsForRule('transaction')` is consumed once (`Transaction.test.ts:158-159`) and immediately `.find(v => v.name === 'minimal-without-is-valid')`, so these bytes are only checked by the metadata assertions in `dijkstraVectors.test.ts`. Did I miss a consumer? If not, would you be up for a `decode → round-trip → assert tx-id` test against it? The bytes are already here, so it should be cheap and it turns the golden from dead data into the anchor that proves the encoder matches ledger output.

packages/core/test/Serialization/dijkstraVectors.ts
```diff
@@ -0,0 +1,169 @@
+/**
+ * Dijkstra-era CBOR test vectors, vendored or rendered from IntersectMBO/cardano-ledger
+ * at commit 9b81d9943ce79b8531e9d8e2c5f9adcb158677cb, or constructed and validated
+ * against that commit's dijkstra.cddl.
+ */
+
+const LEDGER = 'cardano-ledger 9b81d9943ce79b8531e9d8e2c5f9adcb158677cb';
+const GOLDEN_HS = `${LEDGER} eras/dijkstra/impl/testlib/Test/Cardano/Ledger/Dijkstra/Binary/Golden.hs`;
+const EXAMPLES_HS = `${LEDGER} eras/dijkstra/impl/testlib/Test/Cardano/Ledger/Dijkstra/Examples.hs`;
+const CONSTRUCTED = 'constructed, validated against the pinned dijkstra.cddl';
```
rhyslbw, line 10: The provenance for the constructed vectors says "validated against the pinned dijkstra.cddl", and commit `3e605a96`'s message describes a `validate-cddl.mjs` that runs the ruby `cddl` CLI — but I can't find that script (or any `.cddl`) in the tree or in `git ls-files`. Was it meant to be committed? Without it wired into CI, the CDDL-conformance of the ~13 constructed vectors isn't reproducible; it rests on a one-time local run. Would it make sense to land the validator as an optional CI job so this stays verifiable as the CDDL is re-synced before the hardfork?

packages/core/test/Serialization/dijkstraVectors.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  describe.each(dijkstraVectors)('$rule/$name', (vector) => {
+    it('has non-empty lowercase hex of even length', () => {
+      expect(vector.hex.length).toBeGreaterThan(0);
+      expect(vector.hex).toMatch(/^(?:[\da-f]{2})+$/);
+    });
+
+    it('states its provenance', () => {
+      expect(vector.provenance.length).toBeGreaterThan(0);
+      expect(vector.provenance).toMatch(/cardano-ledger 9b81d994|constructed/);
```
rhyslbw, line 78: Minor / naming: this suite asserts hex formatting + a provenance regex + uniqueness, but never decodes or encodes a vector — the real round-trips live in the per-type tests, and only for the vectors those tests import. So a green `dijkstraVectors` run doesn't actually tell us the bytes are serialization-correct (byte-garbage that happened to be valid lowercase hex would pass every assertion here). Given the file's name it reads like a golden-conformance suite; might be worth either a round-trip-every-vector test to match that expectation, or a comment/rename clarifying it's a manifest lint. Not blocking.

packages/core/src/Serialization/Certificates/CertificateKind.ts
```diff
@@ -36,10 +44,18 @@ export enum CertificateKind {
   /**
    * This certificate is used to delegate from a Genesis key to a set of keys. This was primarily used in the early
    * phases of the Cardano network's existence during the transition from the Byron to the Shelley era.
+   *
+   * @deprecated Deprecated since the Conway era and rejected by the Dijkstra decoder. Retained for
```
rhyslbw: These JSDoc lines (kinds 5 and 6) say they're "rejected by the Dijkstra decoder", but `Certificate.fromCbor` still decodes both (`Certificate.ts:83,86`) with no rejection — and the kinds 0/1 JSDoc just above (lines 13/24) says the opposite for the same actual behavior ("Retained for decoding historical chain data"). Is the rejection intended-but-not-yet-implemented, or should 5/6 read like 0/1 ("retained for decoding")? The PR description also says "the dijkstra decoder rejects them outright" for 0/1, which doesn't match what the decoder does today — just want doc and code to agree so nobody relies on a rejection that isn't there.

packages/core/src/Serialization/TransactionBody/AccountBalanceInterval.ts
```diff
@@ -0,0 +1,169 @@
...
+  /**
+   * Serializes an AccountBalanceInterval into CBOR format.
+   *
+   * @returns The AccountBalanceInterval in CBOR format.
+   */
+  toCbor(): HexBlob {
+    if (this.#originalBytes) return this.#originalBytes;
+
+    if (this.#inclusiveLowerBound === undefined && this.#exclusiveUpperBound === undefined)
+      throw new InvalidStateError(BOTH_NIL_MESSAGE);
```
rhyslbw, line 43: The "at most one nil bound" invariant is enforced here on encode (and on `fromCbor`), but not in the constructor / `fromCore` — `new AccountBalanceInterval()` builds a both-nil object that stays invalid until `toCbor()` throws. Would it be cleaner to validate in the constructor so the invariant holds by construction and fails at the boundary rather than at serialization time? Minor.

packages/core/src/Serialization/Update/ProtocolParamUpdate.ts
```diff
@@ -444,6 +510,10 @@ export class ProtocolParamUpdate {
       poolRetirementEpochBound: this.#maxEpoch,
       poolVotingThresholds: this.#poolVotingThresholds?.toCore(),
       prices: this.#executionCosts?.toCore(),
+      refScriptCostMultiplier: this.#refScriptCostMultiplier
+        ? this.#refScriptCostMultiplier.toFloat().toString()
```
rhyslbw, line 514: `toCore` stringifies `toFloat()` and `fromCore` (`:594-595`) does `UnitInterval.fromFloat(Number(...))`, so the exact numerator/denominator isn't preserved across `toCore → fromCore → toCbor` for this rational. The wire decode→encode path is safe via the originalBytes cache, so this only affects the core round-trip. I see it matches the existing `treasuryExpansion` / `minFeeRefScriptCostPerByte` pattern, so this may be a deliberate accepted tradeoff — but since key 37 is new and consensus-affecting, is float fidelity good enough here, or worth carrying num/denom like some other rationals do? Non-blocking, just checking.

packages/core/src/Serialization/TransactionBody/SubTransactionBody.ts
```diff
@@ -0,0 +1,1064 @@
+/* eslint-disable sonarjs/cognitive-complexity, complexity, max-statements */
```
rhyslbw, line 1: Big-picture maintainability question rather than a bug: this file is largely a parallel copy of `TransactionBody.ts` — the per-field encoders/decoders, and notably the canonical mint / withdrawal / credential-map sorts. Since both feed the same transaction hash, a future fix to a canonical-ordering or set-tag edge case in `TransactionBody.ts` wouldn't automatically reach here, and nothing couples the two. Is a shared per-field codec feasible (sub-tx = top-level minus keys 2/13/16/17/23, plus key 24), or is the duplication a deliberate call given the field-set differences? If full extraction is too invasive for this PR, even a cross-body test asserting identical CBOR for the shared fields would convert silent drift into a red test.

rhyslbw (commented):
Thorough pass on the serialization + test-vector side, reviewed with a few focused lenses (wire correctness, test-vector coverage, type integrity, maintainability). Overall this is careful, high-quality work: the guards split/merge is reversibly correct and doesn't drop script-hash guards, the empty-multiasset normalization is consistent and lossless, the sub-tx id derivation is sound, the strict sub-transaction-body key admission matches the CDDL, and I hand-decoded a sample of the vectors and they conform to their rules. Most of my comments are questions rather than assertions — I may be missing intent in places.

**The one I'd like to resolve before merge** is the `is_valid=false` decode behavior (inline on `Transaction.ts`). As I read it the rejection isn't era-gated, and this decoder is the universal path for reading chain data, so it looks like it would throw/drop legitimate phase-2-failed mainnet transactions from earlier eras. If that holds it's potentially merge-blocking; if I'm misreading the ingest paths I'm happy to be corrected.

**On test vectors** (the axis I weighted most heavily, per the review's golden-standard framing): the vectors are all committed, but the single byte-vendored ledger golden appears to never actually be decoded (inline comment), and the `validate-cddl.mjs` referenced in the commit history isn't in the tree — so "validated against CDDL" isn't reproducible in CI. Wiring the golden into a real round-trip and landing the validator as an optional CI job would raise confidence that the encoders match cardano-ledger and not just themselves.

**One I couldn't anchor inline** (the file isn't touched by this PR): the description says ordered sets "reject duplicates" and empty sets are rejected, but `packages/core/src/Serialization/Common/CborSet.ts#fromCbor` doesn't dedup or check non-empty. It's shared pre-existing code so not a regression, but the credential-form guards oset and the other Dijkstra sets inherit that. Is the duplicate / non-empty guarantee enforced elsewhere, or should that claim be scoped a bit more narrowly?

None of this diminishes the scope here — thanks for the detailed writeup, it made review much easier. Flagging mostly to confirm intent.

packages/core/src/Serialization/Transaction.ts
rhyslbw: Followed up on my own question by checking the CDDL at the pinned commit + provider behavior, and I think you're right about the direction — with one qualifier.

**You're correct for Dijkstra.** [dijkstra.cddl](https://github.com/IntersectMBO/cardano-ledger/blob/9b81d9943ce79b8531e9d8e2c5f9adcb158677cb/eras/dijkstra/impl/cddl/data/dijkstra.cddl) makes the on-chain `transaction` 3-element (`[body, witness_set, auxiliary_data/nil]`, no flag); validity moves to `block_body.invalid_transactions`; the 4-element `transaction_mempool` is `true`-only and the flag is stripped once in a block. So hardcoding `is_valid=true` for authoring/encoding is right, and it's safe to make that the default now.

**The qualifier is pre-Dijkstra decode.** [conway.cddl](https://github.com/IntersectMBO/cardano-ledger/blob/9b81d9943ce79b8531e9d8e2c5f9adcb158677cb/eras/conway/impl/cddl/data/conway.cddl) keeps `transaction = [transaction_body, transaction_witness_set, bool, auxiliary_data/nil]` — the standalone Conway/Babbage/Alonzo serialization genuinely carries `false` for a phase-2 failure, and those are permanent on-chain (collateral-loss txs). Since `Transaction.fromCbor` is the era-blind decoder used to read historical chain data, rejecting `false` on decode would reject/drop real pre-Dijkstra history until the hardfork lands.

Two ways to keep the new default without the read-path regression: (A) default `true` on `fromCore`/encode now but keep **decode permissive** (accept + surface `false`) until Dijkstra is live, or (B) era-gate the `false` rejection so pre-Dijkstra frames still decode.

Caveat on my end: I confirmed the *format* definitively but didn't capture a concrete byte payload of a real failed Conway tx flowing through ogmios `--include-cbor` / Blockfrost `/txs/{hash}/cbor`. Worth noting the current suite doesn't decode a `false` flag from chain data in either direction (Blockfrost tests only cover `valid_contract: true`; the ogmios chain-sync fixtures don't exercise the `cbor` path), so a regression test decoding a known mainnet phase-2-failed Conway tx and asserting `isValid: false` would settle it. Happy to be corrected if the providers turn out to emit something else.

AngelCastilloB: You are right, this would reject phase-2 failed txs from older eras, not the intent. Went with your option A: decode now accepts false and surfaces it through toCore (fromCore also accepts it back so chain history consumers can re serialize), authoring stays hardcoded to true, there is no setter and the flag is only settable by fromCbor. Added the regression tests you suggested. done in 8651ea5171d

packages/core/test/Serialization/dijkstraVectors.ts
AngelCastilloB: Good catch, added a decode -> round trip -> tx id test against the golden, and it immediately paid off: the full decode surfaced a strictness bug, the SDK typed the pool metadata hash as 32 bytes but the CDDL says plain bytes, and the quickcheck golden carries a 2 byte one. Relaxed the decoder, the golden now fully decodes, round trips byte exact and maps to core, with the tx id pinned. done in 6beab7baf78

packages/core/test/Serialization/dijkstraVectors.ts
AngelCastilloB: That was a dev only helper to validate the constructed vectors against the pinned CDDL, dont think it needs to live in the repo, the vectors are already locked in with round trip tests so there is nothing to re run, and if we ever need it again when re syncing the CDDL is cheap to recreate with an LLM.

packages/core/test/Serialization/dijkstraVectors.test.ts
AngelCastilloB: Fair, the suite now round trips every vector that has a decoder (the golden included), the rules that dont have a standalone class are listed explicitly so the gap is visible. done in 6beab7baf78

packages/core/src/Serialization/Certificates/CertificateKind.ts
AngelCastilloB: The rejection is the ledgers decoder, not ours, we decode all four kinds for historical data. Reworded 5/6 to match the 0/1 phrasing and clarified the PR description. done in 91e2e1978a4

packages/core/src/Serialization/TransactionBody/AccountBalanceInterval.ts
AngelCastilloB: Added the check to fromCore. Left the constructor lenient on purpose, these classes follow the construct then set pattern everywhere else in the serialization layer, toCbor remains the backstop. done in 290e9a61e51

packages/core/src/Serialization/Update/ProtocolParamUpdate.ts
AngelCastilloB: Deliberate tradeoff, all the interval params go through the same float path (key 33, treasuryExpansion et al) and the wire round trip is exact via the cached bytes. Agree exact num/denom would be better for a consensus affecting param, but IMO thats a follow up migrating all of them together rather than changing the representation for one key only.

packages/core/src/Serialization/TransactionBody/SubTransactionBody.ts
AngelCastilloB: The duplication was a deliberate call, the key admission and required fields differ enough that parameterizing TransactionBody looked worse. But you are right that nothing coupled them, added a cross body test that feeds the same core input to both and asserts byte identical CBOR for all 18 shared fields (and that neither encoder emits anything outside its expected key set), so drift turns into a red test. The shared per field codec refactor is better as a follow up. done in 2059843ef97

AngelCastilloB:
> the description says ordered sets "reject duplicates" and empty sets are rejected, but CborSet.ts#fromCbor doesn't dedup or check non-empty

CborSet is intentionally permissive (pre existing, shared with all the conway sets), the guarantees live in the field decoders: key 23 rejects duplicate sub tx ids by id, keys 14/24/25/26 reject empty on decode. Enforcing it in CborSet itself would change decode behavior for every pre conway field that uses it. The description claim is about those field decoders, scoped the wording now.

packages/core/src/Serialization/Transaction.ts
rhyslbw: Verified — option A looks right. Decode reads and stores the flag (no throw), `toCbor`/`toCore` surface it, and `fromCore` accepts it back, so a decoded phase-2-failed Conway/Babbage tx now round-trips with `isValid: false` while authoring stays hardcoded to true (no public setter). The three new tests cover exactly the historical-read case. Resolving.

packages/core/test/Serialization/dijkstraVectors.ts
rhyslbw: Nice — and good find on the `PoolMetadata` strictness bug. Confirmed against the pinned CDDL: `pool_metadata = [url, bytes]` (plain unbounded bytes, no `.size 32`), so relaxing the decoder is correct and the golden’s short hash is CDDL-valid. This is exactly why anchoring to the vendored golden was worth it. Resolving.

packages/core/src/Serialization/Update/ProtocolParamUpdate.ts
rhyslbw: Tracked in #1726 (the existing protocol-param precision-loss issue). I extended it with a comment noting this PR adds `refScriptCostMultiplier` (key 37) as another instance of the same float→rational path, so key 37 gets folded into the scope-1 (string→rational) fix there rather than opening a duplicate. Resolving here since the tradeoff is deliberate and consistent with the other interval params.

packages/core/src/Serialization/TransactionBody/SubTransactionBody.ts
rhyslbw: Opened #1736 to track the remaining duplication debt. The cross-body parity test added here is the interim guardrail; #1736 covers the eventual shared-codec extraction.

rhyslbw (approved):
Approving. The blocker is properly resolved and the follow-ups landed cleanly.

**Blocker (is_valid=false):** fixed via option A — `Transaction.fromCbor` now preserves the on-chain flag, `toCbor`/`toCore` surface it, and `fromCore` accepts it back, so historical phase-2-failed Conway/Babbage transactions round-trip with `isValid: false` while authoring stays hardcoded to `true` (no public setter). Backed by targeted decode/round-trip tests. Confirmed against the pinned CDDL: Dijkstra's on-chain `transaction` is 3-element (validity moved to `block_body.invalid_transactions`), so hardcoding true for authoring is correct for the new era, and permissive decode keeps pre-Dijkstra reads lossless.

**Test vectors:** the vendored ledger golden is now actually decoded (stable tx-id + round-trip), and that immediately surfaced and fixed a real `PoolMetadata` strictness bug — the decoder forced a 32-byte hash but the CDDL is `pool_metadata = [url, bytes]` (unbounded). Every decodable vector now round-trips, with unmapped rules listed explicitly.

**Everything else** addressed or declined with sound rationale (cert 5/6 wording, ABI both-nil in `fromCore`, cross-body byte-identity parity test for the SubTransactionBody duplication, CborSet guarantees living in the field decoders).

**Deferred, tracked:**
- Exact rational fidelity for `refScriptCostMultiplier` (key 37) → folded into #1726.
- SubTransactionBody / TransactionBody shared-codec extraction → #1736 (parity test is the interim guardrail).

Note on verification: this review was by code inspection + CDDL cross-check against the pinned ledger commit, not a local suite run — CI remains the test-execution gate. As the PR notes, the era is still under upstream development (peras/leios placeholders, Plutus V4 script context stubbed), so a re-sync pass before the hardfork is expected.

mkazlauskas (approved):

PR #1739, opened 2026-07-22
feat: Dijkstra batch groundwork - hydrated sub transactions and batch-aware inspection

AngelCastilloB:
# Context

Follow-up to the Dijkstra serialization work (LW-15103). Serialization round-trips CIP-0118 batches, guards, direct deposits and account balance intervals, but nothing above that layer surfaces them: the hydrated representation cannot carry sub transactions, the Blockfrost CBOR path silently drops the new body fields, and the tx inspectors compute wrong wallet amounts for a batch (no single level of a batch is balanced on its own). Wallets need this stack ready before the era activates on preview.

# Proposed Solution

Five commits, each independently reviewable:

1. **feat(util-dev): add Dijkstra batch transaction fixture** - deterministic CIP-0118 batch built through the real serialization layer with coherent economics: the batch balances while the top level and both sub transactions are individually unbalanced. Covers an intra-batch input (sub tx 2 spends a sub tx 1 output), guards (key + script), a direct deposit, an account balance interval and a sub tx mint. Ships resolved-input companions for wiring input resolvers in tests.
2. **feat(core): add hydrated Dijkstra sub transaction types** - `HydratedTxBody.subTransactions`, `HydratedSubTransaction` with an explicit `id` (a hydrated body cannot be re-serialized to derive it, and sibling intra-batch inputs reference it).
3. **feat(util-dev): add hydrated view to Dijkstra batch fixture** - the same batch as a `Cardano.HydratedTx`, the exact shape chain-history consumers and inspector tests work with.
4. **fix(cardano-services-client): surface Dijkstra body fields from the Blockfrost CBOR path** - `guards`, `directDeposits` and `accountBalanceIntervals` were deserialized and then dropped by `mapTxBody`; now passed through. Tested against the fixture CBOR through the real deserialize-and-map pipeline.
5. **feat(core): make transactionSummaryInspector Dijkstra batch aware** - `toMergedBatchView` flattens a batch for value arithmetic (intra-batch spends cancel naturally), `createBatchInputResolver` resolves inputs from the batch's own outputs before the caller's resolver, and the summary formula becomes `own outputs + own direct deposits - (own inputs + withdrawals)` with the direct deposit total joining the fee in the unaccounted-funds check. Adds `directDepositsInspector` and `guardsInspector` primitives.

- All changes are additive; no breaking API changes. Existing (non-batch) transactions take the exact same code path as before (`toMergedBatchView` is an identity without sub transactions).
- Sub transaction hydration in `BlockfrostChainHistoryProvider` is deliberately out of scope: it depends on how Blockfrost will represent batch utxos (`txs/{id}/utxos`) for Dijkstra, which is not published yet. The seam is ready once that lands.
- The util-dev fixture doubles as the shared test vector set for downstream work (wallet, Lace rendering).

### DIFF
--- a/packages/cardano-services-client/src/ChainHistoryProvider/BlockfrostChainHistoryProvider.ts
+++ b/packages/cardano-services-client/src/ChainHistoryProvider/BlockfrostChainHistoryProvider.ts
@@ -442,10 +442,13 @@ export class BlockfrostChainHistoryProvider extends BlockfrostProvider implement
 
     const body: Cardano.HydratedTxBody = this.mapTxBody(
       {
+        accountBalanceIntervals: txFromCBOR.body.accountBalanceIntervals,
         certificates: await this.processCertificates(txContent, txFromCBOR.body.certificates),
         collateralReturn: txFromCBOR.body.collateralReturn,
         collaterals,
+        directDeposits: txFromCBOR.body.directDeposits,
         fee: txFromCBOR.body.fee,
+        guards: txFromCBOR.body.guards,
         inputs,
         mint: txFromCBOR.body.mint ? new Map([...txFromCBOR.body.mint].sort()) : undefined,
         outputs,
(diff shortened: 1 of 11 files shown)


AngelCastilloB:
@rhyslbw @mkazlauskas this PR closes a few more gaps to prepare for dijkstra, we will need these to render the TX properly on Lace platform, no of this still activates until we atart receiving Dijkstra era TXs from blockfrost tho

